PR #37, opened 2020-03-21
Interceptor refactoring

mfateev:
Refactored workflow interceptor into:
- WorkflowInterceptor
- WorkflowCallsInterceptor
- WorkflowInvocationInterceptor
- WorkflowInvoker

### DIFF
--- a/src/main/java/io/temporal/internal/sync/DeterministicRunnerImpl.java
+++ b/src/main/java/io/temporal/internal/sync/DeterministicRunnerImpl.java
@@ -92,6 +92,10 @@ class DeterministicRunnerImpl implements DeterministicRunner {
   private final Lock lock = new ReentrantLock();
   private final ExecutorService threadPool;
   private final SyncDecisionContext decisionContext;
+
+  // Note that threads field is a set. So we need to make sure that getPriority never returns the
+  // same value for different threads. We use addedThreads variable for this.
+  //
   // protected by lock
   private final Set<WorkflowThread> threads =
       new TreeSet<>((t1, t2) -> Ints.compare(t1.getPriority(), t2.getPriority()));
--- a/src/main/java/io/temporal/internal/sync/DeterministicRunnerImpl.java
+++ b/src/main/java/io/temporal/internal/sync/DeterministicRunnerImpl.java
@@ -19,6 +19,7 @@
 
 package io.temporal.internal.sync;
 
+import com.google.common.primitives.Ints;
 import com.uber.m3.tally.Scope;
 import io.temporal.context.ContextPropagator;
 import io.temporal.converter.DataConverter;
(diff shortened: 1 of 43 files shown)


src/main/java/io/temporal/internal/sync/POJOWorkflowImplementationFactory.java
```diff
@@ -318,24 +296,76 @@ public void processSignal(String signalName, byte[] input, long eventId) {
...
+      @Override
+      public void init(WorkflowCallsInterceptor interceptor) {
+        WorkflowInternal.getRootDecisionContext().setHeadInterceptor(interceptor);
+      }
+
+      @Override
+      public void processSignal(String signalName, Object[] arguments, long eventId) {
+        Method signalMethod = signalHandlers.get(signalName);
+        try {
+          newInstance();
```
samarabbas: I'm confused on why newInstance is created for both processSignal and execute calls.  Is it expected that only one of them will ever be called?

src/main/java/io/temporal/workflow/WorkflowInterceptor.java
```diff
@@ -19,89 +19,12 @@
...
-
-  void continueAsNew(
-      Optional<String> workflowType, Optional<ContinueAsNewOptions> options, Object[] args);
-
-  void registerQuery(String queryType, Type[] argTypes, Functions.Func1<Object[], Object> callback);
-
-  UUID randomUUID();
-
-  void upsertSearchAttributes(Map<String, Object> searchAttributes);
+  WorkflowInvoker interceptExecuteWorkflow(
```
samarabbas: Is that only going to be used by root context?  Why signal is missing here?

src/main/java/io/temporal/internal/sync/DeterministicRunnerImpl.java
```diff
@@ -71,6 +71,10 @@
 /** Throws Error in case of any unexpected condition. It is to fail a decision, not a workflow. */
 class DeterministicRunnerImpl implements DeterministicRunner {
 
+  private static final int ROOT_THREAD_PRIORITY = 0;
+  private static final int CALLBACK_THREAD_PRIORITY = 10;
+  private static final int WORKFLOW_THREAD_PRIORITY = 20000000;
```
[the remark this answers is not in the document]
mfateev: These priorities are not Java thread priorities. They define the order of cooperative scheduling to ensure that signals are processed before the main method. And for the root thread we need it to ensure that it runs before the first signal is delivered in case of signalWithStart.

src/main/java/io/temporal/internal/sync/POJOWorkflowImplementationFactory.java
mfateev: It is a noop if the instance is already created. But you are right that after my latest changes it belongs in init call.

src/main/java/io/temporal/workflow/WorkflowInterceptor.java
mfateev: WorkflowInterceptor.interceptExecuteWorkflow returns WorkflowInvoker which has:

- init() called when workflow class is instantiated
- execute() called when main workflow method is executed
- processSignal() called for each signal.

PR #45, opened 2020-03-27
Added missing fields to local activity task

mfateev:
Now Activity.getTask().getWorkflowExecution() and getWorkflowDomain() return data.


src/main/java/io/temporal/internal/replay/ExecuteLocalActivityParameters.java
```diff
@@ -213,10 +216,31 @@ public void setElapsedTime(long startTime) {
     this.elapsedTime = startTime;
   }
 
+  public String getWorkflowDomain() {
```
samarabbas: Probably this needs to be Namespace now.

PR #54, opened 2020-04-01
Updates due to WorkflowExecutionStatus proto enum change

mfateev:


src/main/java/io/temporal/internal/common/WorkflowExecutionUtils.java
```diff
@@ -499,20 +499,20 @@ public static WorkflowExecutionCloseStatus waitForWorkflowInstanceCompletion(
     return getCloseStatus(closeEvent);
   }
 
-  public static WorkflowExecutionCloseStatus getCloseStatus(HistoryEvent event) {
+  public static WorkflowExecutionStatus getCloseStatus(HistoryEvent event) {
```
alexshtin, line 502: Method name probably should be `getStatus` now.

PR #58, opened 2020-04-04
Added @WorkflowInterface

mfateev:
Added @WorkflowInterface. 
The main advantage of having @WorkflowInterface is to support polymorphic invocation of workflows, signals and queries. For example this is valid code after this change:

 ```
 public interface WorkflowBase {
    @WorkflowMethod
    String execute(String arg);
  }

  @WorkflowInterface
  public interface WorkflowA extends WorkflowBase {}

  @WorkflowInterface
  public interface WorkflowB extends WorkflowBase {}

  public static class WorkflowAImpl implements WorkflowA {
    @Override
    public String execute(String arg) {
      return "WorkflowAImpl" + arg;
    }
  }

  public static class WorkflowBImpl implements WorkflowB {
    @Override
    public String execute(String arg) {
      return "WorkflowBImpl" + arg;
    }
  }

  @Test
  public void testPolymorphicStart() {
    startWorkerFor(WorkflowBImpl.class, WorkflowAImpl.class);
    WorkflowOptions options = newWorkflowOptionsBuilder(taskList).build();
    WorkflowBase[] stubs =
        new WorkflowBase[] {
          workflowClient.newWorkflowStub(WorkflowA.class, options),
          workflowClient.newWorkflowStub(WorkflowB.class, options),
        };
    String results = stubs[0].execute("0") + ", " + stubs[1].execute("1");
    assertEquals("WorkflowAImpl0, WorkflowBImpl1", results);
  }

```
Fixes https://github.com/temporalio/temporal-java-sdk/issues/55 and https://github.com/temporalio/temporal-java-sdk/issues/43

### CODE REVIEW

File: src/main/java/io/temporal/internal/sync/POJOWorkflowInterfaceMetadata.java [hunk 1]
```diff
@@ -0,0 +1,247 @@
...
+    }
+    return result;
+  }
+
+  public List<POJOWorkflowMethodMetadata> getMethodsMetadata() {
+    return new ArrayList<>(this.methods.values());
+  }
+
+  private void initWorkflowInterface(Class<?> anInterface) {
+    if (!anInterface.isInterface()) {
```
shawnhathaway, line 161: nit: init and constructor have some similar code that might be able to be turned into util

mfateev: Good catch. It is actually leftover of me moving this code to constructor :). Removing this method completely.

PR #61, opened 2020-04-07
Activity wait for cancellation

mfateev:
Added ability to specify behaviour of activity and child workflow in case of the enclosing CancellationScope cancellation. After this change it is possible to specify if a workflow should wait  for activity confirming cancellation, cancelling it asynchronously or not cancelling it at all. For child workflows there is one more option to specify if the parent has to wait for the child confirming that it received the cancellation request.


src/main/java/io/temporal/internal/replay/StartChildWorkflowExecutionParameters.java
```diff
@@ -232,42 +245,46 @@ public ParentClosePolicy getParentClosePolicy() {
...
+        && Objects.equal(workflowType, that.workflowType)
         && workflowIdReusePolicy == that.workflowIdReusePolicy
-        && Objects.equals(retryParameters, that.retryParameters)
-        && Objects.equals(cronSchedule, that.cronSchedule)
-        && Objects.equals(context, that.context)
-        && Objects.equals(parentClosePolicy, that.parentClosePolicy);
+        && Objects.equal(retryParameters, that.retryParameters)
+        && Objects.equal(cronSchedule, that.cronSchedule)
+        && Objects.equal(context, that.context)
+        && parentClosePolicy == that.parentClosePolicy;
```
shawnhathaway, line 269: Not sure if move to reference equality was intended?

src/main/java/io/temporal/internal/replay/WorkflowDecisionContext.java
```diff
@@ -89,6 +111,8 @@ public void accept(Exception cause) {
   // key is initiatedEventId
   private final Map<Long, OpenChildWorkflowRequestInfo> scheduledExternalWorkflows =
       new HashMap<>();
+  /** Maps cancellationInitiatedEventId to child initiatedEventId */
+  private final Map<Long, Long> scheduledExternalCancellations = new HashMap<>();
```
shawnhathaway, line 115: Does this need `Long` rather than `long`?

src/main/java/io/temporal/internal/replay/StartChildWorkflowExecutionParameters.java
mfateev: For enums == is a way to go: https://intellipaat.com/community/1254/comparing-java-enum-members-or-equals

BTW this method is autogenerated by my IDE.

PR #69, opened 2020-04-19
Added support for consistent query

mfateev:


src/main/java/io/temporal/internal/replay/QueryWorkflowParameters.java
```diff
@@ -105,26 +122,35 @@ public QueryWorkflowParameters withQueryRejectCondition(
...
     result.setQueryType(queryType);
     result.setWorkflowId(workflowId);
     result.setQueryRejectCondition(queryRejectCondition);
+    result.setQueryConsistencyLevel(queryConsistencyLevel);
     return result;
   }
+
+  @Override
+  public String toString() {
+    return "QueryWorkflowParameters{"
```
shawnhathaway, line 138: Could we just use a json serializer for .ToString?

PR #77, opened 2020-05-01
Added gRPC service call metrics

mfateev:


src/test/java/io/temporal/workflow/MetricsTest.java
```diff
@@ -335,4 +354,63 @@ public WorkflowInvoker interceptExecuteWorkflow(
...
+    }
+
+    // Wait for reporter
+    Thread.sleep(600);
+
+    verify(reporter, times(1))
+        .reportCounter("temporal-StartWorkflowExecution.temporal-request", new HashMap<>(), 1);
+    verify(reporter, times(1))
+        .reportCounter(
+            "temporal-StartWorkflowExecution.temporal-invalid-request", new HashMap<>(), 1);
```
shawnhathaway, line 413: Are these defined somewhere as consts or is there a common way to generate them?

src/test/java/io/temporal/workflow/MetricsTest.java
```diff
@@ -335,4 +354,63 @@ public WorkflowInvoker interceptExecuteWorkflow(
...
+          testEnvironment.getWorkflowClient().getWorkflowServiceStubs();
+
+      serviceStubs.blockingStub().describeNamespace(DescribeNamespaceRequest.newBuilder().build());
+      fail("failure expected");
+    } catch (StatusRuntimeException e) {
+      assertEquals(Status.Code.UNIMPLEMENTED, e.getStatus().getCode());
+    }
+
+    // Wait for reporter
+    Thread.sleep(600);
```
shawnhathaway, line 377: Where does 600 come from? Is this long enough for most/all developers who will run these?

mfateev: Changed to more reasonable value.

src/test/java/io/temporal/workflow/MetricsTest.java
mfateev: Replaced with constants

src/test/java/io/temporal/workflow/MetricsTest.java
mfateev: It looks like a smaller value leads to intermittent test failures. So keeping it at 600.

PR #78, opened 2020-05-02
Upgraded Gradle to 6.3

mfateev:


settings.gradle
```diff
@@ -1 +1 @@
-rootProject.name = 'temporal-java-client'
+rootProject.name = 'temporal-sdk'
```
shawnhathaway, line 1: Why not keep java? `temporal-java-sdk`

mfateev: This is the name of the library in maven central which is Java by definition. It would be like calling [grpc-api](https://search.maven.org/artifact/io.grpc/grpc-api/1.29.0/jar) library grpc-java-api.

PR #82, opened 2020-05-10
Added poll thread count options

mfateev:
Added options to configure number of polling threads:

- `WorkerFactoryOptions.workflowHostLocalPollThreadCount` for number of poller threads for cached workflows
-` WorkerOptions.workflowPollThreadCount `for number of poller threads on a workflow task list
- `WorkerOptions.activityPollThreadCount `for number of poller threads on an activity task list

Also fixed `WorkerOptions.taskListActivitiesPerSecond` which configures global task list throttling.


src/main/java/io/temporal/internal/sync/SyncActivityWorker.java
```diff
@@ -39,11 +39,14 @@ public SyncActivityWorker(
       WorkflowServiceStubs service,
       String namespace,
       String taskList,
+      double taskListActivitiesPerSecond,
```
shawnhathaway, line 42: Why pass this separately and not use ```SingleWorkerOptions``` value for ```taskListActivitiesPerSecond```? https://github.com/temporalio/temporal-java-sdk/blob/master/src/main/java/io/temporal/internal/worker/SingleWorkerOptions.java#L45

mfateev: Good catch. Removed taskListActivitiesPerSecond from SinglePollerOptions as they don't apply to the workflow worker.

PR #92, opened 2020-05-21
Consume and Update to Latest Protos

shawnhathaway:
Update to latest protos (v0.23.1). 

Mostly renames, but meaningful changes around ActivityCancellation and replacing `activityId` with `scheduledEventId`.


src/main/java/io/temporal/internal/replay/DecisionsHelper.java
```diff
@@ -190,14 +189,15 @@ boolean handleActivityTaskCanceled(HistoryEvent event) {
...
   boolean handleRequestCancelActivityTaskFailed(HistoryEvent event) {
-    RequestCancelActivityTaskFailedEventAttributes attributes =
-        event.getRequestCancelActivityTaskFailedEventAttributes();
-    String activityId = attributes.getActivityId();
-    long scheduledEventId = getActivityScheduledEventId(activityId);
-    DecisionStateMachine decision =
-        getDecision(new DecisionId(DecisionTarget.ACTIVITY, scheduledEventId));
-    decision.handleCancellationFailureEvent(event);
-    return decision.isDone();
+    //    RequestCancelActivityTaskFailedEventAttributes attributes =
```
shawnhathaway, line 192: Looks like RequestCancelActivityTaskFailed is deprecated, but event type still exists.

PR #93, opened 2020-05-22
Activity cancellation change to eventId

mfateev:
Update to reflect RequestCancelActivityTaskDecisionAttributes change from using activityId to scheduledEventId.


src/test/java/io/temporal/internal/testing/WorkflowTestingTest.java
```diff
@@ -306,6 +309,11 @@ public void workflow(
     public String activity1(String input) {
       long start = System.currentTimeMillis();
       while (true) {
+        try {
+          Thread.sleep(500);
```
shawnhathaway, line 313: Does this solve something unrelated to this change? (Failing test?)

PR #101, opened 2020-06-16
Major error handling refactoring

mfateev:
Refactoring to the new [Failure](https://github.com/temporalio/temporal-proto/blob/master/temporal/failure/v1/message.proto) proto API.

This is to support  propagation of chained errors across multiple SDKs. The biggest incompatible change from the application developer point of view that type of failure is returned as a cause of the ActivityFailure or WorkflowFailure instead of using child exceptions.


src/test/java/io/temporal/worker/StickyWorkerTest.java
```diff
@@ -226,7 +226,7 @@ public void workflowCacheEvictionDueToThreads() {
     int count = 100;
     ActivitiesWorkflow[] workflows = new ActivitiesWorkflow[count];
     WorkflowParams w = new WorkflowParams();
-    w.TemporalSleep = Duration.ofSeconds(1);
+    w.TemporalSleepMillis = 1000;
```
shawnhathaway, line 229: For defining time intervals - Do we prefer units on `ints` rather than strongly typed `Duration`?

src/main/java/io/temporal/internal/common/LocalActivityMarkerData.java
```diff
@@ -95,120 +98,115 @@ public Builder setBackoff(Duration backoff) {
...
-    return Strings.isNullOrEmpty(headers.errReason) ? Optional.empty() : result;
+  public Optional<Failure> getFailure() {
+    return failure;
   }
 
   public Optional<Payloads> getResult() {
     return result;
   }
 
   public long getReplayTimeMillis() {
```
shawnhathaway, line 169: Mixed time types? ReplayTimeMillis vs Backoff

src/main/java/io/temporal/internal/common/StartWorkflowExecutionParameters.java
```diff
@@ -348,14 +348,14 @@ public static StartWorkflowExecutionParameters fromWorkflowOptions(WorkflowOptio
       rp.setInitialIntervalInSeconds(roundUpToSeconds(retryOptions.getInitialInterval()));
       rp.setMaximumIntervalInSeconds(roundUpToSeconds(retryOptions.getMaximumInterval()));
       rp.setMaximumAttempts(retryOptions.getMaximumAttempts());
-      List<String> reasons = new ArrayList<>();
+      List<String> types = new ArrayList<>();
       // Use exception type name as the reason
-      List<Class<? extends Throwable>> doNotRetry = retryOptions.getDoNotRetry();
+      String[] doNotRetry = retryOptions.getDoNotRetry();
```
shawnhathaway, line 353: What is reason for using string rather than types? Does previous usage fall under Illegal reflective access issues?

src/main/java/io/temporal/failure/ApplicationFailure.java
```diff
@@ -0,0 +1,132 @@
...
+
+  public Value getDetails() {
+    return details;
+  }
+
+  public void setNonRetryable(boolean nonRetryable) {
+    this.nonRetryable = nonRetryable;
+  }
+
+  public boolean isNonRetryable() {
```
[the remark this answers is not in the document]
mfateev: The pattern I always use is to have default value as the default value of the type. For boolean the default is false.

src/test/java/io/temporal/worker/StickyWorkerTest.java
mfateev: We prefer Duration unless it has to be passed across process boundary and serialized/deserialized. We can add serialization of Duration to the Jackson json serializer, but then it might not play well with Go and other SDKs.

src/main/java/io/temporal/internal/common/LocalActivityMarkerData.java
mfateev: ReplayTimeMillis is a specific time of replay (timestamp). Backoff is a duration which is time interval.

src/main/java/io/temporal/internal/common/StartWorkflowExecutionParameters.java
mfateev: This is to be compatible with other SDKs. An activity can be implemented in Go and return an error type which is not Java exception name.

PR #105, opened 2020-06-17
Activity APIs cleanup.

mfateev:
Exposes ActivityExecutionContext to simplify passing activity info around to other threads. Removed redundant methods from Activity class. Also some renaming for consistency. Changed some getters to return Optional instead of null.

Also fixes https://github.com/temporalio/temporal-java-sdk/issues/98

### DIFF
--- a/src/main/java/io/temporal/client/ActivityCompletionClient.java
+++ b/src/main/java/io/temporal/client/ActivityCompletionClient.java
@@ -19,11 +19,12 @@
 
 package io.temporal.client;
 
+import io.temporal.activity.ActivityExecutionContext;
 import java.util.Optional;
 
 /**
  * Used to complete asynchronously activities that called {@link
- * io.temporal.internal.sync.ActivityExecutionContext#doNotCompleteOnReturn()}.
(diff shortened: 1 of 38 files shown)


src/main/java/io/temporal/client/ActivityCompletionClient.java
```diff
@@ -19,41 +19,37 @@
...
- * <p>Use {@link WorkflowClient#newActivityCompletionClient()} to create an instance. TODO: Throw
- * relevant exceptions like EntityNotExists, etc.
+ * <p>Use {@link WorkflowClient#newActivityCompletionClient()} to create an instance.
  */
 public interface ActivityCompletionClient {
 
   <R> void complete(byte[] taskToken, R result) throws ActivityCompletionException;
 
-  <R> void complete(WorkflowExecution execution, String activityId, R result)
+  <R> void complete(String workflowId, Optional<String> runId, String activityId, R result)
```
samarabbas, line 35: It feels weird to have optional parameter in the middle of signature.  It is considered very bad practice to have required parameters after optional ones.

PR #120, opened 2020-06-22
Added thread interceptor

mfateev:
Added WorkflowInterceptor support for workflow threads. It is to allow support for context propagation without modifying the SDK code directly.


src/main/java/io/temporal/internal/sync/DeterministicRunner.java
```diff
@@ -121,4 +122,14 @@ static DeterministicRunner newRunner(
...
    */
   void executeInWorkflowThread(String name, Runnable r);
+
+  /**
+   * Creates a new instance of a workflow thread. To be called only from another workflow thread.
+   */
+  WorkflowThread newThread(Runnable runnable, boolean detached, String name);
+
+  /** Only to be called by SyncDecisionContext. */
+  WorkflowThread newThreadNoInterceptor(Runnable runnable, boolean detached, String name);
```
samarabbas, line 132: I'm trying to understand why this additional API is needed?  Why can't there be a no-op interceptor which does nothing.  It would be good to provide some context as part of API documentation.

mfateev: Refactored to remove the need for newThreadNoInterceptor.

PR #143, opened 2020-07-06
Activity interceptor

mfateev:


src/main/java/io/temporal/internal/sync/POJOActivityTaskHandler.java
```diff
@@ -234,29 +277,30 @@ public Result handle(
...
+      for (ActivityInterceptor interceptor : interceptors) {
+        inboundCallsInterceptor = interceptor.interceptActivity(inboundCallsInterceptor);
+      }
+      inboundCallsInterceptor.init(context);
       try {
         Object[] args =
             dataConverter.arrayFromPayloads(
                 input, method.getParameterTypes(), method.getGenericParameterTypes());
-        Object result = method.invoke(activity, args);
+        Object result = inboundCallsInterceptor.execute(args);
```
shawnhathaway, line 292: Minor nit - Looks like stack trace in this test JSON may be out of sync after this change:

https://github.com/temporalio/temporal-java-sdk/blob/master/src/test/resources/testAsyncActivityRetryHistory.json#L122

PR #147, opened 2020-07-10
Added tags to request metrics. Eliminated duplicated ones

mfateev:


src/main/java/io/temporal/internal/metrics/MetricsType.java
```diff
@@ -32,18 +31,8 @@
...
-      TEMPORAL_METRICS_PREFIX + "workflow-get-history-failed";
-  public static final String WORKFLOW_GET_HISTORY_SUCCEED_COUNTER =
-      TEMPORAL_METRICS_PREFIX + "workflow-get-history-succeed";
-  public static final String WORKFLOW_GET_HISTORY_LATENCY =
-      TEMPORAL_METRICS_PREFIX + "workflow-get-history-latency";
-  public static final String WORKFLOW_SIGNAL_WITH_START_COUNTER =
-      TEMPORAL_METRICS_PREFIX + "workflow-signal-with-start";
-  public static final String DECISION_TIMEOUT_COUNTER =
-      TEMPORAL_METRICS_PREFIX + "decision-timeout";
+  public static final String DECISION_TASK_REPLAY_LATENCY =
```
samarabbas, line 34: renamed as Workflow_TASK

src/main/java/io/temporal/internal/metrics/MetricsType.java
```diff
@@ -122,32 +77,26 @@
...
       TEMPORAL_METRICS_PREFIX + "corrupted-signals";
 
   public static final String WORKER_START_COUNTER = TEMPORAL_METRICS_PREFIX + "worker-start";
   public static final String POLLER_START_COUNTER = TEMPORAL_METRICS_PREFIX + "poller-start";
 
   public static final String TEMPORAL_REQUEST = TEMPORAL_METRICS_PREFIX + "request";
-  public static final String TEMPORAL_ERROR = TEMPORAL_METRICS_PREFIX + "error";
-  public static final String TEMPORAL_LATENCY = TEMPORAL_METRICS_PREFIX + "latency";
-  public static final String TEMPORAL_INVALID_REQUEST = TEMPORAL_METRICS_PREFIX + "invalid-request";
+  public static final String TEMPORAL_REQUEST_FAILURE = TEMPORAL_REQUEST + "-failure";
```
samarabbas, line 87: Looks like we are deviating from convention we use from go-sdk.  Is there a reason to move away from existing convention.  We don't want to build 2 separate dashboards for each different type of sdk.

src/main/java/io/temporal/internal/metrics/MetricsType.java
```diff
@@ -21,7 +21,6 @@
 
 public class MetricsType {
   public static final String TEMPORAL_METRICS_PREFIX = "temporal-";
```
samarabbas, line 23: '-' does not work well with certain backends like prometheus for instance.  It would be nice to replace '-' with '_' everywhere if possible.  I think we have to plugin name sanitizer to fix that at runtime otherwise.

PR #148, opened 2020-07-13
Updated to the latest protos and added Proto converter

mfateev:


src/main/java/io/temporal/common/converter/EncodingKeys.java
```diff
@@ -31,7 +31,16 @@
   static final String METADATA_ENCODING_RAW_NAME = "raw";
   static final ByteString METADATA_ENCODING_RAW =
       ByteString.copyFrom(METADATA_ENCODING_RAW_NAME, StandardCharsets.UTF_8);
+
   static final String METADATA_ENCODING_JSON_NAME = "json";
   static final ByteString METADATA_ENCODING_JSON =
       ByteString.copyFrom(METADATA_ENCODING_JSON_NAME, StandardCharsets.UTF_8);
+
+  static final String METADATA_ENCODING_PROTOBUF_JSON_NAME = "protobuf-json";
```
samarabbas, line 39: Is this the default now?

We want to make sure interop works out of the box between go-sdk and java-sdk.

mfateev: I will make sure that the Go SDK is updated appropriatly

PR #150, opened 2020-07-14
Decision to WorkflowTask or Command renaming

mfateev:


src/main/java/io/temporal/internal/metrics/MetricsType.java
```diff
@@ -66,15 +60,15 @@
   public static final String ACTIVITY_E2E_LATENCY =
       TEMPORAL_METRICS_PREFIX + "activity_endtoend_latency";
   public static final String ACTIVITY_TASK_ERROR_COUNTER =
-      TEMPORAL_METRICS_PREFIX + "activity_task_error";
+      TEMPORAL_METRICS_PREFIX + "activity_task_error_thrown";
```
samarabbas, line 63: can we just call it activity_task_error? _thrown suffix won't be meaningful for go-sdk.

src/main/java/io/temporal/internal/metrics/MetricsType.java
```diff
@@ -66,15 +60,15 @@
...
   public static final String LOCAL_ACTIVITY_TOTAL_COUNTER =
       TEMPORAL_METRICS_PREFIX + "local_activity_total";
   public static final String LOCAL_ACTIVITY_CANCELED_COUNTER =
       TEMPORAL_METRICS_PREFIX + "local_activity_canceled";
   public static final String LOCAL_ACTIVITY_FAILED_COUNTER =
       TEMPORAL_METRICS_PREFIX + "local_activity_failed";
   public static final String LOCAL_ACTIVITY_ERROR_COUNTER =
-      TEMPORAL_METRICS_PREFIX + "local_activity_panic";
+      TEMPORAL_METRICS_PREFIX + "local_activity_error_thrown";
```
samarabbas, line 71: can we just call it local_activity_task_error? _thrown suffix won't be meaningful for go-sdk.

src/main/java/io/temporal/internal/sync/WorkflowThreadImpl.java
```diff
@@ -55,7 +55,7 @@
   class RunnableWrapper implements Runnable {
 
     private final WorkflowThreadContext threadContext;
-    private final DecisionContext decisionContext;
+    private final ReplayWorkflowContext replayWorkflowContext;
```
samarabbas, line 58: why ReplyWorkflowContext?  Looks like a comment would be helpful here.

mfateev: Added TODO to remove from there.

PR #154, opened 2020-07-24
Updates Proto Submodule reference

mastermanu:


src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -653,20 +652,6 @@ private void processCancelTimer(
       RequestContext ctx, CancelTimerCommandAttributes d, long workflowTaskCompletedId) {
     String timerId = d.getTimerId();
     StateMachine<TimerData> timer = timers.get(timerId);
-    if (timer == null) {
```
mastermanu, line 656: @mfateev - is this safe to remove from the test service? I can also keep the if check and return if timer is null as a protective measure

mfateev: The real service changed from reporting CANCEL_TIMER_FAILED to failing the whole decision. So the test service should replace the content of this if with a workflow task failure.

PR #161, opened 2020-07-29
Ensures that we rely on server-specified activity retry policy if user doesn't set retry options

mastermanu:


src/main/java/io/temporal/activity/ActivityOptions.java
```diff
@@ -173,13 +177,19 @@ public ActivityOptions build() {
     }
 
     public ActivityOptions validateAndBuildWithDefaults() {
+      if (retryOptions != null) {
+        // retryOptions being null means we will explicitly fallback to the default
+        // activity retry behavior as specified on the server
+        retryOptions = RetryOptions.newBuilder(retryOptions).validateBuildWithDefaults();
```
mfateev, line 183: I think we should not add any defaults to the retry options and rely on the service for these.

src/main/java/io/temporal/internal/testservice/StateMachines.java
```diff
@@ -980,15 +975,19 @@ private static void scheduleActivityTask(
...
-    if (d.hasRetryPolicy()) {
-      RetryPolicy retryPolicy = d.getRetryPolicy();
-      long expirationInterval = TimeUnit.SECONDS.toMillis(d.getScheduleToCloseTimeoutSeconds());
-      long expirationTime = data.store.currentTimeMillis() + expirationInterval;
-      retryState = new TestServiceRetryState(retryPolicy, expirationTime);
+    RetryPolicy retryPolicy;
+    if (!d.hasRetryPolicy()) {
+      // If no activity retry policy is supplied, assume a default to mimic a standard Temporal
+      // server
+      retryPolicy = toRetryPolicy(RetryOptions.newBuilder().validateBuildWithDefaults()).build();
```
mfateev, line 982: We should compy defaults from the service and do not rely on validateBuildWithDefaults which was intended for the SDK only validation.

src/main/java/io/temporal/internal/sync/SyncWorkflowContext.java
```diff
@@ -351,12 +351,15 @@ private ExecuteActivityParameters constructExecuteActivityParameters(
...
+    String[] doNotRetry = retryOptions.getDoNotRetry();
+
     return RetryPolicy.newBuilder()
         .setInitialIntervalInSeconds(roundUpToSeconds(retryOptions.getInitialInterval()))
         .setMaximumIntervalInSeconds(roundUpToSeconds(retryOptions.getMaximumInterval()))
         .setBackoffCoefficient(retryOptions.getBackoffCoefficient())
         .setMaximumAttempts(retryOptions.getMaximumAttempts())
-        .addAllNonRetryableErrorTypes(Arrays.asList(retryOptions.getDoNotRetry()));
+        .addAllNonRetryableErrorTypes(
+            Arrays.asList(doNotRetry != null ? doNotRetry : new String[0]));
```
mastermanu, line 362: [Nit] i can technically make this a private static variable that represents an empty array

src/main/java/io/temporal/internal/testservice/StateMachines.java
```diff
@@ -1858,4 +1845,23 @@ private static void failExternalCancellation(
             .build();
     ctx.addEvent(event);
   }
+
+  // Mimics the default activity retry policy of a standard Temporal server.
+  static RetryPolicy ensureDefaultFieldsForActivityRetryPolicy(RetryPolicy originalPolicy) {
+    return RetryPolicy.newBuilder()
+        .setInitialIntervalInSeconds(
+            originalPolicy.getInitialIntervalInSeconds() == 0
+                ? 1
```
mfateev: We already have constants for some defaults in this file. I would convert these to constants as well.

src/main/java/io/temporal/internal/sync/SyncWorkflowContext.java
mastermanu: good point, i will omit that

src/main/java/io/temporal/internal/testservice/StateMachines.java
```diff
@@ -110,14 +110,7 @@
...
 import io.temporal.internal.testservice.TestWorkflowStore.WorkflowTask;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.OptionalLong;
-import java.util.UUID;
+import java.util.*;
```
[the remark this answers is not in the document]
mastermanu: i did run gradlew, but i guess it doesn't undo the collapse if intellij does it. i changed my IDE settings and reverted

src/main/java/io/temporal/internal/testservice/StateMachines.java
mastermanu: good point. added statics at the top of the file

PR #164, opened 2020-07-31
Renamed EncodedValue to EncodedValues to support multiple details

mfateev:
* Updated DataConverter API to support extracting individual values from the serialized list.
* Renamed EncodedValue to EncodedValues.
* Added ApplicationFailure factories for clarity.

### DIFF
--- a/src/main/java/io/temporal/common/converter/DataConverter.java
+++ b/src/main/java/io/temporal/common/converter/DataConverter.java
@@ -75,7 +75,8 @@ public interface DataConverter {
       Type[] genericParameterTypes)
       throws DataConverterException {
     if (parameterTypes != null
-        && (genericParameterTypes == null || parameterTypes.length != genericParameterTypes.length)) {
+        && (genericParameterTypes == null
+            || parameterTypes.length != genericParameterTypes.length)) {
       throw new IllegalArgumentException(
           "parameterTypes don't match length of valueTypes: "
               + Arrays.toString(parameterTypes)
--- a/src/main/java/io/temporal/common/converter/DataConverter.java
+++ b/src/main/java/io/temporal/common/converter/DataConverter.java
@@ -75,7 +75,7 @@ public interface DataConverter {
       Type[] genericParameterTypes)
       throws DataConverterException {
     if (parameterTypes != null
-        && (parameterTypes == null || parameterTypes.length != genericParameterTypes.length)) {
+        && (genericParameterTypes == null || parameterTypes.length != genericParameterTypes.length)) {
       throw new IllegalArgumentException(
           "parameterTypes don't match length of valueTypes: "
               + Arrays.toString(parameterTypes)
--- a/src/main/java/io/temporal/common/converter/DataConverter.java
+++ b/src/main/java/io/temporal/common/converter/DataConverter.java
@@ -19,9 +19,11 @@
 
 package io.temporal.common.converter;
 
+import com.google.common.base.Defaults;
 import io.temporal.api.common.v1.Payload;
 import io.temporal.api.common.v1.Payloads;
 import java.lang.reflect.Type;
+import java.util.Arrays;
 import java.util.Optional;
 
 /**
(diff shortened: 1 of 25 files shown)


src/main/java/io/temporal/failure/ApplicationFailure.java
```diff
@@ -51,49 +51,47 @@
  */
 public final class ApplicationFailure extends TemporalFailure {
   private final String type;
-  private final Value details;
+  private final Values details;
   private boolean nonRetryable;
 
   /**
+   * New ApplicationFailure with {@link #isNonRetryable()} flag set to false. Note that this
```
mastermanu, line 58: didn't quite follow the 2nd sentence of this comment

src/main/java/io/temporal/failure/ApplicationFailure.java
```diff
@@ -51,49 +51,47 @@
...
    *     io.temporal.common.RetryOptions#addDoNotRetry(String...)}.
    * @param details optional details about the failure. They are serialized using the same approach
-   *     as arguments and results and can be accessed through {@link #getDetails()}
-   * @param cause failure cause. Each element of the cause chain is converted to ApplicationFailure
-   *     if it doesn't extend {@link TemporalFailure}.
+   *     as arguments and results.
    */
-  public ApplicationFailure(String message, String type, Object details, Exception cause) {
-    this(message, type, new EncodedValue(details), false, cause);
+  public static ApplicationFailure newFailure(String message, String type, Object... details) {
```
mastermanu, line 68: why not just call it newRetryableFailure?

src/main/java/io/temporal/failure/ApplicationFailure.java
```diff
@@ -51,49 +51,47 @@
...
+  public static ApplicationFailure newNonRetryableFailure(
+      String message, String type, Object... details) {
+    return new ApplicationFailure(message, type, true, new EncodedValues(details), null);
   }
 
-  /** * @param message optional error message */
-  public ApplicationFailure(String message) {
-    this(message, null);
+  static ApplicationFailure newFromValues(
+      String message, String type, boolean nonRetryable, Values details, Throwable cause) {
```
mastermanu, line 89: [Nit], having boolean names that are negative (nonRetryable vs retryable) can be kind of confusing

src/main/java/io/temporal/internal/sync/ActivityExecutionContextImpl.java
```diff
@@ -130,7 +130,7 @@
         return result;
       }
       Optional<Payloads> details = info.getHeartbeatDetails();
-      return Optional.ofNullable(dataConverter.fromPayloads(details, detailsClass, detailsType));
+      return Optional.ofNullable(dataConverter.fromPayloads(0, details, detailsClass, detailsType));
```
mastermanu, line 133: we didn't want this function to have an overload without the index that internally assumes the first element?

src/test/java/io/temporal/workflow/WorkflowTest.java
```diff
@@ -3995,7 +3996,10 @@ public void testWorkflowFailureNonRetryableFlag() {
     } catch (WorkflowException e) {
       assertTrue(e.getCause() instanceof ApplicationFailure);
       assertEquals("foo", ((ApplicationFailure) e.getCause()).getType());
-      assertEquals("details", ((ApplicationFailure) e.getCause()).getDetails().get(String.class));
+      assertEquals(
+          "details1", ((ApplicationFailure) e.getCause()).getDetails().get(0, String.class));
```
mastermanu, line 4000: and we still have a test that validates get without the index specified works fine?

src/main/java/io/temporal/common/converter/DataConverter.java
```diff
@@ -51,30 +53,56 @@ static DataConverter getDefaultInstance() {
...
-      Optional<Payloads> content, Class<?>[] parameterTypes, Type[] genericParameterTypes)
-      throws DataConverterException;
+  static Object[] arrayFromPayloads(
+      DataConverter converter,
+      Optional<Payloads> content,
+      Class<?>[] parameterTypes,
+      Type[] genericParameterTypes)
+      throws DataConverterException {
+    if (parameterTypes != null
+        && (parameterTypes == null || parameterTypes.length != genericParameterTypes.length)) {
```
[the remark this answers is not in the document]
mfateev: good catch!

src/main/java/io/temporal/common/converter/DataConverter.java
```diff
@@ -51,30 +53,56 @@ static DataConverter getDefaultInstance() {
...
+        result[i] = Defaults.defaultValue((Class<?>) genericParameterTypes[i]);
+      }
+      return result;
+    }
+    Payloads payloads = content.get();
+    int count = payloads.getPayloadsCount();
+    for (int i = 0; i < parameterTypes.length; i++) {
+      Class<?> pt = parameterTypes[i];
+      Type gt = genericParameterTypes[i];
+      if (i >= count) {
```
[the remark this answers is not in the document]
mfateev: Good question. The current answer is that it is needed to be able to add and remove arguments of functions.

src/main/java/io/temporal/common/converter/EncodedValues.java
```diff
@@ -58,22 +58,35 @@ public void setDataConverter(DataConverter converter) {
...
+    return get(index, parameterType, parameterType);
+  }
+
+  @Override
+  public <T> T get(int index, Class<T> parameterType, Type genericParameterType)
       throws DataConverterException {
-    return converter.fromPayloads(payloads, parameterType, genericParameterType);
+    if (values != null) {
+      @SuppressWarnings("unchecked")
+      T result = (T) values[index];
```
mastermanu, line 83: if the user exceeds the bounds of the values/payload, no issues right?

src/main/java/io/temporal/common/converter/EncodedValues.java
```diff
@@ -58,22 +58,35 @@ public void setDataConverter(DataConverter converter) {
...
-      @SuppressWarnings("unchecked")
-      T result = (T) value.orElse(null);
-      return result;
+  public int getSize() {
+    if (values != null) {
+      return values.length;
     } else {
-      if (converter == null) {
-        throw new IllegalStateException("converter not set");
+      if (payloads.isPresent()) {
```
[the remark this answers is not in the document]
mfateev: In a lot of cases, Payloads are not set at all if there are no any values. But we probably need to audit for this.

src/main/java/io/temporal/common/converter/EncodedValues.java
```diff
@@ -58,22 +58,35 @@ public void setDataConverter(DataConverter converter) {
...
-    return converter.fromPayloads(payloads, parameterType, genericParameterType);
+    if (values != null) {
+      @SuppressWarnings("unchecked")
+      T result = (T) values[index];
+      return result;
+    } else {
+      if (converter == null) {
+        throw new IllegalStateException("converter not set");
+      }
+      return converter.fromPayloads(index, payloads, parameterType, genericParameterType);
```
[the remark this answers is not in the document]
mfateev: The idea of Values is that it can be created with objects or Payloads. I'm not sure if caching is really going to help here as every value accessed only once I believe.

src/main/java/io/temporal/failure/ApplicationFailure.java
mfateev: Good catch.

src/main/java/io/temporal/failure/ApplicationFailure.java
mfateev: I called it initially. But then I found that it is very confusing that it can be not retried if the policy says so.

src/main/java/io/temporal/failure/ApplicationFailure.java
mfateev: I almost always follow the pattern that the default value should be the default value for the type. So the default of the boolean type is false. So the field is called nonRetryable as its default is false. But I agree that it looks weird in the parameter name.

src/main/java/io/temporal/internal/sync/ActivityExecutionContextImpl.java
mfateev: This class is used only internally by the SDK, so I think having an explicit index makes sense to indicate that it can hold multiple values.

src/test/java/io/temporal/workflow/WorkflowTest.java
mfateev: I checked and found at least two unit tests that call it.

PR #170, opened 2020-08-04
Fixes overzealous validation logic and adds specific check for public interfaces on registration

mastermanu:
- Fails registration if there are Activity/Workflow interfaces which are non-public
- Does not perform validation on interfaces that would not be leveraged / used by the Temporal SDK

### DIFF
--- a/src/main/java/io/temporal/internal/sync/POJOWorkflowImplMetadata.java
+++ b/src/main/java/io/temporal/internal/sync/POJOWorkflowImplMetadata.java
@@ -19,7 +19,6 @@
 
 package io.temporal.internal.sync;
 
-import java.lang.reflect.Modifier;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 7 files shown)


src/main/java/io/temporal/internal/sync/POJOWorkflowImplMetadata.java
```diff
@@ -87,6 +88,11 @@ private POJOWorkflowImplMetadata(Class<?> implClass, boolean listener) {
     Class<?>[] interfaces = implClass.getInterfaces();
     for (int i = 0; i < interfaces.length; i++) {
       Class<?> anInterface = interfaces[i];
+
+      if (!Modifier.isPublic(anInterface.getModifiers())) {
+        continue;
+      }
```
mfateev, line 94: I know it would require a larger change, but ideally, we would throw an exception on nonpublic interface that is annotated with @ActivityInterface or @WorkflowInterface.

mastermanu: done!

PR #172, opened 2020-08-07
StateMachine rewrite

mfateev:
Complete rewrite of the workflow task handling state machine. It was caused by the need to fix immediate cancellation issues that weren't possible to solve using previous design.


src/main/java/io/temporal/activity/ActivityOptions.java
```diff
@@ -185,7 +185,7 @@ public ActivityOptions validateAndBuildWithDefaults() {
           taskQueue,
           retryOptions,
           contextPropagators,
-          cancellationType);
+          cancellationType == null ? ActivityCancellationType.TRY_CANCEL : cancellationType);
```
mastermanu, line 188: [Question] - Should this type of "default" be set on the Server side instead?

src/main/java/io/temporal/internal/grpc/GrpcTracingInterceptor.java
```diff
@@ -49,7 +49,7 @@
         next.newCall(method, callOptions)) {
       @Override
       public void sendMessage(ReqT message) {
-        log.trace("Invoking " + method.getFullMethodName() + "with input: " + message);
+        log.trace("Invoking \"" + method.getFullMethodName() + "\" with input: " + message);
```
mastermanu, line 52: [Nit] dumb question (based on similar change above), when do we use single quote (') vs double quote (") for these types of messages?

src/main/java/io/temporal/internal/metrics/MetricsType.java
```diff
@@ -42,13 +42,19 @@
       TEMPORAL_METRICS_PREFIX + "workflow_task_schedule_to_start_latency";
   public static final String WORKFLOW_TASK_EXECUTION_LATENCY =
       TEMPORAL_METRICS_PREFIX + "workflow_task_execution_latency";
+  // Total latency of a workflow task which can include multiple synchronous decision tasks
```
mastermanu, line 45: even with the comment, it is still not obvious to me what the difference is between execution latency and execution total latency?

mastermanu: [Nit] also styling question on when to use "//" comment vs "/** */" comment delimiters, as is done below on lines 48, 51

src/main/java/io/temporal/internal/replay/ExecuteLocalActivityParameters.java
```diff
@@ -24,19 +24,17 @@
...
 
   public PollActivityTaskQueueResponse.Builder getActivityTask() {
     return activityTask;
   }
 
-  public long getElapsedTime() {
-    return elapsedTime;
+  @Override
+  public String toString() {
+    return "ExecuteLocalActivityParameters{" + "activityTask=" + activityTask + '}';
```
mastermanu, line 38: [Nit] no single or double quote delimiters around activityTask here?

mastermanu: Also you can remove the first plus sign and concatenate the first two strings (e.g. "ExecuteLocalActivityParameters{activityTask=" + activityTask + "}"

src/main/java/io/temporal/internal/replay/ReplayWorkflowContext.java
```diff
@@ -104,39 +104,39 @@
...
-  Consumer<Exception> startChildWorkflow(
+  Functions.Proc1<Exception> startChildWorkflow(
       StartChildWorkflowExecutionParameters parameters,
-      Consumer<WorkflowExecution> executionCallback,
-      BiConsumer<Optional<Payloads>, Exception> callback);
+      Functions.Proc1<WorkflowExecution> executionCallback,
+      Functions.Proc2<Optional<Payloads>, Exception> callback);
 
-  Consumer<Exception> signalWorkflowExecution(
+  Functions.Proc1<Exception> signalExternalWorkflowExecution(
```
mastermanu, line 131: Curious why some of these methods have full comments (such as startChildWorkflow), but then some like this one (signalExternalWorkflowExecution) don't

src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
```diff
@@ -104,10 +108,27 @@ public ReplayWorkflowTaskHandler(
...
+                                .setFailure(((WorkflowExecutionException) e).getFailure()))
+                        .build())
+                .build();
+        return new WorkflowTaskHandler.Result(workflowType, response, null, null, null, false);
+      }
       // Only fail workflow task on the first attempt, subsequent failures of the same workflow task
       // should timeout. This is to avoid spin on the failed workflow task as the service doesn't
       // yet increase the retry interval.
-      if (workflowTask.getAttempt() > 0) {
+      if (workflowTask.getAttempt() > 1) {
```
mastermanu, line 131: hopefully we added a unit test to cover this as this should have been caught when I was syncing to Alex's latest proto changes

src/main/java/io/temporal/internal/replay/WorkflowExecutor.java
```diff
@@ -24,45 +24,53 @@
...
 
 public interface WorkflowExecutor {
 
-  WorkflowTaskResult handleWorkflowTask(PollWorkflowTaskQueueResponseOrBuilder workflowTask)
-      throws Throwable;
+  /**
+   * Handles a single workflow task.
+   *
+   * @param workflowTask task to handle
+   * @return true if new task should be created synchronously as local activities are still running.
```
mastermanu, line 38: [Nit] what does it mean by return "true if new task should be created synchronously"? Does this mean not null?

src/main/java/io/temporal/internal/replay/WorkflowExecutor.java
```diff
@@ -24,45 +24,53 @@
...
     private final boolean finalCommand;
     private final Map<String, WorkflowQueryResult> queryResults;
+    private final boolean forceWorkflowTask;
 
     public WorkflowTaskResult(
         List<Command> commands,
         Map<String, WorkflowQueryResult> queryResults,
-        boolean forceCreateNewWorkflowTask,
-        boolean finalCommand) {
+        boolean finalCommand,
```
mastermanu, line 59: we swapped the order to two arguments with the same type...hopefully all the callers have been updated

src/main/java/io/temporal/internal/statemachines/ActionOrEventType.java
```diff
@@ -0,0 +1,77 @@
...
+ *  permissions and limitations under the License.
+ */
+
+package io.temporal.internal.statemachines;
+
+import com.google.common.base.Objects;
+import io.temporal.api.enums.v1.CommandType;
+import io.temporal.api.enums.v1.EventType;
+
+class ActionOrEventType<Action> {
```
mastermanu, line 26: Name of class implies two possible configurations, but it looks like its ActionOrEventTypeOrCommandType (three)

src/main/java/io/temporal/activity/ActivityOptions.java
mfateev: This is purely client side option. It is not passed to the service anywhere.

src/main/java/io/temporal/internal/grpc/GrpcTracingInterceptor.java
mfateev: There is no guideline here. I would support coming up with one and making sure that all the log messages obey it.

src/main/java/io/temporal/internal/statemachines/ActivityStateMachine.java
```diff
@@ -0,0 +1,338 @@
...
+            State.CANCELED)
+        .add(
+            State.CREATED,
+            Action.SCHEDULE,
+            State.SCHEDULE_COMMAND_CREATED,
+            ActivityStateMachine::createScheduleActivityTaskCommand)
+        .add(
+            State.SCHEDULE_COMMAND_CREATED,
+            CommandType.COMMAND_TYPE_SCHEDULE_ACTIVITY_TASK,
+            State.SCHEDULE_COMMAND_CREATED)
```
mastermanu, line 85: is this a typo / copy paste error (SCHEDULE_COMMAND_CREATED twice)? Maybe I'm interpreting this incorrectly? I assumed first line is current state, subsequent line is potential transition, second line is final state and last line is an additional callback to run

src/main/java/io/temporal/internal/statemachines/ActivityStateMachine.java
```diff
@@ -0,0 +1,338 @@
...
+            EventType.EVENT_TYPE_ACTIVITY_TASK_CANCEL_REQUESTED,
+            State.STARTED_ACTIVITY_CANCEL_EVENT_RECORDED,
+            ActivityStateMachine::notifyCanceledIfTryCancel)
+        .add(
+            State.STARTED,
+            Action.CANCEL,
+            State.STARTED_ACTIVITY_CANCEL_COMMAND_CREATED,
+            ActivityStateMachine::createRequestCancelActivityTaskCommand)
+        .add(
+            State.STARTED_ACTIVITY_CANCEL_COMMAND_CREATED,
```
mastermanu, line 157: for this state and the cancel_command_Created state, we don't have transitions for a timed out event?

src/main/java/io/temporal/internal/statemachines/ActivityStateMachine.java
```diff
@@ -0,0 +1,338 @@
...
+    STARTED,
+    COMPLETED,
+    FAILED,
+    TIMED_OUT,
+    CANCELED,
+    SCHEDULED_ACTIVITY_CANCEL_COMMAND_CREATED,
+    SCHEDULED_ACTIVITY_CANCEL_EVENT_RECORDED,
+    STARTED_ACTIVITY_CANCEL_COMMAND_CREATED,
+    STARTED_ACTIVITY_CANCEL_EVENT_RECORDED,
+    COMPLETED_CANCEL_REQUESTED,
```
mastermanu, line 64: where are these 3 states used?

src/main/java/io/temporal/internal/statemachines/ActivityStateMachine.java
```diff
@@ -0,0 +1,338 @@
...
+            .setActivityType(scheduleAttr.getActivityType())
+            .setRetryState(timedOut.getRetryState())
+            .setScheduledEventId(timedOut.getScheduledEventId())
+            .setStartedEventId(timedOut.getStartedEventId())
+            .build();
+    Failure failure =
+        Failure.newBuilder()
+            .setActivityFailureInfo(failureInfo)
+            .setCause(timedOut.getFailure())
+            .setMessage("Activity task timedOut")
```
mastermanu, line 296: should we have string constants for these so that its easy for a technical writer to review or so we can do translations (in the future)

src/main/java/io/temporal/internal/metrics/MetricsType.java
mfateev: Fixed comments.

To understand this comment you need to know about workflow task heartbeat feature. Basically a workflow task when reporting completion can set ForceCreateNewWorkflowTask property to true on RespondWorkflowTaskCompletedRequest. In this case, the new workflow task is created immediately and returned as a response to the RespondWorkflowTaskCompletedRequest call. Such call acts as a heartbeat that extends the duration of a workflow task. Thus a single total execution latency of a workflow task can be larger than latencies of multiple such synchronously created tasks.

src/main/java/io/temporal/internal/statemachines/CancelExternalStateMachine.java
```diff
@@ -0,0 +1,127 @@
...
+    completionCallback.apply(null, null);
+  }
+
+  private void notifyFailed() {
+    WorkflowExecution execution =
+        WorkflowExecution.newBuilder()
+            .setWorkflowId(requestCancelAttributes.getWorkflowId())
+            .setRunId(requestCancelAttributes.getRunId())
+            .build();
+    completionCallback.apply(null, new CancelExternalWorkflowException(execution, "", null));
```
mastermanu, line 121: we don't have any information as to why cancellation failed?

src/main/java/io/temporal/internal/statemachines/ChildWorkflowStateMachine.java
```diff
@@ -0,0 +1,289 @@
...
+            State.START_COMMAND_CREATED,
+            EventType.EVENT_TYPE_START_CHILD_WORKFLOW_EXECUTION_INITIATED,
+            State.START_EVENT_RECORDED)
+        .add(
+            State.START_COMMAND_CREATED,
+            Action.CANCEL,
+            State.CANCELED,
+            ChildWorkflowStateMachine::cancelStartChildCommand)
+        .add(
+            State.START_EVENT_RECORDED,
```
mastermanu, line 99: can't get a cancel event type here?

src/main/java/io/temporal/internal/statemachines/ChildWorkflowStateMachine.java
```diff
@@ -0,0 +1,289 @@
...
+        currentEvent.getStartChildWorkflowExecutionFailedEventAttributes();
+    Exception failure =
+        new ChildWorkflowTaskFailedException(
+            currentEvent.getEventId(),
+            WorkflowExecution.newBuilder().setWorkflowId(attributes.getWorkflowId()).build(),
+            attributes.getWorkflowType(),
+            RetryState.RETRY_STATE_NON_RETRYABLE_FAILURE,
+            null);
+    failure.initCause(
+        new WorkflowExecutionAlreadyStarted(
```
mastermanu, line 220: is it guaranteed that this is the only reason why we could not start a workflow?

src/main/java/io/temporal/internal/statemachines/DynamicTransitionTarget.java
```diff
@@ -0,0 +1,57 @@
...
+
+  @Override
+  public State apply(Data data) {
+    state = callback.apply(data);
+    for (State s : expectedStates) {
+      if (s.equals(state)) {
+        return state;
+      }
+    }
+    throw new IllegalStateException(
```
mastermanu, line 49: shouldn't we do this check before applying the callback?

src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
mfateev: Good catch. We had a test that ensured that the worker doesn't spin on the task failure. But after I added an additional check that the task failure is reported exactly once I found that it wasn't doing so. [Here is](https://github.com/temporalio/java-sdk/pull/172/commits/250f70dc022bffc0c2ea350a082fb8b46e6067ec) the new unit test and the fix. The unit test fails with attempts condition changed.

src/main/java/io/temporal/internal/replay/WorkflowExecutor.java
mfateev: See my earlier comment about force creating workflow tasks.

src/main/java/io/temporal/internal/replay/WorkflowExecutor.java
mfateev: Agree that it is error-prone. [Changed](https://github.com/temporalio/java-sdk/pull/172/commits/68db6c6c2da1c4fce870ff2c41cbfdade02928cd) WorkflowTaskResult to use the Builder pattern.

src/main/java/io/temporal/internal/statemachines/ActionOrEventType.java
mfateev: [Refactored](https://github.com/temporalio/java-sdk/pull/172/commits/1b1254b636ad9dd6895e8dc2179bede07b2153ae) StateMachine nomenclature to match the standard one.

src/main/java/io/temporal/internal/statemachines/ActivityStateMachine.java
mfateev: This is practically saying that this event is not causing a state transition. We could introduce another event here, but it doesn't really bring any value as no action is executed.

src/main/java/io/temporal/internal/statemachines/ActivityStateMachine.java
```diff
@@ -0,0 +1,404 @@
...
+            EventType.EVENT_TYPE_ACTIVITY_TASK_TIMED_OUT,
+            State.TIMED_OUT,
+            ActivityStateMachine::notifyTimedOut)
+        .add(
+            State.SCHEDULED_EVENT_RECORDED,
+            ExplicitEvent.CANCEL,
+            State.SCHEDULED_ACTIVITY_CANCEL_COMMAND_CREATED,
+            ActivityStateMachine::createRequestCancelActivityTaskCommand)
+        .add(
+            State.SCHEDULED_ACTIVITY_CANCEL_COMMAND_CREATED,
```
mastermanu, line 127: [Nit] swap the order of these two state transitions just for consistency

src/main/java/io/temporal/internal/statemachines/ActivityStateMachine.java
```diff
@@ -0,0 +1,404 @@
...
+        .add(
+            State.STARTED_ACTIVITY_CANCEL_COMMAND_CREATED,
+            EventType.EVENT_TYPE_ACTIVITY_TASK_CANCEL_REQUESTED,
+            State.STARTED_ACTIVITY_CANCEL_EVENT_RECORDED,
+            ActivityStateMachine::notifyCanceledIfTryCancel)
+        .add(
+            State.STARTED_ACTIVITY_CANCEL_COMMAND_CREATED,
+            CommandType.COMMAND_TYPE_REQUEST_CANCEL_ACTIVITY_TASK,
+            State.STARTED_ACTIVITY_CANCEL_COMMAND_CREATED)
+        .add(
```
mastermanu, line 170: can we group the transitions by state? You have a bunch of transitions from state.STARTED above...feels like you should move this cancel up there as well

src/main/java/io/temporal/internal/statemachines/ActivityStateMachine.java
```diff
@@ -0,0 +1,404 @@
...
+            State.SCHEDULED_ACTIVITY_CANCEL_COMMAND_CREATED,
+            CommandType.COMMAND_TYPE_REQUEST_CANCEL_ACTIVITY_TASK,
+            State.SCHEDULED_ACTIVITY_CANCEL_COMMAND_CREATED)
+        /*
+        This state transition is not possible.
+        It looks like it is valid when an event, handling of which requests activity
+        cancellation, precedes EVENT_TYPE_ACTIVITY_TASK_STARTED event.
+        But as all code execution happens in the event loop the STARTED event is
+        applied to the sate machine (as it is done for all command events before
```
mastermanu, line 140: [Nit] typo = 'sate machine'

src/main/java/io/temporal/internal/statemachines/ActivityStateMachine.java
```diff
@@ -0,0 +1,404 @@
...
+  public void createScheduleActivityTaskCommand() {
+    addCommand(
+        Command.newBuilder()
+            .setCommandType(CommandType.COMMAND_TYPE_SCHEDULE_ACTIVITY_TASK)
+            .setScheduleActivityTaskCommandAttributes(parameters.getAttributes())
+            .build());
+  }
+
+  public void cancel() {
+    if (parameters.getCancellationType() == ActivityCancellationType.ABANDON) {
```
mastermanu, line 246: (more for my own knowledge than an actual comment), what does "Abandon" mean vs some of the other semantics and why would a developer choose one over the other

src/main/java/io/temporal/internal/statemachines/ActivityStateMachine.java
```diff
@@ -0,0 +1,404 @@
...
+  private void notifyCanceled() {
+    ScheduleActivityTaskCommandAttributes.Builder scheduleAttr = parameters.getAttributes();
+    Failure canceledFailure =
+        Failure.newBuilder()
+            .setSource(JAVA_SDK)
+            .setCanceledFailureInfo(CanceledFailureInfo.getDefaultInstance())
+            .build();
+    ActivityFailureInfo activityFailureInfo =
+        ActivityFailureInfo.newBuilder()
+            .setActivityId(scheduleAttr.getActivityId())
```
mastermanu, line 279: [Nit] these fields seem more specific to ActivityInfo vs. ActivityFailureInfo

src/main/java/io/temporal/internal/statemachines/ActivityStateMachine.java
```diff
@@ -0,0 +1,404 @@
...
+              .setActivityId(scheduleAttr.getActivityId())
+              .setActivityType(scheduleAttr.getActivityType())
+              .setScheduledEventId(canceledAttr.getScheduledEventId())
+              .setStartedEventId(canceledAttr.getStartedEventId())
+              .build();
+      Failure failure =
+          Failure.newBuilder()
+              .setActivityFailureInfo(failureInfo)
+              .setCause(canceledFailure)
+              .setMessage("Activity task timedOut")
```
mastermanu, line 369: should this error message indicate the activity was canceled rather than timed out?

src/main/java/io/temporal/internal/statemachines/ActivityStateMachine.puml
```diff
@@ -0,0 +1,54 @@
...
+STARTED --> TIMED_OUT: ACTIVITY_TASK_TIMED_OUT
+SCHEDULED_EVENT_RECORDED --> SCHEDULED_ACTIVITY_CANCEL_COMMAND_CREATED: CANCEL
+SCHEDULED_ACTIVITY_CANCEL_COMMAND_CREATED --> SCHEDULED_ACTIVITY_CANCEL_EVENT_RECORDED: ACTIVITY_TASK_CANCEL_REQUESTED
+SCHEDULED_ACTIVITY_CANCEL_COMMAND_CREATED --> SCHEDULED_ACTIVITY_CANCEL_COMMAND_CREATED: REQUEST_CANCEL_ACTIVITY_TASK
+SCHEDULED_ACTIVITY_CANCEL_COMMAND_CREATED --> TIMED_OUT: ACTIVITY_TASK_TIMED_OUT
+SCHEDULED_ACTIVITY_CANCEL_EVENT_RECORDED --> CANCELED: ACTIVITY_TASK_CANCELED
+SCHEDULED_ACTIVITY_CANCEL_EVENT_RECORDED --> STARTED_ACTIVITY_CANCEL_EVENT_RECORDED: ACTIVITY_TASK_STARTED
+STARTED_ACTIVITY_CANCEL_COMMAND_CREATED --> STARTED_ACTIVITY_CANCEL_EVENT_RECORDED: ACTIVITY_TASK_CANCEL_REQUESTED
+STARTED_ACTIVITY_CANCEL_COMMAND_CREATED --> STARTED_ACTIVITY_CANCEL_COMMAND_CREATED: REQUEST_CANCEL_ACTIVITY_TASK
+STARTED --> STARTED_ACTIVITY_CANCEL_COMMAND_CREATED: CANCEL
```
mastermanu, line 43: Same comment as above file. Can we sort by source state?

src/main/java/io/temporal/internal/statemachines/CancellableCommand.java
```diff
@@ -0,0 +1,83 @@
...
+
+import io.temporal.api.command.v1.Command;
+import io.temporal.api.enums.v1.CommandType;
+import io.temporal.api.history.v1.HistoryEvent;
+import java.util.Objects;
+
+class CancellableCommand {
+
+  private final Command command;
+  private final EntityStateMachine commands;
```
mastermanu, line 30: this variable name (commands) seems off?

src/main/java/io/temporal/internal/statemachines/CancellableCommand.java
```diff
@@ -0,0 +1,83 @@
...
+      HistoryEvent event, boolean hasNextEvent) {
+    if (canceled) {
+      return WorkflowStateMachines.HandleEventStatus.NOT_MATCHING_EVENT;
+    }
+    return commands.handleEvent(event, hasNextEvent);
+  }
+
+  @Override
+  public String toString() {
+    return "NewCommand{" + "command=" + command + ", canceled=" + canceled + '}';
```
mastermanu, line 77: "NewCommand" --> "CancellableCommand"

src/main/java/io/temporal/internal/statemachines/ChildWorkflowStateMachine.java
```diff
@@ -0,0 +1,308 @@
...
+            attributes.getFailure());
+    completionCallback.apply(Optional.empty(), failure);
+  }
+
+  private void notifyTimedOut() {
+    ChildWorkflowExecutionTimedOutEventAttributes attributes =
+        currentEvent.getChildWorkflowExecutionTimedOutEventAttributes();
+    TimeoutFailure timeoutFailure =
+        new TimeoutFailure(null, null, TimeoutType.TIMEOUT_TYPE_START_TO_CLOSE);
+    timeoutFailure.setStackTrace(new StackTraceElement[0]);
```
mastermanu, line 256: why do we need to set the stack trace here, but not for terminated?

src/main/java/io/temporal/internal/statemachines/DynamicTransitionAction.java
```diff
@@ -0,0 +1,58 @@
...
+  State[] expectedStates;
+  State state;
+
+  DynamicTransitionAction(State[] expectedStates, DynamicCallback<State, Data> callback) {
+    this.expectedStates = expectedStates;
+    this.callback = callback;
+  }
+
+  @Override
+  public String toString() {
```
mastermanu, line 38: [Nit] toString value does not match class name

mastermanu: also might be worth printing out the expected states

src/main/java/io/temporal/internal/statemachines/EntityStateMachineBase.java
```diff
@@ -0,0 +1,88 @@
...
+
+  public EntityStateMachineBase(
+      StateMachine<State, ExplicitEvent, Data> stateMachine,
+      Functions.Proc1<CancellableCommand> commandSink) {
+    this.stateMachine = stateMachine;
+    this.commandSink = commandSink;
+  }
+
+  /**
+   * Notifies that command is included into zthe workflow task completion result.
```
mastermanu, line 43: [Nit] typo in comment

src/main/java/io/temporal/internal/statemachines/EntityStateMachineBase.java
```diff
@@ -0,0 +1,88 @@
...
+    return stateMachine.isFinalState();
+  }
+
+  protected State getState() {
+    return stateMachine.getState();
+  }
+
+  @Override
+  public String toString() {
+    return this.getClass().getSimpleName() + "{" + "state=" + stateMachine + '}';
```
mastermanu, line 86: worth printing out some of the other class members here?

src/main/java/io/temporal/internal/statemachines/FixedTransitionAction.java
```diff
@@ -0,0 +1,53 @@
...
+  final Functions.Proc1<Data> action;
+
+  FixedTransitionAction(State state, Functions.Proc1<Data> action) {
+    this.state = state;
+    this.action = action;
+  }
+
+  @Override
+  public String toString() {
+    return "TransitionDestination{" + "state=" + state + ", callback=" + action + '}';
```
mastermanu, line 40: [Nit] class doesn't match string value here as well. Also action= vs callback=

src/main/java/io/temporal/internal/sync/TestActivityEnvironmentInternal.java
```diff
@@ -115,6 +114,7 @@ public TestActivityEnvironmentInternal(TestEnvironmentOptions options) {
             WorkflowServiceStubsOptions.newBuilder()
                 .setChannel(channel)
                 .setMetricsScope(options.getMetricsScope())
+                .setQueryRpcTimeout(600000)
```
mastermanu, line 117: what is this timeout? is it in ms? should we have a constant/take a duration here?

src/main/java/io/temporal/worker/WorkerOptions.java
```diff
@@ -301,15 +304,15 @@ public int hashCode() {
   public String toString() {
     return "WorkerOptions{"
         + "maxActivitiesPerSecond="
```
mastermanu, line 306: [Nit] let's update the strings for these two variables as well

src/main/java/io/temporal/internal/statemachines/ActivityStateMachine.java
mfateev: I've added unit tests that demonstrate that these transitions are not possible. The basic idea is that the workflow code always executes in the context of an event loop. It causes activity timeouts to be delivered before the code that causes the timeout which generates the cancel command.

src/main/java/io/temporal/internal/statemachines/ActivityStateMachine.java
mfateev: Good catch. Removed.

src/main/java/io/temporal/internal/statemachines/ActivityStateMachine.java
mfateev: Would you file an issue for this? I think it is a great idea, but we want to do it holistically with the whole SDK.

src/main/java/io/temporal/internal/statemachines/CancelExternalStateMachine.java
mfateev: Added message.

src/main/java/io/temporal/internal/statemachines/ChildWorkflowStateMachine.java
mfateev: ChildWorkflow is canceled using CancelExternalStateMachine.

src/main/java/io/temporal/internal/statemachines/ChildWorkflowStateMachine.java
mfateev: Yes, according to [StartChildWorkflowExecutionFailedCause](https://github.com/temporalio/api/blob/4fd842b5ee4275d63fde0849c3b06803a1754131/temporal/api/enums/v1/failed_cause.proto#L60:6)

src/main/java/io/temporal/internal/statemachines/DynamicTransitionTarget.java
mfateev: After any such exception, the whole StateMachines object should be thrown away.

src/main/java/io/temporal/internal/statemachines/ActivityStateMachine.java
mfateev: Done

src/main/java/io/temporal/internal/statemachines/ActivityStateMachine.java
mfateev: Fixed

src/main/java/io/temporal/internal/statemachines/ActivityStateMachine.java
mfateev: Let me know if the [existing JavaDoc ](https://github.com/temporalio/java-sdk/blob/e9361fc01840a9e26e49effe0c52240f23583f94/src/main/java/io/temporal/activity/ActivityCancellationType.java#L30)is not clear.

src/main/java/io/temporal/internal/statemachines/ActivityStateMachine.java
mfateev: I don't understand this comment.

src/main/java/io/temporal/internal/statemachines/ActivityStateMachine.java
mfateev: Fixed

src/main/java/io/temporal/internal/statemachines/ActivityStateMachine.puml
mfateev: Done

src/main/java/io/temporal/internal/statemachines/CancellableCommand.java
mfateev: done

src/main/java/io/temporal/internal/statemachines/CancellableCommand.java
mfateev: done

src/main/java/io/temporal/internal/statemachines/ChildWorkflowStateMachine.java
mfateev: removed

src/main/java/io/temporal/internal/statemachines/DynamicTransitionAction.java
mfateev: done

src/main/java/io/temporal/internal/statemachines/EntityStateMachineBase.java
mfateev: done

src/main/java/io/temporal/internal/statemachines/FixedTransitionAction.java
mfateev: I"m keeping action. For me callback is more like reply thing. In this case it is actual action implementation.

src/main/java/io/temporal/internal/sync/TestActivityEnvironmentInternal.java
mfateev: Done.

src/main/java/io/temporal/worker/WorkerOptions.java
mfateev: done

src/main/java/io/temporal/internal/replay/ReplayWorkflowContextImpl.java
```diff
@@ -214,183 +207,127 @@ public SearchAttributes getSearchAttributes() {
...
+      Functions.Proc1<WorkflowExecution> executionCallback,
+      Functions.Proc2<Optional<Payloads>, Exception> callback) {
+    Functions.Proc cancellationHandler =
+        workflowStateMachines.startChildWorkflow(parameters, executionCallback, callback);
+    return (exception) -> cancellationHandler.apply();
   }
 
   @Override
-  public Consumer<Exception> signalWorkflowExecution(
+  public Functions.Proc1<Exception> signalExternalWorkflowExecution(
```
mastermanu, line 239: Interesting that signalExternalWorkflowExecution is cancelable. What is the scenario for that?

src/main/java/io/temporal/internal/replay/ReplayWorkflowContextImpl.java
```diff
@@ -214,183 +207,127 @@ public SearchAttributes getSearchAttributes() {
...
-    return workflowClock.isReplaying();
+    return workflowStateMachines.isReplaying();
   }
 
   @Override
-  public Consumer<Exception> createTimer(Duration delay, Consumer<Exception> callback) {
-    return workflowClock.createTimer(delay, callback);
+  public Functions.Proc1<RuntimeException> newTimer(
+      Duration delay, Functions.Proc1<RuntimeException> callback) {
+    if (delay == Duration.ZERO) {
```
mastermanu, line 276: Can we do <= here?

src/main/java/io/temporal/internal/replay/ReplayWorkflowContextImpl.java
```diff
@@ -214,183 +207,127 @@ public SearchAttributes getSearchAttributes() {
...
+  private void handleTimerCallback(Functions.Proc1<RuntimeException> callback, HistoryEvent event) {
+    switch (event.getEventType()) {
+      case EVENT_TYPE_TIMER_FIRED:
+        {
+          callback.apply(null);
+          return;
+        }
+      case EVENT_TYPE_TIMER_CANCELED:
+        {
+          CanceledFailure exception = new CanceledFailure("Cancelled by request");
```
mastermanu, line 299: [Super-Nit] Canceled vs Cancelled

src/main/java/io/temporal/internal/replay/ReplayWorkflowExecutor.java
```diff
@@ -295,32 +190,27 @@ private void eventLoop() {
...
 
   private void completeWorkflow() {
     if (failure != null) {
-      commandHelper.failWorkflowExecution(failure);
+      workflowStateMachines.newFailWorkflow(failure.getFailure());
       metricsScope.counter(MetricsType.WORKFLOW_FAILED_COUNTER).inc(1);
     } else if (cancelRequested) {
-      commandHelper.cancelWorkflowExecution();
+      workflowStateMachines.newCancelWorkflow();
       metricsScope.counter(MetricsType.WORKFLOW_CANCELLED_COUNTER).inc(1);
```
mastermanu, line 204: Canceled vs Cancelled

src/main/java/io/temporal/internal/replay/ReplayWorkflowExecutor.java
```diff
@@ -395,96 +242,48 @@ private void handleWorkflowExecutionSignaled(HistoryEvent event) {
...
+  public WorkflowTaskResult handleWorkflowTask(
+      PollWorkflowTaskQueueResponseOrBuilder workflowTask) {
     lock.lock();
     try {
+      long startTime = System.currentTimeMillis();
       queryResults.clear();
-      boolean forceCreateNewWorkflowTask = handleWorkflowTaskImpl(workflowTask, null);
-      return new WorkflowTaskResult(
-          commandHelper.getCommands(), queryResults, forceCreateNewWorkflowTask, completed);
+      handleWorkflowTaskImpl(workflowTask);
```
mastermanu, line 251: [Nit] kinda strange that some of these methods communicate their results via return values while others communicate using instance members of the class.

src/main/java/io/temporal/internal/replay/ReplayWorkflowExecutor.java
```diff
@@ -622,27 +341,76 @@ public void close() {
...
   }
 
-  public Consumer<HistoryEvent> getLocalActivityCompletionSink() {
-    return localActivityCompletionSink;
+  private void processLocalActivityRequests(long startTime) {
+    Duration maxProcessingTime = getWorkflowTaskTimeout().multipliedBy(4).dividedBy(5);
+    while (true) {
+      List<ExecuteLocalActivityParameters> laRequests =
+          workflowStateMachines.takeLocalActivityRequests();
+      long timeoutInterval = (long) ((System.currentTimeMillis() - startTime) * 0.5);
```
mastermanu, line 361: this is a strange name for a variable (timeoutInterval) especially as it just seems to represent the elapsed amount of time?

src/main/java/io/temporal/internal/replay/ReplayWorkflowExecutor.java
```diff
@@ -622,27 +341,76 @@ public void close() {
...
+      for (ExecuteLocalActivityParameters laRequest : laRequests) {
+        // TODO(maxim): In the presence of workflow task heartbeat this timeout doesn't make
+        // much sense. I believe we should add ScheduleToStart timeout for the local activities
+        // as well.
+        long processingTime = Math.min(System.currentTimeMillis() - startTime, timeoutInterval);
+        boolean accepted =
+            localActivityTaskPoller.apply(
+                new LocalActivityWorker.Task(
+                    laRequest, localActivityCompletionSink, 10000 /* TODO: Configurable */),
+                Duration.ofMillis(processingTime));
```
mastermanu, line 371: this represents the amount of time the local activity has to execute?

src/main/java/io/temporal/internal/replay/ReplayWorkflowExecutor.java
```diff
@@ -622,27 +341,76 @@ public void close() {
...
+          return;
+        }
+        ActivityTaskHandler.Result laCompletion;
+        // Do not wait under lock.
+        try {
+          laCompletion =
+              localActivityCompletionQueue.poll(maxWaitAllowed.toMillis(), TimeUnit.MILLISECONDS);
+          if (laCompletion == null) {
+            // Need to force a new task
+            break;
```
mastermanu, line 393: [Nit] double nested while(true) loops with breaks and returns might make this code hard to read

src/main/java/io/temporal/internal/statemachines/EntityStateMachineInitialCommand.java
```diff
@@ -0,0 +1,60 @@
...
+
+  protected final void cancelCommand() {
+    command.cancel();
+  }
+
+  protected long getInitialCommandEventId() {
+    return initialCommandEventId;
+  }
+
+  /** Sets initialCommandEventId ot the currentEvent eventId. */
```
mastermanu, line 56: typo in comment

src/main/java/io/temporal/internal/statemachines/MutableSideEffectStateMachine.java
```diff
@@ -0,0 +1,298 @@
...
+    @Override
+    public WorkflowStateMachines.HandleEventStatus handleEvent(
+        HistoryEvent event, boolean hasNextEvent) {
+      if (event.getEventType() != EventType.EVENT_TYPE_MARKER_RECORDED
+          || !event
+              .getMarkerRecordedEventAttributes()
+              .getMarkerName()
+              .equals(MUTABLE_SIDE_EFFECT_MARKER_NAME)) {
+        explicitEvent(ExplicitEvent.NON_MATCHING_EVENT);
+        return WorkflowStateMachines.HandleEventStatus.NOT_MATCHING_EVENT;
```
mastermanu, line 159: so we have NOT_MATCHING_EVENT and NON_MATCHING_EVENT. Can we align these?

src/main/java/io/temporal/internal/statemachines/MutableSideEffectStateMachine.java
```diff
@@ -0,0 +1,298 @@
...
+      }
+      Map<String, Payloads> detailsMap = event.getMarkerRecordedEventAttributes().getDetailsMap();
+      Optional<Payloads> idPayloads = Optional.ofNullable(detailsMap.get(MARKER_ID_KEY));
+      String expectedId = dataConverter.fromPayloads(0, idPayloads, String.class, String.class);
+      if (Strings.isNullOrEmpty(expectedId)) {
+        throw new IllegalStateException(
+            "Marker details map missing required key: " + MARKER_ID_KEY);
+      }
+      if (!id.equals(expectedId)) {
+        //        explicitEvent(ExplicitEvent.NON_MATCHING_EVENT);
```
mastermanu, line 169: any reason this is commented out?

src/main/java/io/temporal/internal/statemachines/MutableSideEffectStateMachine.java
```diff
@@ -0,0 +1,298 @@
...
+      super.handleEvent(event, hasNextEvent);
+      return WorkflowStateMachines.HandleEventStatus.OK;
+    }
+
+    State createMarker() {
+      State toState;
+      RecordMarkerCommandAttributes markerAttributes;
+      Optional<Payloads> updated;
+      updated = func.apply(result);
+      if (!updated.isPresent()) {
```
mastermanu, line 181: my own learning: i assume that isPresent returns false is the value is identical to the previously stored version, right?

src/main/java/io/temporal/internal/statemachines/SideEffectStateMachine.java
```diff
@@ -0,0 +1,164 @@
...
+
+  private State createMarkerCommand() {
+    RecordMarkerCommandAttributes markerAttributes;
+    if (replaying.apply()) {
+      markerAttributes = RecordMarkerCommandAttributes.getDefaultInstance();
+      return State.MARKER_COMMAND_CREATED_REPLAYING;
+    } else {
+      // executing first time
+      result = func.apply();
+      if (result == null) {
```
mastermanu, line 129: does this mean a side effect cannot return a null value?

src/main/java/io/temporal/internal/statemachines/SideEffectStateMachine.java
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  private void markerResultFromEvent() {
+    MarkerRecordedEventAttributes attributes = currentEvent.getMarkerRecordedEventAttributes();
+    if (!attributes.getMarkerName().equals(SIDE_EFFECT_MARKER_NAME)) {
+      throw new IllegalStateException(
+          "Expected " + SIDE_EFFECT_MARKER_NAME + ", received: " + attributes);
+    }
+    Map<String, Payloads> map = attributes.getDetailsMap();
+    Optional<Payloads> fromMaker = Optional.ofNullable(map.get(MARKER_DATA_KEY));
```
mastermanu, line 157: [Nit] fromMarker

src/main/java/io/temporal/internal/statemachines/SignalExternalStateMachine.java
```diff
@@ -0,0 +1,172 @@
...
+                  State.SIGNAL_EXTERNAL_COMMAND_CREATED,
+                  ExplicitEvent.CANCEL,
+                  State.CANCELED,
+                  SignalExternalStateMachine::cancelSignalExternalCommand)
+              .add(
+                  State.SIGNAL_EXTERNAL_COMMAND_CREATED,
+                  EventType.EVENT_TYPE_SIGNAL_EXTERNAL_WORKFLOW_EXECUTION_INITIATED,
+                  State.SIGNAL_EXTERNAL_COMMAND_RECORDED,
+                  EntityStateMachineInitialCommand::setInitialCommandEventId)
+              .add(
```
mastermanu, line 112: basically, once we have a notification from the server that the external workflow was signaled, then a cancel is basically a no-op, hence this transition back to its own state?

src/main/java/io/temporal/internal/replay/ReplayWorkflowContextImpl.java
mfateev: When workflow (or any part of its code) is canceled we automatically cancel all outstanding requests including signalExternalWorkflowExecution. Note that it is cancellable only until the command was sent out to the service.

src/main/java/io/temporal/internal/replay/ReplayWorkflowContextImpl.java
mfateev: done

src/main/java/io/temporal/internal/replay/ReplayWorkflowExecutor.java
mfateev: It looks like we are using Canceled in the API: https://github.com/temporalio/api/search?q=canceled&unscoped_q=canceled

[Google says ](https://www.google.com/search?q=canceled+vs+cancelled&oq=CANCELED+V&aqs=chrome.1.69i57j0l7.2643j0j7&sourceid=chrome&ie=UTF-8) that Canceled is American and cancelled is British.

mfateev: I'm going to rename cancelled to canceled everywhere.

src/main/java/io/temporal/internal/replay/ReplayWorkflowExecutor.java
mfateev: It would be a pretty large change. So far public methods return values and private ones communicate through fields.

src/main/java/io/temporal/internal/replay/ReplayWorkflowContextImpl.java
mastermanu: yeah, that cancellation window seems very, very tiny tbh.

src/main/java/io/temporal/internal/statemachines/TimerStateMachine.java
```diff
@@ -0,0 +1,179 @@
...
+              .add(
+                  State.START_COMMAND_CREATED,
+                  CommandType.COMMAND_TYPE_START_TIMER,
+                  State.START_COMMAND_CREATED)
+              .add(
+                  State.START_COMMAND_CREATED,
+                  EventType.EVENT_TYPE_TIMER_STARTED,
+                  State.START_COMMAND_RECORDED,
+                  EntityStateMachineInitialCommand::setInitialCommandEventId)
+              .add(
```
mastermanu, line 98: there is zero chance we somehow get a timer fired event before the timer started event? This should be properly serialized on the server, right?

src/main/java/io/temporal/internal/statemachines/TimerStateMachine.java
```diff
@@ -0,0 +1,179 @@
...
+                  State.START_COMMAND_CREATED,
+                  ExplicitEvent.CANCEL,
+                  State.CANCELED,
```
mastermanu, line 101: similarly, is there a chance the timer fired event arrives even after the explicit cancel?

src/main/java/io/temporal/internal/statemachines/TimerStateMachine.java
```diff
@@ -0,0 +1,179 @@
...
+    addCommand(
+        Command.newBuilder()
+            .setCommandType(CommandType.COMMAND_TYPE_CANCEL_TIMER)
+            .setCancelTimerCommandAttributes(
+                CancelTimerCommandAttributes.newBuilder().setTimerId(startAttributes.getTimerId()))
+            .build());
+  }
+
+  private void cancelTimerCommandFireTimer() {
+    cancelCommand();
```
mastermanu, line 176: just curious: what would happen if we did not call cancelCommand() here?

src/main/java/io/temporal/activity/ActivityCancellationType.java
```diff
@@ -24,7 +24,7 @@
 
 /**
  * Defines behaviour of the parent workflow when {@link CancellationScope} that wraps child workflow
- * execution request is cancelled. The result of the cancellation independently of the type is a
+ * execution request is canceled. The result of the cancellation independently of the type is a
```
alexshtin, line 27: We have the same mess on GoSDK and server side.

alexshtin: https://github.com/temporalio/go-sdk/pull/227

src/main/java/io/temporal/internal/statemachines/ActivityStateMachine.java
alexshtin: We have some ActivityInfo-ish fields in `ActivityFailureInfo`. They are needed to find activity in which failure was happened.

src/main/java/io/temporal/internal/statemachines/ActivityStateMachine.java
alexshtin: Hopefully we will never do translations :-)

src/main/java/io/temporal/internal/statemachines/LocalActivityStateMachine.java
```diff
@@ -0,0 +1,304 @@
...
+                  State.CREATED,
+                  ExplicitEvent.CHECK_EXECUTION_STATE,
+                  new State[] {State.REPLAYING, State.EXECUTING},
+                  LocalActivityStateMachine::getExecutionState)
+              .add(
+                  State.EXECUTING,
+                  ExplicitEvent.SCHEDULE,
+                  State.REQUEST_PREPARED,
+                  LocalActivityStateMachine::sendRequest)
+              .add(State.REQUEST_PREPARED, ExplicitEvent.GET_REQUEST, State.REQUEST_SENT)
```
mastermanu, line 147: is there a better name we can come up with than "GET_REQUEST?" maybe "CREATE_REQUEST" or CREATEORGET_REQUEST or something like that?

src/main/java/io/temporal/internal/statemachines/LocalActivityStateMachine.java
```diff
@@ -0,0 +1,304 @@
...
+              .add(
+                  State.REQUEST_SENT,
+                  ExplicitEvent.NON_REPLAY_WORKFLOW_TASK_STARTED,
```
mastermanu, line 150: this basically means local activity execution started?

src/main/java/io/temporal/internal/statemachines/LocalActivityStateMachine.java
```diff
@@ -0,0 +1,304 @@
...
+                  State.REQUEST_SENT)
+              .add(
+                  State.REQUEST_SENT,
+                  ExplicitEvent.HANDLE_RESPONSE,
```
mastermanu, line 154: and is this for explicitly reporting the local activity result? if so, can we have better naming for explicit event here?

src/main/java/io/temporal/internal/statemachines/LocalActivityStateMachine.java
```diff
@@ -0,0 +1,304 @@
...
+                  EventType.EVENT_TYPE_MARKER_RECORDED,
+                  State.MARKER_COMMAND_RECORDED)
+              .add(State.REPLAYING, ExplicitEvent.SCHEDULE, State.WAITING_MARKER_EVENT)
+              .add(
+                  State.WAITING_MARKER_EVENT,
+                  EventType.EVENT_TYPE_MARKER_RECORDED,
+                  State.MARKER_COMMAND_RECORDED,
+                  LocalActivityStateMachine::notifyResultFromEvent)
+              .add(
+                  // This is to cover the following edge case:
```
mastermanu, line 173: had a hard time following this. when calling getExecutionState wouldn't it still be as if the "first time" it ran? OR does the force workflowtask is started create a persistent checkpoint

src/main/java/io/temporal/internal/statemachines/ActivityStateMachine.java
mfateev: Done

src/main/java/io/temporal/internal/replay/ReplayWorkflowExecutor.java
mfateev: Refactored this part of the code.

src/main/java/io/temporal/worker/WorkerOptions.java
```diff
@@ -143,8 +144,8 @@ public Builder setMaxConcurrentLocalActivityExecutionSize(
      * every 10 seconds. This can be used to protect down stream services from flooding. The zero
      * value of this uses the default value. Default is unlimited.
      */
-    public Builder setTaskQueueActivitiesPerSecond(double taskQueueActivitiesPerSecond) {
-      this.taskQueueActivitiesPerSecond = taskQueueActivitiesPerSecond;
+    public Builder setMaxTaskQueueActivitiesPerSecond(double maxTaskQueueActivitiesPerSecond) {
```
mastermanu, line 147: why not have negative check here? Or do we treat that as 0? Interesting that we have different behavior than above (although I guess w/different defaults that makes sense)

src/main/java/io/temporal/internal/replay/ReplayWorkflowExecutor.java
mfateev: This is the maximum time for local retries. I've made it configurable through LocalActivityOptions

src/main/java/io/temporal/internal/replay/ReplayWorkflowExecutor.java
mfateev: Fixed

src/main/java/io/temporal/internal/statemachines/MutableSideEffectStateMachine.java
mfateev: Fixed.

src/main/java/io/temporal/internal/statemachines/MutableSideEffectStateMachine.java
mfateev: right

src/main/java/io/temporal/internal/statemachines/SideEffectStateMachine.java
mfateev: yes

src/main/java/io/temporal/internal/statemachines/SignalExternalStateMachine.java
mfateev: Yes, as there is no cancellation support by the service.

src/main/java/io/temporal/internal/statemachines/TimerStateMachine.java
mfateev: No, as timerStarted is written transactionally with the task addition to the timer queue. And due to buffered events the guarantee is even stronger. Command events always follow WorkflowTaskCompleted event.

src/main/java/io/temporal/internal/statemachines/TimerStateMachine.java
mfateev: It is not possible as the code that calls cancel executes (due to event loop invocation) after fired event applied.

src/main/java/io/temporal/internal/statemachines/TimerStateMachine.java
mfateev: Right, it will be sent to the service.

src/main/java/io/temporal/internal/statemachines/LocalActivityStateMachine.java
mfateev: Renamed to MARK_AS_SENT

src/main/java/io/temporal/internal/statemachines/LocalActivityStateMachine.java
mfateev: This means that WORKFLOW_TASK_STARTED event for a non replay task has started. Not sure what would be the better name.

src/main/java/io/temporal/internal/statemachines/LocalActivityStateMachine.java
mfateev: renamed to HANDLE_RESULT

src/main/java/io/temporal/internal/statemachines/LocalActivityStateMachine.java
mfateev: Forced workflow task is a workflow task that is recorded in the history the same way as a normal workflow task.

src/main/java/io/temporal/worker/WorkerOptions.java
mfateev: The above option is used only locally. So validation should be done by the SDK. The maxTaskQueueActivitiesPerSecond is an option sent to the service. We made a design decision to not validate on the client any server-side options.

src/main/java/io/temporal/internal/statemachines/ChildWorkflowStateMachine.java
```diff
@@ -0,0 +1,305 @@
...
+                  State.TIMED_OUT,
+                  State.TERMINATED)
+              .add(
+                  State.CREATED,
+                  ExplicitEvent.SCHEDULE,
+                  State.START_COMMAND_CREATED,
+                  ChildWorkflowStateMachine::createStartChildCommand)
+              .add(
+                  State.START_COMMAND_CREATED,
+                  CommandType.COMMAND_TYPE_START_CHILD_WORKFLOW_EXECUTION,
```
shawnhathaway, line 88: What are the side effects of not defining the no-op transitions that are command based?

src/main/java/io/temporal/internal/statemachines/EntityStateMachineBase.java
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  /**
+   * Notifies that command is included into the workflow task completion result.
+   *
+   * <p>Is not called for commands generated during replay.
+   */
+  @Override
+  public void handleCommand(CommandType commandType) {
+    stateMachine.handleCommand(commandType, (Data) this);
```
shawnhathaway, line 51: Why are events registered/validated, but commands are not?

src/main/java/io/temporal/internal/statemachines/ChildWorkflowStateMachine.java
mfateev: Only defined transitions are allowed. Workflow task fails if command or event is applied to the state that doesn't have a transition defined for it.

src/main/java/io/temporal/internal/statemachines/EntityStateMachineBase.java
mfateev: It is pretty evolved. The short answer is that this validation is required to support mutable side effect and versions which can have events missing for the given state machine. So if event doesn't match then it still can be valid as it just doesn't apply to this specific state machine. The same doesn't apply to commands as they always have to match the state machine that created them.

PR #179, opened 2020-08-15
Wired upsertSearchAttributes to the WorkflowStateMachines

mfateev:
Also changed state machines to not retain passed arguments after a command is created. It will reduce memory consumption.


src/main/java/io/temporal/internal/statemachines/ActivityStateMachine.java
```diff
@@ -415,5 +404,6 @@ private void createRequestCancelActivityTaskCommand() {
                 RequestCancelActivityTaskCommandAttributes.newBuilder()
                     .setScheduledEventId(getInitialCommandEventId()))
             .build());
+    parameters = null; // avoid retaining large input for the duration of the activity
```
mastermanu, line 407: curious what inspired you to make this change (whether it just occurred to you or you saw something interesting during local debugging / execution)

src/test/java/io/temporal/internal/statemachines/MutableSideEffectStateMachineTest.java
```diff
@@ -34,15 +38,46 @@
...
+        MutableSideEffectStateMachine.STATE_MACHINE_DEFINITION.getUnvisitedTransitions(
+            stateMachineList);
+    if (!missed.isEmpty()) {
+      CommandsGeneratePlantUMLStateDiagrams.writeToFile(
+          "test",
+          MutableSideEffectStateMachine.class,
+          MutableSideEffectStateMachine.STATE_MACHINE_DEFINITION.asPlantUMLStateDiagramCoverage(
+              stateMachineList));
+      fail(
+          "SideEffectStateMachine is missing test coverage for the following transitions:\n"
```
mastermanu, line 76: [Nit] Fix error message to say MutableSideEffectStateMachine

src/test/java/io/temporal/internal/statemachines/TimerStateMachineTest.java
```diff
@@ -34,15 +35,42 @@
...
+    List<Transition> missed =
+        TimerStateMachine.STATE_MACHINE_DEFINITION.getUnvisitedTransitions(stateMachineList);
+    if (!missed.isEmpty()) {
+      CommandsGeneratePlantUMLStateDiagrams.writeToFile(
+          "test",
+          TimerStateMachine.class,
+          TimerStateMachine.STATE_MACHINE_DEFINITION.asPlantUMLStateDiagramCoverage(
+              stateMachineList));
+      fail(
+          "SideEffectStateMachine is missing test coverage for the following transitions:\n"
```
mastermanu, line 69: [Nit] fix error message to say TimerStateMachine

src/main/java/io/temporal/internal/statemachines/ActivityStateMachine.java
mfateev: I was thinking about it in my spare time but was forgetting about it when getting to the computer :).

PR #195, opened 2020-08-23
Retry options

mfateev:
Fix for https://github.com/temporalio/java-sdk/issues/174.

The validation and defaulting of RetryOptions are removed when scheduling activities and child workflows. The service is expected to perform validation as well as provide defaults for them.

Added validation and defaulting of RetryOptions for local activities as service is not involved in their execution.

### DIFF
--- a/src/main/java/io/temporal/common/RetryOptions.java
+++ b/src/main/java/io/temporal/common/RetryOptions.java
@@ -66,10 +66,10 @@ public final class RetryOptions {
       if (o == null) {
         return null;
       }
-      return RetryOptions.newBuilder(o).validateBuildWithDefaults();
+      return o;
     }
     if (o == null) {
-      o = RetryOptions.newBuilder().build();
+      o = RetryOptions.getDefaultInstance();
     }
     Duration initial = merge(r.initialIntervalSeconds(), o.getInitialInterval());
     RetryOptions.Builder builder = RetryOptions.newBuilder();
@@ -89,7 +89,7 @@ public final class RetryOptions {
     return builder
         .setMaximumAttempts(merge(r.maximumAttempts(), o.getMaximumAttempts(), int.class))
         .setDoNotRetry(merge(r.doNotRetry(), o.getDoNotRetry()))
-        .validateBuildWithDefaults();
+        .build();
   }
 
   /** The parameter options takes precedence. */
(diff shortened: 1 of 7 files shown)


src/main/java/io/temporal/common/RetryOptions.java
```diff
@@ -66,10 +66,10 @@ public static RetryOptions merge(MethodRetry r, RetryOptions o) {
       if (o == null) {
         return null;
       }
-      return RetryOptions.newBuilder(o).validateBuildWithDefaults();
+      return o;
```
samarabbas, line 69: previously we used to return a new copy.  Now we return the same instance back.  Can this cause issues?

src/main/java/io/temporal/internal/sync/SyncWorkflowContext.java
```diff
@@ -343,9 +343,8 @@ private ExecuteLocalActivityParameters constructExecuteLocalActivityParameters(
       activityTask.setInput(input.get());
     }
     RetryOptions retryOptions = options.getRetryOptions();
-    if (retryOptions != null) {
-      activityTask.setRetryPolicy(toRetryPolicy(retryOptions));
-    }
+    activityTask.setRetryPolicy(
+        toRetryPolicy(RetryOptions.newBuilder(retryOptions).validateBuildWithDefaults()));
```
samarabbas, line 347: validateBuildWithDefaults needs to match same validation we do on server sides for activities?

src/main/java/io/temporal/internal/sync/WorkflowRetryerInternal.java
```diff
@@ -186,7 +186,6 @@ public static void retry(
   }
 
   private static RetryOptions getRetryOptionsSideEffect(String retryId, RetryOptions options) {
-    options = RetryOptions.newBuilder(options).validateBuildWithDefaults();
```
samarabbas, line 189: I'm confused why no validation here?  I thought this is also all client-side.

src/main/java/io/temporal/common/RetryOptions.java
mfateev: It is immutable, so no issues expected.

src/main/java/io/temporal/internal/sync/SyncWorkflowContext.java
mfateev: Not necessary as it is for local activity only. But I believe they do now.

src/main/java/io/temporal/internal/sync/WorkflowRetryerInternal.java
mfateev: This records retry options as they are, to ensure determinism if they change.

PR #219, opened 2020-09-23
Verify that next page is not empty in the history iterator

vitarb:
Next page token may point to an empty page. This results in a possibility that hasNext() may return true and then next() throws NoSuchElementException. This sequence is invalid and breaks iterator contract.

Here is example of an exception that we've observed:
```
Caused by: io.grpc.StatusRuntimeException: INVALID_ARGUMENT: java.util.NoSuchElementException
	at java.base/java.util.Collections$EmptyIterator.next(Collections.java:4277)
	at io.temporal.internal.replay.WorkflowHistoryIterator.next(WorkflowHistoryIterator.java:116)
	at io.temporal.internal.replay.WorkflowHistoryIterator.next(WorkflowHistoryIterator.java:40)
	at io.temporal.internal.replay.ReplayWorkflowRunTaskHandler.handleWorkflowTaskImpl(ReplayWorkflowRunTaskHandler.java:179)
	at io.temporal.internal.replay.ReplayWorkflowRunTaskHandler.handleQueryWorkflowTask(ReplayWorkflowRunTaskHandler.java:255)
	at io.temporal.internal.replay.ReplayWorkflowTaskHandler.handleQueryOnlyWorkflowTask(ReplayWorkflowTaskHandler.java:241)
	at io.temporal.internal.replay.ReplayWorkflowTaskHandler.handleWorkflowTask(ReplayWorkflowTaskHandler.java:110)
	at io.temporal.internal.worker.WorkflowWorker$TaskHandlerImpl.handle(WorkflowWorker.java:309)
	at io.temporal.internal.worker.WorkflowWorker$TaskHandlerImpl.handle(WorkflowWorker.java:275)
	at io.temporal.internal.worker.PollTaskExecutor.lambda$process$0(PollTaskExecutor.java:73)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1130)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:630)
	at java.base/java.lang.Thread.run(Thread.java:832)
```

This PR addresses the problem by making a call to the server in the hasNext method and checking if the next page is not empty.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/replay/WorkflowHistoryIterator.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/replay/WorkflowHistoryIterator.java
@@ -49,7 +49,7 @@ class WorkflowHistoryIterator implements Iterator<HistoryEvent> {
   private final Scope metricsScope;
   private final PollWorkflowTaskQueueResponseOrBuilder task;
   private Iterator<HistoryEvent> current;
-  private ByteString nextPageToken;
+  ByteString nextPageToken;
 
   WorkflowHistoryIterator(
       WorkflowServiceStubs service,
--- a/temporal-sdk/src/main/java/io/temporal/internal/replay/WorkflowHistoryIterator.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/replay/WorkflowHistoryIterator.java
@@ -34,6 +34,7 @@ import io.temporal.internal.common.RpcRetryOptions;
 import io.temporal.serviceclient.WorkflowServiceStubs;
 import java.time.Duration;
 import java.util.Iterator;
+import java.util.NoSuchElementException;
 import java.util.Objects;
 
 /** Supports iteration over history while loading new pages through calls to the service. */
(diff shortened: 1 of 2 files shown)


temporal-sdk/src/main/java/io/temporal/internal/replay/WorkflowHistoryIterator.java
```diff
@@ -77,6 +112,24 @@ public HistoryEvent next() {
       return current.next();
     }
 
+    GetWorkflowExecutionHistoryResponse response = queryWorkflowExecutionHistory(
```
mfateev, line 115: I think you can avoid this call here if line 111 calls this.hasNext

vitarb: That call is to current.hasNext(), which is just cached history.
We still should make a server call if current is empty and page token is available.
This would make it possible to safely call it.next() without calling it.hasNext() first.
That being said I've updated implementation, added tests and verified that it's working in bench, please take a look

temporal-sdk/src/main/java/io/temporal/internal/replay/WorkflowHistoryIterator.java
```diff
@@ -66,17 +66,44 @@
...
   @Override
   public HistoryEvent next() {
-    if (current.hasNext()) {
+    // If current has next element then simply return it, otherwise check if we have next page token
+    // If no page token available then we should call next() which throws NoSuchElementException.
+    if (current.hasNext() || nextPageToken.isEmpty()) {
       return current.next();
     }
 
+    GetWorkflowExecutionHistoryResponse response = queryWorkflowExecutionHistory();
```
mfateev, line 98: I don't think it is correct. As it is still going to blow up with the response contains an empty history and hasNext is not called from outside.

My proposal is to change the line 94 to:
```java
if (this.hasNext()) {
   return current.next();
}
```
and ensure that hasNext either caches non empty events in current or returns false.

mfateev (changes requested):

temporal-sdk/src/main/java/io/temporal/internal/replay/WorkflowHistoryIterator.java
vitarb: If there is no next element then next() should throw a NoSuchElementException. This is exactly what it will do.

vitarb: Simplified the code a bit, but kept behavior same.

mfateev (approved):

PR #220, opened 2020-09-24
Include WorkflowID into workflow thread name

mfateev:


temporal-sdk/src/main/java/io/temporal/internal/sync/DeterministicRunnerImpl.java
```diff
@@ -424,6 +424,9 @@ private void checkClosed() {
 
   /** To be called only from another workflow thread. */
   public WorkflowThread newThread(Runnable runnable, boolean detached, String name) {
+    if (name == null) {
+      name = "workflow[" + workflowContext.getContext().getWorkflowId() + "]-" + addedThreads;
+    }
```
vitarb, line 429: Is it possible to rename thread back to "idle-XXX" once the execution is complete?

vitarb (approved):

temporal-sdk/src/main/java/io/temporal/internal/sync/DeterministicRunnerImpl.java
mfateev: It already sets the name to the original here: https://github.com/temporalio/java-sdk/blob/30420212a26283cf7cbe6339512058a3186074e8/temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowThreadImpl.java#L126

PR #230, opened 2020-10-21
Add default metric tags to the metric scope

vitarb:
This change adds default values for every metric tag that we use, making it compatible with prometheus metrics publisher which requires same sets of tags to be used on every metric.


mfateev (approved):

Spikhalskiy:
This is caused by https://github.com/micrometer-metrics/micrometer/issues/877

PR #238, opened 2020-10-29
Copy headers from factory options into grpc requests

vitarb:
This PR contains changes required to populate headers on grpc requests to the server.
Previously headers set on factory options were simply ignored, here we fix it.

### DIFF
--- a/temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsImpl.java
+++ b/temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsImpl.java
@@ -32,8 +32,6 @@ import io.grpc.stub.MetadataUtils;
 import io.temporal.api.workflowservice.v1.WorkflowServiceGrpc;
 import java.io.IOException;
 import java.time.Duration;
-import java.util.Map;
-import java.util.Map.Entry;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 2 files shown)


temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsImpl.java
```diff
@@ -138,7 +140,7 @@ public WorkflowServiceStubsImpl(
         new GrpcMetricsInterceptor(options.getMetricsScope());
     ClientInterceptor deadlineInterceptor = new GrpcDeadlineInterceptor(options);
     GrpcTracingInterceptor tracingInterceptor = new GrpcTracingInterceptor();
-    Metadata headers = new Metadata();
+    Metadata headers = headersFrom(options.getHeaders());
```
mfateev, line 143: It looks like this wouldn't allow binary headers. Should we just change options.headers type to io.grpc.Metadata?

vitarb: Good call, let me do that, it would give more flexibility on the client side. Also factory options are tied to grpc anyways so exposing some grpc objects externally wouldn't be too bad.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsImpl.java
```diff
@@ -138,7 +140,7 @@ public WorkflowServiceStubsImpl(
         new GrpcMetricsInterceptor(options.getMetricsScope());
     ClientInterceptor deadlineInterceptor = new GrpcDeadlineInterceptor(options);
     GrpcTracingInterceptor tracingInterceptor = new GrpcTracingInterceptor();
-    Metadata headers = new Metadata();
+    Metadata headers = options.getHeaders();
```
mfateev, line 143: I would merge options headers here. Mutating ones from the options is not clean.

mfateev (approved):

mfateev (approved):

temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsImpl.java
vitarb: Done.

PR #246, opened 2020-11-07
Added dump of all workflow thread to the deadlock detector

mfateev:


temporal-sdk/src/main/java/io/temporal/internal/sync/DeterministicRunnerImpl.java
```diff
@@ -297,6 +297,18 @@ public void runUntilAllBlocked(long deadlockDetectionTimeout) {
...
         }
       } while (progress && !threads.isEmpty());
+    } catch (PotentialDeadlockException e) {
+      StringBuilder dump = new StringBuilder();
+      for (WorkflowThread t : threads) {
+        if (t.getWorkflowThreadContext() != e.getWorkflowThreadContext()) {
+          if (dump.length() > 0) {
+            dump.append("\n");
+          }
+          dump.append(t.getStackTrace());
```
vitarb, line 307: Should we add thread names too?

vitarb (approved):

PR #251, opened 2020-11-13
Stop using illegal reflective access for setting cause during exception

vitarb:
This change removes illegal reflective access to the cause field in the `CheckedExceptionWrapper` replacing it with shallow unwrapping of the exception plus full unwrapping during serialization.

The reason we need to keep shallow unwrapping instead of getting rid of it entirely is because there are a few places in the pre-serialization code that either log or conduct actions based on the client exception. Look at [POJOWorkflowImplementationFactory](https://github.com/temporalio/sdk-java/blob/30420212a26283cf7cbe6339512058a3186074e8/temporal-sdk/src/main/java/io/temporal/internal/sync/POJOWorkflowImplementationFactory.java#L297) and [POJOActivityTaskHandler](https://github.com/temporalio/sdk-java/blob/8fba0454392a79d7371f103bc8fb084c1be8ae68/temporal-sdk/src/main/java/io/temporal/internal/sync/POJOActivityTaskHandler.java#L330) as examples.

One tradeoff that we have to make here is whether or not we want to eliminate a possibility of client seeing CheckedExceptionWrapper as current approach doesn't guarantee that and if there are multiple layers of wrappers inner layers might be visible in the log, logged by POJOActivityTaskHandler. Preferred way to solve it could be by transforming that exception to Failure before logging, alternatively we could reimplement unwrap by serializing to failure and back, which would eliminate all exception wrappers in the process.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/failure/FailureConverter.java
+++ b/temporal-sdk/src/main/java/io/temporal/failure/FailureConverter.java
@@ -180,6 +180,9 @@ public class FailureConverter {
   }
 
   public static Failure exceptionToFailure(Throwable e) {
+    if (e instanceof CheckedExceptionWrapper) {
+      return exceptionToFailure(e.getCause());
+    }
     String message;
     if (e instanceof TemporalFailure) {
       TemporalFailure tf = (TemporalFailure) e;
@@ -189,9 +192,6 @@ public class FailureConverter {
       message = tf.getOriginalMessage();
     } else {
       message = e.getMessage() == null ? "" : e.getMessage();
-    }
-    if (e instanceof CheckedExceptionWrapper) {
-      return exceptionToFailure(e.getCause());
     }
     String stackTrace = serializeStackTrace(e);
     Failure.Builder failure =
--- a/temporal-sdk/src/main/java/io/temporal/failure/FailureConverter.java
+++ b/temporal-sdk/src/main/java/io/temporal/failure/FailureConverter.java
@@ -180,11 +180,6 @@ public class FailureConverter {
   }
 
   public static Failure exceptionToFailure(Throwable e) {
-    e = CheckedExceptionWrapper.unwrap(e);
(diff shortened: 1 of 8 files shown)


temporal-sdk/src/main/java/io/temporal/failure/FailureConverter.java
```diff
@@ -195,11 +190,14 @@ public static Failure exceptionToFailureNoUnwrapping(Throwable e) {
     } else {
       message = e.getMessage() == null ? "" : e.getMessage();
     }
+    if (e instanceof CheckedExceptionWrapper) {
+      return exceptionToFailure(e.getCause());
+    }
```
mfateev, line 195: I would move it to the top of the method to avoid unnecessary e.getMessage call.

temporal-sdk/src/main/java/io/temporal/internal/common/CheckedExceptionWrapper.java
```diff
@@ -69,25 +69,11 @@ public static RuntimeException wrap(Throwable e) {
   }
 
   /**
-   * Removes CheckedException wrapper from the whole chain of Exceptions. Assumes that wrapper
+   * Removes CheckedExceptionWrapper from the top of the chain of Exceptions. Assumes that wrapper
```
mfateev, line 72: Remove causeField and the related initialization code above.
Remove setThrowableCause method below.

temporal-sdk/src/main/java/io/temporal/failure/FailureConverter.java
vitarb: Done

temporal-sdk/src/main/java/io/temporal/internal/common/CheckedExceptionWrapper.java
vitarb: Done

mfateev (approved):

PR #260, opened 2020-11-18
Add getLastFailure method to Workflow API

Sushisource:
See https://github.com/temporalio/sdk-java/issues/257

First PR so please fill me in on anything I'm missing arch or style wise. Thanks!

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflow.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflow.java
@@ -24,6 +24,7 @@ import static io.temporal.internal.sync.DeterministicRunner.DEFAULT_DEADLOCK_DET
 import io.temporal.api.common.v1.Payloads;
 import io.temporal.api.common.v1.WorkflowType;
 import io.temporal.api.enums.v1.EventType;
+import io.temporal.api.failure.v1.Failure;
 import io.temporal.api.history.v1.HistoryEvent;
 import io.temporal.api.history.v1.WorkflowExecutionStartedEventAttributes;
 import io.temporal.api.query.v1.WorkflowQuery;
@@ -101,8 +102,13 @@ class SyncWorkflow implements ReplayWorkflow {
         startEvent.hasLastCompletionResult()
             ? Optional.of(startEvent.getLastCompletionResult())
             : Optional.empty();
+    Optional<Failure> lastFail =
+        startEvent.hasContinuedFailure()
+            ? Optional.of(startEvent.getContinuedFailure())
+            : Optional.empty();
+    log.info("Huh? {}", lastFail);
     SyncWorkflowContext syncContext =
-        new SyncWorkflowContext(context, dataConverter, contextPropagators, result);
+        new SyncWorkflowContext(context, dataConverter, contextPropagators, result, lastFail);
 
     workflowProc = new WorkflowExecuteRunnable(syncContext, workflow, startEvent);
     // The following order is ensured by this code and DeterministicRunner implementation:
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflow.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflow.java
@@ -102,12 +102,12 @@ class SyncWorkflow implements ReplayWorkflow {
         startEvent.hasLastCompletionResult()
             ? Optional.of(startEvent.getLastCompletionResult())
             : Optional.empty();
-    Optional<Failure> lastFail =
+    Optional<Failure> lastFailure =
         startEvent.hasContinuedFailure()
             ? Optional.of(startEvent.getContinuedFailure())
             : Optional.empty();
     SyncWorkflowContext syncContext =
-        new SyncWorkflowContext(context, dataConverter, contextPropagators, result, lastFail);
+        new SyncWorkflowContext(context, dataConverter, contextPropagators, result, lastFailure);
 
     workflowProc = new WorkflowExecuteRunnable(syncContext, workflow, startEvent);
     // The following order is ensured by this code and DeterministicRunner implementation:
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflow.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflow.java
@@ -106,7 +106,6 @@ class SyncWorkflow implements ReplayWorkflow {
         startEvent.hasContinuedFailure()
             ? Optional.of(startEvent.getContinuedFailure())
             : Optional.empty();
-    log.info("Huh? {}", lastFail);
     SyncWorkflowContext syncContext =
         new SyncWorkflowContext(context, dataConverter, contextPropagators, result, lastFail);
 
--- a/temporal-sdk/src/main/java/io/temporal/internal/testservice/StateMachines.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/testservice/StateMachines.java
@@ -199,6 +199,7 @@ class StateMachines {
     String originalExecutionRunId;
     Optional<String> continuedExecutionRunId;
     Functions.Proc runTimerCancellationHandle;
+    Failure lastFailure;
 
     WorkflowData(
         Optional<TestServiceRetryState> retryState,
@@ -206,13 +207,15 @@ class StateMachines {
         String cronSchedule,
         Payloads lastCompletionResult,
         String originalExecutionRunId,
-        Optional<String> continuedExecutionRunId) {
+        Optional<String> continuedExecutionRunId,
+        Failure lastFailure) {
       this.retryState = retryState;
       this.backoffStartInterval = backoffStartInterval;
       this.cronSchedule = cronSchedule;
       this.lastCompletionResult = lastCompletionResult;
       this.originalExecutionRunId = originalExecutionRunId;
       this.continuedExecutionRunId = continuedExecutionRunId;
+      this.lastFailure = lastFailure;
     }
 
     @Override
@@ -829,6 +832,9 @@ class StateMachines {
     }
     if (data.lastCompletionResult != null) {
       a.setLastCompletionResult(data.lastCompletionResult);
+    }
+    if (data.lastFailure != null) {
+      a.setContinuedFailure(data.lastFailure);
     }
     if (request.hasMemo()) {
       a.setMemo(request.getMemo());
--- a/temporal-sdk/src/main/java/io/temporal/internal/testservice/StateMachines.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/testservice/StateMachines.java
@@ -140,6 +140,7 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.OptionalLong;
 import java.util.UUID;
@@ -215,7 +216,7 @@ class StateMachines {
       this.lastCompletionResult = lastCompletionResult;
       this.originalExecutionRunId = originalExecutionRunId;
       this.continuedExecutionRunId = continuedExecutionRunId;
-      this.lastFailure = lastFailure;
+      this.lastFailure = Objects.requireNonNull(lastFailure);
     }
 
     @Override
@@ -833,7 +834,7 @@ class StateMachines {
     if (data.lastCompletionResult != null) {
       a.setLastCompletionResult(data.lastCompletionResult);
     }
-    if (data.lastFailure != null && data.lastFailure.isPresent()) {
+    if (data.lastFailure.isPresent()) {
       a.setContinuedFailure(data.lastFailure.get());
     }
     if (request.hasMemo()) {
--- a/temporal-sdk/src/main/java/io/temporal/internal/testservice/StateMachines.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/testservice/StateMachines.java
@@ -196,19 +196,19 @@ class StateMachines {
     Duration backoffStartInterval;
     String cronSchedule;
     Payloads lastCompletionResult;
+    Optional<Failure> lastFailure;
     String originalExecutionRunId;
     Optional<String> continuedExecutionRunId;
     Functions.Proc runTimerCancellationHandle;
-    Failure lastFailure;
 
     WorkflowData(
         Optional<TestServiceRetryState> retryState,
         Duration backoffStartInterval,
         String cronSchedule,
         Payloads lastCompletionResult,
+        Optional<Failure> lastFailure,
         String originalExecutionRunId,
-        Optional<String> continuedExecutionRunId,
-        Failure lastFailure) {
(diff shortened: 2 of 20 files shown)


temporal-sdk/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
```diff
@@ -276,7 +277,8 @@ StartWorkflowExecutionResponse startWorkflowExecutionImpl(
           parent,
           parentChildInitiatedEventId,
           signalWithStartSignal,
-          workflowId);
+          workflowId,
+          retryState.flatMap(TestServiceRetryState::getLastFailure).orElse(null));
```
mfateev, line 281: I think a simple if condition at line 267 would be much more readable.

temporal-sdk/src/main/java/io/temporal/internal/testservice/StateMachines.java
```diff
@@ -199,20 +199,23 @@
...
 
     WorkflowData(
         Optional<TestServiceRetryState> retryState,
         Duration backoffStartInterval,
         String cronSchedule,
         Payloads lastCompletionResult,
         String originalExecutionRunId,
-        Optional<String> continuedExecutionRunId) {
+        Optional<String> continuedExecutionRunId,
+        Failure lastFailure) {
```
mfateev, line 211: Shouldn't this be Optional?

temporal-sdk/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -183,7 +171,8 @@
       OptionalLong parentChildInitiatedEventId,
       Optional<String> continuedExecutionRunId,
       TestWorkflowService service,
-      TestWorkflowStore store) {
+      TestWorkflowStore store,
+      Failure lastFailure) {
```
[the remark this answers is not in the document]
mfateev: And make it Optional

temporal-sdk/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -1191,7 +1181,13 @@ private void processFailWorkflowExecution(
...
 
     if (!Strings.isNullOrEmpty(data.cronSchedule)) {
-      startNewCronRun(ctx, workflowTaskCompletedId, identity, data, data.lastCompletionResult);
+      startNewCronRun(
+          ctx,
+          workflowTaskCompletedId,
+          identity,
+          data,
+          data.lastCompletionResult,
+          Optional.of(d.getFailure()));
```
mfateev, line 1190: gRPC doesn't support nulls.
So d.getFailure never returns null and Optional.of of "d" that has d.hasFailure==false will be an optional of an empty Failure structure.

temporal-sdk/src/main/java/io/temporal/workflow/Workflow.java
```diff
@@ -1181,6 +1182,18 @@ public static Logger getLogger(String name) {
...
 
+  /**
+   * Extract the latest failure from some previous run for this cron workflow. This is used in
+   * combination with cron schedule. A workflow can be started with an optional cron schedule. If
+   * any previous workflow has failed, this function returns that failure. If no previous workflows
+   * have failed, an empty optional is returned.
+   *
+   * @return The last {@link Failure} that occurred in this workflow, if there has been one.
+   */
+  public static Optional<Failure> getLastFailure() {
```
mfateev, line 1193: Change the result type to Optional<RuntimeException> as neither Failure nor Payload are exposed to the workflow code. Use [failureToException](https://github.com/temporalio/sdk-java/blob/4bb23c3a660de6a2630901d0dc6c049d9742ebe1/temporal-sdk/src/main/java/io/temporal/failure/FailureConverter.java#L84) method to convert.

temporal-sdk/src/test/java/io/temporal/workflow/WorkflowTest.java
```diff
@@ -4264,6 +4267,10 @@ public String execute(String testName) {
 
   @Test
   public void testWorkflowWithCronSchedule() {
+    // REVIEW: Min interval can be less than 1min when running against a real server - what explains
+    // that discrepancy?
+    //   Why doesn't "@every 1s" work here? Different cron parsing libs?
```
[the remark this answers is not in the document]
mfateev: We need to update the documentation. The discrepancy is because the service uses https://github.com/robfig/cron and the java unit test service uses https://github.com/jmrozanec/cron-utils.

The time skipping is the major feature of the unit testing framework that supports unit testing long running workflows.

temporal-sdk/src/main/java/io/temporal/internal/testservice/StateMachines.java
Sushisource: Could be, for sure. Java is killing me a bit with the whole `Optional<T>` actually has three states problem, where it can still be null. IntelliJ has a lint that whines at you when you use it as a parameter, which I've always found a bit odd, but there are reasons: https://stackoverflow.com/questions/31922866/why-should-java-8s-optional-not-be-used-in-arguments

I don't really have much of a preference. Since we're already using them as arguments here though, I should probably stay consistent. Curious to hear your thoughts though @mfateev

temporal-sdk/src/test/java/io/temporal/workflow/WorkflowTest.java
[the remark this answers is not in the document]
Sushisource: Thanks for the answer! Can you point me at info/code on the time skipping stuff?

temporal-sdk/src/main/java/io/temporal/workflow/Workflow.java
Sushisource: Changed. Should we change the method name as well? Pertains to my question here:https://github.com/temporalio/sdk-go/pull/293#discussion_r528050388

temporal-sdk/src/main/java/io/temporal/internal/testservice/StateMachines.java
```diff
@@ -830,6 +833,9 @@ private static void startWorkflow(
     if (data.lastCompletionResult != null) {
       a.setLastCompletionResult(data.lastCompletionResult);
     }
+    if (data.lastFailure != null && data.lastFailure.isPresent()) {
```
mfateev, line 836: We shouldn't allow null values for Optional fields. They should be either Optiona.empty or populated.

temporal-sdk/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -37,20 +37,7 @@
...
-import io.temporal.api.command.v1.FailWorkflowExecutionCommandAttributes;
-import io.temporal.api.command.v1.RecordMarkerCommandAttributes;
-import io.temporal.api.command.v1.RequestCancelActivityTaskCommandAttributes;
-import io.temporal.api.command.v1.RequestCancelExternalWorkflowExecutionCommandAttributes;
-import io.temporal.api.command.v1.ScheduleActivityTaskCommandAttributes;
-import io.temporal.api.command.v1.SignalExternalWorkflowExecutionCommandAttributes;
-import io.temporal.api.command.v1.StartChildWorkflowExecutionCommandAttributes;
-import io.temporal.api.command.v1.StartTimerCommandAttributes;
-import io.temporal.api.command.v1.UpsertWorkflowSearchAttributesCommandAttributes;
+import io.temporal.api.command.v1.*;
```
mfateev, line 40: Revert. Change your IDE settings to not use *.

temporal-sdk/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
```diff
@@ -259,10 +260,14 @@ StartWorkflowExecutionResponse startWorkflowExecutionImpl(
         }
       }
       Optional<TestServiceRetryState> retryState;
+      Optional<Failure> lastFailure = Optional.empty();
```
mfateev, line 263: I would move empty() initialization to else block:

Optional<Failure> lastFailure;
if (...) {
   lastFailure = ...
} else {
   lastFailure = Optional.empty();
}

temporal-sdk/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -1307,8 +1299,11 @@ private void startNewCronRun(
             .setTaskQueue(startRequest.getTaskQueue())
             .setBackoffStartInterval(ProtobufTimeUtils.toProtoDuration(backoffInterval))
             .setRetryPolicy(startRequest.getRetryPolicy())
-            .setLastCompletionResult(lastCompletionResult)
-            .build();
+            .setLastCompletionResult(lastCompletionResult);
+    if (lastFailure != null) {
```
mfateev, line 1303: It looks like we use both Optional and null for lastFailure inside the testservice. Let's make this consistent.

temporal-sdk/src/main/java/io/temporal/workflow/Workflow.java
```diff
@@ -1181,6 +1183,19 @@ public static Logger getLogger(String name) {
...
+   * Extract the latest failure from some previous run for this cron workflow. This is used in
+   * combination with cron schedule. A workflow can be started with an optional cron schedule. If
+   * any previous workflow has failed, this function returns that failure. If no previous workflows
+   * have failed, an empty optional is returned.
+   *
+   * @return The last {@link Exception} that occurred in this workflow, if there has been one.
+   */
+  public static Optional<Exception> getLastFailure() {
+    return WorkflowInternal.getLastFailure()
+        .map(f -> FailureConverter.failureToException(f, DataConverter.getDefaultInstance()));
```
mfateev, line 1196: I would move this logic into WorkflowInternal. The idea is that Workflow is just a public layer over internal stuff without any business logic.

temporal-sdk/src/main/java/io/temporal/internal/testservice/StateMachines.java
Sushisource: Maybe I'm still stuck in Rust land but, I worry about the idea that we can only rely on our own consistency to ensure Optionals are non-null. The null check is cheap and not too noisy, and allows us to consider the method by itself without ensuring `data` is always well constructed.

So, I agree with the theory, but in practice I like to check because it decreases mental load. Thoughts?

mfateev: Java has nice mechanism called NullPointerException to report such bugs. So adding repetitive checks all other the code for null is not serving any purpose besides adding clutter. In this particular case if  you want to add the check do in the constructor of Data.

 ```this.lastFailure = Objects.requireNotNull(lastFailure);```

I would also change all Data fields that are not expected to be changed to `final`.

temporal-sdk/src/main/java/io/temporal/workflow/Workflow.java
```diff
@@ -1181,6 +1183,19 @@ public static Logger getLogger(String name) {
...
 
+  /**
+   * Extract the latest failure from some previous run for this cron workflow. This is used in
+   * combination with cron schedule. A workflow can be started with an optional cron schedule. If
+   * any previous workflow has failed, this function returns that failure. If no previous workflows
+   * have failed, an empty optional is returned.
+   *
+   * @return The last {@link Exception} that occurred in this workflow, if there has been one.
+   */
+  public static Optional<Exception> getLastFailure() {
```
mfateev, line 1194: Let's rename to the getPreviousRunFailure. It is not really clear what this failure is about just by the name.

temporal-sdk/src/main/java/io/temporal/workflow/Workflow.java
```diff
@@ -1181,6 +1183,19 @@ public static Logger getLogger(String name) {
     return WorkflowInternal.getLastCompletionResult(resultClass, resultClass);
   }
 
+  /**
+   * Extract the latest failure from some previous run for this cron workflow. This is used in
```
mfateev, line 1187: This doesn't apply to cron workflows only. It applies to any workflow retries (at least this is my understanding that is worth testing). So I would say something like:

Extract the latest failure from some previous run for this workflow. This run can be created as a retry of the previous failed run or as a next cron invocation for cron workflows.

mfateev:
Earlier you asked about time skipping logic. It is encapsulated in the [SelfAdvancingTime](https://github.com/temporalio/sdk-java/blob/30420212a26283cf7cbe6339512058a3186074e8/temporal-sdk/src/main/java/io/temporal/internal/testservice/SelfAdvancingTimer.java#L50).

temporal-sdk/src/main/java/io/temporal/internal/testservice/StateMachines.java
Sushisource: Ah, that's a nice way to handle it. Thanks.

temporal-sdk/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
Sushisource: This would require us to set it to empty in two branching paths to ensure it's non-null, thus more verbose. Any particular reason not to leave it this way?

temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflow.java
```diff
@@ -101,8 +102,12 @@ public void start(HistoryEvent event, ReplayWorkflowContext context) {
         startEvent.hasLastCompletionResult()
             ? Optional.of(startEvent.getLastCompletionResult())
             : Optional.empty();
+    Optional<Failure> lastFail =
```
vitarb, line 105: nit. Fail/Failure

temporal-sdk/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -1193,7 +1197,13 @@ private void processFailWorkflowExecution(
     }
 
     if (!Strings.isNullOrEmpty(data.cronSchedule)) {
-      startNewCronRun(ctx, workflowTaskCompletedId, identity, data, data.lastCompletionResult);
+      startNewCronRun(
+          ctx,
+          workflowTaskCompletedId,
+          identity,
+          data,
+          data.lastCompletionResult,
```
vitarb, line 1205: nit. I know you haven't changed this, but do we really need to pass both data and data.lastCompletionResult separately?

vitarb (approved):
Minor comments, otherwise LGTM.

temporal-sdk/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
Sushisource: We do - at a different site the last completion result comes from another source.

Issue #258, opened 2020-11-18
Add opentracing context propagator


mfateev:
https://github.com/uber/cadence-java-client/pull/478

bowsersenior:
Since opentracing is now part of [opentelemetry](https://opentelemetry.io), it would be great if you could check that the new context propagator will work with opentelemetry as well.

Spikhalskiy:
This can be closed. It's addressed in https://github.com/temporalio/sdk-java/pull/418

PR #263, opened 2020-11-21
Opentracing Support

sdwr98:
Addresses #258 - bringing over the ContextPropagator improvements and adding a default OpenTracing implementation of context propagation


temporal-sdk/src/main/java/io/temporal/common/context/OpenTracingContextPropagator.java
```diff
@@ -0,0 +1,207 @@
...
+      HashMapTextMap contextTextMap = new HashMapTextMap(contextAsMap);
+      setCurrentOpenTracingSpanContext(
+          currentTracer.extract(Format.Builtin.TEXT_MAP, contextTextMap));
+    }
+  }
+
+  @Override
+  public void setUp() {
+    Tracer openTracingTracer = GlobalTracer.get();
+    Tracer.SpanBuilder builder = openTracingTracer.buildSpan("cadence.workflow");
```
mfateev, line 117: I don't think it is going to work. The problem is that during replay a new span is going to be created. 
Unfortunately, there is no way to create a span passing deterministically generated ID to it. 

Another option would be using [SideEffect](https://github.com/temporalio/sdk-java/blob/4bb23c3a660de6a2630901d0dc6c049d9742ebe1/temporal-sdk/src/main/java/io/temporal/workflow/Workflow.java#L920) to store the span, but that implies additional MarkerEvent for every workflow.

temporal-sdk/src/main/java/io/temporal/common/context/ContextPropagator.java
```diff
@@ -126,4 +130,27 @@
 
   /** Sets the current context */
   void setCurrentContext(Object context);
+
```
mfateev, line 133: Looking at this and the implementation I'm getting more inclined to just using interceptors for this integration and getting rid of the context propagator interface completely as it has very limited use.

sdwr98: I’d be happy to look into that - are there any docs or examples of interceptors anywhere? I didn’t see a way of propagating information from worker/client to the server and back using interceptors.

mfateev: I think you are right that there is something missing in the interceptors to support this use case. Let me think about this.

temporal-sdk/src/main/java/io/temporal/common/context/OpenTracingContextPropagator.java
sdwr98: Does it really matter if we start multiple spans if we only finish one?  I don't think spans get reported until you finish() them.

temporal-sdk/src/main/java/io/temporal/common/context/OpenTracingContextPropagator.java
```diff
@@ -0,0 +1,207 @@
...
+    }
+  }
+
+  @Override
+  public String getName() {
+    return "OpenTracing";
+  }
+
+  @Override
+  public Map<String, Payload> serializeContext(Object context) {
```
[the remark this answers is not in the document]
sdwr98: That's a good point - I will take a look at the go propagator and work on ensuring cross-platform.

mfateev:
The newly added [WorkerInterceptor](https://github.com/temporalio/sdk-java/blob/master/temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkerInterceptor.java) should be enough to implement open tracing support.

I'm going to close this PR. I would love to review a contribution that relies on the interceptor.

PR #270, opened 2020-12-09
Adding binary checksum to WorkflowClientOptions

vitarb:
Simply set binary checksum in WorkflowClientOptions
```
WorkflowClient client =
        WorkflowClient.newInstance(
            service, WorkflowClientOptions.newBuilder().setBinaryChecksum("12345").build());
```

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/client/WorkflowClientOptions.java
+++ b/temporal-sdk/src/main/java/io/temporal/client/WorkflowClientOptions.java
@@ -33,7 +33,7 @@ public final class WorkflowClientOptions {
 
   private static final WorkflowClientOptions DEFAULT_INSTANCE;
   private static final String DEFAULT_NAMESPACE = "default";
-  private static final String DEFAULT_BINARY_CHECKSUM = "none";
+  private static final String DEFAULT_BINARY_CHECKSUM = "";
 
   static {
     DEFAULT_INSTANCE = newBuilder().build();
--- a/temporal-sdk/src/main/java/io/temporal/client/WorkflowClientOptions.java
+++ b/temporal-sdk/src/main/java/io/temporal/client/WorkflowClientOptions.java
@@ -33,6 +33,7 @@ public final class WorkflowClientOptions {
 
   private static final WorkflowClientOptions DEFAULT_INSTANCE;
   private static final String DEFAULT_NAMESPACE = "default";
+  private static final String DEFAULT_BINARY_CHECKSUM = "none";
 
   static {
     DEFAULT_INSTANCE = newBuilder().build();
@@ -56,6 +57,7 @@ public final class WorkflowClientOptions {
     private DataConverter dataConverter;
     private WorkflowClientInterceptor[] interceptors;
     private String identity;
+    private String binaryChecksum;
     private List<ContextPropagator> contextPropagators;
     private QueryRejectCondition queryRejectCondition;
 
@@ -69,6 +71,7 @@ public final class WorkflowClientOptions {
       dataConverter = options.dataConverter;
       interceptors = options.interceptors;
       identity = options.identity;
+      binaryChecksum = options.binaryChecksum;
       contextPropagators = options.contextPropagators;
       queryRejectCondition = options.queryRejectCondition;
     }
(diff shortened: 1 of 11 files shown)


temporal-sdk/src/main/java/io/temporal/client/WorkflowClientOptions.java
```diff
@@ -33,6 +33,7 @@
 
   private static final WorkflowClientOptions DEFAULT_INSTANCE;
   private static final String DEFAULT_NAMESPACE = "default";
+  private static final String DEFAULT_BINARY_CHECKSUM = "none";
```
mfateev, line 36: Why not an empty string as it is now?

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
```diff
@@ -127,6 +128,7 @@ private Result failureToResult(PollWorkflowTaskQueueResponse workflowTask, Throw
           RespondWorkflowTaskCompletedRequest.newBuilder()
               .setTaskToken(workflowTask.getTaskToken())
               .setIdentity(options.getIdentity())
+              .setBinaryChecksum(options.getBinaryChecksum())
```
mfateev, line 131: I would make the default null and not call this method if it is not set.

temporal-sdk/src/test/java/io/temporal/worker/StickyWorkerTest.java
```diff
@@ -489,13 +490,18 @@ public void workflowsCanBeQueriedAfterEviction() throws Exception {
...
     public TestEnvironmentWrapper(WorkerFactoryOptions options) {
       if (options == null) {
         options = WorkerFactoryOptions.newBuilder().build();
       }
       WorkflowClientOptions clientOptions =
-          WorkflowClientOptions.newBuilder().setNamespace(NAMESPACE).setIdentity(identity).build();
+          WorkflowClientOptions.newBuilder()
+              .setNamespace(NAMESPACE)
+              .setIdentity(identity)
+              .setBinaryChecksum(binaryChecksum)
```
mfateev, line 503: Does test verify that it was indeed added to the history?

mfateev (changes requested):

Sushisource:
How do we expect to typically set the checksum? Is there another item for instructing users about that?

Otherwise pending Max's comments LGTM

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
vitarb: I did that initially, but didn't like the extra if statement and decided to add a default.
Should we have an empty string default instead?

temporal-sdk/src/main/java/io/temporal/client/WorkflowClientOptions.java
vitarb: Having some value could potentially be useful in case when someone just added their own binary checksum and wants to roll back.
I'm okay changing it to empty string if you prefer it.

vitarb:
> How do we expect to typically set the checksum (by that I mean, what ends up calling `setBinaryChecksum`)? Is there another item for instructing users about that?

Users may set it during worker startup and manage the version incrementing it with every new release of their client code.

Sushisource:
> > How do we expect to typically set the checksum (by that I mean, what ends up calling `setBinaryChecksum`)? Is there another item for instructing users about that?
> 
> Users may set it during worker startup and manage the version incrementing it with every new release of their client code.

Sounds good. It might be worth elaborating in the docstring when/why the user might care about this. Additionally it might make some sense to just call it `workerBinaryIdentifier` since nothing is really forcing them to use a checksum here. They could use any ID, really. We should probably make it clear that it should always change if the worker binary changes.

temporal-sdk/src/test/java/io/temporal/worker/StickyWorkerTest.java
vitarb: I've added a separate test for it in the WorkflowTest, this one just adds it because identity was there already.

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
vitarb: I've changed it to empty string by default, let me know if you think that's sufficient.

temporal-sdk/src/main/java/io/temporal/client/WorkflowClientOptions.java
vitarb: Done

vitarb:
>> It might be worth elaborating in the docstring when/why the user might care about this. 
I've added a comment in the code and also it should be something in our user documentation.

>> It might make some sense to just call it workerBinaryIdentifier since nothing is really forcing them to use a checksum here. 
This feature already exists in go sdk and on the server side, diverging from that naming will only create more confusion IMO.

Sushisource (approved):

mfateev (approved):

PR #276, opened 2020-12-15
Adding an option to disable activity pollers

vitarb:
This PR addressed this [feature request](https://github.com/temporalio/sdk-go/issues/284) adding an option to worker options that allows disabling activity poller.


temporal-sdk/src/test/java/io/temporal/worker/ActivityPollerDisabledTest.java
```diff
@@ -0,0 +1,86 @@
...
+    worker.registerWorkflowImplementationTypes(WorkflowImpl.class);
+    env.start();
+    Thread.sleep(1000);
+    Map<String, Long> threads =
+        Thread.getAllStackTraces().keySet().stream()
+            .map((t) -> t.getName().substring(0, Math.min(20, t.getName().length())))
+            .collect(groupingBy(Function.identity(), Collectors.counting()));
+    assertEquals(hostLocalThreadCount, (long) threads.get(workflowHostLocalPollerThreadNamePrefix));
+    assertEquals(workflowPollCount, (long) threads.get(workflowPollerThreadNamePrefix));
+    assertFalse(threads.containsKey(activityPollerThreadNamePrefix));
```
Sushisource, line 83: This is a pretty neat trick. Do we have any business-logic-related thing we can verify here other than there are no threads and the worker is null?

vitarb: It's the last assertion line:
```
    assertNull(worker.activityWorker);
```

Sushisource: Right - I meant is there something besides that we can assert, like something doesn't happen that otherwise would? Perhaps an integ test that verifies the worker doesn't attempt to perform any activities?

If you think it's low value don't bother. I was just curious if there's something we could do here besides "this is null"

Sushisource (approved):

temporal-sdk/src/main/java/io/temporal/worker/Worker.java
```diff
@@ -323,25 +327,38 @@ void start() {
       return;
     }
     workflowWorker.start();
-    activityWorker.start();
+    if (!options.isActivityPollerDisabled()) {
```
mfateev, line 330: I would change all this check to `activityWorker != null` which would make it more obvious.

temporal-sdk/src/main/java/io/temporal/worker/WorkerOptions.java
```diff
@@ -173,6 +177,17 @@ public Builder setActivityPollThreadCount(int activityPollThreadCount) {
       return this;
     }
 
+    /**
+     * If set to true disables activity poller on this worker. Can be useful if you want to handle
+     * workflow tasks only.
```
mfateev, line 182: To have workflow tasks only don't call worker.registerActivity. The reason to have this option is to support only local activities by this activity worker. 
Should we rename it to something more explicit? More like `setLocalActivityWorkerOnly`?

temporal-sdk/src/test/java/io/temporal/worker/ActivityPollerDisabledTest.java
mfateev: Another option is to schedule a local activity and after it completes schedule a normal activity and let it timeout with ScheduleToStart to show that no activity worker is present.

temporal-sdk/src/main/java/io/temporal/worker/Worker.java
vitarb: I wanted to use this option explicitly in order to be able to catch potential bugs more easily in case if activityWorker is null when it shouldn't be.

temporal-sdk/src/main/java/io/temporal/worker/WorkerOptions.java
vitarb: I don't have strong opinion about naming here as long as default false means normal worker, let me rename to `localActivityWorkerOnly` if you like it better.

temporal-sdk/src/test/java/io/temporal/worker/ActivityPollerDisabledTest.java
vitarb: Added two tests for local and normal activities.

Sushisource: :tada: Thanks!

temporal-sdk/src/main/java/io/temporal/worker/WorkerOptions.java
```diff
@@ -183,8 +183,8 @@ public Builder setActivityPollThreadCount(int activityPollThreadCount) {
      *
      * <p>Default is false.
      */
-    public Builder setActivityPollerDisabled(boolean activityPollerDisabled) {
-      this.activityPollerDisabled = activityPollerDisabled;
+    public Builder setLocalActivityWorkerOnly(boolean localActivityWorkerOnly) {
```
mfateev, line 186: Would you add a comment that explains what is the meaning of this flag?

mfateev (approved):

PR #277, opened 2020-12-18
Avoid polling new tasks when handler pool is full

vitarb:


temporal-sdk/src/main/java/io/temporal/internal/worker/ActivityPollTask.java
```diff
@@ -75,14 +75,6 @@ public PollActivityTaskQueueResponse poll() {
...
     }
 
-    if (taskQueueActivitiesPerSecond > 0) {
-      pollRequest.setTaskQueueMetadata(
-          TaskQueueMetadata.newBuilder()
-              .setMaxTasksPerSecond(
-                  DoubleValue.newBuilder().setValue(taskQueueActivitiesPerSecond).build())
-              .build());
-    }
-
```
mfateev, line 85: Why is this removed? Task queue level rate limiting is a very important feature

temporal-sdk/src/main/java/io/temporal/internal/worker/PollTaskExecutor.java
```diff
@@ -52,7 +53,7 @@
     taskExecutor =
         new ThreadPoolExecutor(
             0,
-            options.getTaskExecutorThreadPoolSize(),
+            options.getPollerOptions().getTaskExecutorThreadPoolSize(),
```
mfateev, line 56: What is the reason for this change?

temporal-sdk/src/main/java/io/temporal/internal/worker/PollerOptions.java
```diff
@@ -48,6 +48,8 @@ public static PollerOptions getDefaultInstance() {
 
     private int maximumPollRateIntervalMilliseconds = 1000;
 
+    private int taskExecutorThreadPoolSize;
```
mfateev, line 51: I don't think poller options is the right place for the executor thread pool size.

temporal-sdk/src/main/java/io/temporal/internal/worker/SingleWorkerOptions.java
```diff
@@ -41,7 +41,6 @@ public static Builder newBuilder(SingleWorkerOptions options) {
     private String identity;
     private String binaryChecksum;
     private DataConverter dataConverter;
-    private int taskExecutorThreadPoolSize = 100;
```
mfateev, line 44: I think it should stay here instead of poller options

temporal-sdk/src/main/java/io/temporal/internal/worker/ActivityPollTask.java
vitarb: There were two identical blocks, I've assumed that it was a mistake and removed one.

temporal-sdk/src/main/java/io/temporal/internal/worker/PollerOptions.java
vitarb: We need to access it in the poller, so it's either this or we should be passing more parameters.

temporal-sdk/src/main/java/io/temporal/internal/worker/SingleWorkerOptions.java
vitarb: Should we pass single worker options to the poller then?

vitarb:
Abandoned in favor of #278

PR #289, opened 2020-12-26
Dynamic workflow and activity implementation support

mfateev:
Added ability to implement workflows and activities dynamically. A single implementation of DynamicWorkflow can implement any number of workflow types. A single implementation of DynamicActivity can implement any number of activity types.

Also added DynamicSignalHandler and DynamicQueryHandler that allow processing any signals and queries using a single handler instance.

Solves https://github.com/temporalio/sdk-java/issues/245

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/activity/DynamicActivity.java
+++ b/temporal-sdk/src/main/java/io/temporal/activity/DynamicActivity.java
@@ -27,7 +27,8 @@ import io.temporal.common.converter.EncodedValues;
  * invocation that doesn't have an explicitly registered handler. Only one instance that implements
  * DynamicActivity per {@link io.temporal.worker.Worker} is allowed.
  *
- * <p>The DynamicActivity can be useful for integrations with existing libraries.
+ * <p>The DynamicActivity can be useful for integrations with existing libraries. For example it can
+ * be used to call some external HTTP API with each function exposed as a different activity type.
  *
  * <p>Use {@link Activity#getExecutionContext()} to query information about the activity type that
  * should be implemented dynamically.
--- a/temporal-sdk/src/main/java/io/temporal/activity/DynamicActivity.java
+++ b/temporal-sdk/src/main/java/io/temporal/activity/DynamicActivity.java
@@ -17,11 +17,23 @@
  *  permissions and limitations under the License.
  */
 
-package io.temporal.workflow;
+package io.temporal.activity;
 
 import io.temporal.common.converter.EncodedValues;
 
-/** */
+/**
+ * Use DynamicActivity to implement any number of activity types dynamically. When an activity
+ * implementation that extends DynamicActivity is registered it is called for any activity type
+ * invocation that doesn't have an explicitly registered handler. Only one instance that implements
+ * DynamicActivity per {@link io.temporal.worker.Worker} is allowed.
+ *
+ * <p>The DynamicActivity can be useful for integrations with existing libraries.
+ *
+ * <p>Use {@link Activity#getExecutionContext()} to query information about the activity type that
+ * should be implemented dynamically.
+ *
+ * @see io.temporal.workflow.DynamicWorkflow
+ */
 public interface DynamicActivity {
   Object execute(EncodedValues args);
 }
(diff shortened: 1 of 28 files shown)


temporal-sdk/src/main/java/io/temporal/activity/DynamicActivity.java
```diff
@@ -0,0 +1,39 @@
...
+
+import io.temporal.common.converter.EncodedValues;
+
+/**
+ * Use DynamicActivity to implement any number of activity types dynamically. When an activity
+ * implementation that extends DynamicActivity is registered it is called for any activity type
+ * invocation that doesn't have an explicitly registered handler. Only one instance that implements
+ * DynamicActivity per {@link io.temporal.worker.Worker} is allowed.
+ *
+ * <p>The DynamicActivity can be useful for integrations with existing libraries.
```
Sushisource, line 30: One sentence example might be useful

temporal-sdk/src/main/java/io/temporal/internal/sync/POJOActivityTaskHandler.java
```diff
@@ -83,11 +86,19 @@
...
+  private void registerActivityImplementation(
       Object activity, BiFunction<Method, Object, ActivityTaskExecutor> newTaskExecutor) {
     if (activity instanceof Class) {
       throw new IllegalArgumentException("Activity object instance expected, not the class");
     }
+    if (activity instanceof DynamicActivity) {
+      if (dynamicActivity != null) {
+        throw new IllegalStateException(
+            "An implementation of DynamicActivity is already registered with the worker");
+      }
```
Sushisource, line 98: Would be good to indicate workers can only have one `DynamicActivity` implementation in a docstring somewhere

temporal-sdk/src/main/java/io/temporal/internal/sync/POJOActivityTaskHandler.java
```diff
@@ -287,6 +301,96 @@ public Object execute(Object[] arguments) {
...
+  private class DynamicActivityImplementation implements ActivityTaskExecutor {
+
+    private final DynamicActivity activity;
+
+    public DynamicActivityImplementation(DynamicActivity activity) {
+      this.activity = activity;
+    }
+
+    @Override
+    public ActivityTaskHandler.Result execute(ActivityInfoImpl info, Scope metricsScope) {
```
Sushisource, line 313: Quite a lot of the code in here is quite similar to the above `POJOActivityImplementation` -- would be nice to dedupe as reasonable.

temporal-sdk/src/main/java/io/temporal/internal/sync/POJOWorkflowImplementationFactory.java
```diff
@@ -94,23 +97,33 @@
...
     addWorkflowImplementationFactory(unitTestingOptions, clazz, factory);
   }
 
+  @SuppressWarnings("unchecked")
   <R> void addWorkflowImplementationFactory(
       WorkflowImplementationOptions options, Class<R> clazz, Functions.Func<R> factory) {
+    if (DynamicWorkflow.class.isAssignableFrom(clazz)) {
+      if (dynamicWorkflowImplementationFactory != null) {
+        throw new IllegalStateException(
+            "An implementation of DynamicWorkflow or its factory is already registered with the worker");
```
Sushisource, line 122: Same as other comment - would be nice to expose this restriction to user in docstring.

temporal-sdk/src/main/java/io/temporal/workflow/DynamicWorkflow.java
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * Use DynamicWorkflow to implement any number of workflow types dynamically. When a workflow
+ * implementation type that extends DynamicWorkflow is registered, it is used to implement any
+ * workflow type that is not implicitly registered with the {@link io.temporal.worker.Worker}. Only
+ * one type that implements DynamicWorkflow per worker is allowed.
+ *
+ * <p>The main use case for DynamicWorkflow is an implementation of custom Domain Specific Languages
+ * (DSLs). A single implementation can implement a workflow type which definition is dynamically
+ * loaded from some external source.
+ *
```
Sushisource, line 33: Presumably it's necessary for the dynamic workflow to always be deterministic in the same way that normal workflows are, right? IE: Deterministic w.r.t. `EncodedValues`. Might be worth making reference to that here just so users don't think magically that requirement has gone away.

Sushisource (approved):
LGTM, mostly just suggestions for clarification

PR #291, opened 2020-12-30
Add support for MaxConcurrentActivityExecutionSize for async activities

vitarb:
This PR introduces a new method `useLocalManualCompletion` in the activity context, which marks activity execution as async and returns `ActivityCompletionClient`. By doing so we expect user to manually complete activity execution on the same machine.
Also we do pass a completion handle down to the `ActivityCompletionClientImpl` and execute it only when user triggers the completion. This means that semaphore limit set by `setMaxConcurrentActivityExecutionSize` for activity polling will be respected for async activities.
Behavior is verified by `LocalAsyncCompletionWorkflowTest`.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/ActivityCompletionClientImpl.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/ActivityCompletionClientImpl.java
@@ -23,58 +23,98 @@ import io.temporal.api.common.v1.WorkflowExecution;
 import io.temporal.client.ActivityCompletionClient;
 import io.temporal.client.ActivityCompletionException;
 import io.temporal.internal.external.ManualActivityCompletionClientFactory;
+import io.temporal.workflow.Functions;
 import java.util.Optional;
 
 class ActivityCompletionClientImpl implements ActivityCompletionClient {
 
   private final ManualActivityCompletionClientFactory factory;
+  private final Functions.Proc completionHandle;
 
   public ActivityCompletionClientImpl(
-      ManualActivityCompletionClientFactory manualActivityCompletionClientFactory) {
+      ManualActivityCompletionClientFactory manualActivityCompletionClientFactory,
+      Functions.Proc completionHandle) {
     this.factory = manualActivityCompletionClientFactory;
+    this.completionHandle = completionHandle;
   }
 
   @Override
   public <R> void complete(byte[] taskToken, R result) {
-    factory.getClient(taskToken).complete(result);
+    try {
+      factory.getClient(taskToken).complete(result);
+    } finally {
+      completionHandle.apply();
+    }
   }
 
   @Override
   public <R> void complete(String workflowId, Optional<String> runId, String activityId, R result) {
-    factory.getClient(toExecution(workflowId, runId), activityId).complete(result);
+    try {
+      factory.getClient(toExecution(workflowId, runId), activityId).complete(result);
+    } finally {
+      completionHandle.apply();
+    }
   }
 
   @Override
   public void completeExceptionally(byte[] taskToken, Exception result) {
-    factory.getClient(taskToken).fail(result);
+    try {
+      factory.getClient(taskToken).fail(result);
+    } finally {
+      completionHandle.apply();
+    }
   }
 
   @Override
   public void completeExceptionally(
       String workflowId, Optional<String> runId, String activityId, Exception result) {
-    factory.getClient(toExecution(workflowId, runId), activityId).fail(result);
+    try {
+      factory.getClient(toExecution(workflowId, runId), activityId).fail(result);
+    } finally {
+      completionHandle.apply();
+    }
   }
 
   @Override
   public <V> void reportCancellation(byte[] taskToken, V details) {
-    factory.getClient(taskToken).reportCancellation(details);
+    try {
+      factory.getClient(taskToken).reportCancellation(details);
+    } finally {
+      completionHandle.apply();
+    }
   }
 
   @Override
   public <V> void reportCancellation(
       String workflowId, Optional<String> runId, String activityId, V details) {
-    factory.getClient(toExecution(workflowId, runId), activityId).reportCancellation(details);
+    try {
+      factory.getClient(toExecution(workflowId, runId), activityId).reportCancellation(details);
+    } finally {
+      completionHandle.apply();
+    }
   }
 
   @Override
   public <V> void heartbeat(byte[] taskToken, V details) throws ActivityCompletionException {
-    factory.getClient(taskToken).recordHeartbeat(details);
+    try {
+      factory.getClient(taskToken).recordHeartbeat(details);
+    } finally {
+      completionHandle.apply();
+    }
   }
 
   @Override
   public <V> void heartbeat(String workflowId, Optional<String> runId, String activityId, V details)
       throws ActivityCompletionException {
-    factory.getClient(toExecution(workflowId, runId), activityId).recordHeartbeat(details);
+    try {
+      factory.getClient(toExecution(workflowId, runId), activityId).recordHeartbeat(details);
+    } finally {
+      completionHandle.apply();
+    }
+  }
+
+  Functions.Proc getCompletionHandle() {
+    return completionHandle;
   }
 
   private static WorkflowExecution toExecution(String workflowId, Optional<String> runId) {
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/ActivityCompletionClientImpl.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/ActivityCompletionClientImpl.java
@@ -24,7 +24,6 @@ import io.temporal.client.ActivityCompletionClient;
 import io.temporal.client.ActivityCompletionException;
 import io.temporal.internal.external.ManualActivityCompletionClientFactory;
 import io.temporal.workflow.Functions;
-
 import java.util.Optional;
 
 class ActivityCompletionClientImpl implements ActivityCompletionClient {
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/ActivityCompletionClientImpl.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/ActivityCompletionClientImpl.java
@@ -24,6 +24,7 @@ import io.temporal.client.ActivityCompletionClient;
 import io.temporal.client.ActivityCompletionException;
 import io.temporal.internal.external.ManualActivityCompletionClientFactory;
 import io.temporal.workflow.Functions;
+
 import java.util.Optional;
 
 class ActivityCompletionClientImpl implements ActivityCompletionClient {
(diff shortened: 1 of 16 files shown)


temporal-sdk/src/main/java/io/temporal/internal/sync/ActivityCompletionClientImpl.java
```diff
@@ -23,58 +23,98 @@
...
+    }
   }
 
   @Override
   public <V> void heartbeat(byte[] taskToken, V details) throws ActivityCompletionException {
-    factory.getClient(taskToken).recordHeartbeat(details);
+    try {
+      factory.getClient(taskToken).recordHeartbeat(details);
+    } finally {
+      completionHandle.apply();
```
mfateev: I don't think you want to complete on a heartbeat.

temporal-sdk/src/main/java/io/temporal/internal/sync/ActivityCompletionClientImpl.java
```diff
@@ -23,58 +23,98 @@
...
   }
 
   @Override
   public <V> void heartbeat(String workflowId, Optional<String> runId, String activityId, V details)
       throws ActivityCompletionException {
-    factory.getClient(toExecution(workflowId, runId), activityId).recordHeartbeat(details);
+    try {
+      factory.getClient(toExecution(workflowId, runId), activityId).recordHeartbeat(details);
+    } finally {
+      completionHandle.apply();
```
mfateev: ditto

temporal-sdk/src/main/java/io/temporal/internal/sync/ActivityExecutionContextImpl.java
```diff
@@ -228,6 +239,17 @@ public boolean isDoNotCompleteOnReturn() {
...
 
+  public boolean isUseLocalManualCompletion() {
+    return useLocalManualCompletion;
+  }
+
+  @Override
+  public ActivityCompletionClient useLocalManualCompletion() {
+    doNotCompleteOnReturn();
+    useLocalManualCompletion = true;
+    return new ActivityCompletionClientImpl(manualCompletionClientFactory, completionHandle);
```
mfateev, line 250: Make this method and other methods thread safe. It looks like doNotComplete on return should be also made thread safe.

temporal-sdk/src/main/java/io/temporal/internal/sync/LocalActivityExecutionContextImpl.java
```diff
@@ -72,6 +73,18 @@ public boolean isDoNotCompleteOnReturn() {
         "isDoNotCompleteOnReturn is not supported for local activities");
   }
 
+  @Override
+  public boolean isUseLocalManualCompletion() {
+    throw new UnsupportedOperationException(
+        "isUseLocalManualCompletion is not supported for local activities");
```
mfateev, line 79: Why? it makes perfect sense for local activities implemented asynchronously.

vitarb: We need to rethink our behavior for limiting local activity throughput as currently semaphore and setMaxConcurrentActivityExecutionSize is only applied to normal activities. This would be a bigger change, and I suggest keep it out of scope for this PR.

temporal-sdk/src/main/java/io/temporal/internal/sync/ActivityExecutionContextImpl.java
vitarb: Done

temporal-sdk/src/main/java/io/temporal/internal/sync/ActivityExecutionContextImpl.java
```diff
@@ -72,6 +75,9 @@
   private final Lock lock = new ReentrantLock();
   private ScheduledFuture future;
   private ActivityCompletionException lastException;
+  private final ManualActivityCompletionClientFactoryImpl manualCompletionClientFactory;
```
mfateev, line 78: Does it need to be the Impl or the ManualActivityCompletionClientFactory interface would be enough?

temporal-sdk/src/main/java/io/temporal/internal/sync/ActivityExecutionContextImpl.java
```diff
@@ -220,14 +231,35 @@ private void sendHeartbeatRequest(Object details) {
...
+    try {
+      doNotCompleteOnReturn = true;
+    } finally {
+      lock.unlock();
+    }
   }
 
   @Override
   public boolean isDoNotCompleteOnReturn() {
     return doNotCompleteOnReturn;
```
mfateev, line 244: Needs locking as well

temporal-sdk/src/main/java/io/temporal/internal/sync/ActivityExecutionContextImpl.java
```diff
@@ -220,14 +231,35 @@ private void sendHeartbeatRequest(Object details) {
...
   }
 
+  public boolean isUseLocalManualCompletion() {
+    return useLocalManualCompletion;
```
mfateev, line 248: Needs locking as well

temporal-sdk/src/main/java/io/temporal/internal/sync/LocalActivityExecutionContextImpl.java
mfateev: Let's keep it out of the scope of this PR. 

I believe that async local activities can be an important performance optimization in some cases.

temporal-sdk/src/main/java/io/temporal/internal/worker/ActivityWorker.java
```diff
@@ -210,17 +210,20 @@ public void handle(ActivityTask task) throws Exception {
...
         }
       } finally {
         MDC.remove(LoggerTag.ACTIVITY_ID);
         MDC.remove(LoggerTag.ACTIVITY_TYPE);
         MDC.remove(LoggerTag.WORKFLOW_ID);
         MDC.remove(LoggerTag.RUN_ID);
-        task.getCompletionHandle().apply();
+        // Apply completion handle if task has been completed synchronously or is async and manual
+        // completion hasn't been requested.
+        if (response != null && !response.isManualCompletion()) {
```
mfateev, line 224: Does response != null check mean that if activity throws an exception the counter is not decremented?

temporal-sdk/src/test/java/io/temporal/workflow/LocalAsyncCompletionWorkflowTest.java
```diff
@@ -0,0 +1,145 @@
...
+
+    @ActivityMethod
+    int execute(int value);
+  }
+
+  public static class AsyncActivityWithManualCompletion implements TestActivity {
+
+    @Override
+    public int execute(int value) {
+      int concurrentActivities = concurrentActivitiesCount.incrementAndGet();
```
mfateev, line 107: This doesn't need to be static. Move to the AsyncActivityWithManualCompletion field.

temporal-sdk/src/test/java/io/temporal/workflow/LocalAsyncCompletionWorkflowTest.java
```diff
@@ -0,0 +1,145 @@
...
+      ActivityExecutionContext context = Activity.getExecutionContext();
+      context.heartbeat(value);
+      ActivityCompletionClient completionClient = context.useLocalManualCompletion();
+      try {
+        Thread.sleep(500);
+        completionClient.complete(context.getTaskToken(), value * 2);
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
+        e.printStackTrace();
+        completionClient.completeExceptionally(context.getTaskToken(), e);
```
mfateev, line 122: I would move completion to a separate thread just to increase code coverage. See [HelloAsyncActivityCompletion](https://github.com/temporalio/samples-java/blob/8218f4114e52417f8d04175b67027ff0af4fb73c/src/main/java/io/temporal/samples/hello/HelloAsyncActivityCompletion.java#L99).

temporal-sdk/src/main/java/io/temporal/internal/sync/ActivityExecutionContextImpl.java
vitarb: Interface is enough.

temporal-sdk/src/main/java/io/temporal/internal/sync/ActivityExecutionContextImpl.java
vitarb: Added.

temporal-sdk/src/main/java/io/temporal/internal/sync/ActivityExecutionContextImpl.java
vitarb: Added.

temporal-sdk/src/test/java/io/temporal/workflow/LocalAsyncCompletionWorkflowTest.java
vitarb: Done

temporal-sdk/src/test/java/io/temporal/workflow/LocalAsyncCompletionWorkflowTest.java
vitarb: Done

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
```diff
@@ -220,7 +220,8 @@ public Throwable wrapFailure(Task task, Throwable failure) {
       metricsScope.counter(MetricsType.LOCAL_ACTIVITY_TOTAL_COUNTER).inc(1);
 
       Stopwatch sw = metricsScope.timer(MetricsType.LOCAL_ACTIVITY_EXECUTION_LATENCY).start();
-      ActivityTaskHandler.Result result = handler.handle(activityTask.build(), metricsScope, true);
+      ActivityTaskHandler.Result result =
+          handler.handle(new ActivityTask(activityTask.build(), () -> {}), metricsScope, true);
```
Sushisource, line 224: nit: Might be worth having a constructor overload for empty completion handles

temporal-sdk/src/test/java/io/temporal/internal/statemachines/LocalActivityStateMachineTest.java
```diff
@@ -172,7 +172,8 @@ protected void buildWorkflow(AsyncWorkflowBuilder<Void> builder) {
               RespondActivityTaskCompletedRequest.newBuilder().setResult(result2).build(),
               null,
               null,
-              null);
+              null,
```
Sushisource, line 175: nit: Constructor overload might be nice for these `Result` types as well, looks like they're frequently constructed w/ a bunch of empty params.

Sushisource (approved):

temporal-sdk/src/main/java/io/temporal/internal/worker/ActivityWorker.java
vitarb: No, it should be null only if activity is async.

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
vitarb: I don't think this warrants a separate constructor. Would prefer to keep as is.

mfateev (approved):

PR #292, opened 2021-01-06
Move test code into testing module

vitarb:
Moving all test code, including test rules and testservice into a test module.

This change also introduces recently developed internal TestWorkflowRule to the public, which can now be used to write unit tests for workflows more easily. See documentation of the TestWorkflowRule for details.

Users are expected to take a test dependency on our new testing module like:
```
    testImplementation group: 'io.temporal', name: 'temporal-testing', version: 'X.Y.Z'
```


temporal-testing/build.gradle
```diff
@@ -0,0 +1,187 @@
...
+    targetCompatibility = JavaVersion.VERSION_1_8
+    withJavadocJar()
+    withSourcesJar()
+}
+
+dependencies {
+    errorproneJavac('com.google.errorprone:javac:9+181-r4173-1')
+    errorprone('com.google.errorprone:error_prone_core:2.4.0')
+
+    api project(':temporal-sdk')
```
mfateev, line 41: I believe most of the other dependencies are already part of temporal-sdk. Can we remove them from here?

temporal-testing/build.gradle
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    implementation group: 'ch.qos.logback', name: 'logback-classic', version: '1.2.3'
+    implementation group: 'com.googlecode.junit-toolbox', name: 'junit-toolbox', version: '2.4'
+    implementation group: 'junit', name: 'junit', version: '4.13.1'
+}
+
+configurations.all {
+    // Conflicts with logback
+    exclude group: 'org.slf4j', module: 'slf4j-simple'
```
mfateev, line 60: I don't think the testing framework needs a concrete slf4j implementation.

temporal-testing/build.gradle
```diff
@@ -0,0 +1,187 @@
...
+
+license {
+    header rootProject.file('license-header.txt')
+    exclude '**/*.puml'
+}
+
+task initSubmodules(type: Exec) {
+    description = 'Initializes submodules'
+    commandLine 'git', 'submodule', 'init'
+}
```
mfateev, line 71: testing doesn't use submodules

temporal-testing/build.gradle
vitarb: We need errorprone here as we are using it directly in that build file. We can probably configure it in the root project instead, but that would take some time.

temporal-testing/build.gradle
vitarb: Cleaned up.

temporal-testing/build.gradle
vitarb: Cleaned up.

temporal-sdk/src/main/java/io/temporal/internal/statemachines/ActivityStateMachine.puml
```diff
@@ -50,5 +50,5 @@ COMPLETED --> [*]
 FAILED --> [*]
 TIMED_OUT --> [*]
 CANCELED --> [*]
-center footer Copyright (C) 2020 Temporal Technologies, Inc. All Rights Reserved.
+center footer Copyright (C) 2021 Temporal Technologies, Inc. All Rights Reserved.
```
mfateev, line 53: Keep 2020 year: https://stackoverflow.com/a/2391555/1664318

mfateev (approved):

PR #293, opened 2021-01-06
Enforce formatting on buildkite

vitarb:


docker/buildkite/docker-compose.yaml
```diff
@@ -57,7 +57,13 @@ services:
...
       context: ../../
       dockerfile: ./docker/buildkite/Dockerfile
-    command: "./gradlew --no-daemon test"
+    command: >
+      bash -c "./gradlew --no-daemon test
+      if [ ! -z $(git status --porcelain) ]; then 
+        echo 'Some files were improperly formatted. Please run build locally and check in all changes.'
+        git status;
+        exit 1;
+      fi"
```
Sushisource, line 66: I'd break this out into it's own little file - I find reading these inlined shell scripts can be a bit of a challenge

vitarb: Good call. Done.

Sushisource (approved):

PR #296, opened 2021-01-08
Adding debug mode

vitarb:
We've recently added deadlock detector, which fails workflow tasks that run longer than 1 second.
This is great for production use, but it can be very annoying when debugging your application or tests.
Added environment variable `TEMPORAL_DEBUG_MODE` which can be set to true during debugging.
Current implementation simply disables deadlock detector, in future we might also adjust some timeouts automatically.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/DeterministicRunner.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/DeterministicRunner.java
@@ -32,7 +32,13 @@ import java.util.function.Supplier;
  */
 interface DeterministicRunner {
 
+  boolean debugMode = Boolean.parseBoolean(System.getenv("TEMPORAL_DEBUG_MODE"));
+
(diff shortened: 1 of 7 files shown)


temporal-sdk/src/main/java/io/temporal/internal/sync/DeterministicRunner.java
```diff
@@ -32,8 +32,14 @@
  */
 interface DeterministicRunner {
 
+  boolean debugMode = Boolean.parseBoolean(System.getenv("TEMPORAL_DEBUG_MODE"));
```
mfateev, line 35: I would make it static. Getting the variable and parsing it on every workflow sounds excessive. But making it static would require some synchronization.

vitarb: It's defined in the interface, so it's final static by default.

mfateev (approved):

PR #304, opened 2021-01-12
Don't swallow replay workflow exceptions

### LINKED ISSUE #301: Worker.replayWorkflowExecution does not report replay failures
## Overview & Expected Behavior

This issue is a "follow-up" for issue #300. With some manual tweaks described in that ticket, I got a workflow history JSON that Java SDK was able to parse, and the test below succeeded. I expected the test to fail if I made any significant or non-backward-compatible changes to the workflow.

```java
import io.temporal.testing.WorkflowReplayer;
import org.junit.jupiter.api.Test;

public class SubmissionWorkflowReplayTest {

    @Test
    public void replayTest() throws Exception {
        WorkflowReplayer.replayWorkflowExecutionFromResource(
            "submission_workflow.json",
            SubmissionWorkflowImpl.class
        );
    }
}
```

## Actual Behavior

The replay test with Java SDK never fails given it can successfully parse the workflow history, regardless of what changes are made to the workflow or history JSON.

This seems to be caused by the [`Worker.replayWorkflowExecution()`](https://github.com/temporalio/sdk-java/blob/24a55be08d8cfaf0b8bb2bb49aab38e8344ce08c/temporal-sdk/src/main/java/io/temporal/worker/Worker.java#L386-L397) method that swallows all exceptions, including those that should signal that the replay has failed.

```java
public void replayWorkflowExecution(WorkflowExecutionHistory history) throws Exception {
  try {
    workflowWorker.queryWorkflowExecution(
        history,
        WorkflowClient.QUERY_TYPE_REPLAY_ONLY,
        String.class,
        String.class,
        new Object[] {});
  } catch (Exception e) {

  }
}
```

  1. Create any workflow and execute it on a local Temporal server.
  2. Export workflow execution history as JSON.
  3. Tweak it so Java SDK could read it (required until #300 is fixed).
  4. Run replay test and check that it succeeds.
  5. Make any breaking change to the workflow logic.
  6. Run the test again.
  7. Observe that it still succeeds.

## Specifications

  - Version: Temporal Java SDK 1.0.4
  - Platform: macOS

GreyTeardrop:
## Overview

Allow `Worker.replayWorkflowExecution()` to throw an exception if the workflow history replay fails due to event mismatch.

Also includes a minimal set of changes to `testAsyncActivityRetryHistory.json` and `testChildWorkflowRetryHistory.json` so `WorkflowTest.testAsyncActivityRetryReplay` and `WorkflowTest.testChildWorkflowRetryReplay` to pass - seems like the replays used to fail before due to some expected history changes but tests still succeeded due to exception swallow block.

Fixes #301

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/worker/Worker.java
+++ b/temporal-sdk/src/main/java/io/temporal/worker/Worker.java
@@ -384,16 +384,12 @@ public final class Worker implements Suspendable {
    * @throws Exception if replay failed for any reason
    */
   public void replayWorkflowExecution(WorkflowExecutionHistory history) throws Exception {
-    try {
-      workflowWorker.queryWorkflowExecution(
-          history,
-          WorkflowClient.QUERY_TYPE_REPLAY_ONLY,
-          String.class,
-          String.class,
-          new Object[] {});
-    } catch (Exception e) {
-
-    }
+    workflowWorker.queryWorkflowExecution(
+        history,
+        WorkflowClient.QUERY_TYPE_REPLAY_ONLY,
+        String.class,
+        String.class,
+        new Object[] {});
   }
 
   /**
--- a/temporal-sdk/src/test/java/io/temporal/workflow/WorkflowTest.java
+++ b/temporal-sdk/src/test/java/io/temporal/workflow/WorkflowTest.java
@@ -3575,6 +3575,18 @@ public class WorkflowTest {
     }
   }
 
+  public static class AlteredTestChildWorkflowRetryWorkflow
+      extends TestChildWorkflowAsyncRetryWorkflow {
+
+    public AlteredTestChildWorkflowRetryWorkflow() {}
+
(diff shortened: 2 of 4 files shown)


temporal-sdk/src/main/java/io/temporal/worker/Worker.java
```diff
@@ -384,16 +384,12 @@ public String toString() {
...
-    try {
-      workflowWorker.queryWorkflowExecution(
-          history,
-          WorkflowClient.QUERY_TYPE_REPLAY_ONLY,
-          String.class,
-          String.class,
-          new Object[] {});
-    } catch (Exception e) {
-
-    }
```
mfateev, line 396: Wow, not sure how a such ignore all exceptions cath ended up in the code :(.

mfateev (approved):
I really appreciate the contribution!

PR #307, opened 2021-01-14
Add a test for completion of async activity with error

vitarb:
Attempting to repro #308

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/common/converter/EncodedValues.java
+++ b/temporal-sdk/src/main/java/io/temporal/common/converter/EncodedValues.java
@@ -37,7 +37,6 @@ public final class EncodedValues implements Values {
 
   public EncodedValues(Object... values) {
     this.values = values;
-    this.converter = DefaultDataConverter.getDefaultInstance();
     this.payloads = null;
   }
 
--- a/temporal-sdk/src/main/java/io/temporal/common/converter/EncodedValues.java
+++ b/temporal-sdk/src/main/java/io/temporal/common/converter/EncodedValues.java
@@ -37,6 +37,7 @@ public final class EncodedValues implements Values {
 
   public EncodedValues(Object... values) {
     this.values = values;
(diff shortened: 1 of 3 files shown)


temporal-sdk/src/main/java/io/temporal/common/converter/EncodedValues.java
```diff
@@ -37,6 +37,7 @@ public EncodedValues(Optional<Payloads> payloads, DataConverter converter) {
 
   public EncodedValues(Object... values) {
     this.values = values;
+    this.converter = DefaultDataConverter.getDefaultInstance();
```
mfateev, line 40: This is not going to work as it is going to ignore custom DataConverters set through options.

mfateev (changes requested):

temporal-sdk/src/main/java/io/temporal/common/converter/EncodedValues.java
vitarb: Yes, I didn't see an option of setting it in the ManualCompletionClient and thought that this would be a better tradeoff than changing exceptionToFailure method signature.
Changed to the option that would work with custom converters too.

temporal-sdk/src/test/java/io/temporal/workflow/AsyncActivityCompleteWithErrorTest.java
```diff
@@ -0,0 +1,111 @@
...
+
+  public static class TestWorkflowImpl implements TestWorkflow {
+
+    @Override
+    public String execute(String taskQueue) {
+      TestActivity activity =
+          Workflow.newActivityStub(
+              TestActivity.class,
+              ActivityOptions.newBuilder()
+                  .setScheduleToStartTimeout(Duration.ofSeconds(1))
```
mfateev, line 60: No need to have scheduleToStartTimeout.

mfateev (approved):

PR #311, opened 2021-01-18
Interceptor refactoring to support headers

mfateev:
Extend the interceptor framework to support use cases that require context propagation. For example, opentrace integration should be implementable through an interceptor.


temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkerInterceptor.java
```diff
@@ -20,10 +20,12 @@
 package io.temporal.common.interceptors;
 
 /**
- * Intercepts workflow execution.
+ * Intercepts workflow and activity executions.
  *
  * <p>TODO(maxim): JavaDoc with sample
  */
-public interface WorkflowInterceptor {
+public interface WorkerInterceptor {
```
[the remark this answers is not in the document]
mfateev: Interceptors are a new feature that I believe is not being widely used.

temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkflowOutboundCallsInterceptor.java
```diff
@@ -50,12 +51,180 @@
  */
 public interface WorkflowOutboundCallsInterceptor {
 
-  final class WorkflowResult<R> {
+  final class ActivityInput<R> {
+    private final String activityName;
+    private final Class<R> resultClass;
+    private final Type resultType;
+    private final Object[] args;
```
Sushisource, line 58: Rather than using `Object` for these args/payloads it might be nice to introduce some kind of `Serializable` interface the DataConverter could use, pushing some type safety a little higher up.

That said I looked a bit and that'd probably be a huge change so feel free to ignore.

temporal-sdk/src/main/java/io/temporal/internal/sync/QueryDispatcher.java
```diff
@@ -0,0 +1,111 @@
...
+      } else {
+        throw new IllegalStateException("Unknown query type: " + queryName);
+      }
+    } else {
+      Object[] args =
+          DataConverter.arrayFromPayloads(
+              converter, input, handler.getArgTypes(), handler.getGenericArgTypes());
+      result =
+          inboundCallsInterceptor
+              .handleQuery(new WorkflowInboundCallsInterceptor.QueryInput(queryName, args))
```
Sushisource, line 93: nit: Call to `handleQuery` could be lifted out of if branches which would create the `QueryInput`

temporal-sdk/src/main/java/io/temporal/internal/sync/SignalDispatcher.java
```diff
@@ -0,0 +1,162 @@
...
+  private WorkflowInboundCallsInterceptor inboundCallsInterceptor;
+  private final DataConverter converter;
+
+  private final Map<String, WorkflowOutboundCallsInterceptor.SignalRegistrationRequest>
+      signalCallbacks = new HashMap<>();
+
+  private DynamicSignalHandler dynamicSignalHandler;
+
+  /** Buffers signals which don't have a registered listener. */
+  private final Queue<SignalData> signalBuffer = new ArrayDeque<>();
```
Sushisource, line 77: Does this need a size limit?

Sushisource (approved):
Just a few minor things

temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkflowOutboundCallsInterceptor.java
mfateev: Such interface already exists. See [Values](https://github.com/temporalio/sdk-java/blob/master/temporal-sdk/src/main/java/io/temporal/common/converter/Values.java). The reason we pass objects through the interceptor is to give it chance to look at them. As the interceptor doesn't know the types of arguments, passing through arguments as Values is useless.

temporal-sdk/src/main/java/io/temporal/internal/sync/SignalDispatcher.java
mfateev: The history sent by the service has a hard limit on a number of signals it can contain.

temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkflowInboundCallsInterceptor.java
```diff
@@ -19,11 +19,99 @@
...
  * Intercepts calls to the workflow execution. Executes under workflow context. So all the
  * restrictions on the workflow code should be obeyed.
  */
 public interface WorkflowInboundCallsInterceptor {
+
+  final class WorkflowInput {
+    private final Map<String, Payload> header;
+    private final Object[] arguments;
+
+    public WorkflowInput(Map<String, Payload> header, Object[] arguments) {
```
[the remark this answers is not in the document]
mfateev: Good point. Introduced Header class.

PR #320, opened 2021-01-25
Add helper methods for setting cloud sslContext

vitarb:
Previously users were required to set verbose configuration for the SSL context in order to communicate with temporal backend in the cloud, e.g.
```
SslContext sslContext =
    SslContextBuilder.forClient()
        .keyManager(clientCert, clientCertKey))
        .trustManager(
            InsecureTrustManagerFactory.INSTANCE)
        .applicationProtocolConfig(
            new ApplicationProtocolConfig(
                ApplicationProtocolConfig.Protocol.ALPN,
                ApplicationProtocolConfig.SelectorFailureBehavior.NO_ADVERTISE,
                ApplicationProtocolConfig.SelectedListenerFailureBehavior.ACCEPT,
                "h2"))
        .build();

WorkflowServiceStubs service =
    WorkflowServiceStubs.newInstance(
            .setSslContext(sslContext)
            .setTarget(temporalServerAddr)
            .build());
```

This change greatly simplifies the experience allowing users to simply do:
```
WorkflowServiceStubs service =
        WorkflowServiceStubs.newInstance(
                        .setSslContext(
                                new TemporalSslContextBuilder(
                                     clientCert, clientCertKey).build())
                        .setTarget(temporalServerAddr)
                        .build());

```

Also this addresses natural temptation of cutting corners and using InsecureTrustManagerFactory as a simpler option instead of checking certificate fingerprints or using system trust store as well as sets correct default parameters for the ApplicationProtocolConfig.
In addition this change removes the need for clients to directly depend on 
```
    compile group: 'io.grpc', name: 'grpc-netty-shaded', version: 'XXX'
```

As part of the implementation we've added a few methods:
* setSslContextWith - that takes keyCertChainInputStream, keyInputStream, keyPassword and trustManager.
* Its overloads that take fewer parameters and can derive defaults, most notable one is the one that creates default TrustManager.
* Insecure counterparts that utilize insecure trust manager, which can be useful during testing or prototyping.

Note that users that require more sophisticated setup may still use setSslContext method and manually populate all parameters.

### DIFF
--- a/temporal-serviceclient/src/main/java/io/temporal/serviceclient/SimpleSslContextBuilder.java
+++ b/temporal-serviceclient/src/main/java/io/temporal/serviceclient/SimpleSslContextBuilder.java
@@ -47,8 +47,8 @@ public class SimpleSslContextBuilder {
           // gRPC requires http2 protocol.
           ApplicationProtocolNames.HTTP_2);
 
-  private InputStream keyCertChain;
-  private InputStream key;
+  private final InputStream keyCertChain;
+  private final InputStream key;
   private TrustManager trustManager;
   private boolean useInsecureTrustManager;
   private String keyPassword;
--- a/temporal-serviceclient/src/main/java/io/temporal/serviceclient/SimpleSslContextBuilder.java
+++ b/temporal-serviceclient/src/main/java/io/temporal/serviceclient/SimpleSslContextBuilder.java
@@ -1,5 +1,5 @@
 /*
- *  Copyright (C) 2021 Temporal Technologies, Inc. All Rights Reserved.
+ *  Copyright (C) 2020 Temporal Technologies, Inc. All Rights Reserved.
  *
  *  Copyright 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
  *
--- a/temporal-serviceclient/src/main/java/io/temporal/serviceclient/SimpleSslContextBuilder.java
+++ b/temporal-serviceclient/src/main/java/io/temporal/serviceclient/SimpleSslContextBuilder.java
@@ -33,7 +33,7 @@ import javax.net.ssl.TrustManager;
 import javax.net.ssl.TrustManagerFactory;
 import javax.net.ssl.X509TrustManager;
 
-public class TemporalSslContextBuilder {
+public class SimpleSslContextBuilder {
 
   // Default TLS protocol config that is used to communicate with TLS enabled temporal backend.
   private static final ApplicationProtocolConfig DEFAULT_APPLICATION_PROTOCOL_CONFIG =
@@ -47,21 +47,23 @@ public class TemporalSslContextBuilder {
           // gRPC requires http2 protocol.
           ApplicationProtocolNames.HTTP_2);
 
-  private final InputStream keyCertChainInputStream;
-  private final InputStream keyInputStream;
+  private InputStream keyCertChain;
+  private InputStream key;
   private TrustManager trustManager;
   private boolean useInsecureTrustManager;
   private String keyPassword;
 
   /**
-   * @param keyCertChainInputStream - an input stream for an X.509 client certificate chain in PEM
-   *     format.
-   * @param keyInputStream - an input stream for a PKCS#8 client private key in PEM format.
+   * @param keyCertChain - an input stream for an X.509 client certificate chain in PEM format.
+   * @param key - an input stream for a PKCS#8 client private key in PEM format.
    */
-  public TemporalSslContextBuilder(
-      InputStream keyCertChainInputStream, InputStream keyInputStream) {
-    this.keyCertChainInputStream = keyCertChainInputStream;
-    this.keyInputStream = keyInputStream;
+  public static SimpleSslContextBuilder newBuilder(InputStream keyCertChain, InputStream key) {
+    return new SimpleSslContextBuilder(keyCertChain, key);
+  }
+
+  private SimpleSslContextBuilder(InputStream keyCertChain, InputStream key) {
+    this.keyCertChain = keyCertChain;
+    this.key = key;
   }
 
   /**
@@ -84,7 +86,7 @@ public class TemporalSslContextBuilder {
                 : useInsecureTrustManager
                     ? InsecureTrustManagerFactory.INSTANCE.getTrustManagers()[0]
                     : getDefaultTrustManager())
-        .keyManager(keyCertChainInputStream, keyInputStream, keyPassword)
+        .keyManager(keyCertChain, key, keyPassword)
         .applicationProtocolConfig(DEFAULT_APPLICATION_PROTOCOL_CONFIG)
         .build();
   }
--- a/temporal-serviceclient/src/main/java/io/temporal/serviceclient/TemporalSslContextBuilder.java
+++ b/temporal-serviceclient/src/main/java/io/temporal/serviceclient/TemporalSslContextBuilder.java
@@ -74,6 +74,9 @@ public class TemporalSslContextBuilder {
    * @throws SSLException - when it was unable to build the context.
    */
   public SslContext build() throws SSLException {
+    if (trustManager != null && useInsecureTrustManager)
+      throw new IllegalArgumentException(
+          "Can not use insecure trust manager if custom trust manager is set.");
     return SslContextBuilder.forClient()
         .trustManager(
             trustManager != null
@@ -88,7 +91,7 @@ public class TemporalSslContextBuilder {
 
   /**
    * @param trustManager - custom trust manager that should be used with the SSLContext for
-   *     verifying server CA * authority.
+   *     verifying server CA authority.
    */
   public void setTrustManager(TrustManager trustManager) {
     this.trustManager = trustManager;
@@ -131,7 +134,7 @@ public class TemporalSslContextBuilder {
       }
     }
     throw new UnknownDefaultTrustManagerException(
-        "Unable to find X509TrustManager in the list of default trust managers");
+        "Unable to find X509TrustManager in the list of default trust managers.");
   }
 
   /**
--- a/temporal-serviceclient/src/main/java/io/temporal/serviceclient/TemporalSslContextBuilder.java
+++ b/temporal-serviceclient/src/main/java/io/temporal/serviceclient/TemporalSslContextBuilder.java
@@ -1,0 +1,149 @@
+/*
+ *  Copyright (C) 2021 Temporal Technologies, Inc. All Rights Reserved.
+ *
+ *  Copyright 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ *  Modifications copyright (C) 2017 Uber Technologies, Inc.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License"). You may not
+ *  use this file except in compliance with the License. A copy of the License is
+ *  located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ *  or in the "license" file accompanying this file. This file is distributed on
+ *  an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
+package io.temporal.serviceclient;
+
+import io.grpc.netty.shaded.io.netty.handler.ssl.ApplicationProtocolConfig;
+import io.grpc.netty.shaded.io.netty.handler.ssl.ApplicationProtocolNames;
+import io.grpc.netty.shaded.io.netty.handler.ssl.SslContext;
+import io.grpc.netty.shaded.io.netty.handler.ssl.SslContextBuilder;
+import io.grpc.netty.shaded.io.netty.handler.ssl.util.InsecureTrustManagerFactory;
+import java.io.InputStream;
+import java.security.KeyStore;
+import java.security.KeyStoreException;
+import java.security.NoSuchAlgorithmException;
+import javax.net.ssl.SSLException;
+import javax.net.ssl.TrustManager;
+import javax.net.ssl.TrustManagerFactory;
+import javax.net.ssl.X509TrustManager;
+
+public class TemporalSslContextBuilder {
+
+  // Default TLS protocol config that is used to communicate with TLS enabled temporal backend.
+  private static final ApplicationProtocolConfig DEFAULT_APPLICATION_PROTOCOL_CONFIG =
+      new ApplicationProtocolConfig(
+          // HTTP/2 over TLS mandates the use of ALPN to negotiate the use of the protocol.
+          ApplicationProtocolConfig.Protocol.ALPN,
+          // NO_ADVERTISE is the only mode supported by both OpenSsl and JDK providers.
+          ApplicationProtocolConfig.SelectorFailureBehavior.NO_ADVERTISE,
+          // ACCEPT is the only mode supported by both OpenSsl and JDK providers.
+          ApplicationProtocolConfig.SelectedListenerFailureBehavior.ACCEPT,
+          // gRPC requires http2 protocol.
+          ApplicationProtocolNames.HTTP_2);
+
+  private final InputStream keyCertChainInputStream;
+  private final InputStream keyInputStream;
+  private TrustManager trustManager;
+  private boolean useInsecureTrustManager;
+  private String keyPassword;
+
+  /**
+   * @param keyCertChainInputStream - an input stream for an X.509 client certificate chain in PEM
+   *     format.
+   * @param keyInputStream - an input stream for a PKCS#8 client private key in PEM format.
+   */
+  public TemporalSslContextBuilder(
+      InputStream keyCertChainInputStream, InputStream keyInputStream) {
+    this.keyCertChainInputStream = keyCertChainInputStream;
+    this.keyInputStream = keyInputStream;
+  }
+
+  /**
+   * Builds {@link SslContext} from specified parameters. If trust manager is set then it will be
+   * used to verify server authority, otherwise system default trust manager (or if {@link
+   * #useInsecureTrustManager} is set then insecure trust manager) is going to be used.
+   *
+   * @return - {@link SslContext} that can be used with the {@link
+   *     WorkflowServiceStubsOptions.Builder#setSslContext(SslContext)}
+   * @throws SSLException - when it was unable to build the context.
+   */
+  public SslContext build() throws SSLException {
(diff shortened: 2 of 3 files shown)


temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsOptions.java
```diff
@@ -222,6 +230,17 @@ public Scope getMetricsScope() {
    */
   public static class Builder {
 
+    // Default TLS protocol config that is used to communicate with temporal backend in the cloud.
```
mastermanu, line 233: instead of "temporal backend in the cloud," can we say "TLS-enabled Temporal Server" instead? That seems more generic for OSS Client

temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsOptions.java
```diff
@@ -273,6 +292,83 @@ public Builder setSslContext(SslContext sslContext) {
       return this;
     }
 
+    /**
+     * Sets default SSL context parameters that can be used with the temporal cloud service. Users
```
mastermanu, line 296: same question around "temporal cloud service" here

temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsOptions.java
```diff
@@ -273,6 +292,83 @@ public Builder setSslContext(SslContext sslContext) {
...
+     * that require additional customization may use {@link #setSslContext(SslContext)} directly.
+     *
+     * @param keyCertChainInputStream - an input stream for an X.509 certificate chain in PEM
```
mastermanu, line 299: we should state that this is the "client" certificate chain to avoid confusion. similar for keyInputStream, we should mention it is the client certificate's private key

temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsOptions.java
```diff
@@ -273,6 +292,83 @@ public Builder setSslContext(SslContext sslContext) {
...
+     *     format.
+     * @param keyPassword - the password of the keyFile, or null if it's not password-protected.
+     * @param keyInputStream - an input stream for a PKCS#8 private key in PEM format.
+     * @param fingerprints - a list of SHA1 fingerprints in hexadecimal form.
```
mastermanu, line 303: can we give more detail on what these SHA1 fingerprints represent (does it represent acceptable Temporal Server certificate? Or acceptable CA that has issued Temporal Server Certificate?)

temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsOptions.java
```diff
@@ -273,6 +292,83 @@ public Builder setSslContext(SslContext sslContext) {
...
+     * @throws SSLException - when it was unable to build the context.
+     */
+    public Builder setCloudSslContext(
+        InputStream keyCertChainInputStream,
+        String keyPassword,
```
mastermanu, line 308: the SslContextBuilder order is: cert chain, key, key password. Our helper methods seem to saw the keyPassword and KeyInput from what the ContextBuilder does. Can we have all the helper methods have the keyPassword after the keyInput/file for consistency?

temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsOptions.java
```diff
@@ -273,6 +292,83 @@ public Builder setSslContext(SslContext sslContext) {
       return this;
     }
 
+    /**
+     * Sets default SSL context parameters that can be used with the temporal cloud service. Users
+     * that require additional customization may use {@link #setSslContext(SslContext)} directly.
+     *
+     * @param keyCertChainInputStream - an input stream for an X.509 certificate chain in PEM
+     *     format.
+     * @param keyPassword - the password of the keyFile, or null if it's not password-protected.
```
mastermanu, line 301: do you mean keyInputStream here?

temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsOptions.java
```diff
@@ -273,6 +292,83 @@ public Builder setSslContext(SslContext sslContext) {
...
+     * @param keyInputStream - an input stream for a PKCS#8 private key in PEM format.
+     * @param fingerprints - a list of SHA1 fingerprints in hexadecimal form.
+     * @throws SSLException - when it was unable to build the context.
+     */
+    public Builder setCloudSslContext(
```
mastermanu, line 306: might be worth having a non-default helper option where the user doesn't care about doing TLS server validation (such as for development purposes). We have the same thing for the GoSDK (InsecureSkipVerify) that has been really useful.

mastermanu (approved):
signed off with comments

temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsOptions.java
vitarb: No, otherwise we'll be dealing with ambiguous signatures, (IS, IS, String, String..), in which case user would need to pass a string array as the last parameter. In my opinion it's better to add password in this position than having to deal with that issue.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsOptions.java
vitarb: Parameters and comments map 1:1 with what we have in netty, see [netty documentation](https://netty.io/4.0/api/io/netty/handler/ssl/SslContextBuilder.html) for details.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsOptions.java
vitarb: This should be a fingerprint of the server's root CA. Updated the comment.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsOptions.java
vitarb: This is something we should discuss. First I thought about calling this method `setDefaultSslContext` or something along those lines, but then I thought that this will likely be used only by our cloud service users (who else would use TLS with certificates heh...) and decided to call it `setCloudSslContext` and that's why comments reflect it as well.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsOptions.java
vitarb: Not sure what you mean here, could you explain?

temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsOptions.java
vitarb: There is a discussion point about this, which we should address. Namely what should be the default behavior in case if fingerprints are not set. This could default to Insecure or our default root CA fingerprint. I think if we hardcode our root CA fingerprint then using insecure will not be required at all. In other words I don't want to create wrong incentives and provide an easy way of using insecure as it may eventually leak into production code.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsOptions.java
mastermanu: " @param keyPassword - the password of the keyFile" --> this specific function doesn't have a keyFile parameter

temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsOptions.java
```diff
@@ -273,6 +292,87 @@ public Builder setSslContext(SslContext sslContext) {
...
+     * @param keyCertChainFile - an X.509 client certificate chain file in PEM format.
+     * @param keyPassword - the password of the keyFile, or null if it's not password-protected.
+     * @param keyFile - a PKCS#8 client private key file in PEM format.
+     * @param fingerprints - a list of SHA1 fingerprints in hexadecimal form of the server's root
+     *     certificate.
+     */
+    public Builder setCloudSslContext(
+        File keyCertChainFile, String keyPassword, File keyFile, String... fingerprints)
+        throws IOException {
+      // TODO consider using default fingerprints from the root CA if user didn't pass any.
```
mastermanu, line 352: if we are connecting to a TLS service with a production LetsEncrypt certificate, and the user doesn't pass any fingerprints, are we saying that the connection will then fail? Or does the FingerprintTrustManagerFactory automatically fallback to the machine's trusted cert store?

temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsOptions.java
```diff
@@ -273,6 +292,87 @@ public Builder setSslContext(SslContext sslContext) {
...
+     *     format.
+     * @param keyPassword - the password of the keyFile, or null if it's not password-protected.
+     * @param keyInputStream - an input stream for a PKCS#8 client private key in PEM format.
+     * @param fingerprints - a list of SHA1 fingerprints in hexadecimal form of the server's root
+     *     certificate.
+     * @throws SSLException - when it was unable to build the context.
+     */
+    public Builder setCloudSslContext(
+        InputStream keyCertChainInputStream,
+        String keyPassword,
```
Sushisource, line 309: FYI you probably want to use `char[]` rather than `String` for passing plaintext secrets around in memory: https://www.baeldung.com/java-storing-passwords

Not hugely important but should be a fairly easy change

Sushisource (approved):
LGTM with small alteration

temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsOptions.java
vitarb: We need to pass this as a string to the key manager anyways, so it's not worth it to require char[] as it will get translated to the string anyways.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsOptions.java
vitarb: It does not fallback to anything, underlying implementation is pretty dumb and it simply matches strings.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsOptions.java
vitarb: On the second thought, let me add an insecure method, and just call it insecure explicitly. I think this would be a better user experience so worth doing.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsOptions.java
vitarb: I think I've found a better name `setSslContextWith(...)` and `setInsecureSslContextWith(...)`.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsOptions.java
```diff
@@ -273,6 +293,160 @@ public Builder setSslContext(SslContext sslContext) {
...
+     *
+     * @param keyCertChainInputStream - an input stream for an X.509 client certificate chain in PEM
+     *     format.
+     * @param keyPassword - the password of the key, or null if it's not password-protected.
+     * @param keyInputStream - an input stream for a PKCS#8 client private key in PEM format.
+     * @param fingerprints - a list of SHA1 fingerprints in hexadecimal form of the server's root
+     *     certificate.
+     * @throws SSLException - when it was unable to build the context.
+     */
+    public Builder setSslContextWith(
```
mfateev, line 308: This doesn't follow the standard builder pattern. The code should be:

setSslContext(SslContext.newBuilder()....build())

and all the variations of the methods should be done through the builder.

mfateev (changes requested):

temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsOptions.java
vitarb: I agree, we have way too many variations now, it's better to move it into a separate builder. The only reason I didn't do it initially is cause there is already SslContextBuilder in netty, so I didn't want to overload the name.

vitarb: Addressed.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/TemporalSslContextBuilder.java
```diff
@@ -0,0 +1,149 @@
...
+  /**
+   * @param keyCertChainInputStream - an input stream for an X.509 client certificate chain in PEM
+   *     format.
+   * @param keyInputStream - an input stream for a PKCS#8 client private key in PEM format.
+   */
+  public TemporalSslContextBuilder(
+      InputStream keyCertChainInputStream, InputStream keyInputStream) {
+    this.keyCertChainInputStream = keyCertChainInputStream;
+    this.keyInputStream = keyInputStream;
+  }
```
mfateev, line 65: I would create a static method and hide the constructor:

TemporalSslContextBuilder.newBuilder(cert, key)

I would also rename the parameters. Having "InputStream" postfix doesn't provide any value.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/TemporalSslContextBuilder.java
```diff
@@ -0,0 +1,149 @@
...
+import java.io.InputStream;
+import java.security.KeyStore;
+import java.security.KeyStoreException;
+import java.security.NoSuchAlgorithmException;
+import javax.net.ssl.SSLException;
+import javax.net.ssl.TrustManager;
+import javax.net.ssl.TrustManagerFactory;
+import javax.net.ssl.X509TrustManager;
+
+public class TemporalSslContextBuilder {
```
mfateev, line 36: That would be the only place where we use Temporal as a prefix for a class name. I think we would be better without it.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/TemporalSslContextBuilder.java
vitarb: Done

temporal-serviceclient/src/main/java/io/temporal/serviceclient/TemporalSslContextBuilder.java
vitarb: Replaced with SimpleSslContextBuilder

temporal-serviceclient/src/main/java/io/temporal/serviceclient/SimpleSslContextBuilder.java
```diff
@@ -47,21 +47,23 @@
...
   private TrustManager trustManager;
   private boolean useInsecureTrustManager;
   private String keyPassword;
 
   /**
-   * @param keyCertChainInputStream - an input stream for an X.509 client certificate chain in PEM
-   *     format.
-   * @param keyInputStream - an input stream for a PKCS#8 client private key in PEM format.
+   * @param keyCertChain - an input stream for an X.509 client certificate chain in PEM format.
+   * @param key - an input stream for a PKCS#8 client private key in PEM format.
```
mfateev, line 58: May be rename to "privateKey" or "privateKeyPem?

temporal-serviceclient/src/main/java/io/temporal/serviceclient/SimpleSslContextBuilder.java
```diff
@@ -47,21 +47,23 @@
...
+  private InputStream key;
   private TrustManager trustManager;
   private boolean useInsecureTrustManager;
   private String keyPassword;
 
   /**
-   * @param keyCertChainInputStream - an input stream for an X.509 client certificate chain in PEM
-   *     format.
-   * @param keyInputStream - an input stream for a PKCS#8 client private key in PEM format.
+   * @param keyCertChain - an input stream for an X.509 client certificate chain in PEM format.
```
mfateev, line 57: Why is it keyCertChain? From the description is sounds like certChain to me. Or maybe certChainPem?

mfateev (approved):

temporal-serviceclient/src/main/java/io/temporal/serviceclient/SimpleSslContextBuilder.java
vitarb: I kept parameter names same as in the netty SslOntextBuilder, which can also take files. I don't think we need to support files as input streams are more generic and it's easy to create FileInputStream from file path. Given that we don't support multiple inputs we can simplify the naming.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/SimpleSslContextBuilder.java
vitarb: Again, borrowed the name from the underlying SslContextBuilder.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/SimpleSslContextBuilder.java
vitarb: I would prefer to keep names consistent with the SslContextBuilder to avoid confusion.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/SimpleSslContextBuilder.java
vitarb: Same as above.

PR #329, opened 2021-02-02
Add abitlity to do async completion without having to specify task token

vitarb:
This change follows up on #291 and removes the need to pass task token and instead of `ActivityCompletionClient` returns `ManualActivityCompletionClient`, which only allows operations on activities run locally.


temporal-sdk/src/main/java/io/temporal/client/ActivityLocalCompletionClient.java
```diff
@@ -0,0 +1,30 @@
...
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
+package io.temporal.client;
+
+public interface ActivityLocalCompletionClient {
+  <R> void complete(R result) throws ActivityCompletionException;
+
+  void completeExceptionally(Exception result) throws ActivityCompletionException;
```
mfateev, line 25: I would rename the parameter to failure

temporal-sdk/src/main/java/io/temporal/client/ActivityLocalCompletionClientImpl.java
```diff
@@ -0,0 +1,70 @@
...
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
+package io.temporal.client;
+
+import io.temporal.internal.external.ManualActivityCompletionClientFactory;
+import io.temporal.workflow.Functions;
+
+public class ActivityLocalCompletionClientImpl implements ActivityLocalCompletionClient {
```
mfateev, line 25: I would make it package visibility and move it to the package that instantiates it.

temporal-sdk/src/main/java/io/temporal/client/ActivityLocalCompletionClientImpl.java
```diff
@@ -0,0 +1,70 @@
...
+  private final ManualActivityCompletionClientFactory factory;
+  private final Functions.Proc completionHandle;
+  private byte[] taskToken;
```
mfateev, line 28: Why is this not final?

temporal-sdk/src/main/java/io/temporal/internal/sync/ActivityExecutionContextImpl.java
```diff
@@ -29,12 +29,7 @@
 import io.temporal.api.common.v1.Payloads;
 import io.temporal.api.workflowservice.v1.RecordActivityTaskHeartbeatRequest;
 import io.temporal.api.workflowservice.v1.RecordActivityTaskHeartbeatResponse;
-import io.temporal.client.ActivityCanceledException;
-import io.temporal.client.ActivityCompletionClient;
-import io.temporal.client.ActivityCompletionException;
-import io.temporal.client.ActivityCompletionFailureException;
-import io.temporal.client.ActivityNotExistsException;
-import io.temporal.client.ActivityWorkerShutdownException;
+import io.temporal.client.*;
```
mfateev, line 32: Undo. Change your IDE settings to avoid such rollups.

mfateev (changes requested):
I believe we can now remove `completionHandle` from `ActivityCompletionClientImpl`.

temporal-sdk/src/main/java/io/temporal/internal/sync/ActivityExecutionContextImpl.java
vitarb: Ah, I've switched from Community to Ultimate and reset my settings. Fixed.

vitarb:
Replaced new interface with ManualActivityCompletionClient since it already does exactly what we need and simply added a new implementation that is completion and task token aware.

Sushisource (approved):

temporal-sdk/src/main/java/io/temporal/internal/external/CompletionAwareManualCompletionClient.java
```diff
@@ -0,0 +1,70 @@
...
+ *  an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
+package io.temporal.internal.external;
+
+import io.temporal.failure.CanceledFailure;
+import io.temporal.workflow.Functions;
+
```
mfateev, line 24: I think this class should wrap another ManualActivityCompletionClient instead of the factory.

temporal-sdk/src/test/java/io/temporal/workflow/LocalAsyncCompletionWorkflowTest.java
```diff
@@ -19,14 +19,10 @@
 
 package io.temporal.workflow;
 
-import io.temporal.activity.Activity;
-import io.temporal.activity.ActivityExecutionContext;
-import io.temporal.activity.ActivityInterface;
-import io.temporal.activity.ActivityMethod;
-import io.temporal.activity.ActivityOptions;
-import io.temporal.client.ActivityCompletionClient;
+import io.temporal.activity.*;
```
mfateev, line 22: Unwrap. And change IDE settings to not add .*.

mfateev (changes requested):
Please move ManualActivityCompletionClient to the public package io.temporal.activity.

temporal-sdk/src/test/java/io/temporal/workflow/LocalAsyncCompletionWorkflowTest.java
vitarb: It did it before I changed the settings :facepalm:

temporal-sdk/src/main/java/io/temporal/internal/external/CompletionAwareManualCompletionClient.java
```diff
@@ -0,0 +1,67 @@
...
+ *  permissions and limitations under the License.
+ */
+
+package io.temporal.internal.external;
+
+import io.temporal.activity.ManualActivityCompletionClient;
+import io.temporal.failure.CanceledFailure;
+import io.temporal.workflow.Functions;
+
+public final class CompletionAwareManualCompletionClient implements ManualActivityCompletionClient {
```
mfateev, line 26: I would move it to the same package that instantiates this class and change visibility to package.

mfateev (approved):

temporal-sdk/src/main/java/io/temporal/internal/external/CompletionAwareManualCompletionClient.java
vitarb: done

PR #336, opened 2021-02-07
Removed InterruptedException from awaitTermination

mfateev:
Having checked exception thrown from `awaitTermination` is very unfriendly for consumers.


temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubs.java
```diff
@@ -61,7 +61,7 @@ static WorkflowServiceStubs newInstance(
 
   boolean isTerminated();
 
-  boolean awaitTermination(long timeout, TimeUnit unit) throws InterruptedException;
+  boolean awaitTermination(long timeout, TimeUnit unit);
```
Sushisource, line 64: Seems like this return value has more meaning than it used to, would be nice to explain in docstring

Sushisource (approved):

PR #344, opened 2021-02-12
Factored out activity tests

vkoby:
The WorkflowTest is very long and takes forever to complete, refactored the first batch of tests into separate files, so that Gradle can run them in parallel. Also made them use our TestWorkflowRule and factored out tracer into a separate file as well.


Sushisource (approved):
LGTM!

vitarb (changes requested):
As discussed offline, let's use our test rule to separate tests. Here is an example of how that can be done for one of them.
https://github.com/temporalio/sdk-java/pull/346/files
We'll need to think about what is the best way to share common things like tracer and consider splitting common TestActivitiesImpl into smaller more specific activity interfaces/implementations.

vkoby:
IMPORTANT NOTE:
These tests fail due to time out
[io.temporal.workflow.LocalActivityRetryTest::testLocalActivityRetry]
[io.temporal.workflow.AsyncUntypedActivityTest::testAsyncUntypedActivity]
[io.temporal.workflow.AsyncUntyped2ActivityTest::testAsyncUntyped2Activity]
[io.temporal.workflow.AsyncActivityTest::testAsyncActivity]

vitarb:
We should also remove tests that were migrated from the WorkflowTest.

temporal-sdk/src/test/java/io/temporal/workflow/AbandonOnCancelActivityTest.java
```diff
@@ -0,0 +1,112 @@
...
+import java.time.Duration;
+import org.junit.Assert;
+import org.junit.Rule;
+import org.junit.Test;
+
+public class AbandonOnCancelActivityTest {
+
+  public static final String NAMESPACE = "UnitTest";
+  private final WorkflowTest.TestActivitiesImpl activitiesImpl =
+      new WorkflowTest.TestActivitiesImpl(null);
```
vitarb, line 43: We should try make this part cleaner. Have you thought about breaking down TestActivitiesImpl into separate more specific activity interfaces (e.g. with one method each)?

vitarb (approved):

temporal-testing-junit4/src/main/java/io/temporal/testing/TestWorkflowRule.java
```diff
@@ -84,14 +85,16 @@ private TestWorkflowRule(
       Object[] activityImplementations,
       WorkerOptions workerOptions,
       long testTimeoutSeconds,
-      boolean doNotStart) {
+      boolean doNotStart,
+      TracingWorkerInterceptor tracer) {
```
vitarb, line 89: Instead of taking specific implementation, let's have a list or var-arg array of WorkerInterceptor here. This would be more generic and will allow users to use any custom interceptors when needed.

PR #351, opened 2021-02-16
Move gRPC connections to idle periodically in order to trigger LB

vitarb:
gRPC abstracts away connection management from the user allowing interactions only on the level of the channel.
One of our issues was that all connections that gRPC is using are long living and would last as long as the server responds on keep alive requests (which practically can mean forever). This results in us being unable to properly load balance connections to the temporal front ends over time.

Consider a simple example:
```
Let's say we have 2 frontend hosts running and 20 workers 
connect to them. At the beginning, since we are using round-robing 
load balancer connections will be distributed evenly (or close to that), 
so we'll have 10 and 10 connections to each front end host on average. 
Now let's assume these front end hosts reboot one after another. 
What would happen is that front end host that comes up first would 
get all 20 worker connections and front end host that comes up later 
would have zero. This situation won't change unless workers restart.
```

To prevent this situation from happening we choose to move gRPC channels on the worker side into the idle state, which triggers a reconnection to the front end and new invocation of the load balancer. This allows us to distribute connections more evenly over time.

Currently we are doing this simply every 5 minutes, but in the future we could add a special header in the response from the server, which would allow front-end host to ask client to reconnect allowing more targeted actions. This is slightly more complex change, which would require propagating additional info through the ring and adding a new header, so we opt for simplicity here.

I've tested this change with aggressive settings for entering idle state (every 1, 10 and 30 seconds) using bench test. Results don't show any impact on request latency and no visible impact on the number of gRPC request errors.

### DIFF
--- a/temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsOptions.java
+++ b/temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsOptions.java
@@ -44,6 +44,10 @@ public class WorkflowServiceStubsOptions {
   private static final Duration DEFAULT_QUERY_RPC_TIMEOUT = Duration.ofSeconds(10);
   /** Default timeout that will be used to reset connection backoff. */
   private static final Duration DEFAULT_CONNECTION_BACKOFF_RESET_FREQUENCY = Duration.ofSeconds(10);
+  /**
+   * Default timeout that will be used to enter idle channel state and reconnect to temporal server.
+   */
+  private static final Duration DEFAULT_ENTER_IDLE_CHANNEL_STATE_FREQUENCY = Duration.ofMinutes(5);
 
   private static final WorkflowServiceStubsOptions DEFAULT_INSTANCE;
 
@@ -85,6 +89,12 @@ public class WorkflowServiceStubsOptions {
   /** Frequency at which connection backoff is going to be reset */
   private final Duration connectionBackoffResetFrequency;
 
+  /**
+   * Frequency at which grpc connection channel will be moved into an idle state, triggering a new
+   * connection to the temporal frontend host.
+   */
+  private final Duration enterIdleChannelStateFrequency;
+
   /** Optional gRPC headers */
   private final Metadata headers;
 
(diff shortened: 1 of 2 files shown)


Sushisource (approved):
This makes sense. I suspect this is something we're going to need to add to Tonic as well...

temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsOptions.java
```diff
@@ -44,6 +44,10 @@
   private static final Duration DEFAULT_QUERY_RPC_TIMEOUT = Duration.ofSeconds(10);
   /** Default timeout that will be used to reset connection backoff. */
   private static final Duration DEFAULT_CONNECTION_BACKOFF_RESET_FREQUENCY = Duration.ofSeconds(10);
+  /**
+   * Default timeout that will be used to enter idle channel state and reconnect to temporal server.
+   */
+  private static final Duration DEFAULT_ENTER_IDLE_CHANNEL_STATE_FREQUENCY = Duration.ofMinutes(5);
```
mfateev, line 50: I think 5 minutes is too large to wait for rebalancing. What about 1 minute?

temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsOptions.java
```diff
@@ -85,6 +89,12 @@ public static WorkflowServiceStubsOptions getDefaultInstance() {
   /** Frequency at which connection backoff is going to be reset */
   private final Duration connectionBackoffResetFrequency;
 
+  /**
+   * Frequency at which grpc connection channel will be moved into an idle state, triggering a new
+   * connection to the temporal frontend host.
+   */
+  private final Duration enterIdleChannelStateFrequency;
```
mfateev, line 96: What about a more user-friendly name? Something like "grpcReconnectFrequency"?

temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsOptions.java
vitarb: I was thinking between 1 and 5, went for a potentially less intrusive option. But we can change it to 1 as well.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsOptions.java
vitarb: I like it, let me rename.

mfateev (approved):

PR #360, opened 2021-02-23
Return supported query types for unsupported queries

vkoby:
Error: Query workflow failed.
Error Details: rpc error: code = InvalidArgument desc = java.lang.IllegalArgumentException: Unknown query type: queryGreeting123, knownTypes=[queryGreeting]

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/QueryDispatcher.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/QueryDispatcher.java
@@ -75,7 +75,8 @@ class QueryDispatcher {
     Object[] args;
     if (handler == null) {
       if (dynamicQueryHandler == null) {
-        throw new IllegalStateException("Unknown query type: " + queryName);
+        throw new IllegalArgumentException(
+            "Unknown query type: " + queryName + ", knownTypes=" + queryCallbacks.keySet());
       }
       args = new Object[] {new EncodedValues(input, converter)};
     } else {
--- a/temporal-sdk/src/test/java/io/temporal/internal/sync/QueryDispatcherTest.java
+++ b/temporal-sdk/src/test/java/io/temporal/internal/sync/QueryDispatcherTest.java
@@ -1,12 +1,12 @@
 /*
- *  Copyright (C) 2021 Temporal Technologies, Inc. All Rights Reserved.
+ *  Copyright (C) 2020 Temporal Technologies, Inc. All Rights Reserved.
  *
  *  Copyright 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
  *
  *  Modifications copyright (C) 2017 Uber Technologies, Inc.
  *
  *  Licensed under the Apache License, Version 2.0 (the "License"). You may not
- *  use this file except in compliance with the License. A copy of the License iss
+ *  use this file except in compliance with the License. A copy of the License is
  *  located at
  *
  *  http://aws.amazon.com/apache2.0
--- a/temporal-sdk/src/test/java/io/temporal/internal/sync/QueryDispatcherTest.java
+++ b/temporal-sdk/src/test/java/io/temporal/internal/sync/QueryDispatcherTest.java
@@ -19,26 +19,65 @@
 
 package io.temporal.internal.sync;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertThrows;
+import static org.junit.Assert.*;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
 
+import io.temporal.api.common.v1.Payloads;
+import io.temporal.common.converter.DataConverter;
+import io.temporal.common.interceptors.WorkflowInboundCallsInterceptor;
 import io.temporal.common.interceptors.WorkflowOutboundCallsInterceptor;
+import java.lang.reflect.Type;
+import java.util.Optional;
 import org.junit.Test;
 
 public class QueryDispatcherTest {
 
+  private QueryDispatcher initializeDispatcher() {
+    // Initialize the dispatcher.
+    QueryDispatcher dispatcher = new QueryDispatcher(DataConverter.getDefaultInstance());
+    WorkflowOutboundCallsInterceptor.RegisterQueryInput requestA =
+        new WorkflowOutboundCallsInterceptor.RegisterQueryInput(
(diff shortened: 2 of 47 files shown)


temporal-sdk/src/test/java/io/temporal/internal/sync/QueryDispatcherTest.java
```diff
@@ -0,0 +1,44 @@
...
+import static org.junit.Assert.assertThrows;
+
+import io.temporal.common.interceptors.WorkflowOutboundCallsInterceptor;
+import org.junit.Test;
+
+public class QueryDispatcherTest {
+
+  @Test
+  public void testQueryDispatcherException() {
+    QueryDispatcher qd = new QueryDispatcher(null);
```
vitarb, line 32: You can pass `DefaultDataConverter.getDefaultInstance()` instead of `null` to make instance functional.

temporal-sdk/src/test/java/io/temporal/internal/sync/QueryDispatcherTest.java
```diff
@@ -0,0 +1,44 @@
...
+        new WorkflowOutboundCallsInterceptor.RegisterQueryInput("QueryA", null, null, null);
+    qd.registerQueryHandlers(requestA);
+    Throwable exception =
+        assertThrows(
+            IllegalArgumentException.class,
+            () -> {
+              qd.handleQuery("Query", null);
+            });
+    assertEquals("Unknown query type: Query, knownTypes=[QueryA]", exception.getMessage());
+  }
```
vitarb, line 43: Can we also add a positive scenario test? (e.g. when query can be resolved successfully)

temporal-sdk/src/test/java/io/temporal/internal/sync/QueryDispatcherTest.java
```diff
@@ -0,0 +1,44 @@
...
+  public void testQueryDispatcherException() {
+    QueryDispatcher qd = new QueryDispatcher(null);
+    WorkflowOutboundCallsInterceptor.RegisterQueryInput requestA =
+        new WorkflowOutboundCallsInterceptor.RegisterQueryInput("QueryA", null, null, null);
+    qd.registerQueryHandlers(requestA);
+    Throwable exception =
+        assertThrows(
+            IllegalArgumentException.class,
+            () -> {
+              qd.handleQuery("Query", null);
```
vitarb, line 40: nit. Should we call this QueryB?

temporal-sdk/src/test/java/io/temporal/internal/sync/QueryDispatcherTest.java
vkoby: getDefaultInstance() is package private. To use it I'll have to make it public. Would that be OK?

vitarb: You should use `DataConverter` interface instead which has 
```static DataConverter getDefaultInstance() {
    return DefaultDataConverter.getDefaultInstance();
  }
```

temporal-serviceclient/build.gradle
```diff
@@ -77,19 +77,21 @@ dependencies {
     errorproneJavac('com.google.errorprone:javac:9+181-r4173-1')
     errorprone('com.google.errorprone:error_prone_core:2.5.1')
 
-    api 'io.grpc:grpc-protobuf:1.35.0'
-    api 'io.grpc:grpc-stub:1.35.0'
-    api 'io.grpc:grpc-core:1.35.0'
-    api group: 'com.google.protobuf', name: 'protobuf-java-util', version: '3.14.0'
-    api group: 'com.uber.m3', name: 'tally-core', version: '0.6.1'
+    api 'io.grpc:grpc-protobuf:1.36.0'
```
vitarb, line 80: I suggest making dependency updates in the separate PR.
This way if something goes wrong problems will be easier to locate and roll back.

temporal-sdk/src/test/java/io/temporal/internal/sync/QueryDispatcherTest.java
```diff
@@ -0,0 +1,83 @@
...
+import io.temporal.common.converter.DataConverter;
+import io.temporal.common.interceptors.WorkflowInboundCallsInterceptor;
+import io.temporal.common.interceptors.WorkflowOutboundCallsInterceptor;
+import java.lang.reflect.Type;
+import java.util.Optional;
+import org.junit.Test;
+
+public class QueryDispatcherTest {
+
+  private QueryDispatcher initializeDispatcher() {
```
vitarb, line 37: You can do this in a @Before method and assign variable to the field.
This way you can avoid having to do
```
     QueryDispatcher dispatcher = initializeDispatcher();
```
multiple times.

temporal-sdk/src/test/java/io/temporal/internal/sync/QueryDispatcherTest.java
```diff
@@ -0,0 +1,83 @@
...
+            "QueryB", new Class[] {}, new Type[] {}, null);
+    dispatcher.registerQueryHandlers(requestA);
+    dispatcher.registerQueryHandlers(requestB);
+
+    return dispatcher;
+  }
+
+  @Test
+  public void testQuerySuccess() {
+    QueryDispatcher dispatcher = initializeDispatcher();
```
vitarb, line 54: nit. You can create a dispatcher in the @Before method and assign it to the field.

temporal-sdk/src/test/java/io/temporal/internal/sync/QueryDispatcherTest.java
vitarb, line 37: Might be a good idea to pass a list of supported queries here, e.g.
```
initializeDispatcher(String... queries) {
...
```
This would make tests a bit more readable.

vitarb (approved):

temporal-sdk/src/main/java/io/temporal/internal/sync/QueryDispatcher.java
```diff
@@ -75,7 +75,8 @@ public void setInboundCallsInterceptor(WorkflowInboundCallsInterceptor inboundCa
     Object[] args;
     if (handler == null) {
       if (dynamicQueryHandler == null) {
-        throw new IllegalStateException("Unknown query type: " + queryName);
+        throw new IllegalArgumentException(
+            "Unknown query type: " + queryName + ", knownTypes=" + queryCallbacks.keySet());
```
feedmeapples, line 79: could you comment a sample string of this output? I want to test with the web

PR #368, opened 2021-02-26
Workflowtest refactoring

vkoby:
Part of Workflowtest Refactor task: This test file was 7k lines long and contained dozens of tests. In addition to readability breaking it up will also enable Gradle to run all those tests in parallel.
- Added an SDKTestWorkflowRule that wraps around TestWorkflowRule for SDK-specific testing 
- Refactored common implementation into shared package


mfateev (commented):
I haven't finished the review yet. But here are some comments.

temporal-sdk/src/test/java/io/temporal/workflow/TestOptions.java
```diff
@@ -0,0 +1,82 @@
...
+package io.temporal.workflow;
+
+import io.temporal.activity.LocalActivityOptions;
+import io.temporal.client.WorkflowOptions;
+import java.time.Duration;
+
+class TestOptions {
+  // When set to true increases test, activity and workflow timeouts to large values to support
+  // stepping through code in a debugger without timing out.
+  public static final boolean DEBUGGER_TIMEOUTS = false;
```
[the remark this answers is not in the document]
vkoby: DeterministicRunner is an interface, so it can't have private fields.

temporal-sdk/src/test/java/io/temporal/workflow/TestOptions.java
```diff
@@ -0,0 +1,82 @@
...
+          .setTaskQueue(taskQueue);
+    } else {
+      return WorkflowOptions.newBuilder()
+          .setWorkflowRunTimeout(Duration.ofHours(30))
+          .setWorkflowTaskTimeout(Duration.ofSeconds(5))
+          .setTaskQueue(taskQueue);
+    }
+  }
+
+  public static io.temporal.activity.ActivityOptions newActivityOptions1(String taskQueue) {
```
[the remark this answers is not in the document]
vkoby: How about:
newActivityOptionsForTaskQueue 
newActivityOptions20sScheduleToClose 
newLocalActivityOptions ?

temporal-sdk/src/test/java/io/temporal/workflow/AbandonOnCancelActivityTest.java
```diff
@@ -30,42 +30,45 @@
...
   @Rule
-  public TestWorkflowRule testWorkflowRule =
-      TestWorkflowRule.newBuilder()
-          .setWorkflowTypes(TestAbandonOnCancelActivity.class)
-          .setActivityImplementations(activitiesImpl)
-          .setUseExternalService(Boolean.parseBoolean(System.getenv("USE_DOCKER_SERVICE")))
-          .setTarget(System.getenv("TEMPORAL_SERVICE_ADDRESS"))
-          .build();
+  public SDKTestWorkflowRule testWorkflowRule =
+      (SDKTestWorkflowRule)
```
mfateev, line 49: I would try to avoid casts in normal code. In this particular case, I would change SDKTestWorkflowRule.Builder to not extend TestWorkflowRule.Builder but delegate to it.

temporal-sdk/src/test/java/io/temporal/workflow/AbandonOnCancelActivityTest.java
```diff
@@ -30,42 +30,45 @@
...
+          SDKTestWorkflowRule.newBuilder()
+              .setWorkflowTypes(TestAbandonOnCancelActivity.class)
+              .setActivityImplementations(activitiesImpl)
+              .setUseExternalService(Boolean.parseBoolean(System.getenv("USE_DOCKER_SERVICE")))
```
mfateev, line 53: I would move this logic (and Target property as well) into SDKTestWorkflowRule.Builder itself. By default, it should initialize the value of these properties to the value of the environment variable.

temporal-sdk/src/test/java/io/temporal/worker/WorkerStressTests.java
```diff
@@ -19,7 +19,7 @@
 
 package io.temporal.worker;
 
-import static io.temporal.workflow.WorkflowTest.NAMESPACE;
+import static io.temporal.testing.SDKTestWorkflowRule.NAMESPACE;
```
mfateev, line 22: I would make this constant private and use `testRule.getTestEnvironment().getNamespace()` instead

Also, consider adding the most used methods to the rule itself. For example `testRule.getNamespace()`, `testRule.newWorkflowStub()`, `testRule.currentTimeMillis()`, etc.

temporal-sdk/src/test/java/io/temporal/workflow/ActivityApplicationFailureRetryTest.java
```diff
@@ -46,12 +49,12 @@
...
   public void testActivityApplicationFailureRetry() {
-    WorkflowTest.TestWorkflow1 workflowStub =
+    TestWorkflows.TestWorkflow1 workflowStub =
         testWorkflowRule
             .getWorkflowClient()
             .newWorkflowStub(
-                WorkflowTest.TestWorkflow1.class,
-                WorkflowTest.newWorkflowOptionsBuilder(testWorkflowRule.getTaskQueue()).build());
+                TestWorkflows.TestWorkflow1.class,
+                TestOptions.newWorkflowOptionsBuilder(testWorkflowRule.getTaskQueue()).build());
```
mfateev, line 57: Consider adding `testWorklfowRule.newWorkflowStub(<class>)` that sets options to the task queue from `testWorkflowRule.` So the above code becomes:
```java
TestWorkflows.TestWorkflow1 workflowStub = testWorkflowRule.newWorklfowStub(TestWorkflows.TestWorkflow1.class);
```

mfateev (commented):
Great progress!

temporal-sdk/src/test/java/io/temporal/worker/WorkerStressTests.java
vkoby: NAMESPACE is used in shared/TestActivities.java for example. If we make it private, then how should it be accessed there?

temporal-sdk/src/test/java/io/temporal/workflow/AbandonOnCancelActivityTest.java
mfateev: Yes, by delegation I mean SDKTestWorkflowRule keeping TestWorkflowRule as a field and forwarding all method calls. 

BTW in IntelliJ, you can generate all these forwarding methods easily:

1. Command->N
1. Delegate Methods...
1. Select field which methods should be delegated
1. Select methods that should be delegated.

temporal-sdk/src/test/java/io/temporal/worker/WorkerStressTests.java
mfateev: You can pass namespace to a TestActivityImpl constructor.

temporal-testing-junit4/src/main/java/io/temporal/testing/TestWorkflowRule.java
```diff
@@ -59,13 +61,19 @@
  */
 public class TestWorkflowRule implements TestRule {
 
+  public static final String TEMPORAL_SERVICE_ADDRESS = System.getenv("TEMPORAL_SERVICE_ADDRESS");
+  // Only enable when USE_DOCKER_SERVICE is true
+  public static final Boolean USE_EXTERNAL_SERVICE =
+      Boolean.parseBoolean(System.getenv("USE_DOCKER_SERVICE"));
+  public static final String NAMESPACE = "UnitTest";
```
mfateev, line 68: All above fields belong to SDKTestWorkflowRule as they are specific to SDK testing. 
TestWorkflowRule is a general-purpose class to be used by developers that use Temporal SDK.

temporal-sdk/src/test/java/io/temporal/workflow/AbandonOnCancelActivityTest.java
```diff
@@ -30,42 +30,39 @@
...
   @Rule
-  public TestWorkflowRule testWorkflowRule =
-      TestWorkflowRule.newBuilder()
-          .setWorkflowTypes(TestAbandonOnCancelActivity.class)
-          .setActivityImplementations(activitiesImpl)
-          .setUseExternalService(Boolean.parseBoolean(System.getenv("USE_DOCKER_SERVICE")))
-          .setTarget(System.getenv("TEMPORAL_SERVICE_ADDRESS"))
-          .build();
+  public SDKTestWorkflowRule testWorkflowRule =
+      (SDKTestWorkflowRule)
```
mfateev, line 49: This cast is not needed anymore.

mfateev (approved):

temporal-sdk/src/test/java/io/temporal/workflow/shared/SDKTestWorkflowRule.java
```diff
@@ -0,0 +1,275 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.*;
+import org.junit.rules.TestRule;
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class SDKTestWorkflowRule implements TestRule {
```
vitarb, line 54: Do we really need this to be a rule wrapping TestWorkflowRule?
Can we just have a bunch of helpers/constants here and use TestWorkflowRule directly?

temporal-testing-junit4/src/main/java/io/temporal/testing/TestWorkflowRule.java
```diff
@@ -278,9 +283,20 @@ public String getTaskQueue() {
...
   /** Returns client to the Temporal service used to start and query workflows. */
   public WorkflowClient getWorkflowClient() {
-    return testEnvironment.getWorkflowClient();
+    if (useExtnlService) {
+      return WorkflowClient.newInstance(
+          testEnvironment.getWorkflowClient().getWorkflowServiceStubs(),
+          WorkflowClientOptions.newBuilder().setNamespace(testEnvironment.getNamespace()).build());
+    } else {
+      return testEnvironment.getWorkflowClient();
+    }
```
GreyTeardrop, line 299: Unless I miss something, this change removes the option to have `TimeLockingInterceptor` with external service via the `TestWorkflowRule`?

It might still be useful in some scenarios, e.g. TestContainers can be used to spin up a "real" Temporal instance that can still benefit from the time-rewinding logic of the `TestWorkflowEnvironment`.

temporal-testing-junit4/src/main/java/io/temporal/testing/TestWorkflowRule.java
```diff
@@ -59,64 +60,91 @@
  */
 public class TestWorkflowRule implements TestRule {
 
-  public Timeout globalTimeout;
+  private static final long DEFAULT_TEST_TIMEOUT_SECONDS = 10;
 
+  // Only enable when USE_DOCKER_SERVICE is true
+  private static final Boolean useExternalService =
+      Boolean.parseBoolean(System.getenv("USE_DOCKER_SERVICE"));
```
GreyTeardrop, line 67: Just a thought: this rule is used by clients that would be testing more than just Temporal, so in that context, `USE_DOCKER_SERVICE` might not be very meaningful.

temporal-sdk/src/test/java/io/temporal/workflow/shared/SDKTestWorkflowRule.java
mfateev: I believe that wrapping simplifies the unit tests. For example, it eliminates boilerplate code when initializing the test rule.

temporal-testing-junit4/src/main/java/io/temporal/testing/TestWorkflowRule.java
mfateev: The time skipping logic is not really supported when a real service is used. It requires changes to the service itself. We have plans to add such logic to the service. It would allow getting rid of the need to implement per SDK testing servcie.

temporal-testing-junit4/src/main/java/io/temporal/testing/TestWorkflowRule.java
mfateev: I agree. Let's remove `useExternalService` and `temporalServiceAddress` initialization from here and keep only in SDKTestWorkflowRule.

vitarb: Just to clarify, we are talking about removing environment variable and not about removing ability to run against external service per se?

vkoby: Correct

temporal-testing-junit4/src/main/java/io/temporal/testing/TestWorkflowRule.java
```diff
@@ -278,9 +279,20 @@ public String getTaskQueue() {
...
 
+  /** Returns name of the task queue that test worker is polling. */
+  public String getTaskQueue() {
+    return taskQueue;
+  }
+
   /** Returns client to the Temporal service used to start and query workflows. */
   public WorkflowClient getWorkflowClient() {
-    return testEnvironment.getWorkflowClient();
+    if (useExternalService) {
```
vitarb, line 289: Why do we need this? TestWorkflowEnvironment already handles workflow stubs creation and WorkflowClient creation.

temporal-testing-junit4/src/main/java/io/temporal/testing/TestWorkflowRule.java
```diff
@@ -192,32 +223,7 @@ public Builder setDoNotStart(boolean doNotStart) {
...
-      return new TestWorkflowRule(
-          testEnvironment,
-          useExternalService,
-          workflowTypes == null ? new Class[0] : workflowTypes,
-          activityImplementations == null ? new Object[0] : activityImplementations,
-          workerOptions,
-          testTimeoutSeconds == 0 ? DEFAULT_TEST_TIMEOUT_SECONDS : testTimeoutSeconds,
-          doNotStart,
-          workerInterceptors);
+      return new TestWorkflowRule((this));
```
vitarb, line 226: remove redundant parentheses

temporal-testing-junit4/src/main/java/io/temporal/testing/TestWorkflowRule.java
```diff
@@ -236,34 +242,29 @@ public void evaluate() throws Throwable {
...
+    worker.registerWorkflowImplementationTypes(workflowImplementationOptions, workflowTypes);
     worker.registerActivitiesImplementations(activityImplementations);
     return taskQueue;
   }
 
-  /** Returns name of the task queue that test worker is polling. */
-  public String getTaskQueue() {
-    return taskQueue;
+  private void start() {
+    if (!doNotStart) {
```
vitarb, line 255: I see this flag has been added recently by Maxim, is it really needed or can we add ability to `addWorkflowImplementationFactory` in the rule?

temporal-testing-junit4/src/main/java/io/temporal/testing/TestWorkflowRule.java
```diff
@@ -59,52 +60,75 @@
...
-    this.useExternalService = useExternalService;
-    this.workflowTypes = workflowTypes;
-    this.activityImplementations = activityImplementations;
-    this.workerOptions = workerOptions;
-    this.globalTimeout = Timeout.seconds(testTimeoutSeconds);
-    this.doNotStart = doNotStart;
-    this.interceptors = interceptors;
+  private TestWorkflowRule(Builder builder) {
+
+    String nameSpace = (builder.namespace == null) ? "UnitTest" : builder.namespace;
```
vitarb, line 85: nit. Lower case namespace please.

temporal-testing-junit4/src/main/java/io/temporal/testing/TestWorkflowRule.java
```diff
@@ -59,52 +60,75 @@
...
+            : builder.workflowClientOptions;
+    WorkerFactoryOptions factoryOptions =
+        WorkerFactoryOptions.newBuilder().setWorkerInterceptors(interceptors).build();
+    TestEnvironmentOptions testOptions =
+        TestEnvironmentOptions.newBuilder()
+            .setWorkflowClientOptions(clientOptions)
+            .setWorkerFactoryOptions(factoryOptions)
+            .setUseExternalService(useExternalService)
+            .setTarget(
+                builder.target == null ? System.getenv("TEMPORAL_SERVICE_ADDRESS") : builder.target)
```
vitarb, line 119: This environment variable should also be moved outside of this rule.

temporal-testing-junit4/src/main/java/io/temporal/testing/TestWorkflowRule.java
vkoby: I was trying to move this functionality from the individual tests that checked useExternalService (eg testChildWorkflowExecutionPromiseHandler()) into the rule and keep the code as readable as possible. Are you saying testEnvironment.getWorkflowClient() does exactly the same thing as in the other part of this if statement? If so you may be right and we don't need it.

temporal-testing-junit4/src/main/java/io/temporal/testing/TestWorkflowRule.java
mfateev: The idea was that some tests need to register test-specific workflows and activities which is not allowed after the start.

temporal-testing-junit4/src/main/java/io/temporal/testing/TestWorkflowRule.java
vitarb: Please take a look at [this code](https://github.com/temporalio/sdk-java/blob/ddeaa5f23eda9113fa23a32e924c078f1a38ae0c/temporal-testing/src/main/java/io/temporal/testing/TestWorkflowEnvironmentInternal.java#L75)

temporal-testing-junit4/src/main/java/io/temporal/testing/TestWorkflowRule.java
vitarb: I thought that we've agreed that having 1 to 1 mapping between rule and workflow types being tested is a good thing as it promotes clarity. From what I see the only thing that is missing in order to support what `DynamicWorkflowTest` needs is having ability to do `addWorkflowImplementationFactory` on the rule itself.

vitarb (commented):
LGTM

PR #372, opened 2021-03-01
[WIP] Experimental branch using custom doc tags

flossypurse:
This is an experimental branch that is testing the use of custom doc tags that work with a "language agnostic" reference generation tool that can be used with our docs site.

Preview of the resulting reference is available here:
https://github.com/temporalio/documentation/pull/261

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/activity/Activity.java
+++ b/temporal-sdk/src/main/java/io/temporal/activity/Activity.java
@@ -17,6 +17,9 @@
  *  permissions and limitations under the License.
  */
 
+/**
+ * @package io.temporal.activity
+ */
 package io.temporal.activity;
 
 import io.temporal.common.converter.DataConverter;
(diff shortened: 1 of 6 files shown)


temporal-sdk/src/main/java/io/temporal/activity/Activity.java
```diff
@@ -26,230 +29,40 @@
...
- * default implementation uses JSON serializer, but an alternative implementation can be configured
- * through {@link io.temporal.client.WorkflowClientOptions.Builder#setDataConverter(DataConverter)}.
- *
- * <p>Example of an interface that defines four activities:
- *
- * <pre><code>
- * {@literal @}ActivityInterface
- * public interface FileProcessingActivities {
- *
- *     void upload(String bucketName, String localName, String targetName);
```
Sushisource, line 47: Where'd all this move to? The conceptual docs? I would still want a link to that in these docstrings at minimum. Being able to get to there from here is important. I'm not seeing where it went in the linked example page.

flossypurse: This content was moved to https://docs.temporal.io/docs/java-activities
The SDK README will contain an explanation of where to find additional conceptual and implementation information. I am against putting links directly in the comment/doc strings for maintainability reasons.

Sushisource: Yeah, I can get that. We'll want some kind of solution though -- for example it's standard practice to have examples in docstrings in Rust code and they actually get compiled and run as tests to ensure they are always up to date.

So, maybe that means extracting the docs page from these comments if the goal is to avoid maintaining links, but it would be ideal to still have code examples easily accessible from the code. 

Whatever tool is using these annotations, can we just use it to insert backlinks?

Issue #373, opened 2021-03-01
Interceptors: Add missing header to activity and client interceptors


mfateev:
## Expected Behavior
It should be possible to inject context into the header when starting workflow and executing an activity.

## Actual Behavior
There is no way to set header from client interceptor and activity interceptor.

Spikhalskiy:
@mfateev this is fully implemented and probably can be closed now.

PR #375, opened 2021-03-04
InboundActivityInterceptor now has an access to header

Spikhalskiy:
Partially addresses #373.
Header is added into ActivityInboundCallsInterceptor#execute method input.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/common/interceptors/ActivityInboundCallsInterceptor.java
+++ b/temporal-sdk/src/main/java/io/temporal/common/interceptors/ActivityInboundCallsInterceptor.java
@@ -22,7 +22,42 @@ package io.temporal.common.interceptors;
 import io.temporal.activity.ActivityExecutionContext;
 
 public interface ActivityInboundCallsInterceptor {
+  final class ActivityInput {
+    private final Header header;
+    private final Object[] arguments;
+
+    public ActivityInput(Header header, Object[] arguments) {
+      this.header = header;
+      this.arguments = arguments;
+    }
+
+    public Header getHeader() {
+      return header;
+    }
+
+    public Object[] getArguments() {
+      return arguments;
+    }
+  }
+
+  final class ActivityOutput {
+    private final Object result;
+
+    public ActivityOutput(Object result) {
+      this.result = result;
+    }
+
+    public Object getResult() {
+      return result;
+    }
+  }
+
   void init(ActivityExecutionContext context);
 
-  Object execute(Object[] arguments);
+  /**
+   * Intercepts a call to the main activity entry method.
+   *
+   * @return result of the activity execution.
+   */
+  ActivityOutput execute(ActivityInput input);
 }
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/ActivityInfoImpl.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/ActivityInfoImpl.java
@@ -20,7 +20,6 @@
 package io.temporal.internal.sync;
 
 import com.google.protobuf.util.Timestamps;
-import io.temporal.activity.ActivityInfo;
 import io.temporal.api.common.v1.Header;
 import io.temporal.api.common.v1.Payloads;
 import io.temporal.api.workflowservice.v1.PollActivityTaskQueueResponse;
@@ -30,7 +29,7 @@ import java.time.Duration;
 import java.util.Objects;
 import java.util.Optional;
 
-final class ActivityInfoImpl implements ActivityInfo {
+final class ActivityInfoImpl implements ActivityInfoInternal {
   private final PollActivityTaskQueueResponse response;
   private final String activityNamespace;
   private final boolean local;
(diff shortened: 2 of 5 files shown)


temporal-sdk/src/main/java/io/temporal/internal/sync/ActivityInfoImpl.java
```diff
@@ -134,4 +135,8 @@ public boolean isLocal() {
     }
     return Optional.empty();
   }
+
+  public Header getHeader() {
```
vitarb, line 139: I think this should also be added into `ActivityInfo` interface.

temporal-sdk/src/main/java/io/temporal/common/interceptors/ActivityInboundCallsInterceptor.java
```diff
@@ -22,7 +22,25 @@
...
+
+    public Object[] getArguments() {
+      return arguments;
+    }
+  }
+
   void init(ActivityExecutionContext context);
 
-  Object execute(Object[] arguments);
+  Object execute(ActivityInput input);
```
vitarb, line 45: This would introduce backwards incompatible change to the public interface and may break anyone who is already using this interceptor.
I think we should think about alternative options.

temporal-sdk/src/main/java/io/temporal/internal/sync/POJOActivityTaskHandler.java
```diff
@@ -328,8 +332,13 @@ public DynamicActivityImplementation(DynamicActivity activity) {
...
       try {
-        EncodedValues args = new EncodedValues(input, dataConverter);
-        Object result = inboundCallsInterceptor.execute(new Object[] {args});
+        EncodedValues encodedValues = new EncodedValues(input, dataConverter);
+        Object[] args = new Object[] {encodedValues};
+
+        Object result =
+            inboundCallsInterceptor.execute(
+                new ActivityInboundCallsInterceptor.ActivityInput(
+                    new Header(info.getHeader()), args));
```
vitarb, line 341: This approach would only solve the problem of getting headers, but what about setting them?

temporal-sdk/src/main/java/io/temporal/internal/sync/ActivityInfoImpl.java
mfateev: I don't think we want to expose the Header protobuf to the activity code directly.

temporal-sdk/src/main/java/io/temporal/common/interceptors/ActivityInboundCallsInterceptor.java
mfateev: I think it is OK to break this. I doubt it is widely used and the change is localized. I would be against a breaking change that affects the actual workflow code.

temporal-sdk/src/main/java/io/temporal/internal/sync/POJOActivityTaskHandler.java
mfateev: What do you mean by "setting them"? The interceptor has full control over header content before passing it to the next interceptor.

temporal-sdk/src/main/java/io/temporal/common/interceptors/ActivityInboundCallsInterceptor.java
mfateev: If we are going to break this I would add `ActivityOutput` struct as well.

temporal-sdk/src/main/java/io/temporal/internal/sync/ActivityInfoImpl.java
Spikhalskiy: There is already a couple of methods like this (getInput and getCompletionHandle) - what do you guys think about introducing an interface like ActivityInfoInternal where we can define these methods and wouldn't allow this interface to escape TemporalSdk internals?

temporal-sdk/src/main/java/io/temporal/common/interceptors/ActivityInboundCallsInterceptor.java
Spikhalskiy: addressed

temporal-sdk/src/main/java/io/temporal/internal/sync/POJOActivityTaskHandler.java
Spikhalskiy: We already pass headers to Activity execution inside ```ExecuteActivityParameters#ScheduleActivityTaskCommandAttributes.Builder#setHeader```.
For that to actually do any useful work Interceptors or ContextPropagators have to actually add the headers first.

Spikhalskiy:
@vitarb @mfateev Addressed your comments, guys. Please revisit when you get a chance.

Spikhalskiy:
Can't open the Details and take a look at the reason for the test build fail, but I noticed that ActivityRetryOnTimeoutTest#testActivityRetryOnTimeout flakes badly during a full build.

vitarb (approved):

PR #383, opened 2021-03-11
Factored out Signal Tests

vkoby:


temporal-testing/src/main/java/io/temporal/testing/TestWorkflowEnvironmentInternal.java
```diff
@@ -150,7 +150,11 @@ public void close() {
     workerFactory.shutdownNow();
     workerFactory.awaitTermination(10, TimeUnit.SECONDS);
     service.close();
-    workflowServiceStubs.shutdown();
+    if (Boolean.parseBoolean(System.getenv("USE_DOCKER_SERVICE"))) {
+      workflowServiceStubs.shutdown();
+    } else {
+      workflowServiceStubs.shutdownNow();
+    }
```
vkoby, line 157: This change made a difference running builds locally, but doesn't seem to make any difference in Buildkite.

temporal-sdk/src/test/java/io/temporal/workflow/ExceptionInSignalTest.java
```diff
@@ -0,0 +1,100 @@
...
+    } catch (Exception e) {
+      // exception expected here.
+    }
+
+    // Suspend polling so that workflow tasks are not retried. Otherwise it will affect our thread
+    // count.
+    testWorkflowRule.getTestEnvironment().getWorkerFactory().suspendPolling();
+
+    // Wait for workflow task retry to finish.
+    Thread.sleep(5000);
```
vkoby, line 68: Used to be 10000ms. If the global timeout is 10s, how did it ever pass?

vitarb: I don't think that WorkflowTest enforced 10sec timeout.

temporal-testing/src/main/java/io/temporal/testing/TestWorkflowEnvironmentInternal.java
vitarb: We shouldn't do shutdownNow unless we want to drop running tasks. If all we want it to ensure that all inflight work is being processed before the shutdown, we should use `awaitTermination`.

temporal-sdk/src/test/java/io/temporal/workflow/SignalTest.java
```diff
@@ -0,0 +1,273 @@
...
+    try {
+      workflowClient.signalWithStart(batch4);
+      fail("DuplicateWorkflowException expected");
+    } catch (WorkflowExecutionAlreadyStarted e) {
+      assertEquals(execution3.getRunId(), e.getExecution().getRunId());
+    }
+  }
+
+  @Test
+  public void testSignalUntyped() {
```
vitarb, line 172: This test fails when you use docker service configuration because SDK test rule always uses in-memory service and never calls `setUseExternalService`. You need to fix that bug, otherwise the test is looking fine.

temporal-testing/src/main/java/io/temporal/testing/TestWorkflowEnvironmentInternal.java
vkoby: I propose to do shutdownNow only for in-memory tests. We do it after the test has already executed. So do you think it matters that we drop all the running tasks at that point? This way we don't have to explicitly set longer timeouts for a few tests.

vitarb: According to the documentation:
* shutdown() will just tell the executor service that it can't accept new tasks, but the already submitted tasks continue to run
* shutdownNow() will do the same AND will try to cancel the already submitted tasks by interrupting the relevant threads. Note that if your tasks ignore the interruption, shutdownNow will behave exactly the same way as shutdown

In my understanding, what we should try to achieve in the close method is semantics of shutdown + awaitTermination, which would not only reject new requests, but will also wait until all inflight requests finish. Otherwise there would be a chance that you call `close` but some requests are still inflight after `close` returns.

vitarb (approved):

temporal-testing/src/main/java/io/temporal/testing/TestWorkflowEnvironmentInternal.java
vitarb: Decided to land as is, because closing is done after the test. We will adjust accordingly in the future.

Spikhalskiy:
@vkoby After merging of this PR, full ```./gradlew clean build``` on master reliably gives ```Failed tests for temporal-sdk:
[io.temporal.workflow.SignalTest::testSignalUntyped]``` with
```
io.temporal.client.WorkflowFailedException: workflowId='e31a45a9-089d-4fe5-bf52-ea84e415e7c1', runId='72820b7a-34f5-4f48-ab59-c389697a9a01', workflowType='QueryableWorkflow', retryState=RETRY_STATE_TIMEOUT, workflowTaskCompletedEventId=0
	at io.temporal.internal.common.WorkflowExecutionUtils.getResultFromCloseEvent(WorkflowExecutionUtils.java:200)
	at io.temporal.internal.common.WorkflowExecutionUtils.getWorkflowExecutionResult(WorkflowExecutionUtils.java:134)
	at io.temporal.internal.sync.WorkflowStubImpl.getResult(WorkflowStubImpl.java:346)
	at io.temporal.internal.sync.WorkflowStubImpl.getResult(WorkflowStubImpl.java:328)
	at io.temporal.internal.sync.WorkflowStubImpl.getResult(WorkflowStubImpl.java:321)
	at io.temporal.testing.TestWorkflowEnvironmentInternal$TimeLockingInterceptor$TimeLockingWorkflowStub.getResult(TestWorkflowEnvironmentInternal.java:303)
	at io.temporal.workflow.SignalTest.testSignalUntyped(SignalTest.java:197)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at io.temporal.testing.TestWorkflowRule$2.evaluate(TestWorkflowRule.java:237)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:299)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:293)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.lang.Thread.run(Thread.java:834)
Caused by: io.temporal.failure.TimeoutFailure: timeoutType=TIMEOUT_TYPE_START_TO_CLOSE
	... 20 more
```

PR #397, opened 2021-03-24
Add Per Method Activity Options

vkoby:
Support per activity type options when configuring activity stub instance. Consider adding ActivityOptions annotation to the activity method.


temporal-sdk/src/main/java/io/temporal/activity/NamedActivityOption.java
```diff
@@ -0,0 +1,44 @@
...
+
+package io.temporal.activity;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * This annotation applies only to activity interface & interface methods. Not required. Use it to
```
Sushisource, line 28:
```suggestion
 * This annotation may be applied to activity interface & interface methods. Use it to
```

PR #409, opened 2021-03-30
Make child workflowId available for Interceptors

Spikhalskiy:
This PR moves generation of child `workflowId` into `ChildWorkflowStubImpl` the same way it's done in WorkflowStubImpl: 
https://github.com/temporalio/sdk-java/blob/c802327089fe63694e1baca059a750e970d90b72/temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowStubImpl.java#L109

This makes child workflowId known and available for Outbound Interceptors


mfateev (approved):
Thank you!

PR #412, opened 2021-03-31
[Proposal] Decouple WorkflowOutboundCallsInterceptor from WorkflowInboundCallsInterceptor

Spikhalskiy:
We made a significant breaking change in interceptors in the current release and I propose this change as a part of improving and cleaning up interceptors interfaces.
Right now to define WorkflowOutboundCallsInterceptor you need to define and create it inside WorkflowInboundCallsInterceptor#init() which doesn't make too much sense at all, these two interfaces are not even related and shouldn't be coupled in this manner.

This PR decouples WorkflowInboundCallsInterceptor and WorkflowOutboundCallsInterceptor and allows WorkerInterceptor to define them independently, which keeps things cleaner, at the same time keeping an ability for them to be defined in pairs (inbound/outbound) when it's desired in the same place.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkerInterceptor.java
+++ b/temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkerInterceptor.java
@@ -25,7 +25,9 @@ package io.temporal.common.interceptors;
  * <p>TODO(maxim): JavaDoc with sample
  */
 public interface WorkerInterceptor {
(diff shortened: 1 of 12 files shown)


Spikhalskiy:
Please don't review this one yet, it's created as a proposal for discussion.

mfateev:
This change would make it practically impossible to share per workflow execution state between inbound and outbound interceptors.

PR #418, opened 2021-04-01
Add OpenTracing module

### LINKED ISSUE #258: Add opentracing context propagator

Spikhalskiy:
## What was changed:
This PR adds a new temporal-opentracing module with support for OpenTracing context propagation from the client to workflow and to activity execution.

## Why?
We want to be able to tie all the parts of the workflow execution under the same client parent OpenTracing span for tracing and debugging.

## How it can be used?

See README provided with this module: https://github.com/temporalio/sdk-java/pull/418/files#diff-a87181e0e99e6147518982b133e66674da61be50e9898fe9d1eab9dc91ab2cc6R5

## Closes
#258


Spikhalskiy:
There will be a couple more things later today:
1. moving options into optional input parameters
2. failures logging into Span

But this is ready for the main review and feedback.

temporal-opentracing/src/main/java/io/temporal/opentracing/OpenTracingConfig.java
```diff
@@ -0,0 +1,125 @@
...
+ *  or in the "license" file accompanying this file. This file is distributed on
+ *  an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
+package io.temporal.opentracing;
+
+public class OpenTracingConfig {
+  private final String workflowStartOperationNamePrefix;
```
vitarb, line 23: I would put these into the map and make key an enum (e.g. TracingNamePrefix with options like Start, StartWithSignal, etc.). This should make code a bit cleaner both here and on the consumer end.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingContextPropagationUtils.java
```diff
@@ -0,0 +1,99 @@
...
+import io.temporal.common.converter.DataConverter;
+import io.temporal.common.interceptors.Header;
+import java.lang.reflect.Type;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Optional;
+
+public class OpenTracingContextPropagationUtils {
+  private static final String tracerHeaderKey = "_tracer-data";
```
vitarb, line 37: Name it `TRACER_HEADER_KEY` since it's a constant.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingContextPropagationUtils.java
```diff
@@ -0,0 +1,99 @@
...
+import io.temporal.api.common.v1.Payload;
+import io.temporal.common.converter.DataConverter;
+import io.temporal.common.interceptors.Header;
+import java.lang.reflect.Type;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Optional;
+
+public class OpenTracingContextPropagationUtils {
```
vitarb, line 36: I would prefer more specific name here, like `OpenTracingContextAccessor` or something along those lines. Util classes tend to become garbage cans very quickly.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingContextPropagationUtils.java
```diff
@@ -0,0 +1,99 @@
...
+  private static final String tracerHeaderKey = "_tracer-data";
+
+  public static void writeSpanContextToHeader(
```
vitarb, line 39: Did you consider making these non-static and instantiating this accessor as part of init?

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingSpanUtils.java
```diff
@@ -0,0 +1,150 @@
...
+      String spanOperationName,
+      String workflowId,
+      String parentWorkflowId,
+      String parentRunId) {
+    Map<String, String> tags =
+        new HashMap<String, String>() {
+          {
+            put("workflowTag", workflowId);
+            put("parentWorkflowId", parentWorkflowId);
+            put("parentRunId", parentRunId);
```
vitarb, line 56: These tags should probably be a separate enum or a set of constants so it's easy to see what the full set of tags is.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingSpanUtils.java
```diff
@@ -0,0 +1,150 @@
...
+
+public class OpenTracingSpanUtils {
+  public static Tracer.SpanBuilder createWorkflowStartSpan(
+      Tracer tracer, long startTimeMs, String spanOperationName, String workflowId) {
+    Map<String, String> tags =
+        new HashMap<String, String>() {
+          {
+            put("workflowTag", workflowId);
+          }
+        };
```
vitarb, line 40: You can use ImmutableMap.of instead.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingSpanUtils.java
```diff
@@ -0,0 +1,150 @@
...
+      String workflowId,
+      String runId,
+      SpanContext workflowStartSpanContext) {
+    Map<String, String> tags =
+        new HashMap<String, String>() {
+          {
+            put("workflowTag", workflowId);
+            put("runTag", runId);
+          }
+        };
```
vitarb, line 75: Same as above.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingSpanUtils.java
```diff
@@ -0,0 +1,150 @@
...
+
+  public static Tracer.SpanBuilder createActivityStartSpan(
+      Tracer tracer, long startTimeMs, String spanOperationName, String workflowId, String runId) {
+    Map<String, String> tags =
+        new HashMap<String, String>() {
+          {
+            put("workflowTag", workflowId);
+            put("runTag", runId);
+          }
+        };
```
vitarb, line 93: ditto

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingSpanUtils.java
```diff
@@ -0,0 +1,150 @@
...
+      String workflowId,
+      String runId,
+      SpanContext activityStartSpanContext) {
+    Map<String, String> tags =
+        new HashMap<String, String>() {
+          {
+            put("workflowTag", workflowId);
+            put("runTag", runId);
+          }
+        };
```
vitarb, line 110: ditto

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingSpanUtils.java
```diff
@@ -0,0 +1,150 @@
...
+            put("parentRunId", parentRunId);
+          }
+        };
+    return createSpan(tracer, startTimeMs, spanOperationName, tags, null, References.FOLLOWS_FROM);
+  }
+
+  public static Tracer.SpanBuilder createWorkflowRunSpan(
+      Tracer tracer,
+      long startTimeMs,
+      String spanOperationName,
```
vitarb, line 65: Since we are already in a context of a span, I would rename this to just `operationName`.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowInboundCallsInterceptor.java
```diff
@@ -0,0 +1,68 @@
...
+    Tracer tracer = GlobalTracer.get();
+    SpanContext rootSpanContext =
+        OpenTracingContextPropagationUtils.readSpanContextFromHeader(input.getHeader(), tracer);
+    Span workflowRunSpan =
+        OpenTracingSpanUtils.createWorkflowRunSpan(
+                tracer,
+                Workflow.currentTimeMillis(),
+                openTracingConfig.getWorkflowRunOperationNamePrefix()
+                    + "-"
+                    + Workflow.getInfo().getWorkflowType(),
```
vitarb, line 57: Why do we need to add workflow type at the end? Should we add it into tags instead?

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowOutboundCallsInterceptor.java
```diff
@@ -0,0 +1,104 @@
...
+
+  private <R> Tracer.SpanBuilder createChildWorkflowStartSpanBuilder(
+      Tracer tracer, ChildWorkflowInput<R> input) {
+    WorkflowInfo parentWorkflowInfo = Workflow.getInfo();
+    return OpenTracingSpanUtils.createChildWorkflowStartSpan(
+        tracer,
+        System.currentTimeMillis(),
+        openTracingConfig.getChildWorkflowStartOperationNamePrefix()
+            + "-"
+            + input.getWorkflowType(),
```
vitarb, line 99: Same as above, should we put workflow type into tags instead of part of the operation name?

vitarb (changes requested):
Great work Dmitry, I left some comments here, please ping me on slack if you have questions.

temporal-opentracing/src/main/java/io/temporal/opentracing/OpenTracingConfig.java
Spikhalskiy: Good idea, implemented.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingContextPropagationUtils.java
Spikhalskiy: The name OpenTracingContextAccessor should be used for a helper class with non-static methods, but I don't mind, renamed.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingContextPropagationUtils.java
Spikhalskiy: I don't really see too much of a benefit here, this class doesn't really have anything it doesn't take as an input parameter. I can make it a helper if you really prefer, but the method signatures will be totally the same I think.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingSpanUtils.java
Spikhalskiy: Yeah, I thought about it, but I actually don't bring a dependency on guava in this module. I know that `temporal-sdk` has a guava dependency anyway, but I don't see enough reasons to make this module dependant on guava too JUST for this small helper inline map factory method. I will maybe consider if at least something more useful from there will be needed.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowInboundCallsInterceptor.java
Spikhalskiy: Just WorkflowStart or WorkflowRun as an operation name is very undescriptive and pretty meaningless.

A definition/best practice from [OpenTracing Specification](https://github.com/opentracing/specification/blob/master/specification.md):

> An operation name, a human-readable string which concisely represents the work done by the Span (for example, an RPC method name, a function name, or the name of a subtask or stage within a larger computation). The operation name should be the most general string that identifies a (statistically) interesting class of Span instances. That is, "get_user" is better than "get_user/314159".

Also from [Open Tracing Guide](https://opentracing.io/guides/golang/spans/):

> The reason for choosing general operation names is to allow the tracing systems to do aggregations. For example, Jaeger tracer has an option of emitting metrics for all the traffic going through the application. Having a unique operation name for each span would make the metrics useless. For application or library developers who wish to capture the program arguments in the traces to distinguish them, the recommended solution is to annotate spans with tags or logs. These are discussed in the section below.
> 
> Each operation name should have some semantic meaning to it and should be chosen such that its cardinality is neither too high nor too low. The cardinality of an operation name is defined as follows:
> 
> card(operation X) = total number of spans of operation X
> 
> For example, when using http.url (as operation name) the cardinality would be too high, on the other hand if http.method is chosen as the operation name then the cardinality would be too low.

I think that my current solution is most in line with these guidelines.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingSpanUtils.java
```diff
@@ -0,0 +1,150 @@
...
+import io.opentracing.Span;
+import io.opentracing.SpanContext;
+import io.opentracing.Tracer;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+import javax.annotation.Nullable;
+
+public class OpenTracingSpanUtils {
+  public static Tracer.SpanBuilder createWorkflowStartSpan(
```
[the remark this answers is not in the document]
Spikhalskiy: Not exactly. Each of these methods, other than just tags, encodes two things - a reference type and if explicit parent span is expected/allowed in this scenario.
I would prefer to leave them the current way and it was my intention to make these methods in such a rigid and specific way, so each scenario is super clean - what are the expected tags, if we expect explicitly passed parent span, and what is the reference type. And we have a single place where we can see the difference between these scenarios.

Spikhalskiy:
@vitarb Thanks for your review!

- Addressed PR comments
- Implemented logging of workflow/activity failures
- Reworked and exposed OpenTracingOptions

Ready to be revisited.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingContextAccessor.java
```diff
@@ -0,0 +1,95 @@
...
+    Map<String, String> serializedSpanContext =
+        DataConverter.getDefaultInstance()
+            .fromPayload(payload, HashMap.class, HASH_MAP_STRING_STRING_TYPE);
+    return deserializeSpanContextFromMap(serializedSpanContext, tracer);
+  }
+
+  private static SpanContext deserializeSpanContextFromMap(
+      Map<String, String> serializedSpanContext, Tracer tracer) {
+    HashMapBackedTextMap contextTextMap = new HashMapBackedTextMap(serializedSpanContext);
+    return tracer.extract(Format.Builtin.TEXT_MAP, contextTextMap);
```
orastosk, line 69: you can use io.opentracing.propagation.TextMapAdapter from opentracing-api.jar and remove local HashMapBackedTextMap.
```java
return tracer.extract(Format.Builtin.TEXT_MAP, new TextMapAdapter(serializedSpanContext));
```

Spikhalskiy: That's neat, I replaced this custom `HashMapBackedTextMap` class with standard opentracing classes from `io.opentracing.propagation` package. Thank you for pointing this out!

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingSpanUtils.java
vitarb: We have an API dependency on temporal-sdk, which should bring in guava.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowInboundCallsInterceptor.java
vitarb: It actually goes against what you've quoted above, specifically "operation name should be the most general string" as we concatenate specific workflow type to operation.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingContextPropagationUtils.java
vitarb: I generally have a preference towards non-static helpers, they are less prone to misuse. But I'm okay leaving it as is.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingSpanUtils.java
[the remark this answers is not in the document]
vitarb: I could see the benefit if these functions were called multiple times, but they are called only once each, which only unnecessarily complicates the code by adding another level of indirection.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingSpanUtils.java
Spikhalskiy: That's exactly what I'm saying. You don't suppose to use transitive dependencies in your module, it's actually not a good style. And this is why Gradle has a separation between implementation and API scope. And, by the way, guava is defined as implementation scoped (as it should be). Whenever you use a library directly in your module you should have an own direct dependency defined in the project module. And I don't want to bring the whole guava for a small factory method, it's not a good practice.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingSpanUtils.java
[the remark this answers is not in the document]
Spikhalskiy: I think that this is a good way. It's not a level of indirection, it's a level of incapsulation. Some permutations of parameters are incorrect and these methods provide the required incapsulation to ensure the correctness of permutations of parameter values.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowInboundCallsInterceptor.java
Spikhalskiy: But you just took the fragment out of the context. The full phrase is "most general string that identifies a (statistically) interesting class of Span instances". And there are more phrases around that explain what does it mean. If you open the first link, after this phrase there is a specific example with good and bad examples:

Operation Name | Guidance
-- | --
get | Too general
get_account/792 | Too specific
get_account | Good, and account_id=792 would make a nice Span tag

You offer to make an operation name EVEN MORE generic thing than this `get` example.
I'm standing on the fact that I spent a good amount of time investigating the different implementation and open tracing guidelines and my current solution is much close to the guidelines.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingSpanUtils.java
mfateev: Agree with @Spikhalskiy on not bringing large direct dependencies without a good reason.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingSpanUtils.java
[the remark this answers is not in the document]
mfateev: I like @Spikhalskiy  approach more. Mostly because all the ugliness of creating tag hashmaps is encapsulated in one place. I think this is about the style which is very subjective. I would leave it as it is just to avoid bikeshedding.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingContextAccessor.java
```diff
@@ -0,0 +1,70 @@
...
+
+public class OpenTracingContextAccessor {
+  private static final String TRACER_HEADER_KEY = "_tracer-data";
+  private static final Type HASH_MAP_STRING_STRING_TYPE =
+      TypeToken.getParameterized(HashMap.class, String.class, String.class).getType();
+
+  public static void writeSpanContextToHeader(
+      SpanContext spanContext, Header header, Tracer tracer) {
+    Map<String, String> serializedSpanContext = serializeSpanContextToMap(spanContext, tracer);
+    Optional<Payload> payload = DataConverter.getDefaultInstance().toPayload(serializedSpanContext);
```
mfateev, line 42: Shouldn't we use[ tracer.inject ](https://opentracing.io/guides/java/inject-extract/) and `tracer.extract` instead of DataConverter?

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowClientCallsInterceptor.java
```diff
@@ -0,0 +1,74 @@
...
+      WorkflowClientCallsInterceptor next, OpenTracingOptions openTracingOptions) {
+    super(next);
+    this.openTracingOptions = openTracingOptions;
+  }
+
+  @Override
+  public WorkflowStartOutput start(WorkflowStartInput input) {
+    createAndPassWorkflowStartSpan(
+        input, openTracingOptions.getSpanOperationNamePrefix(SpanOperationType.WORKFLOW_START));
+    return super.start(input);
```
mfateev, line 44: I think we should close this span only after the start gRPC call completes.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowClientCallsInterceptor.java
```diff
@@ -0,0 +1,74 @@
...
+    return super.signalWithStart(input);
+  }
+
+  private static void createAndPassWorkflowStartSpan(
+      WorkflowStartInput input, String spanOperationNamePrefix) {
+    Tracer tracer = GlobalTracer.get();
+    Tracer.SpanBuilder workflowSpan =
+        createWorkflowStartSpanBuilder(tracer, input, spanOperationNamePrefix);
+    Span span = workflowSpan.start();
+    span.finish();
```
mfateev, line 62: don't close before start call completed

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowInboundCallsInterceptor.java
```diff
@@ -0,0 +1,74 @@
...
+                Workflow.getInfo().getRunId(),
+                rootSpanContext)
+            .start();
+    try (Scope scope = tracer.scopeManager().activate(workflowRunSpan)) {
+      return super.execute(input);
+    } catch (Throwable t) {
+      OpenTracingSpanUtils.logFail(workflowRunSpan, t);
+      throw t;
+    } finally {
+      workflowRunSpan.finish();
```
mfateev, line 71: This is going to close the workflow span every time a workflow is unloaded from the worker cache. I think it is pretty confusing to see many spans instead of one span for a workflow. I propose the following solution to this:

Any span created inside the workflow should be created inside a `SideEffect` function. It should be serialized using `Tracer.inject` as SideEffect requires its result to be serializable as it is stored as a _marker_ event in the event history. Then it has to be deserialized using `Tracer.extract` before use. 

`workflowRunSpan.finish()` should not be called on exceptions that extend Error, especially `DestroyWorkflowThreadError` which is used to release threads on workflow unloading from memory. The interceptor interface should be extended to give a clear indication if an exception is going to fail workflow or workflow task only.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowOutboundCallsInterceptor.java
```diff
@@ -0,0 +1,107 @@
...
+
+  public OpenTracingWorkflowOutboundCallsInterceptor(
+      WorkflowOutboundCallsInterceptor next, OpenTracingOptions openTracingOptions) {
+    super(next);
+    this.openTracingOptions = openTracingOptions;
+  }
+
+  @Override
+  public <R> ActivityOutput<R> executeActivity(ActivityInput<R> input) {
+    createAndPassActivityStartSpan(input.getActivityName(), input.getHeader());
```
mfateev, line 45: I would change it to a span that wraps activity execution, not only its invocation. This comment applies to all other wrapped APIs.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowClientCallsInterceptor.java
Spikhalskiy: It's go-sdk approach though I believe: https://github.com/temporalio/sdk-go/blob/cec615e1aaf798e1b153ba3918b4d3027edde808/internal/internal_workflow_client.go#L199

I'm not against your proposal, we can close in finally. But can you give more motivation or rasoning here? These two solutions lead to the same result for me, maybe I don't see something.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingContextAccessor.java
Spikhalskiy: Not really. tracer.inject and tracer.extract are working with TEXT_MAPs. So, we convert a span to a TEXT_MAP using tracer.inject inside serializeSpanContextToMap, but after that we need to serialize this map for temporal headers and we use DataConverter for that. tracer.inject is used here, it just happened inside serializeSpanContextToMap.

temporal-opentracing/src/main/java/io/temporal/opentracing/SpanOperationType.java
```diff
@@ -0,0 +1,39 @@
...
+ *  an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
+package io.temporal.opentracing;
+
+public enum SpanOperationType {
+  WORKFLOW_START("StartWorkflow"),
+  WORKFLOW_START_WITH_SIGNAL("SignalWithStartWorkflow"),
```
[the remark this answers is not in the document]
Spikhalskiy: Alright. I thoughtt that it should be ok to keep SpanTypes and their prefix a little bit different and kind of like the current version, but if you think it's better to make enum names match prefixes - will do it.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowInboundCallsInterceptor.java
Spikhalskiy: This is how OpenTelemetry UIs look like:
c6463a2b-trace-300
0_WDqZqICj-V23_1YW
aanuraag_Distributed-Tracing-OpenTelemetry_f1
To get to tags you need to click on a specific span and open the details. How the operator will work with a sequence of indistinguishable ActivityRun spans?

Also, see OpenTracing implementation for JaxRS2:
https://github.com/opentracing-contrib/java-jaxrs/blob/master/opentracing-jaxrs2/src/main/java/io/opentracing/contrib/jaxrs2/server/OperationNameProvider.java#L46

> Default Microprofile operation name <HTTP method>:<package name>.<Class name>.<method name>
<package name>.<Class name>.<method name> is our ActivityType/WorkflowType

They do provide an alternative strategy
```
  /**
   * Returns HTTP method as operation name
   */
  class HTTPMethodOperationName implements OperationNameProvider {
```
but it's not a default

temporal-opentracing/src/main/java/io/temporal/opentracing/SpanOperationType.java
```diff
@@ -0,0 +1,39 @@
...
+  WORKFLOW_RUN("RunWorkflow"),
+  CHILD_WORKFLOW_START("StartChildWorkflow"),
+  ACTIVITY_START("StartActivity"),
+  ACTIVITY_RUN("RunActivity");
+
+  private String defaultPrefix;
+
+  SpanOperationType(String defaultPrefix) {
```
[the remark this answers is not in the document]
Spikhalskiy: Any Enum constructor in java is private, it can't be anything else.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingActivityInboundCallsInterceptor.java
```diff
@@ -0,0 +1,81 @@
...
+    // Workflow Interceptors have access to Workflow.getInfo methods,
+    // but Activity Interceptors don't have access to Activity.getExecutionContext().getInfo()
+    // This is inconsistent and should be addressed, but this is a workaround.
+    this.activityExecutionContext = context;
+    super.init(context);
+  }
+
+  @Override
+  public ActivityOutput execute(ActivityInput input) {
+    Tracer tracer = GlobalTracer.get();
```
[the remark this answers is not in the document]
Spikhalskiy: From what I know - GlobalTracer is actually the conventional way to tell all components which tracer to use. But if we really want such flexibility to be able to provide something else to temporal - we can expose AtomicReference<Tracer> in OpenTracingOptions. It's not pretty, but we should pass a container with a tracer, not the tracer itself to keep a way to replace the Tracer later (like we can do with replacing the tracer in GlobalTracer).

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingContextAccessor.java
```diff
@@ -0,0 +1,70 @@
...
+ *
+ *  or in the "license" file accompanying this file. This file is distributed on
+ *  an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
+package io.temporal.opentracing.internal;
+
+import com.google.gson.reflect.TypeToken;
```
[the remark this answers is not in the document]
Spikhalskiy: Replaced with TypeToken from guava.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingSpanUtils.java
Spikhalskiy: Another suggestion brought guava anyway, so now replaced with ImmutableMap

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowOutboundCallsInterceptor.java
Spikhalskiy: For wrapping the activity execution with have a RunActivity span. It wraps the execution. This is a StartActivity span. It just basically shows that the activity was triggered/scheduled. I think we should close the span either immediately or after invocation to show in the trace that it happened. It can be especially important for WorkflowStart span of long-living activity. You don't want to wait till the end of a workflow to publish a root span.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowInboundCallsInterceptor.java
Spikhalskiy: What do you think about making this a separate change and PR and implement it as an improvement for the current solution?
The current approach is simple, it works and it could be good enough for the majority of cases.

Also, this idea looks very brittle for me and I think it will not work. The first problem that I see is actually in the root of the proposal:
> This is going to close the workflow span every time a workflow is unloaded from the worker cache.
> workflowRunSpan.finish() should not be called on exceptions that extend Error, especially DestroyWorkflowThreadError which is used to release threads on workflow unloading from memory.

We should call finish in all these cases. Otherwise, imagine a workflow started today and finishing a month from now. A workflow run span of this workflow will not be finished till a month from now? It will not be reported to the server in that case and we will not have any trace.

We should actually wrap every "awakening" as a separate span to keep spans published without huge delays.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowClientCallsInterceptor.java
Spikhalskiy: same ^. go-sdk does it the same way too.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingSpanUtils.java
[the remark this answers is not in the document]
Spikhalskiy: Made it a non-static factory and hid the logic of span operation name creation inside to give an easy entry point for providing extensible strategies in the future if needed.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingContextAccessor.java
mfateev: What about using BINARY format to avoid JSON type serialization? You still would need to use DataConverter to create the Payload, but it will use the binary payload as it is.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowClientCallsInterceptor.java
mfateev: My idea was to have a span that includes the start gRPC call latency and possibly retries. I'm not sure if it is useful.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowInboundCallsInterceptor.java
mfateev: OK. Let's keep your approach. I didn't realize that children of non-closed span are not reported. So it is going to be useless for long running workflows.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingContextAccessor.java
Spikhalskiy: I don’t mind, I don’t see problems with both approaches. If you prefer Binary - let’s go with Binary. Will delete guava and rework back to Hash Maps then :) Temporal Go SDK uses TEXT_MAP + converter approach though. You sure you want it differently between SDKs?

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowClientCallsInterceptor.java
Spikhalskiy: Alright, makes sense, let me do it this way.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingContextAccessor.java
mfateev: I don't have a strong opinion on this. If you feel that map is more debuggable let's keep it as it is.

mfateev (approved):

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingContextAccessor.java
Spikhalskiy: Yeah, me too. Let's go with Map. It looks like it's the most used and referenced implementation and it's also in agreement with how things are done in Go SDK.

vitarb (approved):

PR #421, opened 2021-04-01
Testworkflowrule Interceptor Fix

vkoby:
TestWorkflowRule first got interceptor from a setter method on the rule and after that setWorkerFactoryOptions forced rewriting of WorkerFactoryOptions#workerInterceptors with collection provided (or nor provided) in the separate setWorkerInterceptors(). The entire interceptor setting logic was moved to the Rule, and the interceptor field was removed.


temporal-testing-junit4/src/main/java/io/temporal/testing/TestWorkflowRule.java
```diff
@@ -155,7 +158,13 @@ public Builder setWorkerOptions(WorkerOptions options) {
     }
 
     public Builder setWorkerFactoryOptions(WorkerFactoryOptions options) {
-      this.workerFactoryOptions = options;
+      this.workerFactoryOptions =
+          (options.getWorkerInterceptors() == null)
```
mfateev, line 162: I don't think getWorkerInterceptors ever returns null: https://github.com/temporalio/sdk-java/blob/b3797d6f0dcf174731dd1c5596a3a2e4bcaf2dbb/temporal-sdk/src/main/java/io/temporal/worker/WorkerFactoryOptions.java#L182

It looks like there is no unit test that validates both branches of this condition.

vitarb:
I don't like tight coupling of the TestWorkflowRule and TracingWorkerInterceptor, we should consider keeping it separate. I don't understand what warrants removal of the workerInterceptors from TestWorkflowRule.Builder. Let's discuss offline in case if I'm missing some context.

temporal-testing-junit4/src/main/java/io/temporal/testing/TestWorkflowRule.java
vkoby: Note "validate" param in the constructor. By default it's false and workerInterceptors is null

mfateev (approved):

PR #427, opened 2021-04-06
Disable flaky tests until they are fixed

vitarb:
We've accumulated a number of flaky tests in here, while we should investigate and fix them properly, it would take some time and having them enabled is counter productive.
I suggest we disable most influencing ones and create a task for the follow up.
We should be able to reproduce these failures locally using docker compose configuration for inducing network delays that was recently added in our core images.


Sushisource (approved):
LGTM. I suggest we link notion task here -- and put the exact wording you used in the TODOs in the notion task so they can be searched easily

PR #429, opened 2021-04-06
Fix a potential NullPointer in exception processing inside WorkflowStubImpl

Spikhalskiy:
## What was changed:

Address the problem with `WorkflowStubImpl#startWithOptions` and `WorkflowStubImpl#signalWithStartWithOptions` passing `workflowExecution` that could be not initialized into NotNull `workflowExecution` param of `WorkflowServiceException` constructor

## Why?

Preserves an original exception from an Interceptor if it happens instead of replacing it with a meaningless NullPointer.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowStubImpl.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowStubImpl.java
@@ -107,18 +107,30 @@ class WorkflowStubImpl implements WorkflowStub {
   private WorkflowExecution startWithOptions(WorkflowOptions options, Object... args) {
     checkExecutionIsNotStarted();
     String workflowId = getWorkflowIdForStart(options);
+    WorkflowExecution workflowExecution = null;
     try {
       WorkflowClientCallsInterceptor.WorkflowStartOutput workflowStartOutput =
           workflowClientInvoker.start(
               new WorkflowClientCallsInterceptor.WorkflowStartInput(
                   workflowId, workflowType.get(), Header.empty(), args, options));
-      WorkflowExecution workflowExecution = workflowStartOutput.getWorkflowExecution();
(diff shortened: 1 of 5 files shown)


temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowStubImpl.java
```diff
@@ -107,18 +108,25 @@ public void signal(String signalName, Object... args) {
...
       execution.set(workflowExecution);
       return workflowExecution;
     } catch (StatusRuntimeException e) {
       throw wrapStartException(workflowId, workflowType.orElse(null), e);
     } catch (Exception e) {
-      throw new WorkflowServiceException(execution.get(), workflowType.orElse(null), e);
+      throw new WorkflowServiceException(
+          // if start failed with exception - there could be no valid workflow execution.
+          // WorkflowServiceException requires not null workflowExecution, so we have to provide
+          // an empty default WorkflowExecution instance
```
mfateev, line 126: Why empty? WorkflowId is known here.

temporal-sdk/src/test/java/io/temporal/workflow/interceptorsTests/InterceptorsExceptionsTests.java
```diff
@@ -0,0 +1,86 @@
...
+    TestWorkflows.NoArgsWorkflow workflowStub =
+        testWorkflowRule
+            .getWorkflowClient()
+            .newWorkflowStub(
+                TestWorkflows.NoArgsWorkflow.class,
+                TestOptions.newWorkflowOptionsWithTimeouts(testWorkflowRule.getTaskQueue()));
+    try {
+      workflowStub.execute();
+      fail("Workflow call is expected to fail with an exception");
+    } catch (Exception e) {
```
mfateev, line 58: Nit: Why not catch WorkflowServiceException?

mfateev (approved):

Issue #370, opened 2021-03-01
Add per method activity options


mfateev:
**Is your feature request related to a problem? Please describe.**
From a[ forum post:](https://community.temporal.io/t/activities-requiring-different-options/752)

> Hello, Temporal!
I have an activity interface that defines three different (but related) activities, and each requires different options to be specified (e.g., timeouts). From the Java documentation:
Activity options are needed only if some of the required timeouts are not specified through the @ActivityMethod annotation.
This seems to suggest that @ActivityMethod annotations could be used to specify different options for each activity in a common interface. At the moment, it looks like ActivityMethod.java only supports overriding of the activity’s name.
So, is the only option for achieving different timeouts for different activities to instantiate multiple activity stubs within a workflow (each with its own ActivityOptions instance? And in this case, would distinct task queues be required for each stub?
Thanks!
Peter

**Describe the solution you'd like**
Support per activity type options when configuring activity stub instance. Consider adding ActivityOptions annotation to the activity method.

**Describe alternatives you've considered**
Keep the current implementation that would require a separate stub per activity type.

PR #431, opened 2021-04-07
Per Activity Method Options (Part I)

vkoby:
Added ability to pass activity method options to activityStub. Usage:

Map<activityTypeName, ActivityOptions> activityMethodOptions
Map<activityTypeName, LocalActivityOptions> localActivityMethodOptions

Workflow.newActivityStub(Activity.class, options, activityMethodOptions)
Workflow.newLocalActivityStub(Activity.class, options, localActivityMethodOptions)

WIP: Ability to pass activity method options with WorkflowImplementationOptions. 

Fix for issue #370


temporal-sdk/src/main/java/io/temporal/internal/sync/ActivityInvocationHandler.java
```diff
@@ -54,8 +60,13 @@ private ActivityInvocationHandler(
...
       Method method, MethodRetry methodRetry, String activityName) {
     Function<Object[], Object> function;
+    ActivityOptions methodOptions =
+        (activityMethodOptions == null) ? null : activityMethodOptions.get(method.getName());
     ActivityOptions mergedOptions =
-        ActivityOptions.newBuilder(options).mergeMethodRetry(methodRetry).build();
+        ActivityOptions.newBuilder(options)
+            .mergeMethodOptions(methodOptions)
+            .mergeMethodRetry(methodRetry)
+            .build();
```
mfateev, line 69: Move merging to the constructor.

temporal-sdk/src/main/java/io/temporal/internal/sync/ActivityInvocationHandler.java
```diff
@@ -54,8 +60,13 @@ private ActivityInvocationHandler(
   protected Function<Object[], Object> getActivityFunc(
       Method method, MethodRetry methodRetry, String activityName) {
     Function<Object[], Object> function;
+    ActivityOptions methodOptions =
+        (activityMethodOptions == null) ? null : activityMethodOptions.get(method.getName());
```
mfateev, line 64: I think we want to use the activity type name as the key to the map, not the method name.

temporal-sdk/src/main/java/io/temporal/workflow/Workflow.java
```diff
@@ -379,16 +388,21 @@
...
+   * @param activityInterface interface type implemented by activities.
+   * @param options options that together with the properties of {@link
+   *     io.temporal.activity.ActivityMethod} specify the activity invocation parameters.
    */
-  public static <T> T newActivityStub(Class<T> activityInterface) {
-    return WorkflowInternal.newActivityStub(activityInterface, null);
+  public static <T> T newActivityStub(
+      Class<T> activityInterface,
+      ActivityOptions options,
+      Map<String, ActivityOptions> perActivityMethodOptions) {
```
mfateev, line 404: Rename to perActivityTypeActivityOptions

temporal-sdk/src/main/java/io/temporal/activity/ActivityOptions.java
```diff
@@ -175,6 +175,44 @@ public Builder setCancellationType(ActivityCancellationType cancellationType) {
       return this;
     }
 
+    public Builder mergeMethodOptions(ActivityOptions methodOptions) {
```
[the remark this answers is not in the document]
vkoby: There's a mergeMethodRetry below, do you think we should follow the same naming pattern?

temporal-sdk/src/main/java/io/temporal/internal/sync/ActivityInvocationHandler.java
```diff
@@ -26,26 +26,32 @@
...
 public class ActivityInvocationHandler extends ActivityInvocationHandlerBase {
   private final ActivityOptions options;
+  private final Map<String, ActivityOptions> activityMethodOptions;
   private final WorkflowOutboundCallsInterceptor activityExecutor;
 
   @VisibleForTesting
   public static InvocationHandler newInstance(
       Class<?> activityInterface,
       ActivityOptions options,
+      Map<String, ActivityOptions> methodOptions,
```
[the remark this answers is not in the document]
vkoby: Not sure what you meant here. Annotations?

temporal-sdk/src/main/java/io/temporal/internal/sync/ActivityInvocationHandler.java
vkoby: I remember you saying that keying from the actual Method object would be best and it's very conveniently here. Should we do that instead?

temporal-sdk/src/main/java/io/temporal/internal/sync/ActivityInvocationHandler.java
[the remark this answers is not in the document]
mfateev: You can use @nonnull annotation. But it is not enough. I would add

this.methodOptions = Objects.requireNonNull(methodOptions);

Then you don't need this field nullness anywhere in the code that uses it.

temporal-sdk/src/main/java/io/temporal/internal/sync/ActivityInvocationHandler.java
mfateev: It is not as we planned to support the configuration of these options through a config file. In this case activity type name makes much more sense than the method name.

temporal-sdk/src/main/java/io/temporal/activity/ActivityOptions.java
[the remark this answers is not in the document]
mfateev: I see. Then I would call it `mergeActivityOptions` to follow that pattern.

temporal-sdk/src/main/java/io/temporal/worker/WorkflowImplementationOptions.java
```diff
@@ -40,6 +42,8 @@ public static Builder newBuilder() {
   public static final class Builder {
 
     private Class<? extends Throwable>[] failWorkflowExceptionTypes;
+    private Map<String, ActivityOptions> activityOptions;
+    private Map<String, Map<String, ActivityOptions>> activityToMethodOptions;
```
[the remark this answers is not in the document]
vkoby: Yup that's for the Worker implementation

temporal-sdk/src/test/java/io/temporal/activity/ActivityMethodOptionsTest.java
```diff
@@ -20,98 +20,135 @@
...
+        defaultOps.getStartToCloseTimeout(),
+        Duration.parse(method2OpsValues.get("StartToCloseTimeout")));
   }
 
   @ActivityInterface
   public interface TestActivity {
 
     @ActivityMethod
-    void method1();
+    Map<String, String> method1();
```
mfateev, line 118: Change the activity methods to return Map<String, Duration> to remove toString and parse.

temporal-testing/src/main/java/io/temporal/testing/TestActivityEnvironment.java
```diff
@@ -78,6 +81,36 @@ static TestActivityEnvironment newInstance(TestEnvironmentOptions options) {
...
   <T> T newActivityStub(Class<T> activityInterface);
 
+  /**
+   * Creates a stub that can be used to invoke activities registered through {@link
+   * #registerActivitiesImplementations(Object...)}.
+   *
+   * @param <T> Type of the activity interface.
+   * @param activityInterface activity interface class that the object under test implements
+   * @param options options that specify the activity invocation parameters
+   * @param activityMethodOptions activity method-specific invocation parameters
```
mfateev, line 91: Document that the key value is the activity type which is by default is the method name with the first capitalized letter.

temporal-testing/src/main/java/io/temporal/testing/TestActivityEnvironment.java
```diff
@@ -78,6 +81,36 @@ static TestActivityEnvironment newInstance(TestEnvironmentOptions options) {
...
+      Map<String, ActivityOptions> activityMethodOptions);
+
+  /**
+   * Creates a stub that can be used to invoke activities registered through {@link
+   * #registerActivitiesImplementations(Object...)}.
+   *
+   * @param <T> Type of the activity interface.
+   * @param activityInterface activity interface class that the object under test implements
+   * @param options options that specify the activity invocation parameters
+   * @param activityMethodOptions activity method-specific invocation parameters
```
mfateev, line 106: ditto

mfateev (approved):

Issue #135, opened 2020-06-26
Include local activity input into its marker decision


mfateev:
**Is your feature request related to a problem? Please describe.**
For troubleshooting having activity input available in the event history is very useful.

**Describe the solution you'd like**
Include the input into the marker decision/event. Make it configurable through LocalActivityOptions as for some use cases inputs can be large and increase history size beyond acceptable.

PR #441, opened 2021-04-13
Local type

mfateev:
## What was changed:

Changes to the MarkerRecordedEvent details map.
1. Added "input" that contains activity arguments. Can be disabled through `LocalActivityOptions.doNotIncludeArgumentsIntoMarker` to reduce history size.
2. Added "type" that contains the activity type name.
3. Renamed "data" to "result" for activity result.

Also did refactoring of unit tests to reduce repetitive code.

## Why?
Debugging local activities is hard as their type and arguments are not part of the event history.

## Checklist

1. Closes issue: 
Fix for https://github.com/temporalio/sdk-java/issues/135.

2. How was this tested:
Looking for the "input" field in the UI.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/activity/LocalActivityOptions.java
+++ b/temporal-sdk/src/main/java/io/temporal/activity/LocalActivityOptions.java
@@ -51,6 +51,7 @@ public final class LocalActivityOptions {
     private Duration localRetryThreshold;
     private Duration startToCloseTimeout;
     private RetryOptions retryOptions;
+    private boolean doNotIncludeArgumentsIntoMarker;
 
     /** Copy Builder fields from the options. */
     private Builder(LocalActivityOptions options) {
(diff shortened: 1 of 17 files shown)


temporal-sdk/src/main/java/io/temporal/activity/LocalActivityOptions.java
```diff
@@ -126,24 +147,28 @@ public LocalActivityOptions validateAndBuildWithDefaults() {
...
+          RetryOptions.newBuilder(retryOptions).validateBuildWithDefaults(),
+          doNotIncludeArgumentsIntoMarker);
     }
   }
 
   private final Duration scheduleToCloseTimeout;
   private final Duration localRetryThreshold;
   private final Duration startToCloseTimeout;
   private final RetryOptions retryOptions;
+  private boolean doNotIncludeArgumentsIntoMarker;
```
vitarb, line 159: nit. generally positive naming is preferable, can we call it `includeArgumentsIntoMarker` instead and flip the logic?

vitarb (approved):
LGTM, left a small comment, feel free to address or ignore.

temporal-sdk/src/main/java/io/temporal/activity/LocalActivityOptions.java
mfateev: I always follow the pattern that all properties defaults should match the type default. For a boolean the default value is false. That's why I named the property this way to make the false default value correct.

PR #447, opened 2021-04-16
getVersion sometimes works incorrectly when used inside a Timer

GreyTeardrop:
## Overview

This is not a merge request but a bug report. Please do not merge as-is.

We've recently discovered a couple of issues with the way timers and versions interact with each other. With a help of the test from this PR, I seem to be able to isolate one of the issues.

~~The test illustrates a workflow that starts 2 timers, sleeps twice, and potentially calls `getVersion` from timer callbacks. The test includes 2 implementations of the same workflow that differ only in whether they call `Workflow.getVersion` from the callbacks.~~
**Update**: The test illustrates a workflow that calls `getVersion` from the main workflow thread, starts a timer, and calls `getVersion` from the timer callback. The test includes 2 implementations of the same workflow that differ only in whether they call `Workflow.getVersion` from or not.

If no `getVersion` is called, the test results are what we would expect to get. The time at the end of the workflow is at least 2 hours later than the time at the start of the workflow.

However, if timer callbacks include `getVersion` calls then the test fails.

This test seems to be the simplest reproduction case I was able to get. ~~Both timers and both `getVersion` calls are required, otherwise, all tests succeed. Timers also have to be separated by a `Workflow.sleep` call, otherwise, both tests time out (which might be another issue?)~~
**Update**: A call from the main workflow thread and a call from a timer callback thread seems to be enough to break workflow's virtual time. All elements seem to be required now.

I've also one more test to reproduces another test environment issue. Seems like if 2 timers are fired at the same time the workflow hangs. The `TimerTwoConcurrentFiresTest` test fails with `test timed out after 10 seconds`. If timers are set to fire at 2 different intervals the test succeeds.

### DIFF
--- a/temporal-sdk/src/test/java/io/temporal/workflow/TimerTwoConcurrentFiresTest.java
+++ b/temporal-sdk/src/test/java/io/temporal/workflow/TimerTwoConcurrentFiresTest.java
@@ -1,0 +1,56 @@
+/*
+ *  Copyright (C) 2020 Temporal Technologies, Inc. All Rights Reserved.
+ *
+ *  Copyright 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ *  Modifications copyright (C) 2017 Uber Technologies, Inc.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License"). You may not
+ *  use this file except in compliance with the License. A copy of the License is
+ *  located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ *  or in the "license" file accompanying this file. This file is distributed on
+ *  an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
+package io.temporal.workflow;
+
+import io.temporal.workflow.shared.SDKTestWorkflowRule;
+import io.temporal.workflow.shared.TestWorkflows;
+import java.time.Duration;
+import org.junit.Rule;
+import org.junit.Test;
+
+public class TimerTwoConcurrentFiresTest {
+
+  @Rule
+  public SDKTestWorkflowRule testWorkflowRule =
+      SDKTestWorkflowRule.newBuilder()
+          .setWorkflowTypes(TestTimerTwoConcurrentFiresWorkflowImpl.class)
+          .build();
+
+  @Test
+  public void testTimerTwoConcurrentFires() {
+    TestWorkflows.TestWorkflow1 client =
+        testWorkflowRule.newWorkflowStubTimeoutOptions(TestWorkflows.TestWorkflow1.class);
+    client.execute(testWorkflowRule.getTaskQueue());
+  }
+
+  public static class TestTimerTwoConcurrentFiresWorkflowImpl
+      implements TestWorkflows.TestWorkflow1 {
+
+    @Override
+    public String execute(String taskQueue) {
+      Workflow.newTimer(Duration.ofMinutes(1));
+      Workflow.newTimer(Duration.ofMinutes(1));
+
+      Workflow.sleep(Duration.ofMinutes(10));
+
+      return "";
+    }
+  }
+}
--- a/temporal-sdk/src/test/java/io/temporal/workflow/versionTests/GetVersionAndTimerTest.java
+++ b/temporal-sdk/src/test/java/io/temporal/workflow/versionTests/GetVersionAndTimerTest.java
@@ -77,23 +77,16 @@ public class GetVersionAndTimerTest {
   abstract static class TimedWorkflowImpl implements TimedWorkflow {
     @Override
     public Instant startAndWait() {
-      Workflow.newTimer(Duration.ofMinutes(10))
+      getVersion();
+
+      Workflow.newTimer(Duration.ofMinutes(1))
           .thenApply(
               (v) -> {
                 getVersion();
                 return v;
               });
 
-      Workflow.sleep(Duration.ofHours(1));
-
-      Workflow.newTimer(Duration.ofMinutes(10))
-          .thenApply(
-              (v) -> {
-                getVersion();
-                return v;
-              });
-
-      Workflow.sleep(Duration.ofHours(1));
+      Workflow.sleep(Duration.ofHours(2));
 
       return Instant.ofEpochMilli(Workflow.currentTimeMillis());
     }
--- a/temporal-sdk/src/test/java/io/temporal/workflow/versionTests/GetVersionAndTimerTest.java
+++ b/temporal-sdk/src/test/java/io/temporal/workflow/versionTests/GetVersionAndTimerTest.java
@@ -1,0 +1,121 @@
+/*
+ *  Copyright (C) 2020 Temporal Technologies, Inc. All Rights Reserved.
+ *
+ *  Copyright 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ *  Modifications copyright (C) 2017 Uber Technologies, Inc.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License"). You may not
+ *  use this file except in compliance with the License. A copy of the License is
+ *  located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ *  or in the "license" file accompanying this file. This file is distributed on
+ *  an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
+package io.temporal.workflow.versionTests;
+
+import static org.junit.Assert.assertTrue;
+
+import io.temporal.workflow.Workflow;
+import io.temporal.workflow.WorkflowInterface;
+import io.temporal.workflow.WorkflowMethod;
+import io.temporal.workflow.shared.SDKTestWorkflowRule;
+import java.time.Duration;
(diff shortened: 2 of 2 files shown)


vitarb:
I've looked at the issue with versioning and timers and it indeed looks like a bug, time doesn't get moved after the sleep, but only if that timer calls getVersion. Thanks for the repro, we'll need to investigate this issue a bit more.

GreyTeardrop:
> I've looked at the issue with versioning and timers and it indeed looks like a bug, time doesn't get moved after the sleep, but only if that timer calls getVersion. Thanks for the repro, we'll need to investigate this issue a bit more.

Thanks, @vitarb! I've caught a similar issue today, this time not involving timers at all, only versions. We have a replay test for a workflow, and adding a version evaluation at the start of the workflow's main method has broken that replay test. Looking closely, it seems that after the version is evaluated, all signals that workflow gets "happen" at the initial time of the workflow start instead of the time when that signal was recorded.

It might be the manifestation of the same issue, or something closely related to it. I'll try to isolate it and get a reproducer too.

GreyTeardrop:
Seems like something like this also reproduced the issue:

Workflow interface:
```java
@WorkflowInterface
public interface Test1Workflow {
    @WorkflowMethod
    String waitForSignal();

    @SignalMethod
    void signal();
}
```

Workflow impl:
```java
public class Test1WorkflowImpl implements Test1Workflow {

    private static final Logger logger = LoggerFactory.getLogger(Test1WorkflowImpl.class);

    private final WorkflowQueue<String> queue = Workflow.newQueue(1);

    @Override
    public String waitForSignal() {
        logger.info("[{}] Start main method", now());
        //Workflow.getVersion("a", -1, 1);
        queue.take();
        logger.info("[{}] Complete main method", now());
        return now().toString();
    }

    @Override
    public void signal() {
        logger.info("[{}] Got signal", now());
        queue.put("");
        logger.info("[{}] Queued signal", now());
    }

    private Instant now() {
        return Instant.ofEpochMilli(Workflow.currentTimeMillis());
    }
}
```

Workflow execution history:
```json
{
  "events": [{
    "eventId": "1",
    "eventTime": "2021-04-20T10:55:51.244656300Z",
    "eventType": "EVENT_TYPE_WORKFLOW_EXECUTION_STARTED",
    "taskId": "3145728",
    "workflowExecutionStartedEventAttributes": {
      "workflowType": {
        "name": "Test1Workflow"
      },
      "taskQueue": {
        "name": "abandonment",
        "kind": "TASK_QUEUE_KIND_NORMAL"
      },
      "workflowExecutionTimeout": "100000s",
      "workflowRunTimeout": "100000s",
      "workflowTaskTimeout": "10s",
      "originalExecutionRunId": "7ba02678-343f-4b6f-9154-48065c1adcbe",
      "identity": "tctl@KV0200-Mykola-Rybak",
      "firstExecutionRunId": "7ba02678-343f-4b6f-9154-48065c1adcbe",
      "attempt": 1,
      "workflowExecutionExpirationTime": "2021-04-21T14:42:31.244Z",
      "firstWorkflowTaskBackoff": "0s"
    }
  }, {
    "eventId": "2",
    "eventTime": "2021-04-20T10:55:51.244726500Z",
    "eventType": "EVENT_TYPE_WORKFLOW_TASK_SCHEDULED",
    "taskId": "3145729",
    "workflowTaskScheduledEventAttributes": {
      "taskQueue": {
        "name": "abandonment",
        "kind": "TASK_QUEUE_KIND_NORMAL"
      },
      "startToCloseTimeout": "10s",
      "attempt": 1
    }
  }, {
    "eventId": "3",
    "eventTime": "2021-04-20T10:55:51.262662200Z",
    "eventType": "EVENT_TYPE_WORKFLOW_TASK_STARTED",
    "taskId": "3145734",
    "workflowTaskStartedEventAttributes": {
      "scheduledEventId": "2",
      "identity": "21655@KV0200-Mykola-Rybak",
      "requestId": "7eac83af-30e4-4dec-bc09-1614400d0ef6"
    }
  }, {
    "eventId": "4",
    "eventTime": "2021-04-20T10:55:51.497046900Z",
    "eventType": "EVENT_TYPE_WORKFLOW_TASK_COMPLETED",
    "taskId": "3145737",
    "workflowTaskCompletedEventAttributes": {
      "scheduledEventId": "2",
      "startedEventId": "3",
      "identity": "21655@KV0200-Mykola-Rybak",
      "binaryChecksum": "dc377a5ccfe2b18507cf5d6ab6c64e535580ff34"
    }
  }, {
    "eventId": "5",
    "eventTime": "2021-04-20T10:57:06.027261Z",
    "eventType": "EVENT_TYPE_WORKFLOW_EXECUTION_SIGNALED",
    "taskId": "3145739",
    "workflowExecutionSignaledEventAttributes": {
      "signalName": "signal",
      "identity": "tctl@KV0200-Mykola-Rybak"
    }
  }, {
    "eventId": "6",
    "eventTime": "2021-04-20T10:57:06.027300100Z",
    "eventType": "EVENT_TYPE_WORKFLOW_TASK_SCHEDULED",
    "taskId": "3145741",
    "workflowTaskScheduledEventAttributes": {
      "taskQueue": {
        "name": "21655@KV0200-Mykola-Rybak:fd3c8d91-2806-4b82-ae6f-c5ab7a4e3740",
        "kind": "TASK_QUEUE_KIND_STICKY"
      },
      "startToCloseTimeout": "10s",
      "attempt": 1
    }
  }, {
    "eventId": "7",
    "eventTime": "2021-04-20T10:57:06.042617Z",
    "eventType": "EVENT_TYPE_WORKFLOW_TASK_STARTED",
    "taskId": "3145745",
    "workflowTaskStartedEventAttributes": {
      "scheduledEventId": "6",
      "identity": "fd3c8d91-2806-4b82-ae6f-c5ab7a4e3740",
      "requestId": "2d0b30ea-b1db-46cc-bca6-31dd1b316838"
    }
  }, {
    "eventId": "8",
    "eventTime": "2021-04-20T10:57:06.115538900Z",
    "eventType": "EVENT_TYPE_WORKFLOW_TASK_COMPLETED",
    "taskId": "3145748",
    "workflowTaskCompletedEventAttributes": {
      "scheduledEventId": "6",
      "startedEventId": "7",
      "identity": "21655@KV0200-Mykola-Rybak",
      "binaryChecksum": "dc377a5ccfe2b18507cf5d6ab6c64e535580ff34"
    }
  }, {
    "eventId": "9",
    "eventTime": "2021-04-20T10:57:06.115629300Z",
    "eventType": "EVENT_TYPE_WORKFLOW_EXECUTION_COMPLETED",
    "taskId": "3145749",
    "workflowExecutionCompletedEventAttributes": {
      "result": {
        "payloads": [{
          "metadata": {
            "encoding": "anNvbi9wbGFpbg=="
          },
          "data": "IjIwMjEtMDQtMjBUMTA6NTc6MDYuMDQyWiI="
        }]
      },
      "workflowTaskCompletedEventId": "8"
    }
  }]
}
```

Replaying that history gives
```
[workflow-method] INFO  c.t.n.t.workflow.Test1WorkflowImpl - [2021-04-20T10:55:51.262Z] Start main method
[signal signal] INFO  c.t.n.t.workflow.Test1WorkflowImpl - [2021-04-20T10:57:06.042Z] Got signal
[signal signal] INFO  c.t.n.t.workflow.Test1WorkflowImpl - [2021-04-20T10:57:06.042Z] Queued signal
[workflow-method] INFO  c.t.n.t.workflow.Test1WorkflowImpl - [2021-04-20T10:57:06.042Z] Complete main method
```

Uncommenting the `Workflow.getVersion` and then replaying again will give
```
[workflow-method] INFO  c.t.n.t.workflow.Test1WorkflowImpl - [2021-04-20T10:55:51.262Z] Start main method
[signal signal] INFO  c.t.n.t.workflow.Test1WorkflowImpl - [2021-04-20T10:55:51.262Z] Got signal
[signal signal] INFO  c.t.n.t.workflow.Test1WorkflowImpl - [2021-04-20T10:55:51.262Z] Queued signal
[workflow-method] INFO  c.t.n.t.workflow.Test1WorkflowImpl - [2021-04-20T10:55:51.262Z] Complete main method
```
(note that the timestamps are all the same now)

Also, should replaying a workflow assert that workflow's results stays the same as recorded?

vitarb:
Okay, I think I understand what's going on, but I'm not sure what the right fix should be. So the problem is that when we [process a marker command](https://github.com/temporalio/sdk-java/blob/30420212a26283cf7cbe6339512058a3186074e8/temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java#L797) that gets created by the versioning state machine, we trigger an event loop [here](https://github.com/temporalio/sdk-java/blob/30420212a26283cf7cbe6339512058a3186074e8/temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java#L670) which executes the workflow before we get a chance to set the right time [here](https://github.com/temporalio/sdk-java/blob/30420212a26283cf7cbe6339512058a3186074e8/temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java#L808).
I think we should either set time earlier or not trigger event loop from the version callback. Let me discuss this with other folks on the team and we should have a solution soon.

mfateev:
I agree that moving `setCurrentTimeMillis` before is the right way to fix this. I'm not 100% sure but it can be as simple as moving it to the [line 792](https://github.com/temporalio/sdk-java/blob/30420212a26283cf7cbe6339512058a3186074e8/temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java#L792).

GreyTeardrop:
Thank you! Version-related tests seem to be fixed by #458. Seems like the "two concurrent timers" might be a different issue as it's not affected by that PR.

vitarb:
Closing in favor of #458

PR #451, opened 2021-04-26
Change log level from ERROR to INFO for failed workflow task caused by signal race condition

vkoby:
## What was changed:
Log level: error -> info

## Why?
Worker received a task to close the workflow from the service, while the worker is working on it, the service received a signal to the workflow. When the worker tries to close the workflow, the service will reject the request to close the workflow. The worker receives this rejection and logging it as ERROR. But everything will be fine after the signal is processed and workflow will be closed.

Closes issue: https://github.com/temporalio/sdk-java/issues/446


temporal-sdk/src/main/java/io/temporal/internal/worker/PollerOptions.java
```diff
@@ -135,7 +137,19 @@ public Builder setPollThreadNamePrefix(String pollThreadNamePrefix) {
...
     public PollerOptions build() {
       if (uncaughtExceptionHandler == null) {
-        uncaughtExceptionHandler = (t, e) -> log.error("uncaught exception", e);
+        uncaughtExceptionHandler =
+            (t, e) -> {
+              if (e instanceof RuntimeException
+                  && e.getCause() != null
+                  && e.getCause() instanceof StatusRuntimeException) {
+                StatusRuntimeException sre = (StatusRuntimeException) e.getCause();
+                if (sre.getStatus().getCode() == Status.Code.INVALID_ARGUMENT) {
```
mfateev, line 146: Just to be on the safe side let's add a check for the message. It should be "UnhandledCommand".

temporal-sdk/src/main/java/io/temporal/internal/worker/PollerOptions.java
```diff
@@ -135,7 +137,19 @@ public Builder setPollThreadNamePrefix(String pollThreadNamePrefix) {
...
+                  log.info("Uncaught exception", e);
```
mfateev, line 147: As we know the exact meaning of the exception I would change the log message to something more descriptive.

temporal-sdk/src/main/java/io/temporal/internal/worker/PollerOptions.java
```diff
@@ -135,7 +137,19 @@ public Builder setPollThreadNamePrefix(String pollThreadNamePrefix) {
 
     public PollerOptions build() {
       if (uncaughtExceptionHandler == null) {
-        uncaughtExceptionHandler = (t, e) -> log.error("uncaught exception", e);
+        uncaughtExceptionHandler =
+            (t, e) -> {
+              if (e instanceof RuntimeException
+                  && e.getCause() != null
+                  && e.getCause() instanceof StatusRuntimeException) {
```
vitarb, line 144: instanceof is null safe, so you don't need line 143

temporal-sdk/src/main/java/io/temporal/internal/worker/PollerOptions.java
vitarb, line 147: Why do we choose to ignore all other status codes? Isn't it possible for these exceptions to occur in other situations?

vkoby: See [https://github.com/temporalio/sdk-java/issues/446](url)

temporal-sdk/src/main/java/io/temporal/internal/worker/PollerOptions.java
```diff
@@ -135,8 +137,22 @@ public Builder setPollThreadNamePrefix(String pollThreadNamePrefix) {
...
       if (uncaughtExceptionHandler == null) {
-        uncaughtExceptionHandler = (t, e) -> log.error("uncaught exception", e);
+        uncaughtExceptionHandler =
+            (t, e) -> {
+              if (e instanceof RuntimeException && e.getCause() instanceof StatusRuntimeException) {
+                StatusRuntimeException sre = (StatusRuntimeException) e.getCause();
+                if (sre.getStatus().getCode() == Status.Code.INVALID_ARGUMENT
+                    && sre.getMessage().startsWith("INVALID_ARGUMENT: UnhandledCommand")) {
+                  log.info(
+                      "Failed workflow task caused by signal race condition. This error likely is recoverable.",
```
Sushisource, line 147: I believe it's it's possible for this to happen for reasons other than just a signal race. It might make more sense to just say "Failed workflow task due to unhandled command. ..."

Sushisource: Perhaps also lift this into a static so the test doesn't repeat the message when asserting.

Sushisource (approved):

temporal-sdk/src/main/java/io/temporal/internal/worker/PollerOptions.java
vkoby: Could you please explain what you mean? I'm not sure if it makes sense to make the builder method static, but perhaps you meant something else?

Sushisource: I mean lift the string literal up into a constant

PR #453, opened 2021-04-26
Cleaning up Activity Javadocs

shingjay:
## What was changed:
Javadocs cleaned up in several classes in the Activity package

## Why?
We didn't want implementation and duplicate docs (that exist in the main doc site) to be in the Javadocs as well.

## Checklist

1. Closes issue: 
https://www.notion.so/temporalio/Refresh-Java-SDK-reference-comments-819b790f4acd46c3a5d908dc0226aa15

2. How was this tested:
Tested locally, looking at the Javadocs view feature in IntelliJ

3. Any docs updates needed?
There were a few small edits I made to the documentation page after comparing Javadocs and documentation
https://github.com/temporalio/documentation/pull/378

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/activity/Activity.java
+++ b/temporal-sdk/src/main/java/io/temporal/activity/Activity.java
@@ -36,7 +36,7 @@ public final class Activity {
 
   /**
    * Can be used to get information about Activity invocation and Heartbeats.
-   * Note: This static method relies on a thread local and works only in the original Activity thread.
+   * This static method relies on a local thread and works only in the original Activity thread.
    */
   public static ActivityExecutionContext getExecutionContext() {
     return ActivityInternal.getExecutionContext();
--- a/temporal-sdk/src/main/java/io/temporal/activity/Activity.java
+++ b/temporal-sdk/src/main/java/io/temporal/activity/Activity.java
@@ -26,178 +26,7 @@ import io.temporal.internal.sync.ActivityInternal;
 import io.temporal.internal.sync.WorkflowInternal;
 
 /**
- * An activity is the implementation of a particular task in the business logic.
- *
- * <h2>Activity Interface</h2>
- *
- * <p>Activities are defined as methods of a plain Java interface annotated with {@literal @}{@link
- * ActivityInterface}. Each method defines a single activity type. A single workflow can use more
- * than one activity interface and call more than one activity method from the same interface. The
- * only requirement is that activity method arguments and return values are serializable to a byte
- * array using the provided {@link io.temporal.common.converter.DataConverter} implementation. The
- * default implementation uses JSON serializer, but an alternative implementation can be configured
- * through {@link io.temporal.client.WorkflowClientOptions.Builder#setDataConverter(DataConverter)}.
- *
- * <p>Example of an interface that defines four activities:
- *
- * <pre><code>
- * {@literal @}ActivityInterface
- * public interface FileProcessingActivities {
- *
- *     void upload(String bucketName, String localName, String targetName);
- *
- *     String download(String bucketName, String remoteName);
- *
- *     {@literal @}ActivityMethod(name = "Transcode")
- *     String processFile(String localName);
- *
- *     void deleteLocalFile(String fileName);
- * }
- *
- * </code></pre>
- *
- * An optional {@literal @}{@link ActivityMethod} annotation can be used to specify activity name.
- * By default the method name with the first letter capitalized is used as an activity name. So the
- * above interface defines the following activities: "Upload", "Download", "Transcode" and
- * "DeleteLocalFile".
- *
- * <h2>Activity Implementation</h2>
- *
- * <p>Activity implementation is an implementation of an activity interface. A single instance of
- * the activity's implementation is shared across multiple simultaneous activity invocations.
- * Therefore, the activity implementation code must be <i>thread safe</i>.
- *
- * <p>The values passed to activities through invocation parameters or returned through a result
- * value are recorded in the execution history. The entire execution history is transferred from the
- * Temporal service to workflow workers when a workflow state needs to recover. A large execution
- * history can thus adversely impact the performance of your workflow. Therefore, be mindful of the
- * amount of data you transfer via activity invocation parameters or return values. Other than that,
- * no additional limitations exist on activity implementations.
- *
- * <pre><code>
- * public class FileProcessingActivitiesImpl implements FileProcessingActivities {
- *
- *     private final AmazonS3 s3Client;
- *
- *     private final String localDirectory;
- *
- *     void upload(String bucketName, String localName, String targetName) {
- *         File f = new File(localName);
- *         s3Client.putObject(bucket, remoteName, f);
- *     }
- *
- *     String download(String bucketName, String remoteName, String localName) {
- *         // Implementation omitted for brevity.
- *         return downloadFileFromS3(bucketName, remoteName, localDirectory + localName);
- *     }
- *
- *     String processFile(String localName) {
- *         // Implementation omitted for brevity.
- *         return compressFile(localName);
- *     }
- *
- *     void deleteLocalFile(String fileName) {
- *         File f = new File(localDirectory + fileName);
- *         f.delete();
- *     }
- * }
- * </code></pre>
- *
- * <h3>Accessing Activity Info</h3>
- *
- * <p>The {@link Activity#getExecutionContext()} returns {@link ActivityExecutionContext} which
- * provides static getters to access information about the workflow that invoked it. Note that the
- * activity context information is stored in a thread-local variable. Therefore, calls to {@link
- * Activity#getExecutionContext()} succeeds only in the thread that invoked the activity function.
- *
- * <pre><code>
- * public class FileProcessingActivitiesImpl implements FileProcessingActivities {
- *
- *      {@literal @}Override
- *      public String download(String bucketName, String remoteName, String localName) {
- *         ActivityExecutionContext ctx = Activity.getExecutionContext();
- *         ActivityInfo info = ctx.getInfo();
- *         log.info("namespace=" +  info.getActivityNamespace());
- *         log.info("workflowId=" + info.getWorkflowId());
- *         log.info("runId=" + info.getRunId());
- *         log.info("activityId=" + info.getActivityId());
- *         log.info("activityTimeout=" + info.getStartToCloseTimeoutSeconds());
- *         return downloadFileFromS3(bucketName, remoteName, localDirectory + localName);
- *      }
- *      ...
- *  }
- * </code></pre>
- *
- * <h3>Asynchronous Activity Completion</h3>
- *
- * <p>Sometimes an activity lifecycle goes beyond a synchronous method invocation. For example, a
- * request can be put in a queue and later a reply comes and is picked up by a different worker
- * process. The whole request-reply interaction can be modeled as a single Temporal activity.
- *
- * <p>To indicate that an activity should not be completed upon its method return, call {@link
- * ActivityExecutionContext#doNotCompleteOnReturn()} from the original activity thread. Then later,
- * when replies come, complete the activity using {@link
- * io.temporal.client.ActivityCompletionClient}. To correlate activity invocation with completion
- * use either {@code TaskToken} or workflow and activity IDs.
- *
- * <pre><code>
- * public class FileProcessingActivitiesImpl implements FileProcessingActivities {
- *
- *      public String download(String bucketName, String remoteName, String localName) {
- *          ActivityExecutionContext ctx = Activity.getExecutionContext();
- *          byte[] taskToken = ctx.getInfo().getTaskToken(); // Used to correlate reply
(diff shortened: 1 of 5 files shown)


temporal-sdk/src/main/java/io/temporal/activity/Activity.java
```diff
@@ -217,39 +45,14 @@ public static ActivityExecutionContext getExecutionContext() {
...
-   * try {
-   *     return someCall();
-   * } catch (Exception e) {
-   *     throw Activity.wrap(e);
-   * }
-   * </pre>
-   *
-   * If wrap returned void it wouldn't be possible to write <code>
-   * throw Activity.wrap</code> and compiler would complain about missing return.
+   * Use this to rethrow a checked exception from an Activity instead of adding the exception to a method signature.
```
mfateev, line 48: I would add this information to the documentation before removing it from the JavaDoc.

flossypurse (changes requested):
1.  Let's make sure each sentence has a period.
2. Let's come up with some rules/standard for when a comment sentence goes to a new line vs starting a new sentence on the same line and having it broken up by a newline.
3. I think we should capitalize "Heartbeat".
4. Let's find out the requirements for links. I see them in a bunch of different formats and I want to standardize on one.
5.  Let's remove html from the comments. There is no need to have them here if we use the tags appropriately.
6. Let's standardize on using 
```
/**
 * Comment
 */
```
vs
```
/** Comment */
```
For all Javadoc comments.

temporal-sdk/src/main/java/io/temporal/activity/Activity.java
flossypurse: It is there, though could be more discoverable: https://docs.temporal.io/docs/java/activities/#throwing-activity-errors

temporal-sdk/src/main/java/io/temporal/activity/Activity.java
```diff
@@ -26,178 +26,7 @@
 import io.temporal.internal.sync.WorkflowInternal;
 
 /**
- * An activity is the implementation of a particular task in the business logic.
- *
- * <h2>Activity Interface</h2>
```
flossypurse, line 31: This is duplicate of docs: https://docs.temporal.io/docs/java/activities

tsurdilo (commented):
Just wanted to understand the reason for this change. I understand the docs might have this information as well (or should have) but why is that a reason to remove the comments completely? You could in the comments put a link to the docs that describe this instead but leaving them or shortening them to me at least seems to be better than removing. 

Update - nevermind :) I spoke too soon on this, didn't see comments were not fully deleted.

temporal-sdk/src/main/java/io/temporal/activity/ActivityInfo.java
```diff
@@ -24,34 +24,34 @@
...
   String getWorkflowId();
 
-  /** RunId of the workflow that scheduled the activity. */
+  /** RunId of the workflow that scheduled the Activity. */
   String getRunId();
 
   /**
-   * ID of the activity. This ID can be used to complete the activity asynchronously through {@link
+   * ID of the Activity. This ID can be used to complete the Activity asynchronously through {@link
    * io.temporal.client.ActivityCompletionClient#complete(String, Optional, String, Object)}.
```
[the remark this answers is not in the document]
shingjay: Yeah the link works, it's like this because of method overloading. There are multiple methods with the same name.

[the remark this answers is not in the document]
shingjay: Ohhhh - I realized it doesn't matter. Will research more and figure this out.

temporal-sdk/src/main/java/io/temporal/activity/ActivityMethod.java
```diff
@@ -25,19 +25,19 @@
...
+ * interface methods. Not required. Use it to override default Activity type name.
  */
 @Retention(RetentionPolicy.RUNTIME)
 @Target(ElementType.METHOD)
 public @interface ActivityMethod {
 
   /**
-   * Name of the activity type. Default is method name. Also consider using {@link
+   * Name of the Activity type. Default is method name. Also consider using {@link
    * ActivityInterface#namePrefix()}. Note that the prefix is ignored if the name is specified.
```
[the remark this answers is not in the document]
shingjay: > 2. Let's come up with some rules/standard for when a comment sentence goes to a new line vs starting a new sentence on the same line and having it broken up by a newline.
Sounds good - perhaps stick to same line only if both sentences can fit in one line and not multiple? Else, always have a new line? Like this:
Screen Shot 2021-05-01 at 11 09 58 PM


> 4. Let's find out the requirements for links. I see them in a bunch of different formats and I want to standardize on one.
So [this](https://stackoverflow.com/a/5916021) best explains why they're different in the current state of things. It's basically because some classes are imported in the class already, therefore they do not need the package name defined in this form `{@link com.xxx.yyy.Class}`. 
I prefer not importing a class just because we need it for documentation purposes and to stick to the `{@link com.xxx.yyy.Class}` convention instead. 

Another thing I noticed was that even if the package is defined, it's not printed in the Javadocs page itself. For example `{@link io.temporal.client.ActivityCompletionClient#complete(byte[], Object)}` prints [this](https://www.javadoc.io/doc/io.temporal/temporal-sdk/latest/io/temporal/activity/ActivityInfo.html).

Might be counterproductive to have a convention for this case tbh as it's based on the code. If anything it should be for the Javadocs readers, but it looks like it 'works'.  


> 6. Let's standardize on using
> 
> ```
> /**
>  * Comment
>  */
> ```
> 
> vs
> 
> ```
> /** Comment */
> ```
> 
> For all Javadoc comments.
I'm cool with standardizing, but I think the former is for multi-line comments and the latter may be used for single-line comments (it's slightly cleaner to look at). This is mentioned in the Google Style Guide too https://google.github.io/styleguide/javaguide.html#s7.1.1-javadoc-multi-line.

temporal-sdk/src/main/java/io/temporal/activity/Activity.java
```diff
@@ -206,50 +35,24 @@
 public final class Activity {
 
   /**
-   * Activity execution context. It can be used to get information about activity invocation as well
-   * for heartbeating.
-   *
-   * <p>Note: This static method relies on a thread local and works only in the original activity
-   * thread.
+   * Can be used to get information about Activity invocation and Heartbeats.
+   * Note: This static method relies on a thread local and works only in the original Activity thread.
```
flossypurse, line 39:
```suggestion
   * This static method relies on a local thread and works only in the original Activity thread.
```

temporal-sdk/src/main/java/io/temporal/activity/ActivityCancellationType.java
```diff
@@ -23,23 +23,25 @@
 import io.temporal.workflow.CancellationScope;
 
 /**
- * Defines behaviour of the parent workflow when {@link CancellationScope} that wraps child workflow
- * execution request is canceled. The result of the cancellation independently of the type is a
- * {@link CanceledFailure} thrown from the child workflow method.
+ * Defines behaviour of the parent Workflow when the {@link CancellationScope} that wraps child Workflow
```
flossypurse, line 26:
```suggestion
 * Defines behaviour of the parent Workflow when the {@link CancellationScope} that wraps the child Workflow
```

temporal-sdk/src/main/java/io/temporal/activity/LocalActivityOptions.java
```diff
@@ -75,8 +83,8 @@ public Builder setScheduleToCloseTimeout(Duration timeout) {
     }
 
     /**
-     * Maximum time to retry locally keeping workflow task open through heartbeat. Default is 6
-     * workflow task timeout.
+     * Maximum time to retry locally, while keeping the Workflow Task open via a Heartbeat.
+     * Default value is 6 workflow task timeout.
```
flossypurse, line 87: This comment is not clear.
6 what? 6 seconds? 6 minutes? 6 attempts?
Also, capitalize `Workflow Task`

temporal-sdk/src/main/java/io/temporal/activity/LocalActivityOptions.java
```diff
@@ -120,17 +128,17 @@ public Builder mergeActivityOptions(LocalActivityOptions override) {
     }
 
     /**
-     * RetryOptions that define how activity is retried in case of failure. Default is null which is
-     * no retries.
+     * {@link RetryOptions} that define how an Activity is retried in case of failure.
+     * Default is null which is no retries.
```
flossypurse, line 132: I don't think this is true any more?
@mfateev shouldn't this change to something like "Activities use a default Retry Policy if these Retry options are not provided" ?

temporal-sdk/src/main/java/io/temporal/activity/LocalActivityOptions.java
```diff
@@ -140,14 +148,12 @@ public Builder setMethodRetry(MethodRetry r) {
...
-     * <p>The serialized arguments are included only for human troubleshooting as they are never
-     * read by the SDK code. So in some cases it is worth not including them to reduce the history
-     * size.
-     *
-     * <p>Default is false.
+     * When set to true, the serialized arguments of the local Activity are not included in the
+     * Marker Event that stores the local Activity's invocation result.
+     * The serialized arguments are included only for human troubleshooting as they are never
+     * read by the SDK code.
+     * In some cases, it is better not including them to reduce the history size.
```
flossypurse, line 155:
```suggestion
     * In some cases, it is better  to not include them to reduce the history size.
```

temporal-sdk/src/main/java/io/temporal/activity/LocalActivityOptions.java
shingjay: Took me ages to find this, but it's basically the `Workflow Task`'s timeout multiplied by 6.
https://github.com/temporalio/sdk-java/blob/master/temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflowContext.java#L373

temporal-sdk/src/main/java/io/temporal/activity/ActivityCancellationType.java
```diff
@@ -23,23 +23,27 @@
...
  */
 public enum ActivityCancellationType {
   /**
-   * Wait for activity cancellation completion. Note that activity must heartbeat to receive a
-   * cancellation notification. This can block the cancellation for a long time if activity doesn't
-   * heartbeat or chooses to ignore the cancellation request.
+   * Wait for the Activity to complete its cancellation.
+   * Note that an Activity must Heartbeat to receive a cancellation notification.
+   * This can block the cancellation for a long time if the Activity doesn't Heartbeat
+   * or chooses to ignore the cancellation request.
```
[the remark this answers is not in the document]
shingjay: You are right! I wasn't doing it at first - but I have now.

temporal-sdk/src/main/java/io/temporal/activity/LocalActivityOptions.java
shingjay: Ya looks like it's returning `DEFAULT_INSTANCE` with some basic initialized values. I've updated this.
https://github.com/temporalio/sdk-java/blob/master/temporal-sdk/src/main/java/io/temporal/common/RetryOptions.java#L57

temporal-sdk/src/main/java/io/temporal/activity/ActivityCancellationType.java
```diff
@@ -19,27 +19,21 @@
...
 
-import io.temporal.failure.CanceledFailure;
-import io.temporal.workflow.CancellationScope;
-
-/**
- * Defines behaviour of the parent workflow when {@link CancellationScope} that wraps child workflow
- * execution request is canceled. The result of the cancellation independently of the type is a
- * {@link CanceledFailure} thrown from the child workflow method.
- */
+/** Defines activity cancellation behavior. */
```
flossypurse, line 22:
```suggestion
/** Defines Activity cancellation behavior. */
```

flossypurse (approved):
This is looking good to me.
Let's get at least one more approval though.

temporal-sdk/src/main/java/io/temporal/activity/ActivityCancellationType.java
```diff
@@ -19,27 +19,21 @@
...
- * execution request is canceled. The result of the cancellation independently of the type is a
- * {@link CanceledFailure} thrown from the child workflow method.
- */
+/** Defines Activity cancellation behavior. */
 public enum ActivityCancellationType {
   /**
-   * Wait for activity cancellation completion. Note that activity must heartbeat to receive a
-   * cancellation notification. This can block the cancellation for a long time if activity doesn't
-   * heartbeat or chooses to ignore the cancellation request.
+   * Wait for the Activity to complete its cancellation. Note that an Activity must Heartbeat to
```
Sushisource, line 25: Something like below might be a bit more illuminating
```suggestion
   * Wait for the Activity to confirm any requested cancellation. Note that an Activity must Heartbeat to
```

Sushisource (approved):

Issue #413, opened 2021-03-31
Workflow#newActivityStub(activityInterface) method produces broken activity stubs


Spikhalskiy:
ActivityOptions has a requirement that either `startToClose` or `scheduleToClose` has to be defined. ActivityOptions doesn't have any defaults for both of them at the same time. So default ActivityOptions is actually invalid. 

`Workflow#newActivityStub` has two versions: `Workflow#newActivityStub(activityInterface)` and `Workflow#newActivityStub(activityInterface, activityOptions)`. The first version without `activityOptions` param produces an Activity stub with default broken ActivityOptions, so this method never really can be used right now.

vitarb:
On one hand, since activities are very use case specific, I imagine, it would be hard to pick the "right" value, but on on the other I agree that having invalid default options is bad experience. @mfateev what do you think? Should we set schedule to close to something like 10 seconds by default?

Spikhalskiy:
Also probably if we enforce "`startToClose` or `scheduleToClose`" rule, ActivityOptions#Builder#validateAndBuildWithDefaults should throw an exception when both of them are not initialized.

vkoby:
There is no good default value for these timeouts and you can now set default activity options through WorkflowImplementationOptions. But good point about throwing an exception when they haven't been set.

PR #457, opened 2021-04-29
Workflow#newActivityStub(activityInterface) method produces broken activity stubs

vkoby:
## What was changed:
Deprecated method that creates default ActivityOptions and added an exception if both required options aren't set.

## Why?
ActivityOptions has a requirement that either startToClose or scheduleToClose has to be defined. There's no good default value that would fit most workflows because activity would either seem like it is stuck (if it's too long) or is unable to execute in time (too short). 

Closes issue: [#413](https://github.com/temporalio/sdk-java/issues/413)


temporal-sdk/src/test/java/io/temporal/activity/ActivityOptionsTest.java
```diff
@@ -41,6 +53,34 @@
...
+    TestActivity activity =
+        testEnv.newActivityStub(
+            TestActivity.class,
+            ActivityOptions.newBuilder().setScheduleToCloseTimeout(Duration.ofDays(1)).build(),
+            null);
+
+    // Check that options were set correctly
+    Map<String, Duration> optionsValues = activity.activity1();
+    Assert.assertEquals(Duration.ofDays(1), optionsValues.get("ScheduleToCloseTimeout"));
+    Assert.assertEquals(Duration.ofDays(1), optionsValues.get("StartToCloseTimeout"));
```
mfateev, line 81: I"m confused. Where does this timeout come from if it is not set?

temporal-sdk/src/test/java/io/temporal/internal/testing/ActivityTestingTest.java
```diff
@@ -68,7 +75,7 @@ public String activity1(String input) {
   @Test
   public void testSuccess() {
     testEnvironment.registerActivitiesImplementations(new ActivityImpl());
-    TestActivity activity = testEnvironment.newActivityStub(TestActivity.class);
+    TestActivity activity = testEnvironment.newActivityStub(TestActivity.class, activityOptions);
```
mfateev, line 78: I don't understand why we are making these changes here. I think we should add the ability to set ActivityOptions on the default TestActivityEnvironment.

temporal-sdk/src/test/java/io/temporal/activity/ActivityOptionsTest.java
vkoby: I wasn't sure about the reasoning behind this logic, but this is where it comes from:
https://github.com/temporalio/sdk-java/blob/4c9bdd9c445360c0b3cffb080fb8e28527910e70/temporal-sdk/src/main/java/io/temporal/activity/ActivityOptions.java#L256

temporal-sdk/src/test/java/io/temporal/internal/testing/ActivityTestingTest.java
vkoby: l thought it to be misleading to have different behavior in prod and in tests. If you don't set required timeouts in the stub in prod and expect things to break, this is what should happen in tests. Consistent behavior is important for testing for all coming engineers who would have to learn the product.
Alternatively we can rename the method to something like TestActivityEnvironment#newActivityStubWithDefaults which would hopefully help with confusion.

temporal-sdk/src/test/java/io/temporal/activity/ActivityOptionsTest.java
mfateev: That logic seems outdated. We didn't have sane defaults on the service some time ago, so it was introduced. Let's remove it from the ActivityOptions completely.

temporal-sdk/src/test/java/io/temporal/internal/testing/ActivityTestingTest.java
mfateev: I think for TestActivityEnvironment we can provide default ActivityOptions with a large timeout.

temporal-testing/src/main/java/io/temporal/testing/TestActivityEnvironment.java
```diff
@@ -39,7 +39,7 @@
  *   private static class ActivityImpl implements TestActivity {
  *    {@literal @}Override
  *     public String activity1(String input) {
- *       return Activity.getTask().getActivityType().getName() + "-" + input;
+ *       return Activity.getExecutionContext().getInfo().getActivityType().getName() + "-" + input;
```
mfateev, line 42: I think .getName() should be removed.

mfateev (approved):

PR #458, opened 2021-04-29
Set current time before processing commands

vitarb:
This should address the problem described [here](https://github.com/temporalio/sdk-java/pull/447#issuecomment-828912253) by setting current workflow time before command processing happens. Before this change it was possible to trigger event loop before correct time was set, which means that workflow code would see the previous time instead of a new one.
Here we simply move `setCurrentTimeMillis(currentTimeMillis)` to the top of the `workflowTaskStarted` function. Thanks to  @GreyTeardrop for contributing a test that reproduces an issue, which I've included here.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -9,14 +9,4 @@ steps:
       - docker-compose#v3.0.0:
           run: unit-test-test-service
           config: docker/buildkite/docker-compose.yaml
-  - label: ":java: Unit test with docker service"
-    agents:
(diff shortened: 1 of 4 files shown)


GreyTeardrop:
Thank you, @vitarb! I can confirm this change fixes the `GetVersionAndTimerTest` as well as the issue with [`getVersion` breaking workflow replay](https://github.com/temporalio/sdk-java/pull/447#issuecomment-823188316).

Sushisource (approved):

PR #464, opened 2021-05-03
Make the deadlock detection timeout configurable

uvifian:
## What was changed:
Make the fixed deadlock detection timeout configurable.

## Why?
We see that some code in our workflows takes longer to execute during the first run, most likely due to hotspot compilation combined with other containers starting up.

This may happen to any workflow and is not related to calling Thread.sleep or anything similar. Making the default timeout configurable gives users the option to adapt the threshold. Setting TEMPORAL_DEBUG=true is not desirable as it would disable a useful feature.

## Checklist

1. Closes issue: None, but it's related to https://github.com/temporalio/sdk-java/pull/296

2. How was this tested:
Workflow with a >1000ms wait time

3. Any docs updates needed?
Possibly https://docs.temporal.io/docs/java/testing-and-debugging/ is you want to mention it there

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/DeterministicRunner.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/DeterministicRunner.java
@@ -35,9 +35,12 @@ interface DeterministicRunner {
   boolean debugMode = System.getenv("TEMPORAL_DEBUG") != null;
 
   long DEFAULT_DEADLOCK_DETECTION_TIMEOUT = 1000;
+  long deadlockDetectionTimeout = Long.parseLong(System.getenv().getOrDefault(
+      "DEADLOCK_DETECTION_TIMEOUT",
+      String.valueOf(DEFAULT_DEADLOCK_DETECTION_TIMEOUT)));
 
   static long getDeadlockDetectionTimeout() {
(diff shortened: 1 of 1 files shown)


vitarb:
I wonder if it's worth to readjust this value at all, even if you see some errors, they shouldn't be fatal for the workflow and should be gracefully retried.
It's hard to come up with an ideal number here, if you had ability to override this value what would you set it to? @mfateev what do you think?

uvifian:
@vitarb thanks for having a look!

You're correct, the workflows are re-run and complete on a subsequent attempt. So this is more of a nice-to-have and I'd agree that having this as a configuration option on the workflow would be more user-friendly.

I altered this PR to extend the timeout to 3000ms and remove the configuration option via env var. I'd appreciate if you would consider the option to make this configurable on the workflow options.

vitarb:
Do you have some data to prove that 3 seconds would be better than 1? If no I would prefer to keep it a 1 second. There is nothing wrong with adding a worker option to make this configurable in my opinion. Would you like to contribute this change as part of this PR or do you want us to do it?

uvifian:
3 seconds would be better for our case but there's no way to determine if that's right for everybody, that's why I think allowing configuration is the way forward.

Until then we'll do some code-prewarming, this alleviates the issue for us. I'll see if we can add the worker config option in a PR but it's not a big priority for us at this moment.

PR #470, opened 2021-05-06
Out-of-process test server

drewhoskins-stripe:
Signed-off-by: Drew Hoskins <drewhoskins@stripe.com>

## What was changed:

This wraps the existing Java Test Service in a process so that it can be used by other SDKs that don't have high-fidelity test environments.
* NOTE: I haven't been allowed to sign a temporal CLA yet, but I'd like an initial review.
* NOTE: this is my first time with gradle and my first Java in 20 years.  View with suspicion.
* It creates a script that starts a server that responds to gRPC requests to localhost:7233
* mini-refactoring to isolate Client-specific fields within `TestWorkflowService`, which encompasses both client and server concerns.
Still to do in future PRs:
* Exposing test-specific endpoints through GRPC.
* Any fixes from more thorough testing of integration tests against this.
* Better logging/debuggability.

## Why?

See https://github.com/temporalio/temporal/issues/1521

## Checklist

1. Closes issue: 
https://github.com/temporalio/temporal/issues/1521

2. How was this tested:

Build and test the new server script:
```
./gradlew installDist :temporal-testing:testServiceServer
# Run the server
./temporal-testing/build/install/temporal-testing/bin/test-service-server 7233
# Poll it with ruby workers
temporal-ruby/examples $ ./bin/worker
# Trigger various workflows from temporal-ruby
./bin/trigger BranchingWorkflow
./bin/trigger SimpleTimerWorkflow 10
# Observe that it completes quickly rather than sleeping for 10s
```

Regression testing:
```
./gradlew test
```
3. Any docs updates needed?

Not until this is more baked.


temporal-testing/src/main/java/io/temporal/internal/testservice/TestServiceServer.java
```diff
@@ -0,0 +1,30 @@
...
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
+package io.temporal.internal.testservice;
+
+public class TestServiceServer {
+
+  public static void main(String[] args) throws InterruptedException {
+    TestWorkflowService service = new TestWorkflowService(7233);
```
mfateev, line 25: I would make the port configurable through a command-line argument.

temporal-testing/src/main/java/io/temporal/internal/testservice/TestServiceServer.java
```diff
@@ -0,0 +1,30 @@
...
+    if (service != null) {
```
mfateev, line 26: I don't understand this check and the shutdown part. I think it should call shutdown on a shutdown hook.

mfateev (commented):
Looks good with some minor nits.

temporal-testing/src/main/java/io/temporal/internal/testservice/TestServiceServer.java
vitarb: Also it should never ever be `null` after we've just initialized it.

temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
```diff
@@ -145,23 +162,46 @@ public TestWorkflowService(boolean lockTimeSkipping) {
...
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+    outOfProcessServer = null;
+  }
+
+  // Creates an out-of-process rather than in-process server, and does not set up a client.
+  public TestWorkflowService(int port) {
+    log.info("Server started, listening on " + port);
+    client = null;
```
vitarb, line 181: All functions that are using client like `newClientStub` and `close` will now thrown an NPE.

drewhoskins-stripe: Hmm, on the server side those functions shouldn't ever be called.  I'll hide this constructor and make a factory which clarifies the scenario when this constructor should be used.

mfateev (approved):

temporal-testing/build.gradle
```diff
@@ -1,11 +1,30 @@
...
+task testServiceServer(type: CreateStartScripts) {
+    mainClassName = 'io.temporal.internal.testservice.TestServiceServer'
+    applicationName = 'test-service-server'
+    outputDir = new File(project.buildDir, 'tmp')
+    classpath = startScripts.classpath
+    dependsOn 'sourcesJar'
+}
+
+// Run: ./gradlew installDist :temporal-testing:testServiceServer
+applicationDistribution.into('bin') {
```
drewhoskins-stripe, line 27: I would also like to publish with maven, if that's okay with y'all.  It can be a separate PR; I'd like to get this checked in to avoid a third round of merge conflicts.

Notes:
* The generated script is useful for debugging or if you don't want to run a container.
* In production, we're using a container with only the .jar files in `./temporal-testing/build/install/temporal-testing/lib`.  
* We just uploaded them to our artifactory and want to do something more official.

I don't know anything about maven or java conventions so I'm not sure what you'd want and would need guidance.

temporal-testing/build.gradle
```diff
@@ -1,11 +1,30 @@
+plugins {
+    id 'application'
+}
+
 description = '''Temporal Workflow Java SDK testing'''
 
 dependencies {
+    api group: 'org.slf4j', name: 'slf4j-simple', version: '1.7.30'
```
robzienert, line 8: I'm not sure adding this as an `api` dependency was the right move? Exposing `slf4j-api` as an api-scoped dependency makes sense, though.

drewhoskins-stripe: Very little understanding went into this, feel free to change it!

Issue #471, opened 2021-05-07
WorkflowExecutionUtils#getInstanceCloseEvent could get into infinite loop after timeout


Spikhalskiy:
# Actual Behavior

WorkflowExecutionUtils#getInstanceCloseEvent is implemented in a way that could get into an infinite loop after a timeout.

`WorkflowExecutionUtils#getInstanceCloseEvent` has the following simplified implementation:

```
GetWorkflowExecutionHistoryResponse response = null;
long start = System.currentTimeMillis();
do {
      long elapsed = System.currentTimeMillis() - start;
      long expiration = unit.toMillis(timeout) - elapsed;
      if (expiration > 0) {
        response = ...
      }
      if (response == null || !response.hasHistory()) {
          //if we are here and it's a timeout - we will never return from this cycle, we will never get to the following timeout check
          continue;
      }
      if (timeout != 0 && System.currentTimeMillis() - start > unit.toMillis(timeout)) {throw new TimeoutException(...);}
      ...
      }
    } while (true);
```

This simplified structure shows that if we got an empty response and we used up our allocated timeout - we can fall into an infinite `continue` loop which never will enforce the timeout.

I found investigating an incident when after a network blip some temporal stubs stuck into "retries" of WorkflowExecutionUtils#getInstanceCloseEvent performed every 1ms

PR #472, opened 2021-05-07
Fixed implementation of WorkflowExecutionUtils#getInstanceCloseEvent that previously could get into infinite loop after a timeout

Spikhalskiy:
## What was changed:
Execution flow in `WorkflowExecutionUtils#getInstanceCloseEvent` has been fixed to don't allow falling into an infinite `continue` cycle described in Issue #471.

## Closes issue: 
Issue #471

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/common/WorkflowExecutionUtils.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/common/WorkflowExecutionUtils.java
@@ -98,11 +98,10 @@ public class WorkflowExecutionUtils {
 
   private static final Logger log = LoggerFactory.getLogger(WorkflowExecutionUtils.class);
 
-  private static final RpcRetryOptions retryParameters =
+  private static final RpcRetryOptions GET_INSTANCE_CLOSE_EVENT_RETRY_OPTIONS =
       RpcRetryOptions.newBuilder()
-          .setBackoffCoefficient(2)
-          .setInitialInterval(Duration.ofMillis(500))
-          .setMaximumInterval(Duration.ofSeconds(30))
+          .setBackoffCoefficient(1)
+          .setInitialInterval(Duration.ofMillis(1))
           .setMaximumAttempts(Integer.MAX_VALUE)
           .addDoNotRetry(Status.Code.INVALID_ARGUMENT, null)
           .addDoNotRetry(Status.Code.NOT_FOUND, null)
@@ -205,7 +204,12 @@ public class WorkflowExecutionUtils {
     }
   }
 
-  /** Returns an instance closing event, potentially waiting for workflow to complete. */
+  /**
+   * @param timeout timeout to retrieve InstanceCloseEvent in {@code unit} units. If 0 - MAX_INTEGER
+   *     will be used
+   * @param unit time unit of {@code timeout}
+   * @return an instance closing event, potentially waiting for workflow to complete.
+   */
   public static HistoryEvent getInstanceCloseEvent(
       WorkflowServiceStubs service,
       String namespace,
@@ -215,10 +219,9 @@ public class WorkflowExecutionUtils {
       TimeUnit unit)
       throws TimeoutException {
     ByteString pageToken = ByteString.EMPTY;
-    GetWorkflowExecutionHistoryResponse response = null;
+    GetWorkflowExecutionHistoryResponse response;
     // TODO: Interrupt service long poll call on timeout and on interrupt
     long start = System.currentTimeMillis();
-    HistoryEvent event;
     do {
       GetWorkflowExecutionHistoryRequest r =
           GetWorkflowExecutionHistoryRequest.newBuilder()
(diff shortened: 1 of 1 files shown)


temporal-sdk/src/main/java/io/temporal/internal/common/WorkflowExecutionUtils.java
```diff
@@ -229,15 +234,17 @@ public static HistoryEvent getInstanceCloseEvent(
               .setWaitNewEvent(true)
               .setNextPageToken(pageToken)
               .build();
+
       long elapsed = System.currentTimeMillis() - start;
-      Deadline expiration = Deadline.after(unit.toMillis(timeout) - elapsed, TimeUnit.MILLISECONDS);
-      if (expiration.timeRemaining(TimeUnit.MILLISECONDS) > 0) {
+      long millisRemaining = unit.toMillis(timeout != 0 ? timeout : Integer.MAX_VALUE) - elapsed;
```
vitarb, line 239: `timeout : Integer.MAX_VALUE` this default wasn't there before, why do we need it?

temporal-sdk/src/main/java/io/temporal/internal/common/WorkflowExecutionUtils.java
```diff
@@ -229,15 +234,17 @@ public static HistoryEvent getInstanceCloseEvent(
               .setWaitNewEvent(true)
               .setNextPageToken(pageToken)
               .build();
+
       long elapsed = System.currentTimeMillis() - start;
-      Deadline expiration = Deadline.after(unit.toMillis(timeout) - elapsed, TimeUnit.MILLISECONDS);
```
vitarb, line 233: What there anything wrong with using Deadline?

Spikhalskiy: The creation of Deadline was useless here. We were converting millis into Deadline only to covert Deadline immediately back to millis. Not wrong, just pointless.

temporal-sdk/src/main/java/io/temporal/internal/common/WorkflowExecutionUtils.java
Spikhalskiy: It's kinda a mess right now. This line didn't accept 0 as timeout, but two lines after it we actually have a check timeout != 0:
`if (timeout != 0 && System.currentTimeMillis() - start > unit.toMillis(timeout))`

So, I added javadoc explicitly stating that 0 is supported and added this processing. It's not needed because right now we provide timeout this way usually:
https://github.com/temporalio/sdk-java/blob/fd28e214fe9820e00a3f94883b3faaa2e27ae63e/temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowStubImpl.java#L261

It's also done inconsistently. Some places provide MAX_INT, some MAX_LONG

temporal-sdk/src/main/java/io/temporal/internal/common/WorkflowExecutionUtils.java
```diff
@@ -256,11 +256,10 @@ public static HistoryEvent getInstanceCloseEvent(
                       .withDeadline(expirationInRetry)
                       .getWorkflowExecutionHistory(r);
                 });
-      }
-      if (response == null || !response.hasHistory()) {
-        continue;
-      }
-      if (timeout != 0 && System.currentTimeMillis() - start > unit.toMillis(timeout)) {
+        if (response == null || !response.hasHistory()) {
```
vitarb, line 259: Note that there is a slight behavioral change here, as before we've used to check current time AFTER the call to the server, now we are checking that condition BEFORE. This means that code below will be executed one more time comparing to what it was, and it is possible that we return an event for a workflow that timed out already. I think this is not great, and I would prefer to always check the timeout condition with current time after the service call.

temporal-sdk/src/main/java/io/temporal/internal/common/WorkflowExecutionUtils.java
vitarb: Yeah, there is even that https://github.com/temporalio/sdk-java/blob/fd28e214fe9820e00a3f94883b3faaa2e27ae63e/temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowStubImpl.java#L225 where max int is used intentionally.

temporal-sdk/src/main/java/io/temporal/internal/common/WorkflowExecutionUtils.java
Spikhalskiy: Yeah, I actually made this change intentionally.
It looks pointless to wait for an event, actually get it and verify that it's a real event only to timeout after it. 
I personally think that it's actually a fix for the edge case.

Why do you think it's so critical to throw timeout even if we already spent this time and we already got the event?

vitarb: I think it's okay given that it's not a workflow timeout, just wanted to point out that there is a change in behavior.

Spikhalskiy: It's really a super tiny edge case here.
We enforce and actively work with timeout in GrpcRetryer here:
```
GrpcRetryer.retryWithResult(
                retryOptions,
                () -> {
                  long elapsedInRetry = System.currentTimeMillis() - start;
                  Deadline expirationInRetry =
                      Deadline.after(
                          unit.toMillis(timeout) - elapsedInRetry, TimeUnit.MILLISECONDS);
                 ... 
)
```
So, it's possible that we actually met the deadline, got the event from the Retryer and after that discarded because one 1 more millisecond ticked or something.
So, it's not like this change really makes it possible to significantly overdue the timeout, it's still fully enforced in the Retryer and in the lambda that is passed to the Retryer.

vitarb: Yep, let's keep it like it is then.

vitarb:
Would be nice to also add a test, but it looks like it's not going to be easy, so LGTM.

vitarb (approved):

temporal-sdk/src/main/java/io/temporal/internal/common/WorkflowExecutionUtils.java
```diff
@@ -98,11 +98,10 @@
...
   private static final Logger log = LoggerFactory.getLogger(WorkflowExecutionUtils.class);
 
-  private static final RpcRetryOptions retryParameters =
+  private static final RpcRetryOptions GET_INSTANCE_CLOSE_EVENT_RETRY_OPTIONS =
       RpcRetryOptions.newBuilder()
-          .setBackoffCoefficient(2)
-          .setInitialInterval(Duration.ofMillis(500))
-          .setMaximumInterval(Duration.ofSeconds(30))
+          .setBackoffCoefficient(1)
+          .setInitialInterval(Duration.ofMillis(1))
```
vitarb, line 104: Why did we change this?

vitarb, line 104: Unrelated to this PR, since you've just copied existing values, but is it a good idea to hammer server without backoff?

Spikhalskiy: Thank you for rising it!
It's actually not and this is already on my radar and I probably will send one more PR shortly after I familiarize myself better with code implementations of GrpcRetryer and BackoffThrottler.
This issue was discovered specifically because clients very aggressively hammering the server.

Spikhalskiy:
@vitarb thank you for the review, Vitaly! Yeah, a test would be good here, but a refactoring is needed here to make things mockable first.

Issue #310, opened 2021-01-15
Do not fail stack trace query on a closed workflow. But return empty stack


mfateev:
## Expected Behavior
A human readable string that indicates that workflow is closed.

## Actual Behavior
```
3 INVALID_ARGUMENT: java.lang.Error: closed at
io.temporal.internal.sync.DeterministicRunnerImpl.checkClosed(DeterministicRunnerImpl.java:433) at 
io.temporal.internal.sync.DeterministicRunnerImpl.stackTrace(DeterministicRunnerImpl.java:418) at 
io.temporal.internal.sync.SyncWorkflow.query(SyncWorkflow.java:171) at 
io.temporal.internal.replay.ReplayWorkflowExecutor.query(ReplayWorkflowExecutor.java:137) at 
io.temporal.internal.replay.ReplayWorkflowRunTaskHandler.handleQueryWorkflowTask(ReplayWorkflowRunTaskHandler.java:256) at 
io.temporal.internal.replay.ReplayWorkflowTaskHandler.handleQueryOnlyWorkflowTask(ReplayWorkflowTaskHandler.java:257) at 
io.temporal.internal.replay.ReplayWorkflowTaskHandler.handleWorkflowTask(ReplayWorkflowTaskHandler.java:112) at 
io.temporal.internal.worker.WorkflowWorker$TaskHandlerImpl.handle(WorkflowWorker.java:314) at 
io.temporal.internal.worker.WorkflowWorker$TaskHandlerImpl.handle(WorkflowWorker.java:280) at 
io.temporal.internal.worker.PollTaskExecutor.lambda$process$0(PollTaskExecutor.java:73) at 
java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128) at 
java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628) at 
java.base/java.lang.Thread.run(Thread.java:834)
```

Mhany95:
Any updates on this one? I can see the stack trace as long as the workflow is "Running", but once it's completed, failed or terminated, I get that error that @mfateev is mentioning?

PR #473, opened 2021-05-07
Do not fail stack trace query on a closed workflow

vkoby:
## What was changed:
Stack trace query return a human readable string that indicates that workflow is closed instead of an error.

Closes issue: #310


vitarb:
Is it possible to add a test for this?

vitarb (approved):

Issue #476, opened 2021-05-09
GrpcRetryer#retryWithResultAsync doesn't perform retries in some cases


Spikhalskiy:
## Expected Behavior
GrpcRetryer#retryWithResultAsync work as expected performing retries until expiration time is reached.

## Actual Behavior
GrpcRetryer#retryWithResultAsync returns after the first retry if an exception is thrown from the supplied function instead of being provided in the returned CompletableFuture.

## Cause
CompletionException is not processed and unwrapped correctly.

## Steps to Reproduce the Problem

This unit test fails with the current implementation, while it's expected to pass:
https://github.com/Spikhalskiy/java-sdk/commit/73f678f79a2338f53da394f3a317897953b92951#diff-889cc42980a7287f0e29c43980e54d3c67c343c483105136bf09cb35269e9051R46

Spikhalskiy:
Closed by #477

PR #477, opened 2021-05-09
Fixed GrpcRetryer#retryWithResultAsync implementation

Spikhalskiy:
## What was changed:

`GrpcRetryer#retryWithResultAsync` implementation was fixed to correctly unwrap `CompletionException`
Tests provided for `GrpcRetryer`.
The unused `Retryer` class is deleted.

1. Closes issue: 
Issue #476

2. How was this tested:
Basic unit tests for GrpcRetryer were added


temporal-sdk/src/main/java/io/temporal/internal/common/GrpcRetryer.java
```diff
@@ -205,9 +209,14 @@ public Throwable getException() {
     if (e == null) {
       return new ValueExceptionPair<>(CompletableFuture.completedFuture(r), null);
     }
+    if (e instanceof CompletionException) {
+      e = e.getCause();
```
Spikhalskiy, line 213: This is the fix

vitarb: Can you leave a comment explaining why we do what we do?
Also it's possible for CompletionException to be constructed without cause, should we still do this trick in that case?

Spikhalskiy: > Can you leave a comment explaining why we do what we do?
I added a comment, plus now there are tests for both wrapped and unwrapped exceptions.

> Also it's possible for CompletionException to be constructed without cause, should we still do this trick in that case?
It's a kinda illegal state. CompletionException is intended to be a wrapper over another exception. But it's technically possible and it will work correctly: The next `if` check will be true and we will return without retrying because GrpcRetrier retries only StatusRuntimeException.

vitarb (approved):

Issue #204, opened 2020-09-10
Block of copied code


skrul:
Looks like there was a block of code copied twice here:

https://github.com/temporalio/java-sdk/blob/27ba155bf0684f05d2b26a7777651db83ad4641c/src/main/java/io/temporal/internal/worker/ActivityPollTask.java#L78

Does not look like it is a problem though.

Spikhalskiy:
Closed by #478

PR #478, opened 2021-05-09
Cleanup the way Pollers handle InterruptedExceptions

Spikhalskiy:
## What was changed:

Code in Worker's `Poller` and `PollerTask`s was cleaned up to one unified way of handling `InterruptedException`.

Before `LocalActivityPollTask` and `ActivityPollTask` had a full mix of correct and incorrect handling of `InterruptedException`:
- It was swallowed and ignored
- it was replaced with RuntimeException
- interrupted flag was raised (the only correct way and Pollers got it through GRPC stub implementation)
See [the article](https://dzone.com/articles/how-to-handle-the-interruptedexception for more context.) for more context about appropriate ways of handling `InterruptedException`.

Now `PollerTask`s are refactored to one approach (same as GRPC) - raising interrupted flag. And `Poller` was refactored to the more explicit way of handling interruptions.

## Closes

This PR also includes a small cleanup of ActivityPollTask and closes Issue #204

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/worker/ActivityPollTask.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/worker/ActivityPollTask.java
@@ -78,14 +78,6 @@ final class ActivityPollTask implements Poller.PollTask<ActivityTask> {
               .build());
     }
 
-    if (taskQueueActivitiesPerSecond > 0) {
-      pollRequest.setTaskQueueMetadata(
-          TaskQueueMetadata.newBuilder()
-              .setMaxTasksPerSecond(
-                  DoubleValue.newBuilder().setValue(taskQueueActivitiesPerSecond).build())
-              .build());
-    }
-
     if (log.isTraceEnabled()) {
       log.trace("poll request begin: " + pollRequest);
     }
@@ -95,8 +87,10 @@ final class ActivityPollTask implements Poller.PollTask<ActivityTask> {
     try {
       pollSemaphore.acquire();
     } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
       return null;
     }
+
     try {
       response =
           service
--- a/temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityPollTask.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityPollTask.java
@@ -44,9 +44,9 @@ final class LocalActivityPollTask
         log.trace("LocalActivity Task poll returned: " + task.getActivityId());
       }
       return task;
-
     } catch (InterruptedException e) {
-      throw new RuntimeException("local activity poll task interrupted", e);
+      Thread.currentThread().interrupt();
+      return null;
     }
   }
 
@@ -67,6 +67,7 @@ final class LocalActivityPollTask
       }
       return accepted;
     } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
       return false;
     }
   }
--- a/temporal-sdk/src/main/java/io/temporal/internal/worker/Poller.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/worker/Poller.java
@@ -39,6 +39,15 @@ import org.slf4j.LoggerFactory;
 public final class Poller<T> implements SuspendableWorker {
 
   public interface PollTask<TT> {
+    /**
+     * Pollers should shade all {@code java.lang.InterruptedException}s and raise {@code
+     * Thread.interrupted()} flag if the exceptions happen. This follows GRPC stubs approach, see
+     * {@code io.grpc.stub.ClientCalls#blockingUnaryCall}. Because pollers use GRPC subs anyway,
+     * this implementation was chosen for consistency. The caller of the poll task is responsible
+     * for handling the flag
+     *
+     * @return result of the task
+     */
     TT poll();
   }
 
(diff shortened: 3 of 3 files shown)


temporal-sdk/src/main/java/io/temporal/internal/worker/ActivityPollTask.java
```diff
@@ -78,14 +78,6 @@ public ActivityTask poll() {
               .build());
     }
 
-    if (taskQueueActivitiesPerSecond > 0) {
```
Spikhalskiy, line 81: copy-paste cleanup, Issue #204

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityPollTask.java
```diff
@@ -67,6 +68,7 @@ public Boolean apply(LocalActivityWorker.Task task, Duration maxWaitAllowed) {
       }
       return accepted;
     } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
```
Spikhalskiy, line 71: Here previously we potentially were completely loosing an interrupted flag of a worker thread. [WorkflowWorker -> ReplayWorkflowTaskHandler -> ReplayWorkflowRunTaskHandler -> LocalActivityPollTask#apply]

temporal-sdk/src/main/java/io/temporal/internal/worker/Poller.java
```diff
@@ -232,27 +237,44 @@ public void run() {
...
+      } catch (InterruptedException e) {
+        // we restore the flag here, so it can be checked and processed (with exit) in finally
+        Thread.currentThread().interrupt();
       } catch (Throwable e) {
         pollBackoffThrottler.failure();
-        if (!pollExecutor.isShutdown() && !pollExecutor.isTerminating()
-            || !(e.getCause() instanceof InterruptedException)
-                && !(e instanceof RejectedExecutionException)) {
+        // if we are terminating and getting rejected execution - it's normal
+        if (!pollExecutor.isTerminating() || !(e instanceof RejectedExecutionException)) {
```
Spikhalskiy, line 250: `pollExecutor.isTerminating()` is covering both `pollExecutor.isShutdown()` and `pollExecutor.isTerminating()`.
`|| !(e.getCause() instanceof InterruptedException)` was here for handling what `LocalActtivityPollTask` was doing here:
```
catch (InterruptedException e) {
      throw new RuntimeException("local activity poll task interrupted", e);
}
```
It's now replaced with a proper interrupted flag handling.

temporal-sdk/src/main/java/io/temporal/internal/worker/Poller.java
```diff
@@ -232,27 +237,44 @@ public void run() {
...
+    }
+
+    private boolean shouldInterrupt() {
+      boolean threadIsInterrupted = Thread.interrupted();
+      if (threadIsInterrupted) {
+        // it's a good idea to preserve the flag, because it's not obvious that this method will
+        // flush the flag
+        Thread.currentThread().interrupt();
+      }
+      return pollExecutor.isTerminating() || threadIsInterrupted;
```
Spikhalskiy, line 276: `pollExecutor.isTerminating()` is covering both `pollExecutor.isShutdown()` and `pollExecutor.isTerminating()`

temporal-sdk/src/main/java/io/temporal/internal/worker/Poller.java
```diff
@@ -213,13 +222,7 @@ public String toString() {
     @Override
     public void run() {
       try {
-        if (pollExecutor.isShutdown()) {
```
Spikhalskiy, line 216: This is really not needed here. This was just checked by us in finally AND by the thread pool.

temporal-sdk/src/main/java/io/temporal/internal/worker/Poller.java
```diff
@@ -232,27 +235,45 @@ public void run() {
...
         }
 
-        if (pollExecutor.isShutdown()) {
+        if (shouldTerminate()) {
           return;
         }
+
         task.run();
         pollBackoffThrottler.success();
+      } catch (InterruptedException e) {
```
vitarb, line 244: So, you just made `LocalActivityPollTask` consume `InterruptedException` and return immediately without exception. And this clause would only handle exceptions from `pollBackoffThrottler` and `suspender`, right?

temporal-sdk/src/main/java/io/temporal/internal/worker/Poller.java
```diff
@@ -232,27 +235,45 @@ public void run() {
...
+        // we restore the flag here, so it can be checked and processed (with exit) in finally
+        Thread.currentThread().interrupt();
```
vitarb, line 246: Should we also do `pollBackoffThrottler.failure()` to avoid leaking?

Spikhalskiy: I would say no. InterruptedException is not really something that should trigger a backoff, it's not really a "failure", it's just a sign that we should terminate the thread because something is canceled or shutting down. 
But I don't think it's very important or critical here, we can add it if you think it looks better this way.

temporal-sdk/src/main/java/io/temporal/internal/worker/Poller.java
vitarb: Makes sense.

vitarb (approved):

Spikhalskiy:
@vitarb Thank you for the review!

Issue #490, opened 2021-05-14
Add LocalActivityOption#contextPropagators


Spikhalskiy:
**Is your feature request related to a problem? Please describe.**
Looks like Temporal Java SDK is lacking the support of context propagators for local activities.
Currently, there is no way to make context propagation work for LocalActivities as described in Issue #489.
But even when #489 is resolved, Java SDK API lacks functionality for LocalActivities + ContextPropagators comparable to the one that is available for Activities + ContextPropagators.
`ActivityOption` has `ActivityOption#contextPropagators` that allow specifying custom context propagators for the activity.
You would expect that `LocalActivityOption` will have `LocalActivityOption#contextPropagators` with the same semantic, but it's not there.

**Describe the solution you'd like**
Add `LocalActivityOption#contextPropators` and ensure they are supported the same way as `ActivityOption#contextPropators`

Spikhalskiy:
I will take it after #489 is fixed

Spikhalskiy:
After discussion with @mfateev we decided that `ActivityOptions#contextPropagators` has extremely limited usage. Probably the only reasonable use case would be if we have an activity implemented in a different codebase/language. Otherwise, ContextPropagators should be specified on Worker options and be applied for all activities.
So, it's not a good idea to adapt it for local activities, and event existing `ActivityOptions#contextPropagators` should be used carefully with full understanding that they are applied only to the client/stub side.

PR #492, opened 2021-05-14
Add LocalActivityOptions#contextPropagators

Spikhalskiy:
## What was changed
Add LocalActivityOptions#contextPropagators that work the sam way as ActivityOptions#contextPropagators

1. Closes issue: 
Issue #490

⚠️  This PR is open for discussion. 
It kinda shows that ActivityOption#contextPropagators API looks borderline broken and misleading.

### DIFF
--- a/temporal-sdk/src/test/java/io/temporal/workflow/contextPropagatonTests/CustomActivityContextPropagationTest.java
+++ b/temporal-sdk/src/test/java/io/temporal/workflow/contextPropagatonTests/CustomActivityContextPropagationTest.java
@@ -1,0 +1,157 @@
+package io.temporal.workflow.contextPropagatonTests;
+
+import static org.junit.Assert.assertEquals;
+
+import io.temporal.activity.ActivityInterface;
+import io.temporal.activity.ActivityOptions;
+import io.temporal.activity.LocalActivityOptions;
+import io.temporal.client.WorkflowClient;
+import io.temporal.client.WorkflowClientOptions;
+import io.temporal.client.WorkflowOptions;
+import io.temporal.testing.TestEnvironmentOptions;
+import io.temporal.testing.TestWorkflowEnvironment;
+import io.temporal.worker.Worker;
+import io.temporal.workflow.Workflow;
+import io.temporal.workflow.WorkflowInterface;
+import io.temporal.workflow.WorkflowMethod;
+import java.time.Duration;
+import java.util.Arrays;
+import java.util.Collections;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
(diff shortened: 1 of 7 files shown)


temporal-sdk/src/test/java/io/temporal/workflow/contextPropagatonTests/CustomActivityContextPropagationTest.java
```diff
@@ -0,0 +1,151 @@
...
+    }
+  }
+
+  public static class ContextPropagationActivityWorkflowImpl implements TestWorkflow {
+    @Override
+    public String workflow1(String input) {
+      ActivityOptions options =
+          ActivityOptions.newBuilder()
+              .setScheduleToCloseTimeout(Duration.ofSeconds(5))
+              //              .setContextPropagators(
```
Spikhalskiy, line 90: Specifying any custom context propagators this way doesn't make sense right now.
These propagators HAVE TO BE in WorkflowClientOptions and without that they will not be used for deserialization. 
If they are in WorkflowClientOptions - they are used for serialization anyway for all activities, so specifying them on ActivityOptions doesn't change anything.

temporal-sdk/src/test/java/io/temporal/workflow/contextPropagatonTests/CustomActivityContextPropagationTest.java
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  public static class ContextPropagationLocalActivityWorkflowImpl implements TestWorkflow {
+    @Override
+    public String workflow1(String input) {
+      LocalActivityOptions options =
+          LocalActivityOptions.newBuilder()
+              .setScheduleToCloseTimeout(Duration.ofSeconds(5))
+              .setContextPropagators(
+                  Collections.singletonList(new TestContextPropagator(CUSTOM_KEY)))
```
Spikhalskiy, line 107: Specifying any custom context propagators this way doesn't make sense right now.
These propagators HAVE TO BE in WorkflowClientOptions and without that they will not be used for deserialization on the Activity Worker side.
If they are already in WorkflowClientOptions - they are used for serialization anyway for all activities, so specifying them on ActivityOptions doesn't change anything.

Spikhalskiy:
After discussion with @mfateev we decided that `ActivityOptions#contextPropagators` has extremely limited usage. Probably the only reasonable use case would be if we have an activity implemented in a different codebase/language. Otherwise, ContextPropagators should be specified on Worker options and be applied for all activities.
So, it's not a good idea to adapt it for local activities, and event existing `ActivityOptions#contextPropagators` should be used carefully with full understanding that they are applied only to the client/stub side.

PR #498, opened 2021-05-18
Option to set initial time for workflow test

GreyTeardrop:
## What was changed

Add a new `initialTime` attribute to the `TestEnvironmentOptions` and expose it via `TestWorkflowRule` and `TestWorkflowExtension`. If set, that attribute specifies the initial time to use for the workflow virtual clock.

The current behavior to use current system time as the initial workflow time is kept as a default.

## Why?

This change enables testing wall clock time-sensitive workflow logic, specifically workflows that are sensitive to the time they are started at. At the moment the time skipping is locked until the workflow is started, and the time the workflow time is set to the current system time, so there's no easy way to test how the workflow behaves if it's e.g. started at the midnight etc.

## Checklist

1. Closes issue: -

2. How was this tested: unit test.

3. Any docs updates needed? Javadocs are added for new methods.

### DIFF
--- a/temporal-testing-junit4/src/main/java/io/temporal/testing/TestWorkflowRule.java
+++ b/temporal-testing-junit4/src/main/java/io/temporal/testing/TestWorkflowRule.java
@@ -141,7 +141,7 @@ public class TestWorkflowRule implements TestRule {
     private boolean useExternalService;
     private boolean doNotStart;
     private long testTimeoutSeconds;
-    private long initialTimeMillis = 0;
+    private long initialTimeMillis;
 
     private Class<?>[] workflowTypes;
     private Object[] activityImplementations;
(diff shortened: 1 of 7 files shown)


temporal-testing-junit4/src/main/java/io/temporal/testing/TestWorkflowRule.java
```diff
@@ -139,6 +141,7 @@ public static Builder newBuilder() {
     private boolean useExternalService;
     private boolean doNotStart;
     private long testTimeoutSeconds;
+    private long initialTimeMillis = 0;
```
mfateev, line 144: We follow the pattern of not providing default values to fields as they are already initialized automatically in Java. So I would change it to:
```
private long initialTimeMillis;
```

mfateev (approved):
Thanks a lot for taking lead on this!

temporal-testing-junit4/src/main/java/io/temporal/testing/TestWorkflowRule.java
GreyTeardrop: Thanks! My initial though was to have those zeros kind of align with the `currentTime = initialTime == 0 ? System.currentTimeMillis() : initialTime` logic from the `SelfAdvancingTimerImpl` but now I see it really looks inconsistent with the other code. Updated.

Issue #474, opened 2021-05-08
Child workflows do not receive search attributes (or memo)


skrul:
## Expected Behavior

Just like regular workflows, child workflows should be created with search attributes and memo.

## Actual Behavior

Child workflows are not being created with search attributes and memo. This can be seen when you look at the child workflow's `WorkflowExecutionStarted` event in the history -- no search attributes.

I did some debugging on this and it seems that [SyncWorkflowContext::executeChildWorkflow](https://github.com/temporalio/sdk-java/blob/master/temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflowContext.java#L375) is missing the code to set this stuff.

  - Version: server 1.6.3 / java sdk 1.7.0
  - Platform:

PR #500, opened 2021-05-18
Fixed SearchAttributes and memo lost from ChildWorkflowOptions

vkoby:
## What was changed
Fixed a problem where searchAttributes set on ChildWorkflowOptions were getting lost and didn't show up in history.

## Checklist

1. Closes issue: [#474](https://github.com/temporalio/sdk-java/issues/474)

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowInfoImpl.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowInfoImpl.java
@@ -19,39 +19,14 @@
 
 package io.temporal.internal.sync;
 
-import io.temporal.api.common.v1.Payload;
 import io.temporal.api.common.v1.SearchAttributes;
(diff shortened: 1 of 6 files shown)


temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowInfoImpl.java
```diff
@@ -19,15 +19,40 @@
...
+    CustomStringField,
+    NamespaceId,
+    ExecutionTime,
+    HistoryLength,
+    RunId,
+    StartTime,
+    TaskQueue,
+    WorkflowId,
+    WorkflowType;
+  }
```
mfateev, line 54: I don't think we can hardcode the list of the search attributes as they can be added to a cluster at any time.

mfateev (changes requested):

mfateev (approved):

Issue #481, opened 2021-05-10
Allow configuration of workers to back off completely if the server is not available


eminn:
**Is your feature request related to a problem? Please describe.**
I want the workers to stop trying to connecting to the server if the server is not available using `WorkflowServiceStubsOptions`

**Describe the solution you'd like**
I'd like to configure workers to stop polling/trying to connecting to the server if the server is not available after a specified time period.

**Describe alternatives you've considered**
N/A

**Additional context**
Slack discussion https://temporalio.slack.com/archives/CTT84KXK9/p1620663515093300

Spikhalskiy:
> workers to stop polling/trying
I don't think stat "stopping" is the right semantic here. We still want them to reconnect when the network or the server is back. But it would be good to expose backoffCoefficient for workers.
Also, it would be great to expose backoffCoefficient for client stubs too as we discussed here: 
https://github.com/temporalio/sdk-java/pull/472/files#r627959282

vkoby:
Please take a look at [PollerOptions.java#L54](https://github.com/temporalio/sdk-java/blob/23bccf07d54982dd3048ab32f06c06c4811e74ce/temporal-sdk/src/main/java/io/temporal/internal/worker/PollerOptions.java#L54). We currently have default values for 
`pollBackoffCoefficient = 2`
`pollBackoffInitialInterval = 100ms`
`pollBackoffMaximumInterval = 1m`
In the end, the worker will keep polling the server every minute, which should be fine in most cases.
Does this work for you?

eminn:
@vkoby In my case I'd need to stop them to fail a k8s deployment as keeping the workers running and letting them try to connect non-existing server would create a lot of noise in terms of logging. This way I don't need to implement a custom health check logic (e.g. checking the namespace exists), the workers will die after configured time or attempt count passes.

vkoby:
@eminn Will implement health check. This has already been done in Go SDK see [client.go#L417](https://github.com/temporalio/sdk-go/blob/cea0147023cc324a8adce606d608cb7298e6cfcb/internal/client.go#L417)

PR #504, opened 2021-05-23
Health Check

vkoby:
## What was changed
Added optional health check that's performed by the client by making a request to endpoint to make sure that the server is accessible.

1. Closes issue: [#481](https://github.com/temporalio/sdk-java/issues/481)
2. How was this tested:
Unit test


vitarb (changes requested):
Please consider what we have in go, also look at https://github.com/temporalio/sdk-go/pull/453 where we've added retries for health check.
We need to have same/similar naming and configuration defaults across go and java.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsImpl.java
```diff
@@ -212,6 +227,32 @@ private ScheduledExecutorService grpcConnectionManager() {
...
 
+  /**
+   * Checks service health using gRPC health check:
+   * https://github.com/grpc/grpc/blob/master/doc/health-checking.md
+   *
+   * @throws StatusRuntimeException if the service is unavailable.
+   * @throws RuntimeException if the check returns unhealthy status.
+   * @return true if server is up.
+   */
+  private void checkHealth() {
```
vitarb, line 238: Should we inline it, since it does a single function call.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsOptions.java
```diff
@@ -77,6 +77,24 @@ public static WorkflowServiceStubsOptions getDefaultInstance() {
...
+   * Asks client to perform a health check after gRPC connection to the Server is created by making
+   * a request to endpoint to make sure that the server is accessible.
+   */
+  private final boolean disableHealthCheck;
+
+  /**
+   * HealthCheckAttemptTimeout specifies how to long to wait for service response on each health
+   * check attempt. Default: 5s.
+   */
+  private final Duration HealthCheckAttemptTimeout;
```
vitarb, line 90: s/HealthCheckAttemptTimeout/healthCheckAttemptTimeout

vitarb (approved):
Approved with a small comment.

PR #510, opened 2021-05-24
Workflow retry in Test Service

### LINKED ISSUE #414: Workflow retry after Activity failure works incorrectly with TestService
## Expected Behavior

We have a simple workflow that triggers an activity that fails on its first invocation and succeeds on subsequent calls.
We setup Activity#retryOptions into maxAttempt = 1 and Workflow#retryOptions into maxAttempt = 2.
We expect the workflow to pass after the second run.

## Actual Behavior

Workflow fails in unit tests, there is no second run of the workflow.

The following unit test is failing, while expected to pass:
https://github.com/Spikhalskiy/java-sdk/commit/3ff42f47a9a0c4da359cec07d6cccccf808e6836#diff-1cfa937ab805ee84ebdaf1e67c2f920b3ad72d53cf29e560061aeee0f3d7e012R91
 
If uncomment `//.setUseExternalService(true)` the test passes which points that the issue is probably in TestService implementation.

## Specifications

Current master: 8384e5393eb84be3044fced0b3db3daa8749c75f

vkoby:
## What was changed
Workflow retry logic in Test Service

## Why?
Was working differently than in prod

## Checklist
1. Closes #414 

2. How was this tested:
Unit test


temporal-sdk/src/test/java/io/temporal/workflow/TestRetry.java
```diff
@@ -0,0 +1,88 @@
...
+import io.temporal.failure.ApplicationFailure;
+import io.temporal.testing.TestWorkflowRule;
+import io.temporal.workflow.shared.TestActivities.TestActivity1;
+import io.temporal.workflow.shared.TestWorkflows.TestWorkflow1;
+import java.time.Duration;
+import java.util.concurrent.atomic.AtomicInteger;
+import org.junit.Rule;
+import org.junit.Test;
+
+public class TestRetry {
```
vitarb, line 37: We use naming `FooTest` for tests. Also just Retry sounds too generic, you should be more specific about purpose of this test.

temporal-sdk/src/test/java/io/temporal/workflow/shared/TestActivities.java
```diff
@@ -94,6 +94,11 @@
     Map<String, Duration> activity2();
   }
 
+  @ActivityInterface
```
vitarb, line 97: I think there is no reason to grow `TestActivities` interface, just move declaration to your test.

temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -1149,9 +1151,21 @@ private void processFailWorkflowExecution(
           failureType = Optional.of(failureInfo.getType());
           backoffInterval = rs.getBackoffIntervalInSeconds(failureType, store.currentTime());
         }
-      } else {
+      } else if (d.getFailure().hasTerminatedFailureInfo()
```
vitarb, line 1154: This needs some serious comment explaining what is going on here and why those checks are there.

temporal-sdk/src/test/java/io/temporal/workflow/TestRetry.java
```diff
@@ -0,0 +1,88 @@
...
+
+  @Rule
+  public TestWorkflowRule testWorkflowRule =
+      TestWorkflowRule.newBuilder()
+          .setWorkflowTypes(WorkflowImpl.class)
+          .setActivityImplementations(new FailingActivityImpl())
+          .build();
+
+  @Test
+  public void testActivityFailureFirstWorkflowRun() {
```
vitarb, line 49: Your tests are passing without that block that you've added in the `TestWorkflowMutableStateImpl`. What are they testing?

temporal-sdk/src/test/java/io/temporal/workflow/shared/TestActivities.java
vkoby: Used in more than 1 test, so I moved it here.

temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -1149,10 +1157,22 @@ private void processFailWorkflowExecution(
           failureType = Optional.of(failureInfo.getType());
           backoffInterval = rs.getBackoffIntervalInSeconds(failureType, store.currentTime());
         }
-      } else {
+      } else if (failure.hasTerminatedFailureInfo()
```
vitarb, line 1160: We can probably simplify this to
```
failure.hasTerminatedFailureInfo()
          || failure.hasCanceledFailureInfo()
          || failure.hasServerFailureInfo() && failure.getServerFailureInfo().getNonRetryable()
```
since it's only called for workflow tasks.

vitarb (approved):

PR #511, opened 2021-05-25
Pluggable span naming

### LINKED ISSUE #502: OpenTracing: support pluggable strategy for span naming/tagging
We use DataDog for APM and tracing, and they do support integration with OpenTracing.  However, they expect span naming and tagging to be in a particular format that does not align with the current Temporal OT implementation

Currently the Temporal OT implementation names the spans like `RunWorkflow_<WorkflowName>` and `RunActivity_<ActivityName>`, and adds tags for Workflow/Run ID.   However, DataDog expects a span name like `RunWorkflow` or `RunActivity` and a tag with the workflow/activity name. 

It would be great if the `OpenTracingOptions` provided the ability to completely override the strategy for naming/tagging spans.

sdwr98:
## What was changed
Creates the ability for users of the SDK to configure how OpenTracing spans are named and tagged

## Why?
DataDog APM (for example) supports OpenTracing, but has specific requirements and patterns for naming and tagging spans for them to show up properly

## Checklist

1. Closes #502 

2. How was this tested:
Tested in our development environment, and a unit test was added to ensure both default naming and custom naming worked as expected

3. Any docs updates needed?
No


temporal-opentracing/src/main/java/io/temporal/opentracing/OperationNameAndTagsProvider.java
```diff
@@ -0,0 +1,42 @@
...
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
+package io.temporal.opentracing;
+
+import java.util.Map;
+
+/** Allows a pluggable mechanism to configure an OpenTracing span's name and tags */
+public interface OperationNameAndTagsProvider {
```
vitarb, line 25: I suggest we turn this around a little bit and create pluggable `SpanContextBuilder` and make these functions `getSpanName` and `getSpanTags` available on the span context itself. This way we won't need this new abstraction that extracts tags/name out of the span context.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/SpanFactory.java
```diff
@@ -50,9 +49,10 @@ public SpanFactory(OpenTracingOptions options) {
       String workflowType,
       long startTimeMs,
       String workflowId) {
-    Map<String, String> tags = ImmutableMap.of(StandardTagNames.WORKFLOW_ID, workflowId);
-    String operationName =
-        options.getSpanOperationNamePrefix(operationType) + PREFIX_DELIMITER + workflowType;
+    StartSpanContext context =
```
vitarb, line 52: Passing multiple strings into one function is error prone, I suggest we use builder patter as I've mentioned above and make it more explicit.

temporal-opentracing/src/main/java/io/temporal/opentracing/OpenTracingOptions.java
```diff
@@ -55,13 +60,20 @@ public Tracer getTracer() {
...
+    return operationNameAndTagsProvider;
+  }
+
   public static Builder newBuilder() {
     return new Builder();
   }
 
   public static final class Builder {
     private Tracer tracer;
     private final Map<SpanOperationType, String> customSpanOperationNamePrefixes = new HashMap<>();
```
Spikhalskiy, line 74: I would expect this collection to go away with the introduction of ```OperationNameAndTagsProvider```.

I think this collection should be in ```DefaultOperationNameAndTagsProvider``` and somebody could create this ```DefaultOperationNameAndTagsProvider``` with custom profixes map.

sdwr98: I agree with this, I was just trying to maintain backwards compatibility

temporal-opentracing/src/main/java/io/temporal/opentracing/OperationNameAndTagsProvider.java
sdwr98: I'm a little unsure of what you're suggesting here.  Are you suggesting that we have a `StartSpanContextBuilder` that is set into `OpenTracingOptions` that has methods like `setWorkflowId` and `setRunId` for `SpanFactory` to call (as per your other comment), and also has implementations for `getSpanName` and `getSpanTags` that can be overridden as needed?

Spikhalskiy: I'm not in agreement with this proposal. 

> pluggable SpanContextBuilder

Builder pattern is not a strategy/provider/factory pattern. Builder shouldn't be pluggable. Builder shouldn't provide different implementations. We can use Builder pattern to just keep span name and tag in one basket if we want, but we shouldn't make Builder pluggable.

> getSpanName and getSpanTags available on the span context itself.

That's not what we want to achieve here. We want the same span context to produce different span names and tags depending on the plugged naming strategy.

> This way we won't need this new abstraction that extracts tags/name out of the span context.

I think that the creation of such abstraction is the sole goal of this PR...

temporal-opentracing/src/main/java/io/temporal/opentracing/OpenTracingOptions.java
sdwr98: are you suggesting that `Builder.setSpanOperationNamePrefix()` would pass through the prefixes to the `DefaultOperationNameAndTagsProvider` instead?

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/DefaultOperationNameAndTagsProvider.java
```diff
@@ -0,0 +1,59 @@
...
+package io.temporal.opentracing.internal;
+
+import com.google.common.collect.ImmutableMap;
+import io.temporal.opentracing.OperationNameAndTagsProvider;
+import io.temporal.opentracing.StandardTagNames;
+import io.temporal.opentracing.StartSpanContext;
+import java.util.HashMap;
+import java.util.Map;
+
+public class DefaultOperationNameAndTagsProvider implements OperationNameAndTagsProvider {
```
Spikhalskiy, line 29: Let's rename this class. `Default` doesn't provide any information about this strategy. 
You can put a note that it's the default strategy in the javadoc.
It can be something like `OperationTypeWithActionNameAsOperationName` instead.
Where OperationType means RunActivity, etc, and ActionName means workflow or activity type.
Also, there should be a javadoc explaining the output of this strategy with a couple of examples.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/DefaultOperationNameAndTagsProvider.java
```diff
@@ -0,0 +1,59 @@
...
+
+  private static final String PREFIX_DELIMITER = ":";
+
+  @Override
+  public String getSpanName(StartSpanContext context) {
+    return context.getOptions().getSpanOperationNamePrefix(context.getSpanOperationType())
```
Spikhalskiy, line 35: I think that prefixes probably should be a part of the strategy and should be moved here from options. If we want to give full flexibility - prefixes could go away completely in some other strategy.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/DefaultOperationNameAndTagsProvider.java
```diff
@@ -0,0 +1,59 @@
...
+            StandardTagNames.PARENT_WORKFLOW_ID, context.getParentWorkflowId(),
+            StandardTagNames.PARENT_RUN_ID, context.getParentRunId());
+      case RUN_WORKFLOW:
+      case START_ACTIVITY:
+      case RUN_ACTIVITY:
+        return ImmutableMap.of(
+            StandardTagNames.WORKFLOW_ID, context.getWorkflowId(),
+            StandardTagNames.RUN_ID, context.getRunId());
+    }
+    return new HashMap<>();
```
Spikhalskiy, line 57: I think that instead of this - the default clause in the switch should be provided with probably an exception instead of the empty map.

temporal-opentracing/src/main/java/io/temporal/opentracing/StartSpanContext.java
```diff
@@ -0,0 +1,76 @@
...
+
+  private OpenTracingOptions options;
+  private SpanOperationType spanOperationType;
+  private String typeName;
+  private String workflowId;
+  private String runId;
+  private String parentWorkflowId;
+  private String parentRunId;
+
+  public StartSpanContext(
```
Spikhalskiy, line 32: Maybe I'm not getting something, but why is it called Start*? It's not used to start anything and it's not only for Start operations (There is no RunSpanContext or smth)

Should it be called `SpanNamingContext` with javadoc saying something like:
"Contains all pieces of information needed by OperationNameAndTagsProvider"

temporal-opentracing/src/main/java/io/temporal/opentracing/StartSpanContext.java
```diff
@@ -0,0 +1,76 @@
...
+      OpenTracingOptions options,
+      SpanOperationType spanOperationType,
+      String typeName,
+      String workflowId,
+      String runId,
+      String parentWorkflowId,
+      String parentRunId) {
+    this.options = options;
+    this.spanOperationType = spanOperationType;
+    this.typeName = typeName;
```
Spikhalskiy, line 42: `typeName` is a meaningless name. Yeah, if it's `workflowType` and `activityType` - it's more or less clear what does it mean. But `typeName` is super misleading. Let's invent something like ActionName and add a javadoc clarifying what is expected here (workflowType or activityType)

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/SpanFactory.java
```diff
@@ -63,15 +63,17 @@ public SpanFactory(OpenTracingOptions options) {
...
+        new StartSpanContext(
+            options,
+            SpanOperationType.START_CHILD_WORKFLOW,
+            childWorkflowType,
+            workflowId,
+            null,
+            parentWorkflowId,
+            parentRunId);
+    Map<String, String> tags = options.getOperationNameAndTagsProvider().getSpanTags(context);
```
Spikhalskiy, line 75: Please lookup options.getOperationNameAndTagsProvider() into a local var  once and use it, code will be cleaner. The same s true for other usages like this in this PR

temporal-opentracing/src/main/java/io/temporal/opentracing/OpenTracingOptions.java
Spikhalskiy: No, I think that prefixes should go away from the main `OpenTracingOptions` and `OpenTracingOptionsBuilder` and become a part of the provider itself.
You can create DefaultOperationNameAndTagsProviderOptionsBuilder and make DefaultOperationNameAndTagsProviderOptionsBuilder.setSpanOperationNamePrefix() or provide the map in the DefaultOperationNameAndTagsProvider constructor.

I personally think that breaking of compatibility is alright here, this module was just released and I hope some instability of its API is expected.

Spikhalskiy:
I don't see the new strategy that you wanted to add here. If you just want a strategy that uses only `SpanOperationType` to define the span name and put what I called `action name` in one of the PR comments into tags - I think it's alright to add this strategy into the upstream with a JavaDoc explaining when and why this strategy can be preferred by the user. JaxRS has two or their standard implementations in their distribution and it's one of them.

temporal-opentracing/src/main/java/io/temporal/opentracing/OperationNameAndTagsProvider.java
Spikhalskiy: Ok, I think I maybe got what Vitaly proposes.

We have SpanFactory#createSpan and we can refactor it to call something like:
```
  interface SpanBuilderProvider {
    Tracer.SpanBuilder createSpanBuilder(Tracer tracer, StartSpanContext startSpanContext);
  }
```
and the implementation will be responsible for a population of information from StartSpanContext to Tracer.SpanBuilder. After that SpanFactory can finishing publishing of additional information into this builder.


and we can have a default implementation that will use our new OperationNameAndTagsProvider:
```
  class SpanBuilderFromSpanContentProvider {
    private final OperationNameAndTagsProvider spanContentProvider;

    public SpanBuilderFromSpanContentProvider(OperationNameAndTagsProvider spanContentProvider) {
      this.spanContentProvider = spanContentProvider;
    }

    Tracer.SpanBuilder createSpanBuilder(Tracer tracer, StartSpanContext context) {
      Tracer.SpanBuilder spanBuilder =
              tracer.buildSpan(spanContentProvider.getOperationName(context));


      spanContentProvider.getTags(context).forEach(spanBuilder::withTag);

      return spanBuilder;
    }
  }
```

And we can in that case define an instance of SpanBuilderProvider on OpenTracingOptions to give full flexibility to the span creation process. Not just methods getName and getTags, but this way we can fully hijack the process if we want.

I'm not sure. From one point of view, this way we make things even more flexible, but I'm not a big fan. I think it's a more clean way for temporal users to have a well defined interface with simpler methods like getName and getTags. I will not opposite this solution, but I think it's an overkill here. Maybe we can implement it only if somebody really needs it for something.

vitarb: Yes, my proposal is to put tags/name into span context and have a builder that builds it based on all available parameters. This way you could make default implementation and also allow provide ability to supply a custom builder, which would allow any type of transformation from parameters to tags/name. 
Functionally it will be the same, but will provide a little cleaner abstractions allowing us to get rid of 'OperationNameAndTagsProvider`.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/SpanFactory.java
vitarb: My concern is that it's easy to make a mistake and accidentally reorder parameters during refactoring and on the programmatic level, most likely you would only be able to validate set/not set, but not that what's set is actually what it's supposed to be. Builder eliminates this concern, also there are multiple places where we do pass nulls there...

temporal-opentracing/src/main/java/io/temporal/opentracing/OperationNameAndTagsProvider.java
Spikhalskiy: I like Scott's proposal with an interface with clearly defined `#getSmth` methods because it makes the implementation agnostic of OpenTracing Java classes. It will be a smoother design if we implement OpenTelemetry natively in addition to OpenTracing. Which is effectively the same as OpenTracing, but has different classes structure / packages. 
The current OperationNameAndTagsProvider interface and user's implementations of it can be used for both OpenTelemetry and OpenTracing without any code changes.
So, I want to propose that if we support a fully flexible SpanBuilderFromSpanContentProvider#createSpanBuilder - the default implementation of it should use OperationNameAndTagsProvider and accept it as a constructor parameter. This way majority of users will use our default SpanBuilderFromSpanContentProvider and provide only a custom OperationNameAndTagsProvider. And users who want full access to the span can override the whole SpanBuilderFromSpanContentProvider.

temporal-opentracing/src/main/java/io/temporal/opentracing/StartSpanContext.java
sdwr98: I called it that because it's used when we start a span. But I see the confusion. I don't like `SpanNamingContext` because it's not just about naming, but tagging.  I'm thinking `SpanCreationContext`

Spikhalskiy: Yeah, I think SpanCreationContext is a better name than StartSpanContext

temporal-opentracing/src/main/java/io/temporal/opentracing/OperationNameAndTagsProvider.java
sdwr98: OK, I wound up using the SpanContentProvider concept without the additional OperationNameAndTags provider - it just seemed a little cleaner and it's really easy to just override the `getSpanName` and `getSpanTags` on the default implementation.  Please take a look

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/SpanFactory.java
sdwr98: Builder created.

temporal-opentracing/src/main/java/io/temporal/opentracing/OpenTracingOptions.java
sdwr98: I removed the old operation name prefix - this is a breaking change, but hopefully nobody will mind too much

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/DefaultOperationNameAndTagsProvider.java
sdwr98: Done

temporal-opentracing/src/main/java/io/temporal/opentracing/StartSpanContext.java
sdwr98: Done

temporal-opentracing/src/main/java/io/temporal/opentracing/StartSpanContext.java
sdwr98: Done - I used `operationName` to go along with `operationType`

temporal-opentracing/src/main/java/io/temporal/opentracing/SpanCreationContext.java
```diff
@@ -0,0 +1,128 @@
...
+
+package io.temporal.opentracing;
+
+/**
+ * Used when creating an OpenTracing span and provides contextual information used for naming and
+ * tagging OT spans.
+ */
+public class SpanCreationContext {
+
+  private SpanOperationType spanOperationType;
```
Spikhalskiy, line 28: Can and should be final

temporal-opentracing/src/main/java/io/temporal/opentracing/provider/DataDogOpenTracingSpanBuilderProvider.java
```diff
@@ -0,0 +1,60 @@
...
+import io.temporal.opentracing.SpanCreationContext;
+import io.temporal.opentracing.internal.SpanBuilderFromSpanContentProvider;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * This implementation of {@link SpanBuilderProvider} names and tags the OpenTracing spans in a way
+ * that is compatible with DataDog's APM tooling.
+ */
+public class DataDogOpenTracingSpanBuilderProvider extends SpanBuilderFromSpanContentProvider
```
Spikhalskiy, line 32: Can we rename it from DataDog into something more generic? It doesn't have to be used with DataDog, it doesn't depend on any DataDog classes or super custom logic? So, can we choose a prefix that describes the logic of this class instead of a specific vendor?

temporal-opentracing/src/main/java/io/temporal/opentracing/provider/DataDogOpenTracingSpanBuilderProvider.java
```diff
@@ -0,0 +1,60 @@
...
+    implements SpanBuilderProvider {
+
+  /**
+   * Uses just the operation type as the name, e.g. "StartActivity"
+   *
+   * @param context
+   * @return
```
Spikhalskiy, line 39: Please provide short descriptions for parameters and the return value or delete the "@param context" and "@return", it's meaningless now.

temporal-opentracing/src/main/java/io/temporal/opentracing/provider/DataDogOpenTracingSpanBuilderProvider.java
```diff
@@ -0,0 +1,60 @@
...
+  @Override
+  protected String getSpanName(SpanCreationContext context) {
+    return context.getSpanOperationType().getDefaultPrefix();
+  }
+
+  /**
+   * Includes the default tags but also uses {@link SpanCreationContext#getOperationName()} as a tag
+   * with the key "resource.name"
+   *
+   * @param context
```
Spikhalskiy, line 50: Please provide short descriptions for parameters and the return value or delete the "@param context" and "@return", it's meaningless now.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/SpanBuilderFromSpanContentProvider.java
```diff
@@ -0,0 +1,80 @@
...
+package io.temporal.opentracing.internal;
+
+import com.google.common.collect.ImmutableMap;
+import io.opentracing.Tracer;
+import io.temporal.opentracing.SpanBuilderProvider;
+import io.temporal.opentracing.SpanCreationContext;
+import io.temporal.opentracing.StandardTagNames;
+import java.util.Map;
+
+public class SpanBuilderFromSpanContentProvider implements SpanBuilderProvider {
```
Spikhalskiy, line 29: This class has the logic that defines the name and tags creation inside now. So, let's give it a more meaningful name that describes this logic. Something like OpTypeAndNameAsSpanNameProvider.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/SpanBuilderFromSpanContentProvider.java
```diff
@@ -0,0 +1,80 @@
...
+
+package io.temporal.opentracing.internal;
+
+import com.google.common.collect.ImmutableMap;
+import io.opentracing.Tracer;
+import io.temporal.opentracing.SpanBuilderProvider;
+import io.temporal.opentracing.SpanCreationContext;
+import io.temporal.opentracing.StandardTagNames;
+import java.util.Map;
+
```
Spikhalskiy, line 28: Would be great to add a javadoc with a short description and an example of an input and a result span.

temporal-opentracing/src/main/java/io/temporal/opentracing/provider/DataDogOpenTracingSpanBuilderProvider.java
Spikhalskiy, line 32: Would be great to add a javadoc with a short description and an example of an input and a result span.
It's ok to leave the info that it's compatible with Datadog tooling, but it's not enough to understand what is the result and when the user would prefer to use this provider outside of Datadog context.

temporal-opentracing/src/main/java/io/temporal/opentracing/provider/DataDogOpenTracingSpanBuilderProvider.java
```diff
@@ -0,0 +1,60 @@
...
+   * @return
+   */
+  @Override
+  protected Map<String, String> getSpanTags(SpanCreationContext context) {
+    Map<String, String> tags = new HashMap<>();
+    tags.putAll(super.getSpanTags(context));
+    tags.put("resource.name", context.getOperationName());
```
Spikhalskiy, line 57: Please make constants for text values with self-explanatory names.

Spikhalskiy:
I support the new version of this PR and I think it addresses most of my and @vitarb recommendations. I left some more comments about code style and naming that should be addressed before the merge, but they are localized and should be easy to address now.

temporal-opentracing/src/main/java/io/temporal/opentracing/SpanCreationContext.java
```diff
@@ -0,0 +1,128 @@
...
+  private String operationName;
```
Spikhalskiy, line 29: I think `operationName` should be renamed to something like `actionName`, `resourceName`, or AT LEAST `opName`.
The reason is that `operationName` is already used/reserved by both OpenTracing and OpenTelemetry and it means "span name". Reusing it here looks very ambiguous.

temporal-opentracing/src/main/java/io/temporal/opentracing/provider/DataDogOpenTracingSpanBuilderProvider.java
sdwr98: Well, this is not just the span naming but also datadog-specific tag names.  I feel like somebody coming in here working with DataDog would benefit from having a provider that is known to work with that vendor.

Spikhalskiy: Discussed with @mfateev, let's leave DataDogOpenTracingSpanBuilderProvider name and datadog related constants inside this class.

Spikhalskiy (approved):

Spikhalskiy (approved):
Thanks for the PR and addressing the comments!

temporal-opentracing/src/main/java/io/temporal/opentracing/OpenTracingOptions.java
```diff
@@ -61,23 +58,28 @@ public static Builder newBuilder() {
...
 
     public void setTracer(Tracer tracer) {
       this.tracer = tracer;
     }
 
-    public Builder setSpanOperationNamePrefix(
-        SpanOperationType spanOperationType, String customPrefix) {
-      this.customSpanOperationNamePrefixes.put(spanOperationType, customPrefix);
+    /**
+     * Allows for more control over how the OpenTracing span is created, named, and tagged.
```
Sushisource, line 70:
```suggestion
     * Allows for more control over how OpenTracing spans are created, named, and tagged.
```

Sushisource (approved):
Looking great! Thanks for the contribution!

PR #513, opened 2021-05-26
Refactors gradle build configuration to simplify and remove duplication

cfieber:
## What was changed
Each of the submodules in the project were explicitly configuring build opinions.

This change extracts that common configuration into some shared script files, and
applies those scripts in the top level build.gradle.

The commits in this PR are incremental and independently buildable and generally
follow the route of moving one piece of build configuration at a time first from the
submodules to the top level build.gradle, and then extracting the common config
from build.gradle into separate scripts. (This could certainly be squashed down to
a smaller number of commits if desired)

## Why?
The motivation behind this is that we need to maintain a forked version of
the repo that builds due to internal dependency conflicts related to grpc 
versioning and other legacy library dependency opinions. 

With this config cleanup it will be much easier for us to ensure we don't drift from
upstream with the patches we need to apply to override dependency versions.

## Checklist
1. How was this tested:

* The build still runs and the tests still passes.
* I've performed a ./gradlew publishToMavenLocal to insure the pom files are correct.
* I've used the commented out block in publishing.gradle to ensure the nexus publish plugin
correctly publishes (writing to `$HOME/repos`) and that signing works.
* We've patched on top of this with our own versioning and integrated used the resulting SDK
against our temporal service.


mfateev (approved):
Thanks a lot! Really bit improvement.

PR #518, opened 2021-05-28
Release v1.0.8

vkoby:
Release v1.0.8 highlights:

- Option to set default ActivityOptions and per activityType. Can be done
- through the activityStub, WorkflowImplementationOptions, and on a workflow.
- Added context propagation for local activities
- Option to set initial time for workflow tests
- Implemented a Health Check. By default it's on and checks if the server is up
- every 5s. If the server is not up in 10s, the worker will not start.

### DIFF
--- a/releases/v1.0.8
+++ b/releases/v1.0.8
@@ -5,20 +5,18 @@ through the activityStub, WorkflowImplementationOptions, and on a workflow.
 * Added context propagation for local activities
 * Option to set initial time for workflow tests
 * Implemented a Health Check. By default it's on and checks if the server is up
-every 5s. If the server is not up in 10s, the worker will not sart.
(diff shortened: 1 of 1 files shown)


releases/v1.0.8
```diff
@@ -0,0 +1,39 @@
+Highlights
+
+* Option to set default ActivityOptions and per activityType. Can be done
+through the activityStub, WorkflowImplementationOptions, and on a workflow.
+* Added context propagation for local activities
+* Option to set initial time for workflow tests
+* Implemented a Health Check. By default it's on and checks if the server is up
```
vitarb, line 7: We should explicitly call out that this change would break users that deploy workers while server is not up.

vitarb (approved):

PR #524, opened 2021-06-02
Implement configurable DEFAULT_DEADLOCK_DETECTION_TIMEOUT

### LINKED ISSUE #503: Make DEFAULT_DEADLOCK_DETECTION_TIMEOUT configurable

We have a custom data converter that can get a little slow due to certain cluster conditions and we sometimes trip over this 1s timeout.

Make DEFAULT_DEADLOCK_DETECTION_TIMEOUT configurable, see https://github.com/temporalio/sdk-java/blob/master/temporal-sdk/src/main/java/io/temporal/internal/sync/DeterministicRunner.java#L37

Forking 😅

vkoby:
## What was changed
Implemented configurable deadlock.

1. Closes #503  

2. How was this tested:
Unit tests


temporal-sdk/src/main/java/io/temporal/internal/sync/DeterministicRunnerImpl.java
```diff
@@ -81,42 +81,38 @@
...
-  private static class NamedRunnable {
-    private final String name;
-    private final Runnable runnable;
-
-    private NamedRunnable(String name, Runnable runnable) {
-      this.name = name;
-      this.runnable = runnable;
-    }
-  }
+  static final String WORKFLOW_ROOT_THREAD_NAME = "workflow-method";
```
Sushisource, line 84: Is there any particular reason these fields all got shuffled around? Makes it a little hard to find the relevant changes in the diff, so I generally try to put pure style changes like this in their own PRs.

vkoby: No, just noticed that this class has weird format while debugging. I'll make sure to to make a separate PR next time.

temporal-sdk/src/test/java/io/temporal/workflow/DeadlockDetectorTest.java
```diff
@@ -25,52 +25,76 @@
...
-    @Override
-    public void execute() {
-      Async.procedure(() -> Workflow.await(() -> false));
-      Workflow.sleep(Duration.ofSeconds(1));
-      try {
-        Thread.sleep(2000);
-      } catch (InterruptedException e) {
-        throw Workflow.wrap(e);
+  @Rule
+  public SDKTestWorkflowRule testWorkflowRule2 =
```
Sushisource, line 51: These rules should probably have more descriptive names than rule 1 and rule 2

temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflowWorker.java
```diff
@@ -62,22 +62,22 @@ public SyncWorkflowWorker(
...
 
     factory =
         new POJOWorkflowImplementationFactory(
-            workflowOptions.getDataConverter(),
+            singleWorkerOptions.getDataConverter(),
             workflowThreadPool,
             workerInterceptors,
             cache,
-            workflowOptions.getContextPropagators());
+            singleWorkerOptions);
```
Sushisource, line 80: If the data converter is always taken from the worker options when this is called, we could eliminate it as a separate parameter.

Sushisource (approved):

PR #528, opened 2021-06-04
Add getWorker method to TestWorkflowRule

tsurdilo:

## What was changed

Added getWorker method to TestWorkflowRule.

## Why?

Currently if you have multiple tests where you want to register different activities (one might use real activities impl and another test might want to use a mock for example) you have to do:

      TestWorkflowRule.newBuilder()
                .setWorkflowTypes(GreetingWorkflowImpl.class)
                .setDoNotStart(true)
                .build();

and then in each test method to register activity:

            workflowRule
                    .getTestEnvironment()
                    .getWorkerFactory()
                    .getWorker(workflowRule.getTaskQueue())
                    .registerActivitiesImplementations(new GreetingActivitiesImpl());

This adds a convenience getWorker method so registering activities in test methods can be just

          workflowRule.getWorker().registerActivitiesImplementations(new GreetingActivitiesImpl());

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/temporal-testing-junit4/src/main/java/io/temporal/testing/TestWorkflowRule.java
+++ b/temporal-testing-junit4/src/main/java/io/temporal/testing/TestWorkflowRule.java
@@ -350,7 +350,7 @@ public class TestWorkflowRule implements TestRule {
   /**
    * Returns the default worker created for each test method.
    * This worker listens to the default task queue which is obtainable
-   * via the {@link #getTaskQueue()} method
+   * via the {@link #getTaskQueue()} method.
    */
   public Worker getWorker() {
     return testEnvironment.getWorkerFactory().getWorker(getTaskQueue());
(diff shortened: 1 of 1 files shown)


temporal-testing-junit4/src/main/java/io/temporal/testing/TestWorkflowRule.java
```diff
@@ -346,4 +346,9 @@ public History getWorkflowExecutionHistory(WorkflowExecution execution) {
             .build();
     return this.blockingStub().getWorkflowExecutionHistory(request).getHistory();
   }
+
+  /** Returns the worker */
```
mfateev, line 350: Improve the comment to state which worker as it is possible to have more than one worker per rule.

tsurdilo:
@mfateev @vitarb thanks for review! udated

mfateev (approved):

PR #550, opened 2021-06-22
Prohibit use of WorkflowClient from workflow code

### LINKED ISSUE #402: Prohibit use of WorkflowClient and Activity interfaces in workflow code.
New users might not be aware of nondeterministic requirements and might use WorkflowClient and Activity interfaces from the workflow code.

Add explicit checks that throw exceptions from WorkflowClient and Activity APIs if they are called from workflow threads.

vkoby:
## What was changed
Added explicit checks that throw exception from WorkflowClient and Activity APIs if they are called from workflow threads.

## Why?
New users might not be aware of nondeterministic requirements and might use WorkflowClient and Activity interfaces from the workflow code.

## Checklist
1. Closes #402 

2. How was this tested:
Unit test

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowClientInternal.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowClientInternal.java
@@ -50,6 +50,8 @@ import java.util.Arrays;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 
+import static io.temporal.internal.sync.WorkflowInternal.enforceNonWorkflowThread;
+
 public final class WorkflowClientInternal implements WorkflowClient {
 
   private final GenericWorkflowClientExternalImpl genericClient;
@@ -426,11 +428,5 @@ public final class WorkflowClientInternal implements WorkflowClient {
       A5 arg5,
       A6 arg6) {
     return execute(() -> workflow.apply(arg1, arg2, arg3, arg4, arg5, arg6));
-  }
-
-  private static void enforceNonWorkflowThread() {
-    if (DeterministicRunnerImpl.currentThreadInternalIfPresent().isPresent()) {
-      throw new IllegalStateException("Cannot be called from workflow thread.");
-    }
   }
 }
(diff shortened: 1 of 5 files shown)


temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowClientInternal.java
```diff
@@ -414,4 +427,10 @@ public static WorkflowExecution start(Functions.Proc workflow) {
       A6 arg6) {
     return execute(() -> workflow.apply(arg1, arg2, arg3, arg4, arg5, arg6));
   }
+
+  private static void checkThread() {
+    if (Thread.currentThread().getName().startsWith("workflow")) {
```
Spikhalskiy, line 432: This doesn't look like a good method to check the thread for being a workflow. What if some user code names it's own thread "workflow"?

Looking at currently existing code, the right way to do this is
```
DeterministicRunnerImpl.currentThreadInternalIfPresent().isPresent();
```

I would actually add a method like
`WorkflowInternal#isCurrentThreadWokflow` containing this exact code.

temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowClientInternal.java
```diff
@@ -414,4 +427,10 @@ public static WorkflowExecution start(Functions.Proc workflow) {
       A6 arg6) {
     return execute(() -> workflow.apply(arg1, arg2, arg3, arg4, arg5, arg6));
   }
+
+  private static void checkThread() {
```
Spikhalskiy, line 431: Please name with something more descriptive. Like `enforceWorkflowThread`.

temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowClientInternal.java
vkoby: I'm confused why the above guarantees that this is a wf thread as opposed to just any other internal thread. 
Though I agree that relying on the name of the thread is not very robust. This is where I encountered this way of identifying threads:
[ExceptionInSignalTest.java#L69](https://github.com/temporalio/sdk-java/blob/ce9c1dc5f9105f8631092364f4022407c3ae4ec3/temporal-sdk/src/test/java/io/temporal/workflow/signalTests/ExceptionInSignalTest.java#L69)

Spikhalskiy:
Implemented in #556

PR #552, opened 2021-06-23
Update ProtobufJsonPayloadConverter.java

### LINKED ISSUE #551: ProtobufJsonPayloadConverter not using custom printer
## Expected Behavior
`ProtobufJsonPayloadConverter` using predefined (passed through constructor) `JsonFormat.Printer`.

## Actual Behavior
It uses a default one (`JsonFormat.Printer.printer()`), it is being created on line 60 in `ProtobufJsonPayloadConverter`, shadowing the instance field `printer`.

  1. Create `ProtobufJsonPayloadConverter` with custsom JsonFormat.Printer by passing it through constructor.
  2. The custom printer is not used as the default one is used instead.

## Specifications

  - Version: 1.0.9
  - Platform: N/A

mschielmann:
Regarding #551 - bugfix - use preconfigured printer rather than default JsonFormat.printer();

## What was changed
Bugfix - In JavaSDK, ProtobufJsonPayloadConverter allows you to configure both printer and parser. Yet, the preconfigured printer has not been used to convert Object to data. Default JsonFormat printer has been used instead.

## Why?
Bugfix to allow configurable printer to be used (as designed, it can be passed as constructor param).

## Checklist
1. Closes #551

2. How was this tested:
Passed custom JsonFormat.Printer to constructor and see it being used.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/common/converter/ProtobufJsonPayloadConverter.java
+++ b/temporal-sdk/src/main/java/io/temporal/common/converter/ProtobufJsonPayloadConverter.java
@@ -57,7 +57,7 @@ public final class ProtobufJsonPayloadConverter implements PayloadConverter {
     if (!(value instanceof MessageOrBuilder)) {
       return Optional.empty();
     }
-    JsonFormat.Printer printer = JsonFormat.printer();
+
     try {
       String data = printer.print((MessageOrBuilder) value);
       return Optional.of(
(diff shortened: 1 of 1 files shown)


Spikhalskiy (changes requested):
While this change is totally reasonable and compatible, it looks like it will change the way we serialize photo field names. We use `JsonFormat.printer()` now and we will use `JsonFormat.printer().preservingProtoFieldNames()` if this PR applied as it is. Could you please edit the constructor to preserve the current behavior?

mschielmann:
hi @Spikhalskiy thank you for your review - good catch, I've made a change you've requested (I hope that's what you've meant).

Spikhalskiy (approved):

PR #556, opened 2021-06-25
Prohibit use of WorkflowClient from workflow code

### LINKED ISSUE #402: Prohibit use of WorkflowClient and Activity interfaces in workflow code.

Spikhalskiy:
## What was changed
Now the usage of `WorkflowClient`, `ActivityCompletionClient`, and `WorkflowServiceStubs` is prohibited from workflow code.

It's an alternative PR to the original PR #550. 

The main difference is that this PR doesn't rely on magic thread names to understand that we are inside a workflow thread, but a marker utility class was implemented that is published by DeterministicRunner and available to serviceclient code.

It includes a fix for `TestActivityEnvironmentInternal` to execute activities code in a separate thread pool, not in a main thread of DeterministicRunnerImpl to emulate the behavior of our real worker setup and allow a correct work of implemented checks for ActivityCompletionClient (this class is actively used in activities code)

Also, this PR uses proxies to 

- ensure the right type of a calling thread for each non-static method of `WorkflowClient`, `ActivityCompletionClient`, and `WorkflowServiceStubs` instances
- decrease copy-paste
- minimize mistakes in the future.

## Why?
New users might not be aware of nondeterministic requirements and might use WorkflowClient and Activity interfaces from the workflow code.

Closes #402

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowClientInternal.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowClientInternal.java
@@ -19,6 +19,8 @@
 
 package io.temporal.internal.sync;
 
+import static io.temporal.internal.WorkflowThreadMarker.enforceNonWorkflowThread;
+
 import com.google.common.base.Strings;
 import com.google.common.reflect.TypeToken;
 import com.uber.m3.tally.Scope;
@@ -32,6 +34,7 @@ import io.temporal.client.WorkflowStub;
 import io.temporal.common.converter.DataConverter;
 import io.temporal.common.interceptors.WorkflowClientCallsInterceptor;
 import io.temporal.common.interceptors.WorkflowClientInterceptor;
+import io.temporal.internal.WorkflowThreadMarker;
 import io.temporal.internal.client.RootWorkflowClientInvoker;
 import io.temporal.internal.external.GenericWorkflowClientExternalImpl;
 import io.temporal.internal.external.ManualActivityCompletionClientFactory;
@@ -62,7 +65,8 @@ public final class WorkflowClientInternal implements WorkflowClient {
   private final Scope metricsScope;
 
   /**
-   * Creates client that connects to an instance of the Temporal Service.
+   * Creates client that connects to an instance of the Temporal Service. Cannot be used from within
+   * workflow code.
    *
    * @param service client to the Temporal Service endpoint.
    * @param options Options (like {@link io.temporal.common.converter.DataConverter} override) for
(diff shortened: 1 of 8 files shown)


temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowClientInternal.java
```diff
@@ -62,15 +65,18 @@
...
    *
    * @param service client to the Temporal Service endpoint.
    * @param options Options (like {@link io.temporal.common.converter.DataConverter} override) for
    *     configuring client.
    */
   public static WorkflowClient newInstance(
       WorkflowServiceStubs service, WorkflowClientOptions options) {
-    return new WorkflowClientInternal(service, options);
+    enforceNonWorkflowThread();
+    return WorkflowThreadMarker.protectFromWorkflowThread(
```
vitarb, line 78: We should consider adding an option that would disable this enforcement. If users are already running "bad" workflows in production, then there should be a way to disable this feature until workflows are refactored/updated, otherwise we are forcing users to stay on older SDK versions.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/internal/WorkflowThreadMarker.java
```diff
@@ -0,0 +1,77 @@
...
+  @SuppressWarnings("unchecked")
+  public static <T> T protectFromWorkflowThread(T instance, Class<T> iface) {
+    return (T)
+        Proxy.newProxyInstance(
+            iface.getClassLoader(),
+            new Class<?>[] {iface},
+            (proxy, method, args) -> {
+              // you make your check here
+              if (isWorkflowThread()) {
+                throw new IllegalStateException("Cannot be called from workflow thread.");
```
vitarb, line 67: Let's add details to the error message with which function and on which class has been called.

temporal-testing/src/main/java/io/temporal/testing/TestActivityEnvironmentInternal.java
```diff
@@ -327,11 +326,42 @@ public void close() {
...
+      Future<Result> activityFuture =
+          activityWorkerExecutor.submit(
+              () ->
+                  activityTaskHandler.handle(
+                      new ActivityTask(activityTask, () -> {}),
+                      testEnvironmentOptions.getMetricsScope(),
+                      localActivity));
+
+      try {
+        // 10 seconds is just a "reasonable" wait to don't make an infinite wait
```
vitarb, line 353: nit. s/to don't/to not/

vitarb (approved):
LGTM overall, as mentioned above, I would add a feature flag to provide backwards compatibility option for existing workflows.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/internal/WorkflowThreadMarker.java
Spikhalskiy: Why do we need it in the message if this will be in a stacktrace?

Issue #537, opened 2021-06-11
Open tracing span context not propagated when activity or child workflow invoked asynchronously 


smax48:
## Expected Behavior

Given a workflow that invokes some activities and/or child workflows in an async way, trace context should correctly propagate into those async activities/child WF. Also the result trace should not be broken into multiple unrelated chunks

## Actual Behavior

Result trace for the main WF doesn't contain spans for async activities / child WF - their execution is represented in isolated traces instead.

## Steps to Reproduce the Problem

The easiest way to reproduce is to use SpanContextPropagationTest from temporal-opentracing and change activity invocation to Async.function {...}.get()

## Specifications

  - Version: Temporal Java SDK 1.0.9

Spikhalskiy:
@smax48 thanks for the report, it does look like a bug.

Failing unit test: https://github.com/Spikhalskiy/java-sdk/commit/1811aa0818df841ee97d4d44d857335662e24c83#diff-b6c10900bd776516302d896d09b7946aaf53b70297ec46c5f75eb0d76b212998R124

smax48:
@Spikhalskiy , thanks for confirming this!
Async execution is always a problem for things like tracing where you need to preserve the single context across multiple threads... 
For example, in Armeria framework they use their own request context to store the trace context - so it correctly flows across threads.
For temporal, I think that if "normal" workflow context propagation (`ContextPropagator` interface) works correctly with async (I do hope, but need to check that) - that can be potentially used in the open tracing scenario. 
But it might be an easier way - you definitely know this better! :)

Spikhalskiy:
A little bit more context. If workflow triggers an async activity the following way:
``` 
return Async.function(activity::activity, input).get();
```
Interceptors and propagation actually work correctly. Here is the unit test verifying it:
https://github.com/Spikhalskiy/java-sdk/commit/59d620a2c319659c447e72a6617a74b4cf20c8e9#diff-b6c10900bd776516302d896d09b7946aaf53b70297ec46c5f75eb0d76b212998R146

But if workflow triggers the activity using lambda:
```
return Async.function(() -> activity.activity(input)).get();
```
the context is lost and interceptors don't see the context of the original thread. Here is a failing unit test verifying it:
https://github.com/Spikhalskiy/java-sdk/commit/59d620a2c319659c447e72a6617a74b4cf20c8e9#diff-b6c10900bd776516302d896d09b7946aaf53b70297ec46c5f75eb0d76b212998R186

[...]

PR #557, opened 2021-06-28
Rework how DeterministicRunnerImpl and SyncWorkflowContext handle creation of workflow root and method threads

Spikhalskiy:
Refactored workflow root thread into a separate code path making it explicitly different from workflow-method thread and child workflows.
Added WorkflowInboundCallsInterceptor#newWorkflowMethodThread intercepting creation of the main workflow-method thread.
Added WorkflowInboundCallsInterceptor#newCallbackThread intercepting creation of a callback thread.
WorkflowOutboundCallsInterceptor#newThread is renamed into WorkflowOutboundCallsInterceptor#newChildThread and now it intercepts only new workflow child threads and doesn't intercept creation of the main workflow-method thread.
Fixed situation that the same "workflow-method" name was used for workflow initializer thread (now called "workflow-root") and for the thread that is used to run a workflow method (continued to be "workflow-method").

This PR is a first step to address issue #537

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkflowInboundCallsInterceptor.java
+++ b/temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkflowInboundCallsInterceptor.java
@@ -19,6 +19,8 @@
 
 package io.temporal.common.interceptors;
 
+import javax.annotation.Nullable;
+
 /**
  * Intercepts calls to the workflow execution. Executes under workflow context. So all the
  * restrictions on the workflow code should be obeyed.
@@ -128,4 +130,24 @@ public interface WorkflowInboundCallsInterceptor {
 
   /** Called when a workflow is queried. */
   QueryOutput handleQuery(QueryInput input);
+
+  /**
+   * Intercepts creation of the workflow main method thread
+   *
+   * @param runnable thread function to run
+   * @param name name of the thread, optional
+   * @return created workflow thread. Should be treated as a pass-through object that shouldn't be
+   *     manipulated in any way by the interceptor code.
+   */
+  Object newWorkflowMethodThread(Runnable runnable, @Nullable String name);
+
+  /**
+   * Intercepts creation of a workflow callback thread
+   *
+   * @param runnable thread function to run
+   * @param name name of the thread, optional
+   * @return created workflow thread. Should be treated as a pass-through object that shouldn't be
+   *     manipulated in any way by the interceptor code.
+   */
+  Object newCallbackThread(Runnable runnable, @Nullable String name);
 }
--- a/temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkflowOutboundCallsInterceptor.java
+++ b/temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkflowOutboundCallsInterceptor.java
@@ -22,13 +22,8 @@ package io.temporal.common.interceptors;
 import io.temporal.activity.ActivityOptions;
 import io.temporal.activity.LocalActivityOptions;
 import io.temporal.api.common.v1.WorkflowExecution;
-import io.temporal.workflow.ChildWorkflowOptions;
-import io.temporal.workflow.ContinueAsNewOptions;
-import io.temporal.workflow.DynamicQueryHandler;
-import io.temporal.workflow.DynamicSignalHandler;
-import io.temporal.workflow.Functions;
+import io.temporal.workflow.*;
 import io.temporal.workflow.Functions.Func;
-import io.temporal.workflow.Promise;
 import java.lang.reflect.Type;
 import java.time.Duration;
 import java.util.List;
@@ -483,7 +478,18 @@ public interface WorkflowOutboundCallsInterceptor {
 
   void upsertSearchAttributes(Map<String, Object> searchAttributes);
 
-  Object newThread(Runnable runnable, boolean detached, String name);
+  /**
+   * Intercepts creation of the workflow child thread.
+   *
+   * <p>Please note, that "workflow child thread" and "child workflow" are different and independent
+   * concepts.
+   *
+   * @param runnable thread function to run
+   * @param detached if this thread is detached from the parent {@link CancellationScope}
+   * @param name name of the thread
+   * @return created WorkflowThread
+   */
+  Object newChildThread(Runnable runnable, boolean detached, String name);
 
   long currentTimeMillis();
 }
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/BaseRootWorkflowInboundCallsInterceptor.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/BaseRootWorkflowInboundCallsInterceptor.java
@@ -1,0 +1,67 @@
+/*
+ *  Copyright (C) 2020 Temporal Technologies, Inc. All Rights Reserved.
+ *
+ *  Copyright 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ *  Modifications copyright (C) 2017 Uber Technologies, Inc.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License"). You may not
+ *  use this file except in compliance with the License. A copy of the License is
+ *  located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ *  or in the "license" file accompanying this file. This file is distributed on
+ *  an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
+package io.temporal.internal.sync;
+
+import io.temporal.common.interceptors.WorkflowInboundCallsInterceptor;
+import io.temporal.common.interceptors.WorkflowOutboundCallsInterceptor;
+
+/**
+ * Provides core functionality for a root WorkflowInboundCallsInterceptor that is reused by specific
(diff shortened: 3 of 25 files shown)


temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkflowInboundCallsInterceptor.java
```diff
@@ -128,4 +130,22 @@ public Object getResult() {
 
   /** Called when a workflow is queried. */
   QueryOutput handleQuery(QueryInput input);
+
+  /**
+   * Intercepts creation of the workflow main method thread
+   *
+   * @param runnable thread function to run
+   * @param name name of the thread, optional
+   * @return created WorkflowThread
```
mfateev, line 139: Change comment to smth like:

@return created workflow thread. Treat the result as an opaque object that shouldn't be manipulated in any way by the interceptor code.

temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkflowOutboundCallsInterceptor.java
```diff
@@ -483,7 +478,15 @@ public DynamicSignalHandler getHandler() {
 
   void upsertSearchAttributes(Map<String, Object> searchAttributes);
 
-  Object newThread(Runnable runnable, boolean detached, String name);
+  /**
+   * Intercepts creation of the workflow child thread
```
mfateev, line 482: I fear that can get confused with child workflows. Maybe add to the comment that it is not related to child workflows in any way.

temporal-sdk/src/main/java/io/temporal/internal/sync/DeterministicRunnerImpl.java
```diff
@@ -360,10 +313,10 @@ public void close() {
       return;
     }
     try {
-      for (WorkflowThread c : threadsToAdd) {
+      for (WorkflowThread c : workflowThreadsToAdd) {
         threads.add(c);
       }
-      threadsToAdd.clear();
+      workflowThreadsToAdd.clear();
```
mfateev, line 319: Should we do the same for the callback threads to add?

temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflowContext.java
```diff
@@ -139,15 +141,29 @@ public DeterministicRunner getRunner() {
...
-      this.headInterceptor = head;
+  public WorkflowInboundCallsInterceptor getWorkflowInboundInterceptor() {
+    // This is needed for unit tests that create DeterministicRunner directly.
+    return headInboundInterceptor == null
+        ? new BaseRootWorkflowInboundCallsInterceptor(this)
+        : headInboundInterceptor;
+  }
+
+  public void initHeadOutboundCallsInterceptor(WorkflowOutboundCallsInterceptor head) {
+    if (headOutboundInterceptor == null) {
```
mfateev, line 157: Why this check?

temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflowContext.java
```diff
@@ -139,15 +141,29 @@ public DeterministicRunner getRunner() {
...
+      headOutboundInterceptor = head;
+    }
+  }
+
+  public void initHeadInboundCallsInterceptor(WorkflowInboundCallsInterceptor head) {
+    if (headInboundInterceptor == null) {
```
mfateev, line 163: Why null check?

temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkflowInboundCallsInterceptor.java
```diff
@@ -128,4 +130,22 @@ public Object getResult() {
...
+   */
+  Object newMethodThread(Runnable runnable, @Nullable String name);
```
mfateev, line 141: Rename to newWorkflowMethodThread or newMainWorkflowThread.

temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowThreadImpl.java
```diff
@@ -45,7 +45,7 @@
 import org.slf4j.LoggerFactory;
 import org.slf4j.MDC;
 
-class WorkflowThreadImpl implements WorkflowThread {
+public class WorkflowThreadImpl implements WorkflowThread {
```
mfateev, line 48: Why public?

temporal-sdk/src/test/java/io/temporal/internal/sync/PromiseTest.java
```diff
@@ -55,11 +55,10 @@ public void testFailure() throws Throwable {
             () -> {
               CompletablePromise<Boolean> f = Workflow.newPromise();
               trace.add("root begin");
-              WorkflowInternal.newThread(
-                      false, () -> f.completeExceptionally(new IllegalArgumentException("foo")))
+              WorkflowThread.newThread(
```
mfateev, line 58: The idea of WorkflowInternal is to proxy all the public APIs through it and keep all the implementation details private. Making WorkflowThread and its impl public goes against this idea.

temporal-sdk/src/main/java/io/temporal/internal/sync/DeterministicRunnerImpl.java
Spikhalskiy: It's actually what I did in this PR, I made a collection `callbackThreadsToAdd` that works almost the same way (just traversed in a reversed order when we add)

temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflowContext.java
Spikhalskiy: I actually saved the behavior that already was here. It was a null check inside a setter method which is even worse. Probably it was done to avoid double initialization with different implementation, but I'm not sure. But I agree it looks off. Let me try to rework it to throwing an exception if the interceptor is already initialized and see if anything blows up.

temporal-sdk/src/test/java/io/temporal/internal/sync/PromiseTest.java
Spikhalskiy: Yeah, but it's actually not what's going on here.
We had some leftover stuff when internal WorkflowInternal had methods that were just calling WorkflowThread methods.
So, we had an "internal" API that didn't do anything else other than calling a "public" API. So, I just removed these methods and replaced them with using an existing public API.

temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflowContext.java
Spikhalskiy: changed to a throwing an exception. see the explanation behind.

temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflowContext.java
Spikhalskiy: changed to a throwing an exception. see the explanation behind.

temporal-sdk/src/main/java/io/temporal/internal/sync/BaseRootWorkflowInboundCallsInterceptor.java
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  @Override
+  public void init(WorkflowOutboundCallsInterceptor outboundCalls) {
+    workflowContext.initHeadOutboundCallsInterceptor(outboundCalls);
+  }
+
+  @Override
+  public WorkflowOutput execute(WorkflowInput input) {
+    throw new UnsupportedOperationException();
```
vitarb, line 39: nit. Pass a human readable message as a parameter. Something like "Root workflow should never be executed directly".

temporal-sdk/src/main/java/io/temporal/internal/sync/BaseRootWorkflowInboundCallsInterceptor.java
```diff
@@ -0,0 +1,61 @@
...
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
+package io.temporal.internal.sync;
+
+import io.temporal.common.interceptors.WorkflowInboundCallsInterceptor;
+import io.temporal.common.interceptors.WorkflowOutboundCallsInterceptor;
+
+public class BaseRootWorkflowInboundCallsInterceptor implements WorkflowInboundCallsInterceptor {
```
vitarb, line 25: I think it would be good to explain in the comment what the root workflow is and why/when this interceptor could be useful.

temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowThreadImpl.java
```diff
@@ -394,8 +394,15 @@ public void addStackTrace(StringBuilder result) {
...
     int omitTop = 5;
     int omitBottom = 7;
+    // TODO it's not a good idea to rely on the name to understand the thread type. Instead of that
+    // we would better
+    // assign an explicit enum to the threads. This will be especially important when we refactor
+    // root and workflow-method
+    // thread names into names that will include workflowId
     if (DeterministicRunnerImpl.WORKFLOW_ROOT_THREAD_NAME.equals(getName())) {
       omitBottom = 11;
+    } else if (DeterministicRunnerImpl.WORKFLOW_MAIN_THREAD_NAME.equals(getName())) {
```
vitarb, line 404: nit. You could use `||` instead of `if else if` since then part is the same.

PR #563, opened 2021-06-29
Allow zero value on some worker options values

jonsie:
- maxWorkerActivitiesPerSecond
- maxConcurrentActivityExecutionSize
- maxConcurrentWorkflowTaskExecutionSize
- maxConcurrentLocalActivityExecutionSize

Zero is a valid default value for these configs, and not allowing zero to be passed here makes WorkerOptions both less extensibile and limits our ability to use dynamic configuration when tweaking these values

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/worker/WorkerOptions.java
+++ b/temporal-sdk/src/main/java/io/temporal/worker/WorkerOptions.java
@@ -90,22 +90,21 @@ public final class WorkerOptions {
      * workers.
      */
     public Builder setMaxWorkerActivitiesPerSecond(double maxWorkerActivitiesPerSecond) {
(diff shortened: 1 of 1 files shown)


Spikhalskiy:
Could you please provide more context about why you may want to setup any of this options to 0 for a worker?

jonsie:
@Spikhalskiy These options support 0 as a default, just not when passed in via these setters.  I included in my commit this: 

"Zero is a valid default value for these configs, and not allowing zero to be passed here makes WorkerOptions both less extensibile and limits our ability to use dynamic configuration when tweaking these values"

Does that make sense to you?  Basically, if you're extending this SDK (to produce an in-house SDK, for example) and you want to have your own mechanism for building these options you should be able to pass in 0 and use the default configurations.

Spikhalskiy (approved):

PR #565, opened 2021-07-01
Cleanup javadoc as a follow-up to PR #563

Spikhalskiy:
## What was changed
This PR improves javadocs as a follow up to PR #563 that changes acceptable values of `WorkerOptions.Builder`

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/PotentialDeadlockException.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/PotentialDeadlockException.java
@@ -19,7 +19,13 @@
 
 package io.temporal.internal.sync;
 
-class PotentialDeadlockException extends RuntimeException {
+/**
+ * A workflow tasks are allowed to execute only some limited amount of time without interruption. If
+ * workflow task runs longer than specified interval without yielding (like calling an Activity), it
+ * will fail automatically with this exceptions. This is done to detect deadlocks in workflows and
+ * based on the assumptions that workflow code shouldn't be blocking.
+ */
+public class PotentialDeadlockException extends RuntimeException {
 
   private final WorkflowThreadContext workflowThreadContext;
   private String stackDump;
--- a/temporal-sdk/src/main/java/io/temporal/worker/WorkerOptions.java
+++ b/temporal-sdk/src/main/java/io/temporal/worker/WorkerOptions.java
@@ -81,62 +81,68 @@ public final class WorkerOptions {
     }
 
     /**
-     * Maximum number of activities started per second by this worker. Default is 0 which means
-     * unlimited. If worker is not fully loaded while tasks are backing up on the service consider
-     * increasing {@link #setActivityPollThreadCount(int)}.
-     *
-     * <p>Note that this is a per worker limit. Use {@link
-     * #setMaxTaskQueueActivitiesPerSecond(double)} to set per task queue limit across multiple
-     * workers.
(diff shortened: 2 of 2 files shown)


temporal-sdk/src/main/java/io/temporal/worker/WorkerOptions.java
```diff
@@ -98,9 +100,9 @@ public Builder setMaxWorkerActivitiesPerSecond(double maxWorkerActivitiesPerSeco
     }
 
     /**
-     * Maximum number of activities executed in parallel.
+     * @param maxConcurrentActivityExecutionSize Maximum number of activities executed in parallel. Default is 200. Default if 0.
```
Sushisource, line 103:
```suggestion
     * @param maxConcurrentActivityExecutionSize Maximum number of activities executed in parallel. Default is 200, which is chosen if set to zero.
```

Sushisource: Same in other places with this language

temporal-sdk/src/main/java/io/temporal/worker/WorkerOptions.java
```diff
@@ -190,8 +193,17 @@ public Builder setLocalActivityWorkerOnly(boolean localActivityWorkerOnly) {
       return this;
     }
 
-    public Builder setDefaultDeadlockDetectionTimeout(long defaultDeadlockDetectionTimeout) {
-      this.defaultDeadlockDetectionTimeout = defaultDeadlockDetectionTimeout;
+    /**
+     * @param defaultDeadlockDetectionTimeoutMs time period in ms that will be used the detect workflows deadlock. Default is 1000ms. Default if 0.
```
Sushisource, line 197:
```suggestion
     * @param defaultDeadlockDetectionTimeoutMs time period in ms that will be used to detect a workflow deadlock. Default is 1000ms, which is chosen if set to 0.
```

Sushisource (approved):
Some wording fixes

temporal-sdk/src/main/java/io/temporal/worker/WorkerOptions.java
vitarb, line 197: I would expand it a bit, for example:
"amount of time in milliseconds that workflow tasks are allowed to execute, if workflow task runs longer than specified interval it will fail automatically" 
Also "Default is 1000ms. Default if 0." sounds a little confusing. I don't think you need to mention "default if 0" part at all, but if you prefer to, I would just write something like "default is 1 sec, if set to 0 then default value is used."

temporal-sdk/src/main/java/io/temporal/internal/sync/PotentialDeadlockException.java
```diff
@@ -19,7 +19,7 @@
 
 package io.temporal.internal.sync;
 
-class PotentialDeadlockException extends RuntimeException {
+public class PotentialDeadlockException extends RuntimeException {
```
vitarb, line 22: Should we add a javadoc for the class since it's public now?

vitarb (approved):

temporal-sdk/src/main/java/io/temporal/internal/sync/PotentialDeadlockException.java
Spikhalskiy: Changing it to public is pretty technical, it still stays in internal package, but I will write a javadoc here too, because this exception is something that users can encounter.

PR #569, opened 2021-07-03
WorkflowQueue.take takes element from the tail of the queue instead of its head

GreyTeardrop:
## What was changed

`WorkflowQueue.take` does not preserve queue semantics by taking elements from the tail of the queue (the most recently added ones) instead of its head (the earliest added ones).

## Why?

Let `WorkflowQueue` maintain the expected queue semantics.

## Checklist

1. Closes -
2. How was this tested: new unit test
3. Any docs updates needed? -

### DIFF
--- a/temporal-sdk/src/test/java/io/temporal/internal/sync/WorkflowInternalQueueTest.java
+++ b/temporal-sdk/src/test/java/io/temporal/internal/sync/WorkflowInternalQueueTest.java
@@ -389,23 +389,25 @@ public class WorkflowInternalQueueTest {
 
   @Test
   public void testQueueOrder() throws Throwable {
+    WorkflowQueue<Integer> queue = WorkflowInternal.newQueue(3);
+    int[] result = new int[3];
     DeterministicRunner r =
         DeterministicRunner.newRunner(
             () -> {
-              WorkflowQueue<Integer> queue = WorkflowInternal.newQueue(3);
               queue.put(1);
               queue.put(2);
               queue.put(3);
-              trace.add(queue.take().toString());
-              trace.add(queue.poll().toString());
-              trace.add(queue.poll().toString());
+              result[0] = queue.take();
+              result[1] = queue.poll();
+              result[2] = queue.poll();
             });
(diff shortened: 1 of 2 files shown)


temporal-sdk/src/test/java/io/temporal/internal/sync/WorkflowInternalQueueTest.java
```diff
@@ -386,4 +386,26 @@ public void testMap() throws Throwable {
...
+  @Test
+  public void testQueueOrder() throws Throwable {
+    DeterministicRunner r =
+        DeterministicRunner.newRunner(
+            () -> {
+              WorkflowQueue<Integer> queue = WorkflowInternal.newQueue(3);
+              queue.put(1);
+              queue.put(2);
+              queue.put(3);
+              trace.add(queue.take().toString());
```
Spikhalskiy, line 399: I believe we use Tracer mostly for testing of threads' order and behavior. We can just use a regular collection here to persist the elements and compare the collections at the end of the test for simplicity and to don't depend on Tracer where we don't really need it. 
Not a big issue though, I will leave the PR open for some time if you want to refactor the test just to use a regular collection.

Spikhalskiy (approved):

temporal-sdk/src/test/java/io/temporal/internal/sync/WorkflowInternalQueueTest.java
GreyTeardrop: Totally makes sense, will update it now.

GreyTeardrop: Updated.

One thing I've realized - I worry if this fix might break something? E.g. can it happen that upgrading Temporal SDK would introduce a non-deterministic update to the workflow?

Spikhalskiy:
Thank you for your contribution!

Spikhalskiy:
After more thinking, this probably has to be reversed and tabled till the time when we make backward-incompatible changes. This fix can break workflows replaying that was originally executed on an old version of code.

PR #573, opened 2021-07-06
Add getTestWatcher method to TestWorkflowRule

tsurdilo:

## What was changed
Addition of getTestWatcher method to TestWorkflowRule 
## Why?
Currently if you use testworkflowrule you have to create test watcher rule as such for example:

```java
  @Rule
  public TestWorkflowRule testWorkflowRule = ...

  @Rule
  public TestWatcher watchman =
      new TestWatcher() {
        @Override
        protected void failed(Throwable e, Description description) {
          if (testWorkflowRule.getTestEnvironment() != null) {
            System.err.println(testWorkflowRule.getTestEnvironment().getDiagnostics());
            testWorkflowRule.getTestEnvironment().shutdown();
          }
        }
      };
```

This simplifies it to:

```
  @Rule
  public TestWorkflowRule testWorkflowRule = ...

  @Rule
  public TestWatcher watchman = testWorkflowRule.getTestWatcher();
```         

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/temporal-testing-junit4/src/main/java/io/temporal/testing/TestWorkflowRule.java
+++ b/temporal-testing-junit4/src/main/java/io/temporal/testing/TestWorkflowRule.java
@@ -354,4 +354,27 @@ public class TestWorkflowRule implements TestRule {
   public Worker getWorker() {
     return testEnvironment.getWorkerFactory().getWorker(getTaskQueue());
   }
+
+  /**
+   * Returns the default test watcher impl that can be used is your tests. It simplifies the test watcher
+   * creation to just:
+   * <pre><code>
+   *   @Rule
+   *   public TestWatcher watchman = testWorkflowRule.getTestWatcher();
+   *   </code></pre>
+   *
+   * @return default test watcher impl
+   */
+  public TestWatcher getTestWatcher() {
+    return
+      new TestWatcher() {
+        @Override
+        protected void failed(Throwable e, Description description) {
+          if (getTestEnvironment() != null) {
+            System.err.println(getTestEnvironment().getDiagnostics());
+            getTestEnvironment().shutdown();
+          }
+        }
+      };
+  }
 }


temporal-testing-junit4/src/main/java/io/temporal/testing/TestWorkflowRule.java
```diff
@@ -354,4 +354,27 @@ public History getWorkflowExecutionHistory(WorkflowExecution execution) {
...
+   * Returns the default test watcher impl that can be used is your tests. It simplifies the test watcher
+   * creation to just:
+   * <pre><code>
+   *   @Rule
+   *   public TestWatcher watchman = testWorkflowRule.getTestWatcher();
+   *   </code></pre>
+   *
+   * @return default test watcher impl
+   */
+  public TestWatcher getTestWatcher() {
```
Spikhalskiy, line 368: This method should be called `newTestWatcher` or `supplyTestWatcher`, it's not a getter.

temporal-testing-junit4/src/main/java/io/temporal/testing/TestWorkflowRule.java
```diff
@@ -354,4 +354,27 @@ public History getWorkflowExecutionHistory(WorkflowExecution execution) {
...
+    return
+      new TestWatcher() {
+        @Override
+        protected void failed(Throwable e, Description description) {
+          if (getTestEnvironment() != null) {
+            System.err.println(getTestEnvironment().getDiagnostics());
+            getTestEnvironment().shutdown();
```
Spikhalskiy, line 375: Why do we need this? This code is located inside TestWorkflowRule class that takes care of shutting down the environment here, in shutdown call:
```
  @Override
  public Statement apply(Statement base, Description description) {
    taskQueue = init(description);
    Statement testWorkflowStatement =
        new Statement() {
          @Override
          public void evaluate() throws Throwable {
            start();
            base.evaluate();
            shutdown();
          }
        };
    return watchman.apply(globalTimeout.apply(testWorkflowStatement, description), description);
  }
```
So, the environment shutdown is taken care of already by the rule, why do we need also the watcher to handle it?

temporal-testing-junit4/src/main/java/io/temporal/testing/TestWorkflowRule.java
```diff
@@ -354,4 +354,27 @@ public History getWorkflowExecutionHistory(WorkflowExecution execution) {
...
+   *
+   * @return default test watcher impl
+   */
+  public TestWatcher getTestWatcher() {
+    return
+      new TestWatcher() {
+        @Override
+        protected void failed(Throwable e, Description description) {
+          if (getTestEnvironment() != null) {
+            System.err.println(getTestEnvironment().getDiagnostics());
```
Spikhalskiy, line 374: Please use logger, there are almost 0 valid cases when non-trivial app should be performing input and output directly to System.err or System.out.

tsurdilo: This is ran in tests and from what i see logger statements are swallowed. There is already a System.err in this class on like 87 ;)

temporal-testing-junit4/src/main/java/io/temporal/testing/TestWorkflowRule.java
tsurdilo: This is a separate rule so from what i see needs to be there. On exception the test rule might not shut down.

temporal-testing-junit4/src/main/java/io/temporal/testing/TestWorkflowRule.java
tsurdilo: ok i dont see the big deal but sure. I like that method name.

tsurdilo:
TestWorkflowRule has internal TestWatcher and I think this change is not needed after all :) Will close thanks for review

PR #574, opened 2021-07-06
Fixed broken link in docs + formatting.

vkoby:
## What was changed

- fixed broken link in ActivityExecutionContext
- fixed formatting in ActivityInfo as per https://github.com/temporalio/idea-settings

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/activity/ActivityInfo.java
+++ b/temporal-sdk/src/main/java/io/temporal/activity/ActivityInfo.java
@@ -30,8 +30,8 @@ import java.util.Optional;
 public interface ActivityInfo {
 
   /**
-   * Returns a correlation token that can be used to complete the Activity Execution asynchronously through {@link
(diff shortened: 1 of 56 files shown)


temporal-sdk/src/main/java/io/temporal/activity/ActivityInfo.java
```diff
@@ -57,46 +58,32 @@
...
-   */
+  /** Returns the Heartbeat Timeout setting as a Duration. */
   Duration getHeartbeatTimeout();
 
   Optional<Payloads> getHeartbeatDetails();
 
-  /**
-   * Returns the Workflow Type of the Workflow Execution that executed the Activity.
-   */
+  /** Returns the Workflow Type of the Workflow Execution that executed the Activity. */
```
Spikhalskiy, line 72: If you revisit these docs, let's use Javadoc @return tag instead of plain "Returns"

Spikhalskiy (approved):

PR #581, opened 2021-07-08
Polish JUnit5 extension: per-test initial time, dynamic workflows and activities

GreyTeardrop:
## What was changed

* The `TestWorkflowExtension` is changed to support per-test overrides for the virtual time used by the `TestWorkflowEnvironment`.
* Somewhat unrelated, the extension code is polished a bit: the whole `WorkflowOptions` is stored in the extension context instead of just the task queue name, to reduce some code duplication.
* Extend `TestWorkflowExtension` and `TestActivityExtension` to support dynamic workflow/activity scenarios.

## Why?

One of our workflows has time-sensitive logic and we'd like to test it in multiple scenarios that differ in the initial workflow time. Right now we either have to fall back to manually managing `TestWorkflowEnvironment` or to split those scenarios into multiple test classes.

## Checklist

1. **Closes** —
2. **How was this tested:** extended existing unit test
3. **Any docs updates needed?** Longer-term, both JUnit 4 and JUnit 5 extensions could be documented on the [Java SDK testing and debugging](https://docs.temporal.io/docs/java/testing-and-debugging) page.

### DIFF
--- a/temporal-testing-junit5/src/main/java/io/temporal/testing/TestActivityExtension.java
+++ b/temporal-testing-junit5/src/main/java/io/temporal/testing/TestActivityExtension.java
@@ -19,6 +19,7 @@
 
 package io.temporal.testing;
 
+import io.temporal.activity.DynamicActivity;
 import io.temporal.common.metadata.POJOActivityImplMetadata;
 import io.temporal.common.metadata.POJOActivityInterfaceMetadata;
 import java.lang.reflect.Constructor;
@@ -67,6 +68,8 @@ public class TestActivityExtension
 
   private final Set<Class<?>> supportedParameterTypes = new HashSet<>();
 
+  private boolean includesDynamicActivity;
+
   private TestActivityExtension(Builder builder) {
     testEnvironmentOptions = builder.testEnvironmentOptions;
     activityImplementations = builder.activityImplementations;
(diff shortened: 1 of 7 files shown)


Spikhalskiy (approved):

temporal-testing-junit5/src/main/java/io/temporal/testing/TestActivityExtension.java
```diff
@@ -95,6 +102,10 @@ public boolean supportsParameter(
       return false;
     }
 
+    if (includesDynamicActivity) {
+      return true;
```
Spikhalskiy, line 106: Can we add a short javadoc here explaining the logic behind this branch?

temporal-testing-junit5/src/main/java/io/temporal/testing/TestWorkflowExtension.java
```diff
@@ -134,6 +141,10 @@ public boolean supportsParameter(
       return false;
     }
 
+    if (includesDynamicWorkflow) {
+      return true;
```
Spikhalskiy, line 145: Can we add a short javadoc here explaining the logic behind this branch?

temporal-testing-junit5/src/main/java/io/temporal/testing/TestActivityExtension.java
GreyTeardrop: I was going to write that "No way to check what can be injected if a dynamic activity is registered" but I've realized we can still check that the injected type is a proper activity interface, so I've rewritten the code for that.

temporal-testing-junit5/src/main/java/io/temporal/testing/TestWorkflowExtension.java
GreyTeardrop: Same here - rewritten the code to check that only proper workflow interfaces are injectable.

Spikhalskiy:
Thank you for the contribution!

PR #592, opened 2021-07-17
Change docker hub repo to temporaliotest

alexshtin:
After https://github.com/temporalio/temporal/pull/1740 `temporalio/auto-setup:latest` points to latest release not latest `master` commit.

### DIFF
--- a/docker/buildkite/docker-compose.yaml
+++ b/docker/buildkite/docker-compose.yaml
@@ -19,7 +19,7 @@ services:
       - "8126:8126"
 
   temporal:
-    image: temporalio/auto-setup:latest
+    image: temporaliotest/auto-setup:latest
     logging:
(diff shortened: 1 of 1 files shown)


Spikhalskiy:
Wouldn't it be better to just use the same repo and same image, but a different tag? For example, temporalio/auto-setup:canary or temporalio/auto-setup:master?

alexshtin:
`temporaliotest` account is already there and it contains docker images for every github commit (docker tag same as github sha) and `:latest` points to the latest github commit. We use this docker account for all internal test. `temporalio` is for official releases only. There are images for every tagged release and `:latest` there should point to the latest release. It was misused before just for legacy reason and this PR is to fix it.

vitarb (approved):

PR #610, opened 2021-07-28
Flaky test: testChildWorkflowWithCronSchedule

vkoby:
## What was changed
Added an array that records lastCompletionResult for every run.

## Why?
Couldn't guarantee how many runs the cron workflow would perform, hence relying on timers to get the lastCompletionResult of the correct run was not working consistently

### DIFF
--- a/temporal-sdk/src/test/java/io/temporal/workflow/childWorkflowTests/ChildWorkflowWithCronScheduleTest.java
+++ b/temporal-sdk/src/test/java/io/temporal/workflow/childWorkflowTests/ChildWorkflowWithCronScheduleTest.java
@@ -32,6 +32,7 @@ import io.temporal.workflow.shared.TestWorkflowWithCronScheduleImpl;
 import io.temporal.workflow.shared.TestWorkflows.TestWorkflow1;
 import io.temporal.workflow.shared.TestWorkflows.TestWorkflowWithCronSchedule;
 import java.time.Duration;
+import java.util.HashMap;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TestName;
@@ -57,7 +58,7 @@ public class ChildWorkflowWithCronScheduleTest {
             .getWorkflowClient()
             .newUntypedWorkflowStub(
                 "TestWorkflow1", newWorkflowOptionsWithTimeouts(testWorkflowRule.getTaskQueue()));
-    client.start(testName.getMethodName());
+    client.start(testName.toString());
     testWorkflowRule.getTestEnvironment().sleep(Duration.ofHours(3));
     client.cancel();
 
@@ -69,7 +70,9 @@ public class ChildWorkflowWithCronScheduleTest {
     }
 
     // Run 3 failed. So on run 4 we get the last completion result from run 2.
-    assertEquals("run 2", TestWorkflowWithCronScheduleImpl.lastCompletionResults[4]);
+    HashMap<Integer, String> lastCompletionResults =
+        TestWorkflowWithCronScheduleImpl.lastCompletionResults.get(testName.toString());
+    assertEquals("run 2", lastCompletionResults.get(4));
   }
 
   public static class TestCronParentWorkflow implements TestWorkflow1 {
@@ -77,8 +80,8 @@ public class ChildWorkflowWithCronScheduleTest {
         Workflow.newChildWorkflowStub(TestWorkflowWithCronSchedule.class);
 
     @Override
-    public String execute(String taskQueue) {
-      return cronChild.execute(taskQueue);
+    public String execute(String testName) {
+      return cronChild.execute(testName);
     }
   }
 }
--- a/temporal-sdk/src/test/java/io/temporal/workflow/childWorkflowTests/ChildWorkflowWithCronScheduleTest.java
+++ b/temporal-sdk/src/test/java/io/temporal/workflow/childWorkflowTests/ChildWorkflowWithCronScheduleTest.java
@@ -56,11 +56,7 @@ public class ChildWorkflowWithCronScheduleTest {
         testWorkflowRule
             .getWorkflowClient()
             .newUntypedWorkflowStub(
-                "TestWorkflow1",
-                newWorkflowOptionsWithTimeouts(testWorkflowRule.getTaskQueue())
-                    .toBuilder()
-                    .setWorkflowRunTimeout(Duration.ofHours(10))
-                    .build());
+                "TestWorkflow1", newWorkflowOptionsWithTimeouts(testWorkflowRule.getTaskQueue()));
     client.start(testName.getMethodName());
     testWorkflowRule.getTestEnvironment().sleep(Duration.ofHours(3));
     client.cancel();
@@ -73,7 +69,7 @@ public class ChildWorkflowWithCronScheduleTest {
     }
 
     // Run 3 failed. So on run 4 we get the last completion result from run 2.
-    assertEquals("run 2", TestWorkflowWithCronScheduleImpl.lastCompletionResult);
+    assertEquals("run 2", TestWorkflowWithCronScheduleImpl.lastCompletionResults[4]);
   }
 
   public static class TestCronParentWorkflow implements TestWorkflow1 {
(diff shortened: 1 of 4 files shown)


temporal-sdk/src/test/java/io/temporal/workflow/childWorkflowTests/ChildWorkflowWithCronScheduleTest.java
```diff
@@ -56,11 +56,7 @@ public void testChildWorkflowWithCronSchedule() {
         testWorkflowRule
             .getWorkflowClient()
             .newUntypedWorkflowStub(
-                "TestWorkflow1",
-                newWorkflowOptionsWithTimeouts(testWorkflowRule.getTaskQueue())
-                    .toBuilder()
-                    .setWorkflowRunTimeout(Duration.ofHours(10))
```
vkoby, line 62: Removed because it's already set in the newWorkflowOptionsWithTimeouts and this timeout doesn't matter anyway

temporal-sdk/src/test/java/io/temporal/workflow/shared/TestWorkflowWithCronScheduleImpl.java
```diff
@@ -34,6 +34,7 @@
 public class TestWorkflowWithCronScheduleImpl implements TestWorkflowWithCronSchedule {
 
   public static final Map<String, AtomicInteger> retryCount = new ConcurrentHashMap<>();
+  public static String[] lastCompletionResults = new String[10];
```
vkoby, line 37: Usually there's no more than 5 runs and we start at index 1 instead of 0. 10 to make sure there's enough space

temporal-sdk/src/test/java/io/temporal/workflow/shared/TestWorkflowWithCronScheduleImpl.java
```diff
@@ -34,6 +36,8 @@
 public class TestWorkflowWithCronScheduleImpl implements TestWorkflowWithCronSchedule {
 
   public static final Map<String, AtomicInteger> retryCount = new ConcurrentHashMap<>();
+  public static final Hashtable<String, HashMap<Integer, String>> lastCompletionResults =
+      new Hashtable<>();
   public static String lastCompletionResult;
```
Spikhalskiy, line 41: Please refactor into Map<testName, Result> too

temporal-sdk/src/test/java/io/temporal/workflow/shared/TestWorkflowWithCronScheduleImpl.java
```diff
@@ -34,6 +36,8 @@
 public class TestWorkflowWithCronScheduleImpl implements TestWorkflowWithCronSchedule {
 
   public static final Map<String, AtomicInteger> retryCount = new ConcurrentHashMap<>();
+  public static final Hashtable<String, HashMap<Integer, String>> lastCompletionResults =
```
Spikhalskiy, line 39: Please don't use Hashtables, it's effectively deprecated. See ConcurrentHashMap above.
Also, please convert it into Map<String, Map<Integer, String>>. I'm not sure why the second generic is HashMap here. It's always better to use the most generic type.

temporal-sdk/src/test/java/io/temporal/workflow/childWorkflowTests/ChildWorkflowWithCronScheduleTest.java
```diff
@@ -73,16 +70,18 @@ public void testChildWorkflowWithCronSchedule() {
     }
 
     // Run 3 failed. So on run 4 we get the last completion result from run 2.
-    assertEquals("run 2", TestWorkflowWithCronScheduleImpl.lastCompletionResult);
+    HashMap<Integer, String> lastCompletionResults =
```
Spikhalskiy, line 73: Why does this var have to be a HashMap if Map is more than enough?

temporal-sdk/src/test/java/io/temporal/workflow/shared/TestWorkflowWithCronScheduleImpl.java
```diff
@@ -53,8 +57,11 @@ public String execute(String testName) {
     if (count == null) {
       count = new AtomicInteger();
       retryCount.put(testName, count);
+      lastCompletionResults.put(testName, new HashMap<Integer, String>());
     }
+
     int c = count.incrementAndGet();
+    lastCompletionResults.get(testName).put(c, lastCompletionResult);
```
Spikhalskiy, line 64: Please remove the initialization above and change this `get` onto `computeIfAbsent`. Let's use what Java gives us to the whole extend.

temporal-sdk/src/test/java/io/temporal/workflow/shared/TestWorkflowWithCronScheduleImpl.java
```diff
@@ -34,7 +35,9 @@
 public class TestWorkflowWithCronScheduleImpl implements TestWorkflowWithCronSchedule {
 
   public static final Map<String, AtomicInteger> retryCount = new ConcurrentHashMap<>();
-  public static String lastCompletionResult;
+  public static final Map<String, Map<Integer, String>> lastCompletionResults =
+      new ConcurrentHashMap<>();
+  public static Map<String, String> lastCompletionResult = new ConcurrentHashMap();
```
Spikhalskiy, line 40: Why is this one not final?

temporal-sdk/src/test/java/io/temporal/workflow/shared/TestWorkflowWithCronScheduleImpl.java
```diff
@@ -46,15 +49,16 @@ public String execute(String testName) {
...
-    lastCompletionResult = Workflow.getLastCompletionResult(String.class);
     lastFail = Workflow.getPreviousRunFailure();
+    retryCount.computeIfAbsent(testName, k -> new AtomicInteger());
+    lastCompletionResults.computeIfAbsent(testName, k -> new HashMap<>());
 
-    AtomicInteger count = retryCount.get(testName);
-    if (count == null) {
-      count = new AtomicInteger();
-      retryCount.put(testName, count);
+    int c = retryCount.get(testName).incrementAndGet();
```
Spikhalskiy, line 56: computeIfAbsent() returns the value. Use it instead of the second get.

Also, please use meaningful names than `c`.

temporal-sdk/src/test/java/io/temporal/workflow/shared/TestWorkflowWithCronScheduleImpl.java
```diff
@@ -46,15 +49,16 @@ public String execute(String testName) {
...
+    String result = Workflow.getLastCompletionResult(String.class);
+    if (result != null) {
+      lastCompletionResult.put(testName, Workflow.getLastCompletionResult(String.class));
     }
-    int c = count.incrementAndGet();
+    lastCompletionResults.get(testName).put(c, lastCompletionResult.get(testName));
```
Spikhalskiy, line 61: Why do you need to make a `lastCompletionResult.get(testName)` here if you just published this value? Why not to use `result` directly?

temporal-sdk/src/test/java/io/temporal/workflow/shared/TestWorkflowWithCronScheduleImpl.java
```diff
@@ -46,23 +48,20 @@ public String execute(String testName) {
       return null;
     }
 
-    lastCompletionResult = Workflow.getLastCompletionResult(String.class);
     lastFail = Workflow.getPreviousRunFailure();
+    retryCount.computeIfAbsent(testName, k -> new AtomicInteger());
```
Spikhalskiy, line 52: It should be
```
int count = retryCount.computeIfAbsent(testName, AtomicInteger::new).incrementAndGet();
    lastCompletionResults.computeIfAbsent(testName, HashMap::new).put(count, Workflow.getLastCompletionResult(String.class));
```
instead. Don't make lookups for something `computeIfAbsent` is already returning back.

Spikhalskiy (approved):

PR #611, opened 2021-07-29
Added getMemo to WorkflowInfo

### LINKED ISSUE #512: Missing getMemo functionality in WorkflowInfo
## Expected Behavior
WorkflowInfo interface should have getMemo() to return memos

WorkflowOptions.Builder has method `.setMemo` that allows setting memos. There is no corresponding WorkflowInfo method to retrieve this information from running workflow. This makes memo pretty useless when using Temporal Java SDK.

### Version
1.0.7

Conversation in Slack:
https://temporalio.slack.com/archives/CTT84KXK9/p1622042440064800

vkoby:
## What was changed
Added a getter for Memo in WorkflowInfo

## Why?
Memo is part of WorkflowInfo and thus should be retrievable

## Checklist

1. Closes #512 

2. How was this tested:
Unit Test


Spikhalskiy (changes requested):
Memo is exposed to the user as `setMemo(Map<String, Object> memo)` (https://github.com/temporalio/sdk-java/blob/b642a750259e0187eddee6cc4c28599cb493ffb6/temporal-sdk/src/main/java/io/temporal/client/WorkflowOptions.java#L207)

So, `getMemo` should return the same deserialized structure, not a raw Protobuf entity.

temporal-sdk/src/main/java/io/temporal/internal/replay/WorkflowContext.java
```diff
@@ -149,6 +150,10 @@ String getCurrentRunId() {
     return startedAttributes.getHeader().getFieldsMap();
   }
 
+  Memo getMemo() {
```
vitarb, line 153: Let's return map here and avoid exposing protobufs.

temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowInternal.java
```diff
@@ -428,6 +429,11 @@ public static WorkflowInfo getWorkflowInfo() {
     return new WorkflowInfoImpl(getRootWorkflowContext().getContext());
   }
 
+  public static <T> T getMemo(String key, Class<T> valueClass, Type valueType) {
+    Payload memo = getRootWorkflowContext().getContext().getMemo(key);
+    return DataConverter.getDefaultInstance().fromPayload(memo, valueClass, valueType);
```
Spikhalskiy, line 434: Per our design discussion, customer-defined per client DataConverter should be used (Defined in WorkflowClientOptions), not a global static `DataConverter.getDefaultInstance()`

temporal-sdk/src/main/java/io/temporal/workflow/Workflow.java
```diff
@@ -615,6 +615,14 @@ public static WorkflowInfo getInfo() {
     return WorkflowInternal.getWorkflowInfo();
   }
 
+  public static <T> Object getMemo(String key, Class<T> valueClass) {
+    return getMemo(key, valueClass, valueClass.getGenericSuperclass());
```
Spikhalskiy, line 619: valueClass.getGenericSuperclass() doesn't make too much sense. It works because it effectively is not used if valueClass doesn't have generics. Just pass valueClass here for simplicity.

temporal-sdk/src/main/java/io/temporal/workflow/Workflow.java
```diff
@@ -615,6 +615,14 @@ public static WorkflowInfo getInfo() {
...
+  }
+
+  public static <T> T getMemo(String key, Class<T> valueClass, Type valueType) {
```
Spikhalskiy, line 622: 1. Rename valueType to genericType, it's confusing otherwise.
2. Please provide javadocs for both methods. It's also a good idea to link a doc to DataConverter#fromPayloads to explain why there are two parameters here. DataConverter#fromPayloads is well documented by itself.

PR #583, opened 2021-07-10
A test to reproduce timers and cancelations SDK issue

GreyTeardrop:
## What was changed

This PR is not intended to be merged as it. It provides a test to reproduce some weird corner cases of scheduling/canceling timers and evaluating versions in the scope of a single workflow task.

The test contains a seemingly valid workflow that should successfully run with no issues. It seems, however, that one of the workflow tasks fails with the `Version is already set to 1` error, and on the next retry a signal is delivered to the workflow for the second time.

Here's how workflow's logs look like without version evaluation

```
20:32:09.234 [workflow-method] INFO  i.t.w.v.GetVersionAndCancelTimerTest$ReminderWorkflowImpl - <2020-01-01T00:00:00.555Z> Workflow started
20:32:09.277 [signal scheduleReminder] INFO  i.t.w.v.GetVersionAndCancelTimerTest$ReminderWorkflowImpl - <2020-01-01T00:00:00.677Z> Got signal ScheduleReminderSignal[reminderTime=2020-01-01T00:00:30.554Z, reminderText='Reminder 1']
20:32:09.278 [signal scheduleReminder] INFO  i.t.w.v.GetVersionAndCancelTimerTest$ReminderWorkflowImpl - <2020-01-01T00:00:00.677Z> Enqueued signal ScheduleReminderSignal[reminderTime=2020-01-01T00:00:30.554Z, reminderText='Reminder 1']
20:32:09.278 [signal scheduleReminder] INFO  i.t.w.v.GetVersionAndCancelTimerTest$ReminderWorkflowImpl - <2020-01-01T00:00:00.677Z> Got signal ScheduleReminderSignal[reminderTime=2020-01-01T00:00:10.554Z, reminderText='Reminder 2']
20:32:09.279 [signal scheduleReminder] INFO  i.t.w.v.GetVersionAndCancelTimerTest$ReminderWorkflowImpl - <2020-01-01T00:00:00.677Z> Enqueued signal ScheduleReminderSignal[reminderTime=2020-01-01T00:00:10.554Z, reminderText='Reminder 2']
20:32:09.279 [workflow-method] INFO  i.t.w.v.GetVersionAndCancelTimerTest$ReminderWorkflowImpl - <2020-01-01T00:00:00.677Z> Will process signal ScheduleReminderSignal[reminderTime=2020-01-01T00:00:30.554Z, reminderText='Reminder 1']
[...]

PR #614, opened 2021-08-02
Fix handling of Version events in WorkflowStateMachines in case of cancelled commands in a queue

Spikhalskiy:
## What was changed
`WorkflowStateMachines` now handles a situation with Versioning `EVENT_TYPE_MARKER_RECORDED` matching a commands buffer containing canceled commands correctly (by consuming and skipping them)

## Why?
The current implementation fails to match a Versioning `EVENT_TYPE_MARKER_RECORDED` History Event with a right command if there are canceled commands before it in the commands sink. This is a root cause of an issue described in  Issue #583.

1. Closes
PR #583
Issue #615

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/statemachines/VersionStateMachine.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/statemachines/VersionStateMachine.java
@@ -23,7 +23,6 @@ import static io.temporal.internal.sync.WorkflowInternal.DEFAULT_VERSION;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Strings;
-import io.temporal.api.command.v1.Command;
 import io.temporal.api.command.v1.RecordMarkerCommandAttributes;
 import io.temporal.api.common.v1.Payloads;
 import io.temporal.api.enums.v1.CommandType;
@@ -36,6 +35,7 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import javax.annotation.Nullable;
 
 final class VersionStateMachine {
 
@@ -49,7 +49,7 @@ final class VersionStateMachine {
   private final Functions.Proc1<CancellableCommand> commandSink;
   private final Functions.Proc1<StateMachine> stateMachineSink;
 
-  private Optional<Integer> version = Optional.empty();
+  @Nullable private Integer version;
 
   enum ExplicitEvent {
     CHECK_EXECUTION_STATE,
@@ -137,15 +137,14 @@ final class VersionStateMachine {
     }
 
     private void validateVersion() {
-      if (!version.isPresent()) {
+      if (version == null) {
         throw new IllegalStateException("Version not set");
       }
-      int v = version.get();
-      if ((v < minSupported || v > maxSupported) && v != DEFAULT_VERSION) {
+      if ((version < minSupported || version > maxSupported) && version != DEFAULT_VERSION) {
         throw new UnsupportedVersion(
             String.format(
                 "Version %d of changeId %s is not supported. Supported v is between %d and %d.",
-                v, changeId, minSupported, maxSupported));
+                version, changeId, minSupported, maxSupported));
       }
     }
 
@@ -177,8 +176,7 @@ final class VersionStateMachine {
         // still can be followed by an event with our changeId.
         return WorkflowStateMachines.HandleEventStatus.NON_MATCHING_EVENT;
       }
-      super.handleEvent(event, hasNextEvent);
-      return WorkflowStateMachines.HandleEventStatus.OK;
+      return super.handleEvent(event, hasNextEvent);
     }
 
     @Override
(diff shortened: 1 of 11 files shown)


temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
```diff
@@ -234,44 +231,61 @@ private void handleCommandEvent(HistoryEvent event) {
...
       // handleVersionMarker can skip a marker event if the getVersion call was removed.
       // In this case we don't want to consume a command.
       // That's why peek is used instead of poll.
-      command = commands.peek();
+      CancellableCommand command = commands.peek();
       if (command == null) {
         throw new IllegalStateException("No command scheduled that corresponds to " + event);
       }
+
+      if (command.isCanceled()) {
```
Spikhalskiy, line 244: This is the main fix

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
```diff
@@ -234,44 +231,61 @@ private void handleCommandEvent(HistoryEvent event) {
...
-              "Event "
-                  + event.getEventId()
-                  + " of "
-                  + event.getEventType()
-                  + " does not"
-                  + " match command "
-                  + command.getCommandType());
-        }
+
+      if (command.isCanceled()) {
```
Spikhalskiy, line 254: This is the main fix

Spikhalskiy:
I plan to add one more local Version Machine level unit test later to cover this fix

temporal-sdk/src/test/java/io/temporal/workflow/shared/TestActivities.java
```diff
@@ -206,7 +206,7 @@ public String sleepActivity(long milliseconds, int input) {
         Thread.sleep(milliseconds);
       } catch (InterruptedException e) {
         Thread.currentThread().interrupt();
-        throw Activity.wrap(new RuntimeException("interrupted", new Throwable("simulated")));
+        throw Activity.wrap(e);
```
Spikhalskiy, line 209: unrelated, fixed looking at the output of one of the flaky tests.

temporal-sdk/src/test/java/io/temporal/workflow/versionTests/GetVersionAfterScopeCancellationTest.java
```diff
@@ -0,0 +1,101 @@
...
+
+      Workflow.getVersion("some-change", Workflow.DEFAULT_VERSION, 1);
+
+      activeScope1.cancel();
+
+      // it's critical for this duration to be short for the test to fail originally (4s or less)
+      Duration secondScopeTimerDuration = Duration.ofSeconds(4);
+      CancellationScope activeScope2 =
+          Workflow.newCancellationScope(() -> Workflow.newTimer(secondScopeTimerDuration));
+      activeScope2.run();
```
mfateev, line 66: I don't understand why the second scope is needed. How is it different from just creating a timer without the enclosing scope?

temporal-sdk/src/test/java/io/temporal/workflow/versionTests/GetVersionAfterScopeCancellationTest.java
```diff
@@ -0,0 +1,101 @@
...
+
+  public static final class ReminderWorkflowImpl implements ReminderWorkflow {
+
+    @Override
+    public void start() {
+      Workflow.sleep(Duration.ofDays(1));
+    }
+
+    @Override
+    public void signal() {
```
mfateev, line 53: Why is this code in a signal handler? Can it reproduce the bug in the main workflow thread?

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
```diff
@@ -446,13 +456,13 @@ public long getLastStartedEventId() {
             attributes,
             (p, f) -> {
               callback.apply(p, f);
-              if (f != null && f.getCause() != null && f.getCause().hasCanceledFailureInfo()) {
+              if (f != null && f.getCause().hasCanceledFailureInfo()) {
```
mfateev, line 459: Are you 100% sure that it is safe to expect the cause for all possible failures?

temporal-sdk/src/test/java/io/temporal/internal/statemachines/AsyncWorkflowBuilder.java
```diff
@@ -22,6 +22,15 @@
 import io.temporal.workflow.Functions;
 import java.util.Queue;
 
+/**
+ * Represents a step in a workflow which takes {@code <T>} as an input value from the previous step
+ * and triggers registered callbacks with it's input.
```
mfateev, line 27: with its' output?

temporal-sdk/src/test/java/io/temporal/workflow/versionTests/GetVersionAfterScopeCancellationTest.java
Spikhalskiy: Yes, the second cancellation scope is not needed, just a timer to trigger the end of Workflow Task is enough.

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
Spikhalskiy: It's protobuf, so f.getCause() != null is always true, that check didn't make sense. But... I just added an additional `f.hasCause()` check.

temporal-sdk/src/main/java/io/temporal/internal/statemachines/VersionStateMachine.java
```diff
@@ -36,20 +36,27 @@
...
   static final String MARKER_VERSION_KEY = "version";
   static final String MARKER_CHANGE_ID_KEY = "changeId";
   static final String VERSION_MARKER_NAME = "Version";
 
+  private static final Command FAKE_COMMAND =
+      Command.newBuilder()
+          .setCommandType(CommandType.COMMAND_TYPE_RECORD_MARKER)
+          .setRecordMarkerCommandAttributes(RecordMarkerCommandAttributes.getDefaultInstance())
+          .build();
+
```
Sushisource, line 52: [nit] This nice bit of cleanup you did here and in the MSE state machine - it maybe would be nice to just have a common helper function that accepts `Optional<RecordMarkerCommandAttributes>` and constructs with default if empty, otherwise passed in args.

Would get rid of the command builder preamble in both places

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
```diff
@@ -135,7 +133,7 @@
   private final Map<String, MutableSideEffectStateMachine> mutableSideEffects = new HashMap<>();
 
   /** Key is changeId */
-  private final Map<String, VersionStateMachine> vesions = new HashMap<>();
+  private final Map<String, VersionStateMachine> versions = new HashMap<>();
```
Sushisource, line 136: Thank you lol :pray: this has bugged me reading the code but I hadn't fixed it yet

temporal-sdk/src/test/java/io/temporal/internal/statemachines/AsyncWorkflowBuilder.java
```diff
@@ -22,6 +22,15 @@
 import io.temporal.workflow.Functions;
 import java.util.Queue;
 
+/**
+ * Represents a step in a workflow which takes {@code <T>} as an input value from the previous step
+ * and triggers registered callbacks with it's output.
```
Sushisource, line 27:
```suggestion
 * and triggers registered callbacks with its output.
```

temporal-sdk/src/test/java/io/temporal/workflow/versionTests/GetVersionAfterScopeCancellationTest.java
```diff
@@ -0,0 +1,99 @@
...
+      CancellationScope activeScope1 =
+          Workflow.newCancellationScope(() -> Workflow.newTimer(Duration.ofHours(4)));
+      activeScope1.run();
+
+      Workflow.getVersion("some-change", Workflow.DEFAULT_VERSION, 1);
+
+      activeScope1.cancel();
+
+      // it's critical for this duration to be short for the test to fail originally (4s or less)
+      Duration secondScopeTimerDuration = Duration.ofSeconds(4);
```
Sushisource, line 63: Did you ever figure out exactly why this needed to be short? That's really my only concern here along w/ Max's comment about why the code needs to be in the signal handler.

Sushisource (approved):
This is really nice, love the added docs and the cleanup. The fix itself makes sense to me but I'd love to understand the bit about it needing to be in the signal / the short timer duration too.

mfateev (approved):

PR #616, opened 2021-08-03
Deprecate WorkflowInfo.getSearchAttributes, add new getter to Workflow

### LINKED ISSUE #617: Deserialize Search Attributes
## Expected Behavior
WorkflowInfo.getSearchAttributes return something that can be easily read/interpreted by the user

## Actual Behavior
Returns serialized protobuf

vkoby:
## What was changed
Implemented search attribute deserializer. Changed WorkflowInfo.getSearchAttributes to return Map<String, Object> instead of protobuf.

## Why?
More user friendly

## Checklist
1. Closes #617

2. How was this tested:
Unit tests


temporal-sdk/src/main/java/io/temporal/internal/common/SearchAttributesUtil.java
```diff
@@ -0,0 +1,152 @@
...
+    if (serializedMap == null) {
+      return null;
+    }
+
+    Map<String, Object> deserializedMap = new HashMap<>();
+    for (Entry<String, Payload> attribute : serializedMap.getIndexedFieldsMap().entrySet()) {
+      String key = attribute.getKey();
+      String value = converter.fromPayload(attribute.getValue(), String.class, String.class);
+      if (!parseSearchAttributes(key, value, deserializedMap)) {
+        String type = attribute.getValue().getMetadataMap().get("type").toStringUtf8();
```
Spikhalskiy, line 78: Use your constant from `EncodingKeys`

temporal-sdk/src/main/java/io/temporal/internal/common/SearchAttributesUtil.java
```diff
@@ -0,0 +1,152 @@
...
+
+  public static Map<String, Object> getDeserializedMap(SearchAttributes serializedMap) {
+    if (serializedMap == null) {
+      return null;
+    }
+
+    Map<String, Object> deserializedMap = new HashMap<>();
+    for (Entry<String, Payload> attribute : serializedMap.getIndexedFieldsMap().entrySet()) {
+      String key = attribute.getKey();
+      String value = converter.fromPayload(attribute.getValue(), String.class, String.class);
```
Spikhalskiy, line 76: Instead of deserializing everything to String and after that to a specific type again you should define the type and deserialize into that type directly.

temporal-sdk/src/main/java/io/temporal/common/converter/JacksonJsonPayloadConverter.java
```diff
@@ -82,4 +89,8 @@ public String getEncodingType() {
       throw new DataConverterException(e);
     }
   }
+
+  static String typeToString(Class c) {
```
Spikhalskiy, line 93: We should have a function mapping Java types to server types here

Spikhalskiy:
Please see https://github.com/Spikhalskiy/java-sdk/commit/5b94104f9b4eb69389621f68e6d64dbdc29cb0ce for comments and code structure proposals

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesPayloadConverter.java
```diff
@@ -0,0 +1,52 @@
...
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.google.protobuf.ByteString;
+import io.temporal.api.common.v1.Payload;
+import io.temporal.internal.common.SearchAttributesUtil;
+import java.util.Optional;
+
+public class SearchAttributesPayloadConverter extends JacksonJsonPayloadConverter {
+
+  public SearchAttributesPayloadConverter() {
+    super();
```
Spikhalskiy, line 31: Not needed.

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesPayloadConverter.java
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  @Override
+  public Optional<Payload> toData(Object value) throws DataConverterException {
```
Spikhalskiy, line 35: Please rework this class from inheritance to composition. In general, always prefer composition over inheritance. The fact that this method doesn't delegate to super is one more argument for composition here.
Also consider not using JacksonJsonPayloadConverter at all. This method is fully redefined. Mapper should be defined here too. And fromData should be redefined too. BTW, why is it not?
We don't need anything from the parent class at all.

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesPayloadConverter.java
```diff
@@ -0,0 +1,52 @@
...
+
+package io.temporal.common.converter;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.google.protobuf.ByteString;
+import io.temporal.api.common.v1.Payload;
+import io.temporal.internal.common.SearchAttributesUtil;
+import java.util.Optional;
+
+public class SearchAttributesPayloadConverter extends JacksonJsonPayloadConverter {
```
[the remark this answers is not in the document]
Spikhalskiy: What about Memo?

temporal-sdk/src/main/java/io/temporal/internal/common/InternalUtils.java
```diff
@@ -93,10 +90,10 @@ public static Object getValueOrDefault(Object value, Class<?> valueClass) {
 
   public static SearchAttributes convertMapToSearchAttributes(
       Map<String, Object> searchAttributes) {
-    DataConverter converter = DataConverter.getDefaultInstance();
+    SearchAttributesPayloadConverter converter = new SearchAttributesPayloadConverter();
```
Spikhalskiy, line 93: No, there is no need to create a converter for each serialization. The creation of Jackson mappers is very expensive and losing caches inside it too. Please use the singleton pattern as we discussed.

temporal-sdk/src/main/java/io/temporal/internal/common/SearchAttributesUtil.java
```diff
@@ -0,0 +1,182 @@
...
+    Unspecified,
+    String,
+    Keyword,
+    Int,
+    Double,
+    Bool,
+    Datetime
+  }
+
+  public static Map<String, Payload> serializeToObjectMap(Map<String, Object> searchAttributes) {
```
Spikhalskiy, line 68: Naming doesn't make sense. serializeToObjectMap should return object map, not accept it.

temporal-sdk/src/main/java/io/temporal/internal/common/SearchAttributesUtil.java
```diff
@@ -0,0 +1,182 @@
...
+import java.time.LocalDateTime;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Map.Entry;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class SearchAttributesUtil {
+
+  public static DefaultDataConverter converter = DefaultDataConverter.newDefaultInstance();
```
Spikhalskiy, line 37: why?.. only the special converter should be used here.

temporal-sdk/src/main/java/io/temporal/internal/common/SearchAttributesUtil.java
```diff
@@ -0,0 +1,182 @@
...
+  public static Map<String, Object> deserializeToObjectMap(SearchAttributes serializedMap) {
+    if (serializedMap == null) {
+      return null;
+    }
+
+    Map<String, Object> deserializedMap = new HashMap<>();
+    for (Entry<String, Payload> attribute : serializedMap.getIndexedFieldsMap().entrySet()) {
+      String key = attribute.getKey();
+      if (!parseSearchAttributes(key, attribute.getValue(), deserializedMap)) {
+        String typeString = attribute.getValue().getMetadataMap().get("type").toStringUtf8();
```
Spikhalskiy, line 95: You should use the String constant

temporal-sdk/src/main/java/io/temporal/internal/common/SearchAttributesUtil.java
```diff
@@ -0,0 +1,182 @@
...
+        Type javaType = stringToJavaType(typeString);
```
Spikhalskiy, line 96: this should be a part of our new serializer.

temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflowContext.java
```diff
@@ -242,8 +242,7 @@ public void invoke(Optional<Payloads> output, Failure failure) {
             "activity failure callback",
             () ->
                 result.completeExceptionally(
-                    (RuntimeException)
-                        FailureConverter.failureToException(failure, getDataConverter())));
+                    FailureConverter.failureToException(failure, getDataConverter())));
```
Spikhalskiy, line 245: Please revert changes to this class, they are irrelevant

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesPayloadConverter.java
```diff
@@ -0,0 +1,52 @@
...
+
+  @Override
+  public Optional<Payload> toData(Object value) throws DataConverterException {
+    try {
+      byte[] serialized = mapper.writeValueAsBytes(value);
+      return Optional.of(
+          Payload.newBuilder()
+              .putMetadata(EncodingKeys.METADATA_ENCODING_KEY, EncodingKeys.METADATA_ENCODING_JSON)
+              .putMetadata(
+                  EncodingKeys.METADATA_TYPE_KEY,
```
Spikhalskiy, line 42: Why do we store this field here, but don't use it for deserialization and instead of this use this field in the Util class?

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesPayloadConverter.java
```diff
@@ -0,0 +1,52 @@
...
+                  ByteString.copyFromUtf8(
+                      SearchAttributesUtil.javaTypeToEncodedType(value.getClass())))
```
Spikhalskiy, line 44: The relationship between SearchAttributesUtil and SearchAttributesPayloadConverter is completely unclear, methods and responsibilities are distributed randomly and these two classes are calling each without any clear scheme. There is no even a shadow of "incapsulation" and "low coupling" between these two classes.

temporal-sdk/src/main/java/io/temporal/internal/common/SearchAttributesUtil.java
```diff
@@ -0,0 +1,182 @@
...
+        return Double.class;
+      case Bool:
+        return Boolean.class;
+      case Datetime:
+        return LocalDateTime.class;
+    }
+    return null;
+  }
+
+  public static String javaTypeToEncodedType(Class type) {
```
Spikhalskiy, line 166: Please return SearchAttributeType here. The caller can call .name() on the result. This will make the method more universal.

temporal-sdk/src/main/java/io/temporal/internal/common/SearchAttributesUtil.java
```diff
@@ -0,0 +1,182 @@
...
+    String stringName = type.getSimpleName();
```
Spikhalskiy, line 167: stringName? What does it mean? Names are usually strings. className?

temporal-sdk/src/main/java/io/temporal/internal/common/SearchAttributesUtil.java
```diff
@@ -0,0 +1,182 @@
...
+    switch (stringName) {
+      case "String":
+        return SearchAttributeType.String.name();
+      case "Integer":
+        return SearchAttributeType.Int.name();
+      case "Double":
```
Spikhalskiy, line 173: should we take care of float too?

temporal-sdk/src/main/java/io/temporal/internal/common/SearchAttributesUtil.java
```diff
@@ -0,0 +1,182 @@
...
+      case Datetime:
+        return LocalDateTime.class;
+    }
+    return null;
+  }
+
+  public static String javaTypeToEncodedType(Class type) {
+    String stringName = type.getSimpleName();
+    switch (stringName) {
+      case "String":
```
Spikhalskiy, line 169: Why strings though? We at least can do something like
String.class.getSimpleName to get a bit more compile-time checks here.
Also, it can be a good idea to use fully qualified names here. There is org.joda.time.LocalDateTime for example.

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowContext.java
```diff
@@ -66,7 +66,7 @@
   void setContinueAsNewOnCompletion(ContinueAsNewWorkflowExecutionCommandAttributes attributes);
 
   /**
-   * RunId of the first run in the continue as new chain. Empty if this workflow never called
+   * RunId of the first run in the continue-as-new chain. Empty if this workflow never called
```
Spikhalskiy, line 69: Please revert, completely irrelevant.

Spikhalskiy (changes requested):
Right now this PR is a bit untidy with unclear relationship between classes, absence of reasonable abstraction layers, and incapsulation. Also, we discussed the usage of Singleton pattern and it's not what's going on in this PR right now. Also, Search Attributes code is still using a default data converter that should be fully avoided.

temporal-sdk/src/main/java/io/temporal/internal/common/InternalUtils.java
```diff
@@ -93,10 +90,10 @@ public static Object getValueOrDefault(Object value, Class<?> valueClass) {
 
   public static SearchAttributes convertMapToSearchAttributes(
```
Spikhalskiy, line 91: Why is this method not moved to SearchAttributesUtil? Looks like it belongs there.

temporal-sdk/src/main/java/io/temporal/internal/common/SearchAttributesUtil.java
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    Map<String, Object> deserializedMap = new HashMap<>();
+    for (Entry<String, Payload> attribute : serializedMap.getIndexedFieldsMap().entrySet()) {
+      String key = attribute.getKey();
+      if (!parseSearchAttributes(key, attribute.getValue(), deserializedMap)) {
+        String typeString = attribute.getValue().getMetadataMap().get("type").toStringUtf8();
+        Type javaType = stringToJavaType(typeString);
+        if (javaType == null) {
+          log.error("Error parsing Search Attribute: " + key + " of type " + typeString);
```
Spikhalskiy, line 98: Please use Slf4j placeholder syntax with {}

temporal-sdk/src/main/java/io/temporal/internal/common/SearchAttributesUtil.java
```diff
@@ -0,0 +1,182 @@
...
+import java.util.Map.Entry;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class SearchAttributesUtil {
+
+  public static DefaultDataConverter converter = DefaultDataConverter.newDefaultInstance();
+  private static final Logger log = LoggerFactory.getLogger(SearchAttributesUtil.class);
+
+  public enum SearchAttribute {
```
Spikhalskiy, line 40: This should be renamed to something like `StandardSearchAttributes`. SearchAttribute is a good name for a general-purpose class supporting both default and custom attributes. And we don't have one right now.

temporal-sdk/src/main/java/io/temporal/internal/common/SearchAttributesUtil.java
```diff
@@ -0,0 +1,182 @@
...
+      case StartTime:
+        deserializedMap.put(
+            key, converter.fromPayload(value, LocalDateTime.class, LocalDateTime.class));
+        break;
+      case ExecutionDuration:
+      case HistoryLength:
+      case StateTransitionCount:
+        deserializedMap.put(key, converter.fromPayload(value, Integer.class, Integer.class));
+        break;
+    }
```
Spikhalskiy, line 139: default statement?

temporal-sdk/src/main/java/io/temporal/internal/common/SearchAttributesUtil.java
```diff
@@ -0,0 +1,182 @@
...
+    for (Entry<String, Payload> attribute : serializedMap.getIndexedFieldsMap().entrySet()) {
+      String key = attribute.getKey();
+      if (!parseSearchAttributes(key, attribute.getValue(), deserializedMap)) {
+        String typeString = attribute.getValue().getMetadataMap().get("type").toStringUtf8();
+        Type javaType = stringToJavaType(typeString);
+        if (javaType == null) {
+          log.error("Error parsing Search Attribute: " + key + " of type " + typeString);
+        } else {
+          deserializedMap.put(
+              key, converter.fromPayload(attribute.getValue(), javaType.getClass(), javaType));
```
Spikhalskiy, line 101: Why is this here if it belongs to our specialized converter? Why is a default converted is used here that can be redefined instead of the specialized converter?

temporal-sdk/src/main/java/io/temporal/internal/common/SearchAttributesUtil.java
vkoby: No default statement, because DefaultSearchAttributes.valueOf(key) either resolves into one of the DefaultSearchAttributes or an error occurs.

temporal-sdk/src/main/java/io/temporal/common/converter/JacksonJsonPayloadConverter.java
```diff
@@ -61,7 +61,6 @@ public String getEncodingType() {
               .putMetadata(EncodingKeys.METADATA_ENCODING_KEY, EncodingKeys.METADATA_ENCODING_JSON)
               .setData(ByteString.copyFrom(serialized))
               .build());
-
```
Spikhalskiy, line 64: revert?

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesPayloadConverter.java
```diff
@@ -0,0 +1,93 @@
...
+  private SearchAttributesPayloadConverter() {
+    mapper = new ObjectMapper();
+    mapper.configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false);
+    mapper.registerModule(new JavaTimeModule());
+    mapper.setVisibility(PropertyAccessor.FIELD, JsonAutoDetect.Visibility.ANY);
+  }
+
+  public static SearchAttributesPayloadConverter getInstance() {
+    if (converter == null) {
+      converter = new SearchAttributesPayloadConverter();
```
Spikhalskiy, line 49: Why do we need a lazy initialization here? 
Lazy initialization of singletons in Java is not the prettiest topic. (Read https://www.journaldev.com/1377/java-singleton-design-pattern-best-practices-examples for example for more info).
We will always need this converter always. There is no optimization in keeping the initialization lazy.
Let's just not go there is initialize it in an eager maner.

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesPayloadConverter.java
```diff
@@ -0,0 +1,93 @@
...
+import com.google.protobuf.ByteString;
+import io.temporal.api.common.v1.Payload;
+import java.io.IOException;
+import java.lang.reflect.Type;
+import java.util.Optional;
+
+public final class SearchAttributesPayloadConverter implements PayloadConverter {
+
+  private static ObjectMapper mapper;
+  private static SearchAttributesPayloadConverter converter;
```
Spikhalskiy, line 38: When you store a singleton, there is a general practice that it should be stored in an `instance` field. Or, ideally, INSTANCE if the field is final.

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesPayloadConverter.java
```diff
@@ -0,0 +1,93 @@
...
+  @Override
+  public <T> T fromData(Payload content, Class<T> valueClass, Type valueType)
+      throws DataConverterException {
+    ByteString data = content.getData();
+    if (data.isEmpty()) {
+      return null;
+    }
+    try {
+      @SuppressWarnings("deprecation")
+      JavaType reference = mapper.getTypeFactory().constructType(valueType, valueClass);
```
Spikhalskiy, line 87: Please don't use deprecated APIs.
I know that it's copy-paste. But I'm pretty sure people decided to don't touch it there because of being afraid to break compatibility. Here we don't have this fear, so let's get rid of deprecated API here.

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesPayloadConverter.java
```diff
@@ -0,0 +1,93 @@
...
+                      SearchAttributesUtil.javaTypeToEncodedType(obj.getClass()).name()))
+              .setData(ByteString.copyFrom(serialized))
+              .build());
+    } catch (JsonProcessingException e) {
+      throw new DataConverterException(e);
+    }
+  }
+
+  @Override
+  public <T> T fromData(Payload content, Class<T> valueClass, Type valueType)
```
Spikhalskiy, line 79: Obviously, PayloadConverter interface doesn't play nicely here.
Our general purpose converters don't know the type and have nothing to get the type from.
This `SearchAttributesPayloadConverter` converter actually knows the type and knows where to get it.
It looks for me that the cleanest solution will be to remove `implements PayloadConverter` and implement `fromData(Payload content)` instead of `public <T> T fromData(Payload content, Class<T> valueClass, Type valueType)`

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesUtil.java
```diff
@@ -0,0 +1,187 @@
...
+      return null;
+    }
+    return payloadToObjectMap(searchAttributes.getIndexedFieldsMap());
+  }
+
+  public static Map<String, Payload> objectToPayloadMap(Map<String, Object> map) {
+    if (map == null || map.isEmpty()) {
+      return null;
+    }
+    Map<String, Payload> mapOfByteBuffer = new HashMap<>();
```
Spikhalskiy, line 56: Why ByteBuffer?

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesUtil.java
```diff
@@ -0,0 +1,187 @@
...
+
+  public static Map<String, Object> payloadToObjectMap(Map<String, Payload> serializedMap) {
+    if (serializedMap == null || serializedMap.isEmpty()) {
+      return null;
+    }
+    Map<String, Object> deserializedMap = new HashMap<>();
+    for (Map.Entry<String, Payload> attribute : serializedMap.entrySet()) {
+      String key = attribute.getKey();
+      if (!parseDefaultSearchAttributes(key, attribute.getValue(), deserializedMap)) {
+        String type =
```
Spikhalskiy, line 69: As we discussed, for a good encapsulation, it should be inside the converter. Basically, EncodingKeys.METADATA_TYPE_KEY should be an implementation detail of our converter and nobody else should know/use it.

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesUtil.java
```diff
@@ -0,0 +1,187 @@
...
+      case Datetime:
+        return LocalDateTime.class;
+    }
+    return null;
+  }
+
+  public static SearchAttributeType javaTypeToEncodedType(Class type) {
+    if (String.class.equals(type)) {
+      return SearchAttributeType.String;
+    } else if (Integer.class.equals(type)) {
```
Spikhalskiy, line 148: Long?

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesUtil.java
```diff
@@ -0,0 +1,187 @@
...
+      case ExecutionDuration:
+      case HistoryLength:
+      case StateTransitionCount:
+        deserializedMap.put(key, converter.fromData(value, Integer.class, Integer.class));
+        break;
+    }
+    return true;
+  }
+
+  public static Type encodedTypeToJavaType(String type) {
```
Spikhalskiy, line 122: encodedTypeToJavaType and javaTypeToEncodedType should be a part of the converter and be private (until we don't need it somewhere else - we can move it our later). No other code should/have to know which server values do we use for java types.

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesPayloadConverter.java
```diff
@@ -0,0 +1,93 @@
...
+    try {
+      byte[] serialized = mapper.writeValueAsBytes(obj);
+      return Optional.of(
+          Payload.newBuilder()
+              .putMetadata(EncodingKeys.METADATA_ENCODING_KEY, EncodingKeys.METADATA_ENCODING_JSON)
+              .putMetadata(
+                  EncodingKeys.METADATA_TYPE_KEY,
+                  ByteString.copyFromUtf8(
+                      // Handle NPE
+                      SearchAttributesUtil.javaTypeToEncodedType(obj.getClass()).name()))
```
Spikhalskiy, line 70: What if we get Unspecified here?

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesUtil.java
```diff
@@ -0,0 +1,187 @@
...
+    StartTime,
+    StateTransitionCount,
+    TaskQueue,
+    TemporalChangeVersion,
+    WorkflowId,
+    WorkflowType
+  }
+
+  enum SearchAttributeType {
+    Unspecified,
```
Spikhalskiy, line 179: This class should be a part of DataConverter, I don't see a good reason to leak this detail our from the DataConverter.

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesPayloadConverter.java
vkoby: We should discuss this: on options is to treat is a string

vkoby: Talked to Alex. Unspecified should throw an error

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesPayloadConverter.java
```diff
@@ -0,0 +1,135 @@
...
+      case Datetime:
+        return LocalDateTime.class;
+    }
+    throw new DataConverterException("Unsupported Search Attribute type: " + type);
+  }
+
+  private SearchAttributeType javaTypeToEncodedType(Class type) {
+    if (String.class.equals(type)) {
+      return SearchAttributeType.String;
+    } else if (Integer.class.equals(type)) {
```
Spikhalskiy, line 114: Please add Short

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesPayloadConverter.java
```diff
@@ -0,0 +1,135 @@
...
+        return Double.class;
+      case Bool:
+        return Boolean.class;
+      case Datetime:
+        return LocalDateTime.class;
+    }
+    throw new DataConverterException("Unsupported Search Attribute type: " + type);
+  }
+
+  private SearchAttributeType javaTypeToEncodedType(Class type) {
```
Spikhalskiy, line 111: `Class<?> type` to make it gentrified

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesPayloadConverter.java
```diff
@@ -0,0 +1,135 @@
...
+      return SearchAttributeType.Double;
+    } else if (Boolean.class.equals(type)) {
+      return SearchAttributeType.Bool;
+    } else if (LocalDateTime.class.equals(type)) {
+      return SearchAttributeType.Datetime;
+    }
+    throw new DataConverterException("Unsupported Search Attribute type: " + type);
+  }
+
+  enum SearchAttributeType {
```
Spikhalskiy, line 126: private

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesPayloadConverter.java
```diff
@@ -0,0 +1,135 @@
...
+        @SuppressWarnings("deprecation")
+        JavaType reference = mapper.getTypeFactory().constructType(javaType, javaType.getClass());
+        return mapper.readValue(data.toByteArray(), reference);
+      } catch (IOException e) {
+        throw new DataConverterException(e);
+      }
+    }
+  }
+
+  private Type encodedTypeToJavaType(String type) {
```
Spikhalskiy, line 88: static

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesPayloadConverter.java
```diff
@@ -0,0 +1,135 @@
...
+    SearchAttributeType attributeType;
+    try {
+      attributeType = SearchAttributeType.valueOf(type);
+    } catch (IllegalArgumentException e) {
+      return null;
```
Spikhalskiy, line 93: Why here it's null, but we throw an exception in the switch? It's inconsistent.

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesPayloadConverter.java
```diff
@@ -0,0 +1,135 @@
...
+      case Int:
+        return Integer.class;
+      case Double:
+        return Double.class;
+      case Bool:
+        return Boolean.class;
+      case Datetime:
+        return LocalDateTime.class;
+    }
+    throw new DataConverterException("Unsupported Search Attribute type: " + type);
```
Spikhalskiy, line 108: This exception is not handled anywhere. I don't think it's a good idea to fail everything just because we can't deserialize the search attribute.

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesPayloadConverter.java
```diff
@@ -0,0 +1,135 @@
...
+
+  public <T> T fromData(Payload payload) throws DataConverterException {
+    ByteString data = payload.getData();
+    String type = payload.getMetadataMap().get(EncodingKeys.METADATA_TYPE_KEY).toStringUtf8();
+    Type javaType = encodedTypeToJavaType(type);
+    if (data.isEmpty() || javaType == null) {
+      return null;
+    } else {
+      try {
+        @SuppressWarnings("deprecation")
```
Spikhalskiy, line 79: Please rework without deprecated APIs. We save this code in our main converter because we likely are very conscious of potential incompatibilities, but this is a new code, let's implement it without deprecated APIs.

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesPayloadConverter.java
```diff
@@ -0,0 +1,135 @@
...
+    } catch (JsonProcessingException e) {
+      throw new DataConverterException(e);
+    }
+  }
+
+  public <T> T fromData(Payload payload) throws DataConverterException {
+    ByteString data = payload.getData();
+    String type = payload.getMetadataMap().get(EncodingKeys.METADATA_TYPE_KEY).toStringUtf8();
+    Type javaType = encodedTypeToJavaType(type);
+    if (data.isEmpty() || javaType == null) {
```
Spikhalskiy, line 75: I don't think we should just throw away the attribute without a type (javaType == null case). Let's just return Payload if we can't deserialize it. This will be a gentler handling.

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesPayloadConverter.java
```diff
@@ -0,0 +1,135 @@
...
+    } else if (Integer.class.equals(type)) {
+      return SearchAttributeType.Int;
+    } else if (Double.class.equals(type) || Float.class.equals(type)) {
+      return SearchAttributeType.Double;
+    } else if (Boolean.class.equals(type)) {
+      return SearchAttributeType.Bool;
+    } else if (LocalDateTime.class.equals(type)) {
+      return SearchAttributeType.Datetime;
+    }
+    throw new DataConverterException("Unsupported Search Attribute type: " + type);
```
Spikhalskiy, line 123: I believe this breaks backwards compatibility.
Now we accept any class and we serialize it and it works.
So, if we don't find a match to standard types, SearchAttributeType.Undefined should be returned and it should be handled with our main user-defined DataConverter (delegate the serialization from this converter)

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesUtil.java
```diff
@@ -0,0 +1,129 @@
...
+import java.util.Map;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class SearchAttributesUtil {
+  private static final SearchAttributesPayloadConverter converter;
+  private static final Logger log = LoggerFactory.getLogger(SearchAttributesUtil.class);
+
+  static {
+    converter = SearchAttributesPayloadConverter.getInstance();
```
Spikhalskiy, line 35: Why do you need this? and why a separate static block is needed here?

Just use `SearchAttributesPayloadConverter.getInstance()` or `SearchAttributesPayloadConverter.INSTANCE` whereever you use it. Because everything is final - JIT will optimize this code perfectly and with ease.

Spikhalskiy (changes requested):
Please add a unit test that passes an arbitrary object in search attributes and reads it too. 
We. should support it to preserve backward compatibility.
We can add a log with warn level though telling that Search Attributes are supposed to be of predefined types only.

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesPayloadConverter.java
```diff
@@ -0,0 +1,135 @@
...
+              .setData(ByteString.copyFrom(serialized))
+              .build());
+    } catch (JsonProcessingException e) {
+      throw new DataConverterException(e);
+    }
+  }
+
+  public <T> T fromData(Payload payload) throws DataConverterException {
+    ByteString data = payload.getData();
+    String type = payload.getMetadataMap().get(EncodingKeys.METADATA_TYPE_KEY).toStringUtf8();
```
Spikhalskiy, line 73: From discussion with Alex, we don't need this. Nobody will read this field, so let's not write it.

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesPayloadConverter.java
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  public Optional<Payload> toData(Object obj) throws DataConverterException {
+    try {
+      byte[] serialized = mapper.writeValueAsBytes(obj);
+      String type = javaTypeToEncodedType(obj.getClass()).name();
+      return Optional.of(
+          Payload.newBuilder()
+              .putMetadata(EncodingKeys.METADATA_ENCODING_KEY, EncodingKeys.METADATA_ENCODING_JSON)
+              .putMetadata(EncodingKeys.METADATA_TYPE_KEY, ByteString.copyFromUtf8(type))
```
Spikhalskiy, line 63: From discussion with @alexshtin, we don't need this. Nobody will read this field, so let's not write it.

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesPayloadConverter.java
```diff
@@ -0,0 +1,135 @@
...
+      case Bool:
+        return Boolean.class;
+      case Datetime:
+        return LocalDateTime.class;
+    }
+    throw new DataConverterException("Unsupported Search Attribute type: " + type);
+  }
+
+  private SearchAttributeType javaTypeToEncodedType(Class type) {
+    if (String.class.equals(type)) {
```
Spikhalskiy, line 112: From discussion with @alexshtin, we don't need this javaTypeToEncodedType at all.

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesPayloadConverter.java
vkoby: We need this for in-memory testing

temporal-sdk/src/main/java/io/temporal/common/converter/DataConverterException.java
```diff
@@ -40,6 +40,10 @@ public DataConverterException(Payload content, Type[] valueTypes, Throwable caus
     super(toMessage(null, content, valueTypes), cause);
   }
 
+  public DataConverterException(String message) {
```
Spikhalskiy, line 43: We don't need it anymore

temporal-sdk/src/main/java/io/temporal/common/converter/EncodingKeys.java
```diff
@@ -22,8 +22,9 @@
 import com.google.protobuf.ByteString;
 import java.nio.charset.StandardCharsets;
 
-class EncodingKeys {
-  static final String METADATA_ENCODING_KEY = "encoding";
+public class EncodingKeys {
+  public static final String METADATA_ENCODING_KEY = "encoding";
```
Spikhalskiy, line 26: This is a mess now. Why `METADATA_ENCODING_RAW` is package-private, while `METADATA_ENCODING_JSON` is made public?
Also, this class is not intended for our users, at least now. So, if it's public - it should be in *.internal.* package.

There are three ways:
1. Don't mess with it. Keep this class package-private and use the text constants directly in tests.
2. Make everything public in this class and move it to *.internal.* package
3. Don't mess with it. Keep this class package-private. In unit tests add a class in io.tempotal.common.converter the only purpose of which is to expose these constants with public modifier.

I would just go with 1. Not a big sin taking into account very small amount of tests where we need to use this constants in.

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesPayloadConverter.java
```diff
@@ -0,0 +1,115 @@
...
+import java.time.LocalDateTime;
+import java.time.ZoneOffset;
+import java.time.ZonedDateTime;
+import java.util.Optional;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public final class SearchAttributesPayloadConverter {
+
+  private static ObjectMapper mapper;
```
Spikhalskiy, line 42: this shouldn't be static. You shouldn't initialize static fields in a constructor. It should be a non-static field.
this should be final. We initialize it in the constructor and only there and we are not going to hot-swap the instance.

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesUtil.java
Spikhalskiy: This comment stays valid

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesUtil.java
```diff
@@ -0,0 +1,124 @@
...
+        log.error("Error parsing Search Attribute: {}.", key);
+      } else {
+        deserializedMap.put(key, data);
+      }
+    }
+    return deserializedMap;
+  }
+
+  @Nonnull
+  public static SearchAttributeType javaTypeToEncodedType(Class<?> type) {
```
Spikhalskiy, line 76: It looks like this method in production code is used only by SearchAttributesPayloadConverter, why is it here? It belongs there. We don't need another code to use it for now. If we do ever need to expose it - we will consider relocation to the utils class.

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesPayloadConverter.java
```diff
@@ -0,0 +1,115 @@
...
+import java.io.IOException;
+import java.lang.reflect.Type;
+import java.time.LocalDateTime;
+import java.time.ZoneOffset;
+import java.time.ZonedDateTime;
+import java.util.Optional;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public final class SearchAttributesPayloadConverter {
```
Spikhalskiy, line 40: This should be relocated into `io.temporal.internal.common.converter`. It's not intended for app developers.

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesUtil.java
```diff
@@ -0,0 +1,124 @@
...
+import io.temporal.api.common.v1.Payload;
+import io.temporal.api.common.v1.SearchAttributes;
+import java.time.LocalDateTime;
+import java.util.HashMap;
+import java.util.Map;
+import javax.annotation.Nonnull;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class SearchAttributesUtil {
```
Spikhalskiy, line 31: This should be private or it should be relocated into io.temporal.internal.common.converter. It's not intended for app developers.

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesUtil.java
```diff
@@ -0,0 +1,124 @@
...
+      return SearchAttributeType.Double;
+    } else if (Boolean.class.equals(type)) {
+      return SearchAttributeType.Bool;
+    } else if (LocalDateTime.class.equals(type)) {
+      return SearchAttributeType.Datetime;
+    }
+    return SearchAttributeType.Unspecified;
+  }
+
+  public enum RegisteredSearchAttributes {
```
Spikhalskiy, line 91: `Registered` is misleading here. Registered are the ones that users registered.
We actually use the term "System" and I would stick to what is already in our documentation: https://docs.temporal.io/docs/server/workflow-search/#system-attributes

Spikhalskiy: This enum is and should be legitimately public.
 It should be moved from "io.temporal.common.converter" into `io.temporal.workflow` and be a first-class citizen, not an inner enum.

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowContext.java
```diff
@@ -260,6 +252,14 @@ void getVersion(
   /** @return replay safe UUID */
   UUID randomUUID();
 
+  /**
+   * Used to retrieve search attributes.
+   *
+   * @return SearchAttribute object which can be used by {@code
```
Spikhalskiy, line 258: Why it's `code` and not a `link`? Also, the class `WorkflowUtils` doesn't exist. This comment doesn't make much sense to me.

temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -246,9 +252,70 @@ private StartWorkflowExecutionRequest overrideStartWorkflowExecutionRequest(
...
+      } else {
+        searchAttributes.removeIndexedFields(key);
+        throw Status.INVALID_ARGUMENT.asRuntimeException();
+      }
+    }
+    request.setSearchAttributes(searchAttributes);
     return request.build();
   }
 
+  private SearchAttributeType searchAttributeToType(String key) {
```
Spikhalskiy, line 280: should it be called systemSearchAttributeToType? It looks like it works only with our system/default attributes

temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -246,9 +252,70 @@ private StartWorkflowExecutionRequest overrideStartWorkflowExecutionRequest(
...
+      case BinaryChecksums:
+      case ExecutionStatus:
+      case RunId:
+      case TaskQueue:
+      case TemporalChangeVersion:
+      case WorkflowId:
+      case WorkflowType:
+      case CustomKeywordField:
+      case CustomStringField:
+        return javaTypeToEncodedType(String.class);
```
Spikhalskiy, line 299: Why do we need all these calls to conversions here? Why can't we just return corresponded SearchAttributeType here? It just makes it more complex than it should be.

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesPayloadConverter.java
```diff
@@ -0,0 +1,115 @@
...
+  public Object fromData(Payload payload) throws DataConverterException {
+    ByteString data = payload.getData();
+    ByteString type = payload.getMetadataMap().get(EncodingKeys.METADATA_TYPE_KEY);
+    Type javaType = (type == null) ? null : encodedTypeToJavaType(type.toStringUtf8());
+    if (data.isEmpty() || javaType == null) {
+      log.warn("Something went wrong when parsing payload {}", payload);
+      return payload;
+    } else {
+      try {
+        JavaType reference = mapper.getTypeFactory().constructType(javaType);
```
Spikhalskiy, line 84: reference?

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesPayloadConverter.java
```diff
@@ -0,0 +1,115 @@
...
+      throw new DataConverterException(e);
+    }
+  }
+
+  public Object fromData(Payload payload) throws DataConverterException {
+    ByteString data = payload.getData();
+    ByteString type = payload.getMetadataMap().get(EncodingKeys.METADATA_TYPE_KEY);
+    Type javaType = (type == null) ? null : encodedTypeToJavaType(type.toStringUtf8());
+    if (data.isEmpty() || javaType == null) {
+      log.warn("Something went wrong when parsing payload {}", payload);
```
Spikhalskiy, line 80: Something?
Please provide at least some context that can be at least somehow helpful for an app-developer who sees this message.
Both `data.isEmpty()` and javaType == null are first, distinquishable and different, second, can be explained in a short message.

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesPayloadConverter.java
```diff
@@ -0,0 +1,115 @@
...
+              .setData(ByteString.copyFrom(serialized))
+              .build());
+    } catch (JsonProcessingException e) {
+      throw new DataConverterException(e);
+    }
+  }
+
+  public Object fromData(Payload payload) throws DataConverterException {
+    ByteString data = payload.getData();
+    ByteString type = payload.getMetadataMap().get(EncodingKeys.METADATA_TYPE_KEY);
```
Spikhalskiy, line 77: Per our discussion, please provide a separate code path to deserialize old-style system search attributes in case the server doesn't return `type` field back.

temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -246,9 +252,70 @@ private StartWorkflowExecutionRequest overrideStartWorkflowExecutionRequest(
     if (taskTimeoutMillis != Durations.toMillis(request.getWorkflowTaskTimeout())) {
       request.setWorkflowTaskTimeout(Durations.fromMillis(taskTimeoutMillis));
     }
+
+    SearchAttributes.Builder searchAttributes = request.getSearchAttributes().toBuilder();
```
Spikhalskiy, line 256: First, it definitely should be in a separate function with a good name. 

But second, I'm not sure what's going on here - do we support only system search attributes here?
We need to provide an ability for our users to define their own attributes and I'm not sure it's what I see here.

temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -246,9 +252,70 @@ private StartWorkflowExecutionRequest overrideStartWorkflowExecutionRequest(
...
+        return javaTypeToEncodedType(String.class);
+      case CloseTime:
+      case ExecutionTime:
+      case StartTime:
+      case CustomDatetimeField:
+        return javaTypeToEncodedType(LocalDateTime.class);
+      case ExecutionDuration:
+      case HistoryLength:
+      case StateTransitionCount:
+      case CustomIntField:
```
Spikhalskiy, line 308: What is this? Did we provide hardcoded values for custom attributes that we use in our own unit tests?
No, it shouldn't be here. Our tests should register these attributes in the test environment. Like application developers would do it when they write tests for their workflows that could use absolutely any custom attributes. So, we have to provide a `registerSearchAttribute()` or something like that in our test framework.

temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
vkoby: Custom keys are also there.

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesPayloadConverter.java
vkoby: Referencing [TypeReference](http://javadox.com/org.codehaus.jackson/jackson-core-asl/1.9.13/org/codehaus/jackson/type/TypeReference.html) class since we already used `type` and `javaType` in this scope.

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesPayloadConverter.java
vkoby: But it does! 😃

temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
vkoby: Refer to this [documentation](https://docs.temporal.io/docs/server/workflow-search/#search-attributes) page: 
"Docker auto-setup image has some pre-allowlisted search attributes that are handy for testing, that have their types indicated in their name:... "

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesPayloadConverter.java
vkoby: I didn't elaborate more here because Alex's error message will follow with all the details

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesPayloadConverter.java
Spikhalskiy: It doesn't and I verified it with @alexshtin 
You got this behavior because you now send type field to the server when you save and old servers were saving this metadata and reading it back. That's how you got "type" field back.
If you will run this code against an old server with an old workflow initiated by old JavaSDK - you will break backward compatibility.
Feel free to ask @alexshtin for more details, he has the full context.

temporal-sdk/src/main/java/io/temporal/common/converter/SearchAttributesPayloadConverter.java
Spikhalskiy: Why TypeReference?.. TypeReference is not used here.

temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
Spikhalskiy: @vkoby that's exactly the problem

temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
Spikhalskiy: @vkoby how did this address the need for users to run tests for workflows with their own custom search attributes?

temporal-sdk/src/main/java/io/temporal/workflow/SystemSearchAttributes.java
```diff
@@ -17,21 +17,32 @@
...
+  ExecutionTime,
+  HistoryLength,
+  RunId,
+  StartTime,
+  StateTransitionCount,
+  TaskQueue,
+  TemporalChangeVersion,
+  WorkflowId,
+  WorkflowType,
+  CustomKeywordField,
```
Spikhalskiy, line 42: All custom fields should go away from this enum, it's just custom fields really. And we don't care about compatibility with them, because they are just a part of our test docker, not prod temporal.

temporal-sdk/src/main/java/io/temporal/workflow/SystemSearchAttributes.java
```diff
@@ -17,21 +17,32 @@
...
-    }
-    return jsonConverter.fromPayload(
-        searchAttributes.getIndexedFieldsOrThrow(key), classType, classType);
-  }
+/**
+ * These are the system attributes that are provided by default. Use {@link #name()} to get a key
+ * for our Search Attributes objects map.
+ */
+public enum SystemSearchAttributes {
+  BatcherNamespace,
```
Spikhalskiy, line 27: It looks like actually most of these values are never in the SearchAtttibutes object at all.
Only TemporalChangeVersion, BinaryChecksums, BatcherNamespace, BatcherUser are there.
Everything else is actually not here.
Everything that is defined in this enum on server side: https://github.com/temporalio/temporal/blob/36cd514da6a1333bcebd7c45bb99314a0e786bd0/common/searchattribute/defs.go#L57
is not part of SearchAttributes. I'm already not a big fan of having any logic related to these predefined attributes at all.
Let's maybe kill it altogether, especially if we decided that it's ok to trust "type" metadata?

temporal-sdk/src/main/java/io/temporal/internal/common/converter/SearchAttributesPayloadConverter.java
```diff
@@ -0,0 +1,133 @@
...
+        ZonedDateTime utc = ((LocalDateTime) obj).atZone(ZoneOffset.UTC);
+        serialized = mapper.writeValueAsBytes(utc);
+      } else {
+        serialized = mapper.writeValueAsBytes(obj);
+      }
+      String type = javaTypeToEncodedType(obj.getClass()).name();
+      return Optional.of(
+          Payload.newBuilder()
+              .putMetadata(EncodingKeys.METADATA_ENCODING_KEY, EncodingKeys.METADATA_ENCODING_JSON)
+              .putMetadata(EncodingKeys.METADATA_TYPE_KEY, ByteString.copyFromUtf8(type))
```
Spikhalskiy, line 69: Please write a really huge javadoc here to explain why we decided to have this field here even while we ideally shouldn't be sending it to the server at all.

temporal-sdk/src/main/java/io/temporal/internal/common/converter/SearchAttributesUtil.java
```diff
@@ -0,0 +1,104 @@
...
+    Unspecified,
+    String,
+    Keyword,
+    Int,
+    Double,
+    Bool,
+    Datetime
+  }
+
+  public static class Pair<D, T> {
```
Spikhalskiy, line 83: This class is used and needed only in test code, why is it here? Please put it close to the test types registration logic. And better to actually make a separate TypeRegistry class and this class should be just private there.

temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowInfoImpl.java
```diff
@@ -80,6 +80,7 @@ public long getRunStartedTimestampMillis() {
   }
 
   @Override
+  @Deprecated
```
Spikhalskiy, line 83: We don't need to deprecate it, it's a private class because it's located in internal package. If somebody uses it - they will have to deal with that.

temporal-sdk/src/main/java/io/temporal/workflow/WorkflowInfo.java
```diff
@@ -47,6 +47,7 @@
    */
   long getRunStartedTimestampMillis();
 
+  @Deprecated
```
Spikhalskiy, line 50: Please add a comment on why it was deprecated and which method should be used instead.

temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -245,9 +249,43 @@ private StartWorkflowExecutionRequest overrideStartWorkflowExecutionRequest(
     if (taskTimeoutMillis != Durations.toMillis(request.getWorkflowTaskTimeout())) {
       request.setWorkflowTaskTimeout(Durations.fromMillis(taskTimeoutMillis));
     }
+
+    request.setSearchAttributes(getValidSearchAttributes(request));
     return request.build();
   }
 
+  private SearchAttributes.Builder getValidSearchAttributes(
```
Spikhalskiy, line 257: It's not a getter, please rename. It can be supply, convert, extract, transform, but not get.

temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -245,9 +249,43 @@ private StartWorkflowExecutionRequest overrideStartWorkflowExecutionRequest(
...
+      StartWorkflowExecutionRequest.Builder request) {
+
+    SearchAttributes.Builder searchAttributes = request.getSearchAttributes().toBuilder();
+    Map<String, SearchAttributesUtil.Pair<String, SearchAttributeType>> registeredAttributes =
+        store.getRegisteredSearchAttribute();
```
Spikhalskiy, line 262: one?

temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowStore.java
```diff
@@ -144,6 +147,10 @@ public TaskQueueId getTaskQueueId() {
 
   void registerDelayedCallback(Duration delay, Runnable r);
 
+  void registerSearchAttribute(String name, Class<?> type);
+
+  Map<String, Pair<String, SearchAttributeType>> getRegisteredSearchAttribute();
```
Spikhalskiy, line 152: one?

temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowStoreImpl.java
```diff
@@ -288,6 +293,22 @@ public void registerDelayedCallback(Duration delay, Runnable r) {
...
 
+  @Override
+  public void registerSearchAttribute(String name, Class<?> type) {
+    SearchAttributeType encodedType = javaTypeToEncodedType(type);
+    if (encodedType == SearchAttributeType.Unspecified) {
+      throw Status.INVALID_ARGUMENT
+          .withDescription("Unable to read search attribute type: " + type)
+          .asRuntimeException();
+    }
+    searchAttributes.put(name, new Pair(null, encodedType));
```
Spikhalskiy, line 304: Pair belongs to this class. It's the only class that creates, persists and extracts them. All other classes are only passing existing entities around.

temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -245,9 +249,43 @@ private StartWorkflowExecutionRequest overrideStartWorkflowExecutionRequest(
...
+
+      String key = searchAttribute.getKey();
+      Payload val = searchAttribute.getValue();
+
+      if (!registeredAttributes.containsKey(key)) {
+        throw Status.INVALID_ARGUMENT
+            .withDescription(key + " is not a valid search attribute name")
+            .asRuntimeException();
+      } else {
+        SearchAttributeType type = registeredAttributes.get(key).getType();
```
Spikhalskiy, line 275: This codeblock and logic don't make too much sense to me, I'm not sure that it even has a consumer. Let's discuss on Zoom.

PR #619, opened 2021-08-05
Cleaned up code persisting RetryOptions into side effect

Spikhalskiy:
## What was changed
Some optimizations and cleanup of code persisting RetryOptions into side effects.
Also, some clarifying comments were added.

## Why?
The current code is misleading and will break if SerializableRetryOptions gets a non-default implementation of `equals`.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowRetryerInternal.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowRetryerInternal.java
@@ -37,15 +37,13 @@ final class WorkflowRetryerInternal {
   /** This class is needed as Jackson is not capable to serialize RetryOptions as they are. */
   static class SerializableRetryOptions {
     private long initialIntervalMillis;
-
     private double backoffCoefficient;
-
     private int maximumAttempts;
-
     private long maximumIntervalMillis;
-
     private String[] doNotRetry;
 
+    // required for Jackson serialization
+    @SuppressWarnings("unused")
     public SerializableRetryOptions() {}
 
     public SerializableRetryOptions(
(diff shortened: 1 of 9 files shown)


temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowRetryerInternal.java
```diff
@@ -197,31 +191,42 @@ private static RetryOptions getRetryOptionsSideEffect(String retryId, RetryOptio
...
+    // We generate a callId deterministically once during the initial call to retryAsync and persist
+    // the original retryOptions in the side effect, so it can be looked up during replaying
     SerializableRetryOptions sRetryOptions =
         WorkflowInternal.mutableSideEffect(
-            retryId,
+            retryCallId,
             SerializableRetryOptions.class,
             SerializableRetryOptions.class,
-            Object::equals,
+            // we never want to persist a new value if there is already a previously saved value
```
mfateev, line 203: Why? I think it is fine to update the retry options during retries here. 

If we decide to never update retry options then `sideEffect` can be used instead of `mutableSideEffect`.

Spikhalskiy: Because it's how it works now.

You have SerializableRetryOptions without implemented equals (always returns false effectively in our usage)
Then you pass Object:equals as a function that defines when it should be updated (which is kinda crazy already, at least it should be "not equals"). So, you get false for always for it.
So, currently, we always pass false here effectively. And we always save the original retry options. I just expressed what we already have in an explicit manner and not bug-prone.

Quinn-With-Two-Ns:
Closing as this is a stale PR

PR #629, opened 2021-08-15
Add more unit tests for Kotlin Async

Spikhalskiy:
Add more unit tests for Kotlin `Async` support
Code in `KotlinMethodReferenceDisassemblyService` is restructured to handle Lambda cases more explicitly

### DIFF
--- a/temporal-kotlin/src/main/kotlin/io/temporal/internal/async/KotlinMethodReferenceDisassemblyService.kt
+++ b/temporal-kotlin/src/main/kotlin/io/temporal/internal/async/KotlinMethodReferenceDisassemblyService.kt
@@ -22,24 +22,34 @@ package io.temporal.internal.async
 import io.temporal.internal.async.spi.MethodReferenceDisassemblyService
 import io.temporal.workflow.Functions
 import kotlin.jvm.internal.CallableReference
+import kotlin.jvm.internal.Lambda
 
 class KotlinMethodReferenceDisassemblyService : MethodReferenceDisassemblyService {
   override fun getMethodReferenceTarget(methodReference: Any): Any? {
-    if (methodReference is CallableReference) {
-      return unwrapCallableReference(methodReference)
+    return when (methodReference) {
+      is Functions.TemporalFunctionalInterfaceMarker -> {
+        unwrapTemporalFunctionalInterfaceInKotlin(methodReference)
+      }
+      else -> {
+        unwrapIfCallableReference(methodReference)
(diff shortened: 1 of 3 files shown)


temporal-kotlin/build.gradle
```diff
@@ -1,17 +1,17 @@
...
 }
 
 description = '''Temporal Workflow Java SDK Kotlin'''
 
 ext {
-//    kotlinVersion = '1.3.72'
+    kotlinVersion = '1.3.72'
 //    kotlinVersion = '1.4.21'
-    kotlinVersion = '1.5.21'
+//    kotlinVersion = '1.5.21'
```
Sushisource, line 14: Are these minimums? Why the different commented out ones? Just testing compat?

temporal-kotlin/src/main/kotlin/io/temporal/internal/async/KotlinMethodReferenceDisassemblyService.kt
```diff
@@ -22,24 +22,34 @@ package io.temporal.internal.async
...
+    return when (callableReference) {
+      /**
+       * Strategy 1.1
+       * We unwrap simple native Kotlin [CallableReference] (which is used for method references)
+       */
+      is CallableReference -> unwrapCallableReference(callableReference)
+      // we end up here if lambda is passed instead of a method reference
+      is Lambda<*> -> null
+      // something unexpected that we don't know how to handle
+      else -> null
```
Sushisource, line 49: Higher up does this result in us throwing an error? I'm wondering if maybe we can provide a little additional context or info down here somehow. Not a big deal either way.

Sushisource (approved):

temporal-kotlin/build.gradle
Spikhalskiy: It's the last bugfix releases for all minor versions that are still popular and keeps updating.
The original code I wrote worker on 1.4, but didn't work on 1.5 for example, so yeah, I leave it for easier testing

temporal-kotlin/src/main/kotlin/io/temporal/internal/async/KotlinMethodReferenceDisassemblyService.kt
Spikhalskiy: Nah, it will result in our inability to extract a target and we will schedule it as a separate thread just as an async function, not temporal stub method reference.

PR #633, opened 2021-08-15
Stop enforcing TestWorkflowRule timeout when run in JVM debug mode, respect JUnit(timeout) over TestRule timeout

### LINKED ISSUE #612: JUnit's @test(timeout) annotation should override TestWorkflowRule settings
## Expected Behavior
Junit `@Test(timeout)` should override timeouts set on the SDKTestWorkflowRule, which is 10s by default.
As a workaround, you can set a different timeout in the rule, but it would affect all tests in the class:
`SDKTestWorkflowRule.newBuilder().setTestTimeoutSeconds(15).build();
`
## Actual Behavior
Errors

  1. Write a test that uses SDKTestWorkflowRule
  2. Add @Test(timeout=20000) annotation to the test
  3. Run test and observe fail

### LINKED ISSUE #632: Running in TEMPORAL_DEBUG mode should turn off TestWorkflowRule timeouts
TestWorkflowRule enforces 10 seconds timeout for tests by default. This is useful during running a build with all tests, but it's annoying for debugging and hurts productivity. We should disable this timeout when we have TEMPORAL_DEBUG env variable present.

Spikhalskiy:
## What was changed
We now don't apply `globalTimeout` in `TestWorkflowRule` in case an explicit `@Test(timeout)` is provided.
We now also don't enforce `globalTimeout` if TEMPORAL_DEBUG env variable is set.

## Why?
Now during breakpoints-based debugging, there is no way to turn off the default 10 seconds limit which is annoying and hurts productivity.
Also, there is no reason to enforce global timeout if the test has a timeout specified on a test level, we always should prioritize more specific settings.

Resolves #632, resolves #612

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/common/DebugModeUtils.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/common/DebugModeUtils.java
@@ -1,0 +1,37 @@
+/*
+ *  Copyright (C) 2020 Temporal Technologies, Inc. All Rights Reserved.
+ *
+ *  Copyright 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ *  Modifications copyright (C) 2017 Uber Technologies, Inc.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License"). You may not
+ *  use this file except in compliance with the License. A copy of the License is
+ *  located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ *  or in the "license" file accompanying this file. This file is distributed on
+ *  an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
+package io.temporal.internal.common;
+
+public class  DebugModeUtils {
+  private static final boolean TEMPORAL_DEBUG_MODE = readTemporalDebugModeFromEnvVar();
+
+  public static boolean isTemporalDebugModeOn() {
+    return TEMPORAL_DEBUG_MODE;
+  }
+
+  private static boolean readTemporalDebugModeFromEnvVar() {
+    String temporalDebugValue = System.getenv("TEMPORAL_DEBUG");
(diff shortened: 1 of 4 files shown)


temporal-testing-junit4/src/main/java/io/temporal/testing/TestWorkflowRule.java
```diff
@@ -109,10 +112,12 @@ private TestWorkflowRule(Builder builder) {
             ? WorkflowImplementationOptions.getDefaultInstance()
             : builder.workflowImplementationOptions;
     globalTimeout =
-        Timeout.seconds(
-            builder.testTimeoutSeconds == 0
-                ? DEFAULT_TEST_TIMEOUT_SECONDS
-                : builder.testTimeoutSeconds);
+        !DebugModeUtils.isTemporalDebugModeOn()
```
Spikhalskiy, line 115: All the main logical changes of this PR are in this class

temporal-sdk/src/main/java/io/temporal/internal/common/DebugModeUtils.java
```diff
@@ -0,0 +1,28 @@
...
+ *  or in the "license" file accompanying this file. This file is distributed on
+ *  an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
+package io.temporal.internal.common;
+
+public class DebugModeUtils {
+  private static final boolean TEMPORAL_DEBUG_MODE = System.getenv("TEMPORAL_DEBUG") != null;
```
Sushisource, line 23: I know it's annoying but you probably want to make this false if it's a "falsy" value too. Sometimes people will set to '0' or "false" and expect it to get turned off

Sushisource (approved):

temporal-sdk/src/main/java/io/temporal/internal/common/DebugModeUtils.java
Spikhalskiy: Yeah, makes sense actually, let me do it as a part of this PR.

PR #638, opened 2021-08-16
Do not retry local activity when non-retryable ApplicationFailure is thrown

### LINKED ISSUE #323: Throwing non-retriable failure ignored by local activities
## Expected Behavior

In local activities throwing exception returned by ApplicationFailure.newNonRetryableFailure should result for workflow NOT to retry it. 

## Actual Behavior

throwing with ApplicationFailure.newNonRetryableFailure in Activity (launched via `Workflow.newLocalActivityStub`) results for workflow to keep retrying, which is incorrect behavior. Regular activities (launched via `Workflow.newActivityStub`) are working as expected.

  1. Create new activity A with method m() calling `throw ApplicationFailure.newNonRetryableFailure(...)`
  2. Create workflow and register A as local activity via `Workflow.newLocalActivityStub` with `LocalActivityOptions` setup for some retries and duration.
  3. Call method m() from workflow and ensure that m() is keep retrying.

## Specifications

  - Version: Temporal Java SDK 1.0.4
  - Platform: Temporal Server 1.5.x

vkoby:
## What was changed
Implemented a check if activity failed with a nonretryable ApplicationFailure. 

## Why?
Local Activities ignored 

## Checklist
1. Closes #323 

2. How was this tested:
Unit tests


temporal-sdk/src/test/java/io/temporal/workflow/activityTests/ActivityThrowingErrorTest.java
```diff
@@ -59,22 +67,45 @@ public void activityThrowsError() {
...
 
-    TestWorkflow1 workflow = client.newWorkflowStub(TestWorkflow1.class, options);
     try {
-      workflow.execute(UUID.randomUUID().toString());
+      workflow.execute(name, 1);
     } catch (WorkflowFailedException e) {
       assertTrue(e.getCause() instanceof ActivityFailure);
       assertTrue(e.getCause().getCause() instanceof ApplicationFailure);
-      assertEquals("java.lang.Error", ((ApplicationFailure) e.getCause().getCause()).getType());
+      assertEquals("fail", ((ApplicationFailure) e.getCause().getCause()).getType());
```
Spikhalskiy, line 83: let's put this "fail" in a constant so it's dead-clear where this value is defined and where it comes from.

temporal-sdk/src/test/java/io/temporal/workflow/activityTests/ActivityThrowingErrorTest.java
```diff
@@ -86,15 +117,23 @@ public void activityThrowsError() {
...
   }
 
-  public static class Activity1Impl implements TestActivity1 {
+  public static class ApplicationFailureActivity implements TestActivity4 {
+    public static final Map<String, AtomicInteger> invocations = new ConcurrentHashMap<>();
+
     @Override
-    public String execute(String input) {
-      return Workflow.randomUUID().toString();
+    public void execute(String testName, int retryable) {
```
Spikhalskiy, line 130: Let's make retryable is not boolean, it will be easier to read.

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
```diff
@@ -232,6 +230,14 @@ public Throwable wrapFailure(Task task, Throwable failure) {
       int attempt = activityTask.getAttempt();
       result.setAttempt(attempt);
 
+      // Do not retry for non-retryable ApplicationFailures
+      if (result.getTaskFailed() != null
```
Spikhalskiy, line 234: Let's put this shebang
```
result.getTaskFailed() != null
          && result.getTaskFailed().getFailure() != null
          && result.getTaskFailed().getFailure() instanceof ApplicationFailure
          && ((ApplicationFailure) result.getTaskFailed().getFailure()).isNonRetryable()
```
into a separate function. This will make code self-documenting.

Spikhalskiy (approved):

PR #640, opened 2021-08-17
Prevent confusing NPE when throwing deadlock errors

Sushisource:
## What was changed
Fix NPE that could be generated in the middle of generating a deadlock exception

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -27,12 +27,6 @@ To generate licence headers execute
 ./gradlew licenseFormat
 ```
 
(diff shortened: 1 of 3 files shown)


vitarb (approved):

Sushisource:
Talked with Dmitry a bit and I agree with him we need to try to figure out in more detail why this can even happen in the first place

Spikhalskiy:
Fixed by #642

PR #644, opened 2021-08-17
Turn on unit testing with Docker

vkoby:
## What was changed
Added unit testing with Docker


Sushisource:
I didn't realize we weren't doing this. I think this is really going to help catch a lot of problems sooner... as indicated by the fact that there are quite a bunch of already-broken tests

Spikhalskiy (approved):

PR #645, opened 2021-08-17
Log warns on deadlocks to make it easier to debug them

Sushisource:
## What was changed
Minor changes to formatting of deadlock error message and log them at warning level when they occur

## Why?
We're having some pipeline troubles w/ deadlocks and the important part of the stack traces are never really showing up. This should be helpful to us and to users.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/DeterministicRunnerImpl.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/DeterministicRunnerImpl.java
@@ -231,6 +231,7 @@ class DeterministicRunnerImpl implements DeterministicRunner {
         }
       }
       e.setStackDump(dump.toString());
+      log.warn("Workflow potentially deadlocked", e);
       throw e;
     } finally {
       inRunUntilAllBlocked = false;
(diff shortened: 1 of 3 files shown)


temporal-sdk/src/main/java/io/temporal/internal/sync/DeterministicRunnerImpl.java
```diff
@@ -231,6 +231,7 @@ public void runUntilAllBlocked(long deadlockDetectionTimeout) {
         }
       }
       e.setStackDump(dump.toString());
+      log.warn("Workflow potentially deadlocked", e);
```
Spikhalskiy, line 234: Why do we need this? It will go to logs anyway because it will not be handled.

vitarb (approved):

Spikhalskiy (changes requested):
Let's discuss. You don't typically log an exception and rethrow/throw it after, it's kinda a way for a mess. You log an exception in a place where you actually handle it.

temporal-sdk/src/main/java/io/temporal/internal/sync/DeterministicRunnerImpl.java
mfateev, line 234: Wouldn't it lead to logging the same error twice?

Sushisource: In theory, in practice it's not. I'll figure out why

vitarb:
We can wrap an exception and add a message, otherwise LGTM. #

vitarb:
We can wrap an exception and add a message, otherwise LGTM.

Sushisource:
I'm just being a bit daft. You actually can find the stack trace you want deep in the raw history json

Issue #650, opened 2021-08-19
Add unit tests for WorkflowExecutionUtils#getInstanceCloseEvent 


Spikhalskiy:
WorkflowExecutionUtils right now is completely untestable and unmockable, we should decouple this class into testable pieces to make it possible to cover it with unit tests.
The first unit test worth adding is a unit test for an issue fixed in https://github.com/temporalio/sdk-java/pull/472

Spikhalskiy:
This was mostly covered by extensive GrpcRetryer reworking and testing done in the last couple of months.

Issue #651, opened 2021-08-19
Change the default behavior of retries in WorkflowExecutionUtils#getInstanceCloseEvent


Spikhalskiy:
We have no exponential backoff on `WorkflowExecutionUtils#getInstanceCloseEvent` and we should add it to avoid clients hammering the server in case of server/network outage.
Also, it may be something worth exposing in configuration - TBD.

PR #654, opened 2021-08-23
GrpcRetryer now respects DEADLINE_EXCEEDED as non-retryable

### LINKED ISSUE #653: WorkflowExecutionUtils  and GrpcRetryer don't respect reaching deadline from GRPC context and downstream DEADLINE_EXCEEDED exceptions
GRPC has a concept of gRPC deadline propagation when an absolute deadline is passed and propagated through chained GRPC calls using thread-local `io.grpc.Context`

Read more here:
https://docs.microsoft.com/en-us/aspnet/core/grpc/deadlines-cancellation?view=aspnetcore-5.0#propagating-deadlines

`WorkflowExecutionUtils` calculates the Deadline (https://github.com/temporalio/sdk-java/blob/25da5d2064a6c024148e236e060c34d286283b56/temporal-sdk/src/main/java/io/temporal/internal/common/WorkflowExecutionUtils.java#L250-L253) using a timeout value passed as an argument and ignores the fact that the calculated deadline or even the current moment can already be after the absolute Deadline in the context.
GRPC ignores these new deadlines that are later than the current deadline already presented in the GRPC Context: https://github.com/grpc/grpc-java/blob/master/context/src/main/java/io/grpc/Context.java#L316-L320
 `GrpcRetryer` doesn't check the fact that the current moment passed the absolute deadline from GRPC Context and continues to calculate the new deadline, make a call, ignore DEADLINE_EXCEEDED happening immediately later in the chain of calls, rinse and repeat.

There are two unit tests with the reproduction of this problem.
In one of the tests the deadline is reached before the workflow completes and we ignore this fact and keep retrying till the workflow is actually finished.
In another test, the deadline is reached before the workflow is even started and we end up in a cycle of Integer.MAX_VALUE milliseconds of retries (24 days)

https://github.com/Spikhalskiy/java-sdk/commit/d8d21d3b76edf8af7b4f843115e8f57ec1b831b2#diff-41b99c4ace946461232f6ec43784563e243168821f47d1807ae1962e856cfe0fR40

Spikhalskiy:
## What was changed
GrpcRetryer now respects DEADLINE_EXCEEDED as non-retryable
Refactor GrpcRetryer into GrpcAsyncRetryer and GrpcSyncRetryer
Async rewrite code was rewritten to be much simpler and have fewer wrappers and handlers
Implementation is reworked to be more unit-testable

## Why?
As described in issue #653, right now when we reach DEADLINE from GRPC context we continue to retry while it already doesn't make sense, all requests will continue to fail with DEADLINE_EXCEEDED

1. Closes #653
Refactoring helps also to address #650

2. How was this tested:
Integration tests for workflows with GRPC deadline expiring before the start of the workflow and in the middle of workflow execution


temporal-serviceclient/src/main/java/io/temporal/internal/retryer/GrpcAsyncRetryer.java
```diff
@@ -0,0 +1,140 @@
...
+        .thenAccept(
+            (ignore) -> {
+              // try-catch is because get() call might throw.
+              CompletableFuture<R> result;
+
+              try {
+                result = function.get();
+              } catch (Throwable e) {
+                throttler.failure();
+                // function doesn't suppose to throw exceptions, it should always return a
```
vitarb, line 73: nit. s/doesn't suppose/isn't supposed

temporal-serviceclient/src/main/java/io/temporal/internal/retryer/GrpcRetryerUtils.java
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  /**
+   * @param options retry options
+   * @param startTimeMs timestamp when attempts started
+   * @param attempt number of the last made attempt
+   * @param currentTimeMillis current timestamp
+   * @return true if we out of attempts or time to retry
+   */
+  static boolean ranOutOfRetries(
```
vitarb, line 71: nit. Put start time and current time next to each other in the list of parameters.

vitarb, line 71: Should we name it positively like `canRetry` or `isRetryAllowed`?

temporal-serviceclient/src/main/java/io/temporal/internal/retryer/GrpcSyncRetryer.java
```diff
@@ -0,0 +1,92 @@
...
+      try {
+        throttler.throttle();
+        R result = r.apply();
+        throttler.success();
+        return result;
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
+        throw new CancellationException();
+      } catch (StatusRuntimeException e) {
+        lastException = e;
```
vitarb, line 66: Do deadline exceeded exceptions have any cause? If not, should we init the cause and specify previous error there? This way we'll have context of what has caused deadline exceeded error in the first place.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/DefaultServiceOperationRpcRetryOptions.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+
+package io.temporal.serviceclient;
+
+import io.grpc.Status;
+import io.temporal.api.errordetails.v1.QueryFailedFailure;
+import java.time.Duration;
+
+public class DefaultServiceOperationRpcRetryOptions {
+  public static final Duration RETRY_SERVICE_OPERATION_INITIAL_INTERVAL = Duration.ofMillis(20);
```
vitarb, line 27: Is `RETRY_SERVICE_OPERATION_` really necessary?

temporal-serviceclient/src/main/java/io/temporal/serviceclient/DefaultServiceOperationRpcRetryOptions.java
```diff
@@ -0,0 +1,63 @@
...
+  public static final Duration RETRY_SERVICE_OPERATION_EXPIRATION_INTERVAL = Duration.ofMinutes(1);
+  public static final Duration RETRY_SERVICE_OPERATION_MAXIMUM_INTERVAL;
+  public static final double RETRY_SERVICE_OPERATION_BACKOFF = 1.2;
```
vitarb, line 30: We've recently reduced backoff in go to 2x. Probably would make sense to do same here.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/DefaultServiceOperationRpcRetryOptions.java
vitarb, line 27: Initial interval has been recently increased to 200ms in go to avoid spamming the server too much.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/DefaultServiceOperationRpcRetryOptions.java
```diff
@@ -0,0 +1,63 @@
...
+    // CANCELLED and DEADLINE_EXCEEDED are always considered non-retryable
+    roBuilder
+        .addDoNotRetry(Status.Code.INVALID_ARGUMENT, null)
+        .addDoNotRetry(Status.Code.NOT_FOUND, null)
+        .addDoNotRetry(Status.Code.ALREADY_EXISTS, null)
+        .addDoNotRetry(Status.Code.FAILED_PRECONDITION, null)
+        .addDoNotRetry(Status.Code.PERMISSION_DENIED, null)
+        .addDoNotRetry(Status.Code.UNAUTHENTICATED, null)
+        .addDoNotRetry(Status.Code.UNIMPLEMENTED, null)
+        .addDoNotRetry(Status.Code.INTERNAL, QueryFailedFailure.class);
```
vitarb, line 60: I believe `Internal` should be retryable. See https://www.notion.so/temporalio/Retryable-gRPC-error-codes-e1324307c4a745839b6c4a84a27373b8
We may need to check with @alexshtin to confirm.

Spikhalskiy: I don't want to put too many meaningful changes in this PR, because it's heavy in refactoring. I will make this change in a follow-up PR where I will address https://github.com/temporalio/sdk-java/issues/651

temporal-serviceclient/src/main/java/io/temporal/serviceclient/DefaultServiceOperationRpcRetryOptions.java
Spikhalskiy: Thanks for pointing it out! I don't want to put too many meaningful changes in this PR, because it's heavy in refactoring. I will make this change in a follow-up PR.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/DefaultServiceOperationRpcRetryOptions.java
Spikhalskiy: Thanks for pointing it out! I don't want to put too many meaningful changes and parameters tuning in this PR, because it's heavy in refactoring. I will make this change in a follow-up PR.

temporal-serviceclient/src/main/java/io/temporal/internal/retryer/GrpcRetryerUtils.java
Spikhalskiy: I understand where you are coming from, but from readability perspective "ran our of retries" describes much better what happened and what do we actually check.
canRetry or isRetryAllowed can check an exception if it's retryable for example, these names are not very clear.
Renaming it to notRanOutOfRetries and inverting the result doesn't look like a good idea.

temporal-serviceclient/src/main/java/io/temporal/internal/retryer/GrpcSyncRetryer.java
Spikhalskiy: I feel a little bit uneasy about it.
First of all, StatusRuntimeException actually doesn't have a cause at all. It doesn't expose any constructor accepting cause. I think it's done for it being serializable.
Second, I feel that such usage of `cause` can be very misleading for users. The previous exception didn't really cause DeadlineExceeded, it was caused by a deadline timeout. 
I understand what you are coming from, but to do that we probably should make a new class of exceptions. Let's catch up on that and make it in a separate PR if we really want it.

vitarb (approved):
LGTM, let's fork the discussion about numbers into a separate thread.

Issue #514, opened 2021-05-26
DescribeWorkflowExecution is not implemented for the java test service.


drewhoskins-stripe:
## Expected Behavior
Works

## Actual Behavior
`Method temporal.api.workflowservice.v1.WorkflowService/DescribeWorkflowExecution is unimplemented.`

## Steps to Reproduce the Problem
I reproduced this using the out-of-proc test service (https://github.com/temporalio/sdk-java/pull/470/files) which simply exposes the GRPC, but I'm guessing you'd rather reproduce this in process by writing a test.

## Specifications

  - Version:
  - Platform:

drewhoskins-stripe:
I'd be happy to help contribute if somebody could fast-forward me by telling me some of the places I need to edit and where I should put the test (and perhaps similar tests to pattern-match off of.)

PR #670, opened 2021-08-25
Add DescribeWorkflowExecution to TestWorkflowService

nagl-stripe:
## What was changed
This change adds support for DescribeWorkflowExecution to the test service. It should support everything the real server
supports except for AutoResetPoints, which the test service doesn't otherwise support (for good reason).

Implementation-wise, this is just a port of HistoryEngine.DescribeWorkflowExecution from golang, except that we infer some things from history that the real service has in its database.

## Why?
In Stripe, we use the java test service to help people test business logic that starts/interacts-with workflows at build time without using a real temporal server. Several times, we've wanted to call describe in these code paths (e.g. to get diagnostic information about a workflow) but haven't, because it isn't supported in the test service.

## Checklist
1. Closes https://github.com/temporalio/sdk-java/issues/514

2. How was this tested:
There's a new unit test that exercises Describe against workflows in various states.

```
export USE_DOCKER_SERVICE=true
gradlew test # there are failures here, but they happen with or without this commit
export USE_DOCKER_SERVICE=false
gradlew test # no failures before or after this comimt
```

3. Any docs updates needed?
No


temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -2115,6 +2126,209 @@ public void completeQuery(QueryId queryId, RespondQueryTaskCompletedRequest comp
...
+      // We don't track this in the test environment right now, but we could.
+      builder.setLastWorkerIdentity("test-environment-worker-identity");
+    } else {
+      builder.setAttempt(1);
+    }
+
+    return builder.build();
+  }
+
+  private Timestamp timestampFromMillis(long millis) {
```
Spikhalskiy, line 2248: Please use `com.google.protobuf.util.Timestamps#fromMillis` instead

temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -2115,6 +2126,209 @@ public void completeQuery(QueryId queryId, RespondQueryTaskCompletedRequest comp
...
+    List<HistoryEvent> fullHistory =
+        store
+            .getWorkflowExecutionHistory(executionId, getRequest, null)
+            .getHistory()
+            .getEventsList();
+
+    WorkflowExecutionInfo.Builder executionInfo = WorkflowExecutionInfo.newBuilder();
+    executionInfo
+        .setExecution(this.executionId.getExecution())
+        .setType(this.getStartRequest().getWorkflowType())
```
Spikhalskiy, line 2152: There is some mix of styles here. Accessing private fields through this, without this, using getters. It looks like this class is written mostly using the second style, so let's normalize code style in these new methods.

temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -2115,6 +2126,209 @@ public void completeQuery(QueryId queryId, RespondQueryTaskCompletedRequest comp
...
+    if (pendingActivity.lastHeartbeatTime > 0) {
+      // This may overwrite the heartbeat time we just set - that's fine
+      builder.setLastHeartbeatTime(timestampFromMillis(pendingActivity.lastHeartbeatTime));
+
+      if (pendingActivity.heartbeatDetails != null) {
+        builder.setHeartbeatDetails(pendingActivity.heartbeatDetails);
+      }
+    }
+
+    if (scheduledEvent.hasRetryPolicy()) {
```
Spikhalskiy, line 2234: `scheduledEvent. retryPolicy ` and `pendingActivity.retryState.retryPolicy ` are the same thing. If you replace the first one with the second one here - the code will become cleaner and easier to read.
Also after that, this method can be nicely split into two more granular ones: one is publishing from `scheduledEvent` only, another is publishing from `pendingActivity` only.

temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -2115,6 +2126,209 @@ public void completeQuery(QueryId queryId, RespondQueryTaskCompletedRequest comp
...
+      builder.setAttempt(pendingActivity.retryState.getAttempt());
+      builder.setExpirationTime(pendingActivity.retryState.getExpirationTime());
+      builder.setMaximumAttempts(scheduledEvent.getRetryPolicy().getMaximumAttempts());
+      pendingActivity.retryState.getPreviousRunFailure().ifPresent(builder::setLastFailure);
+      // We don't track this in the test environment right now, but we could.
+      builder.setLastWorkerIdentity("test-environment-worker-identity");
+    } else {
+      builder.setAttempt(1);
```
Spikhalskiy, line 2242: If you replace `builder.setAttempt(pendingActivity.retryState.getAttempt());` with `builder.setAttempt(pendingActivity.getAttempt());` and move it from the if statement - you can kill this else completely, because `retryState != null ? retryState.getAttempt() : 1` logic is already located inside `pendingActivity.getAttempt()`

temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -2115,6 +2126,209 @@ public void completeQuery(QueryId queryId, RespondQueryTaskCompletedRequest comp
...
+    if (state == State.CANCELLATION_REQUESTED) {
+      return PendingActivityState.PENDING_ACTIVITY_STATE_CANCEL_REQUESTED;
+    } else if (pendingActivity.startedEvent != null) {
+      return PendingActivityState.PENDING_ACTIVITY_STATE_STARTED;
+    } else {
+      return PendingActivityState.PENDING_ACTIVITY_STATE_SCHEDULED;
+    }
+  }
+
+  private void setTimestampsBasedOnHistory(
```
Spikhalskiy, line 2267: This looks like a setter. Please consider renaming to something like `fillExecutionInfoTimestampsBasedOnHistory` and making this method, `getStartEvent` and `getCompletionEvent` static.

Spikhalskiy (approved):
Thank you for the amazing contribution.
I left some comments regarding readability, code reuse and duplicate, and general code style.
Otherwise LGTM. 
Ping me when the comments are addressed and I will help with merging.

temporal-testing/src/main/java/io/temporal/internal/testservice/StateMachines.java
```diff
@@ -164,18 +164,39 @@
...
+    CONTINUED_AS_NEW(true),
+    TERMINATED(true);
+
+    private final boolean terminal;
+
+    State(boolean terminal) {
+      this.terminal = terminal;
+    }
+
+    public boolean isTerminal() {
```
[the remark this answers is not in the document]
nagl-stripe: Oh, didn't even see that. I'll switch to using it and remove this.

temporal-testing/src/main/java/io/temporal/internal/testservice/StateMachines.java
```diff
@@ -164,18 +164,39 @@
...
+    INITIATED(false),
+    INITIATED_QUERY_ONLY(false),
+    STARTED(false),
+    STARTED_QUERY_ONLY(false),
+    FAILED(true),
+    TIMED_OUT(true),
+    CANCELLATION_REQUESTED(false),
+    CANCELED(true),
+    COMPLETED(true),
+    // Tricky! _this_ run is terminal if it continued-as-new, even though
```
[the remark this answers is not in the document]
nagl-stripe: It keeps the same workflow id per [docs](https://docs.temporal.io/docs/shared/continue-as-new/)

temporal-testing/src/main/java/io/temporal/internal/testservice/StateMachines.java
```diff
@@ -164,18 +164,39 @@
...
+    State(boolean terminal) {
+      this.terminal = terminal;
+    }
+
+    public boolean isTerminal() {
+      return terminal;
+    }
+  }
+
+  public static boolean historyEventIsTerminal(HistoryEvent e) {
```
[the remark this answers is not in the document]
nagl-stripe: Will switch.

temporal-sdk/src/test/java/io/temporal/workflow/shared/DescribeWorkflowAsserter.java
```diff
@@ -0,0 +1,180 @@
...
+import io.temporal.client.WorkflowOptions;
+import io.temporal.internal.common.ProtobufTimeUtils;
+import java.util.Map;
+import java.util.stream.Collectors;
+import org.junit.Assert;
+
+/*
+ * Fluent assertions (ala truth or assert-j) for DescribeWorkflowResults.
+ */
+public class DescribeWorkflowAsserter {
```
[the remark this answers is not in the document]
nagl-stripe: The two libraries mentioned in the comment give you this style for primitives/collections/exceptions. Consider adopting them, particularly if you're tired of people transposing the args to junit's assert.

Truth in particular has some proto-related features, but I find them only mildly useful.

temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
nagl-stripe: Thanks for catching this - I thought I had switched once I discovered it, but clearly not.

temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
nagl-stripe: I think I can clean this up even further - in `StateMachines.scheduleActivityTask`, `.retryState` is always set and the attempt number starts at 1 and gets mutated from there. 

I'm 99% sure retryState is never null and retryPolicy is never null (but may have been in very old schemas, because the golang code makes allowances for those cases). Let's confirm/deny over slack.

nagl-stripe:
So much cleaner now - thank you for the feedback. I think this is ready to go.

PR #678, opened 2021-08-27
Add support for JWT tokens for Temporal server authorization

### LINKED ISSUE #671: Add an easy way to provide JWT tokens to java-sdk for server calls
If we use JWT tokens authorization in Temporal and need to add Authorization header to the calls, the only way now to do that is:
Implement two Grpc Stubs interceptors that set the right header and specify them in WorkflowServiceStubsOptions#Builder#setBlockingStubInterceptor and WorkflowServiceStubsOptions#Builder#setFutureStubInterceptor

It's a painful way for one of the server's default features and we should come up with a better way.
- We can provide JWT specific GrpcInterceptors out of the box 
- We can allow setting up HeaderProviders on WorkflowClientOptions

Related to forum discussion: https://community.temporal.io/t/how-to-do-workflow-api-authentication-using-jwt-in-java-please-help-me-on-this-i-need-required-code-changes-for-client-ad-server-side-for-this/1459/13

Spikhalskiy:
## What was changed
//Draft

## Why?
//Draft

1. Closes #671


Sushisource:
Overall this approach looks totally fine to me as a way to generically attach JWTs -- really it's even just any authorization header. Maybe we should call these classes authorization, and then we could have `JWTTokenSupplier` be an implementation of `AuthorizationSupplier` or something like that, which handles slapping `bearer` on the front of the token. 

I think adding `JWT` into there makes sense if we have a way to supply the JWT payload & secret info so that we actually handle creating the JWT token for them, but that's substantially more work than this, so maybe something for later. Not sure how far we were looking to go here. The easiest-to-use version of this would probably be allowing the user to specify the namespaces / other info you mentioned is in the JWT payload, and their secret (or an interface to access it) and then creating & signing the JWT for them and then passing it into the code you've added here to attach to the header.

Spikhalskiy:
@Sushisource I discussed it with @sergeybykov and the generation/construction of JWT tokens shouldn't be a part of SDKs. Tokens are usually generated and provided by separate services and SDKs should just pass them through.

Sushisource (approved):
Awesome, LGTM

PR #690, opened 2021-08-31
Call health check on intercepted channel

vitarb:
## What was changed
Health check is now getting called on the intercepted channel, which allows attaching custom headers.

## Why?
From what Emin Demirci wrote:
```
It seem that the newly introduced healthchecks are not honoring the configured stub interceptors. In case of server requires an authentication and those authentication tokens retrieved via client interceptors, healtcheck mechanism gets unauthorized request errors. 
My suggestion would be either:
* Exclude the healthcheck endpoint from the list of authenticated endpoints
* Change java-sdk to use the intercepted channels while setting up the healthcheck grpc stub instead of the plain one.
```

### DIFF
--- a/temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsImpl.java
+++ b/temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsImpl.java
@@ -19,7 +19,6 @@
 
 package io.temporal.serviceclient;
 
-import com.google.common.annotations.VisibleForTesting;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import io.grpc.Channel;
 import io.grpc.ClientInterceptor;
@@ -28,7 +27,6 @@ import io.grpc.Deadline;
 import io.grpc.ManagedChannel;
 import io.grpc.Metadata;
 import io.grpc.Server;
-import io.grpc.StatusRuntimeException;
 import io.grpc.health.v1.HealthCheckRequest;
 import io.grpc.health.v1.HealthCheckResponse;
 import io.grpc.health.v1.HealthGrpc;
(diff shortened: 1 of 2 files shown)


temporal-serviceclient/src/test/java/io/temporal/serviceclient/HealthCheckTest.java
```diff
@@ -30,33 +30,17 @@
...
+        // Stub creation triggers health check by default, unless disableHealthCheck flag is set in
+        // the WorkflowServiceStubsOptions.
+        WorkflowServiceStubs.newInstance();
       } catch (Exception e) {
         Assert.fail("Health check failed");
       }
     }
   }
-
-  @Test
```
Spikhalskiy, line 49: Can we save this test though?
Maybe using the `checkHealth(Channel channel, String serviceName)`

temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsImpl.java
```diff
@@ -226,41 +223,24 @@ private ScheduledExecutorService grpcConnectionManager() {
             .build());
   }
 
-  /**
```
Spikhalskiy, line 229: Why the javadoc is gone? It looks pretty applicable to the new code

Spikhalskiy (approved):
with some comments

temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsImpl.java
vitarb: I've inlined a function and IDE apparently has removed the comment, let me put it back.

temporal-serviceclient/src/test/java/io/temporal/serviceclient/HealthCheckTest.java
vitarb: I don't like this ninja parameter passing, I thought that we should either make service name configurable via service stub options, or not have this test at all. And I didn't see a use case for making this parameter configurable, hence I've decided to delete the test. Note that old test was actually doing health check twice (first during init, then explicitly), which was slightly confusing.

PR #698, opened 2021-09-07
Record last worker identity during heartbeat

### LINKED ISSUE #590: Worker identity is not recorded during heartbeat
## Expected Behavior
lastWorkerIdentity is not empty

## Actual Behavior
lastWorkerIdentity is empty

Screen Shot 2021-07-14 at 3 38 59 PM

vkoby:
## What was changed
Record last worker identity during heartbeat

## Why?
It was missing

1. Closes #590 

2. How was this tested:
UI

### DIFF
--- a/temporal-kotlin/src/test/kotlin/io/temporal/workflow/KotlinAsyncCompanionFunctionTest.kt
+++ b/temporal-kotlin/src/test/kotlin/io/temporal/workflow/KotlinAsyncCompanionFunctionTest.kt
@@ -26,7 +26,7 @@ import io.temporal.common.converter.JacksonJsonPayloadConverter
 import io.temporal.common.converter.KotlinObjectMapperFactory
 import io.temporal.internal.async.FunctionWrappingUtil
 import io.temporal.internal.sync.AsyncInternal
-import io.temporal.testing.TestWorkflowRule
+import io.temporal.testing.internal.SDKTestWorkflowRule
 import org.junit.Assert
 import org.junit.Assert.assertTrue
 import org.junit.Rule
@@ -46,7 +46,7 @@ class KotlinAsyncCompanionFunctionTest {
 
   @Rule
   @JvmField
-  var testWorkflowRule: TestWorkflowRule = TestWorkflowRule.newBuilder()
+  var testWorkflowRule: SDKTestWorkflowRule = SDKTestWorkflowRule.newBuilder()
     .setWorkflowTypes(CompanionFunctionReferenceWorkflowImpl::class.java)
     .setWorkflowClientOptions(
       WorkflowClientOptions.newBuilder()
--- a/temporal-kotlin/src/test/kotlin/io/temporal/workflow/KotlinAsyncCompanionFunctionTest.kt
+++ b/temporal-kotlin/src/test/kotlin/io/temporal/workflow/KotlinAsyncCompanionFunctionTest.kt
@@ -27,8 +27,8 @@ import io.temporal.common.converter.KotlinObjectMapperFactory
 import io.temporal.internal.async.FunctionWrappingUtil
 import io.temporal.internal.sync.AsyncInternal
 import io.temporal.testing.TestWorkflowRule
-import junit.framework.Assert.assertTrue
 import org.junit.Assert
+import org.junit.Assert.assertTrue
 import org.junit.Rule
 import org.junit.Test
 import java.util.concurrent.atomic.AtomicBoolean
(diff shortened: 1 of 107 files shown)


temporal-sdk/src/main/java/io/temporal/internal/sync/POJOActivityTaskHandler.java
```diff
@@ -196,7 +190,7 @@ public Result handle(ActivityTask activityTask, Scope metricsScope, boolean loca
     ActivityTaskExecutor activity = activities.get(activityType);
     if (activity == null) {
       if (dynamicActivity != null) {
-        return dynamicActivity.execute(activityInfo, metricsScope);
+        return dynamicActivity.execute(activityInfo, metricsScope, activityTask.getIdentity());
```
Spikhalskiy, line 193: Looking at the structure of this code, `activityTask.getIdentity()` should go inside `ActivityInfoInternal`

Spikhalskiy:
Any reason why there is no unit test here?

temporal-sdk/src/main/java/io/temporal/internal/worker/ActivityTask.java
```diff
@@ -42,4 +45,8 @@ public PollActivityTaskQueueResponse getResponse() {
   public Functions.Proc getCompletionHandle() {
     return completionHandle;
   }
+
+  public String getIdentity() {
```
Spikhalskiy, line 49: This doesn't look like a correct place to pass this data. A worker is already aware of it's identity. It shouldn't be passed as a part of an object having information about a specific activity task.

temporal-sdk/src/main/java/io/temporal/internal/sync/POJOActivityTaskHandler.java
Spikhalskiy: The structure of the data flow is strange. activityTask should have an information about the task coming from the "outside". Worker identity is an internal information of the worker and should be coming from "inside".
I think that identity should be set in a constructor of POJOActivityTaskHandler and be provided by worker initialization code. Like it's done with `interceptors` or `dataConverter`.

Spikhalskiy (changes requested):
Good catch!
But data flow and abstractions/division of responsibility in this PR look a bit off, see in-line comments.

temporal-kotlin/src/test/kotlin/io/temporal/workflow/KotlinAsyncCompanionFunctionTest.kt
```diff
@@ -27,8 +27,8 @@ import io.temporal.common.converter.KotlinObjectMapperFactory
 import io.temporal.internal.async.FunctionWrappingUtil
 import io.temporal.internal.sync.AsyncInternal
 import io.temporal.testing.TestWorkflowRule
-import junit.framework.Assert.assertTrue
 import org.junit.Assert
+import org.junit.Assert.assertTrue
```
vkoby, line 31: Fixes deprecation warnings

temporal-sdk/src/test/java/io/temporal/workflow/DescribeTest.java
```diff
@@ -154,9 +151,6 @@ public void testSuccessfulActivity() throws InterruptedException {
             // Heads up! We're asserting that heartbeat time == started time, which should be true
             // before the heartbeat
             .setLastHeartbeatTime(actual.getLastStartedTime())
-            // this ends up being a dummy value, but if it weren't, we still wouldn't expect to know
-            // it.
-            .setLastWorkerIdentity(actual.getLastWorkerIdentity())
```
vkoby, line 159: Not supposed to have identity here, because the task hasn't been picked up by the worker yet.

temporal-sdk/src/main/java/io/temporal/internal/sync/POJOActivityTaskHandler.java
Spikhalskiy: As noted in the first comment, it's still unclear why identity deserved it's own place inside
```
    ActivityTaskHandler.Result execute(
        ActivityInfoInternal task, Scope metricsScope, String identity);
```

Actually, it doesn't need to be passed at all. See how `POJOActivityImplementation`accesses `namespace` - it can and should access identity the same way, because why not.

temporal-sdk/src/main/java/io/temporal/internal/external/ManualActivityCompletionClientImpl.java
```diff
@@ -202,45 +192,45 @@ public void fail(Throwable exception) {
...
+            .setIdentity(identity);
+    Optional<Payloads> payloads = dataConverter.toPayloads(details);
+    payloads.ifPresent(request::setDetails);
+    RecordActivityTaskHeartbeatResponse status;
+    status =
+        service
+            .blockingStub()
+            .withOption(METRICS_TAGS_CALL_OPTIONS_KEY, metricsScope)
+            .recordActivityTaskHeartbeat(request.build());
+    if (status.getCancelRequested()) {
```
Spikhalskiy, line 216: From what I see, it's better to move this part out from this method to the callers. Because both users catch this exception and process them anyway. They can process the status directly.

Spikhalskiy (approved):

Spikhalskiy (approved):

PR #700, opened 2021-09-07
Fix getResult hang after termination using test service

### LINKED ISSUE #628: In TestWorkflowEnviornment, getResult() hangs after terminate()
## Expected Behavior
Against the real temporal server, calling WorkflowStub.getResult() for a workflow that has been terminated immediately throws a WorkflowFailedException. The test environment should behave the same way.

## Actual Behavior
getResult() hangs indefinitely or until the user-supplied timeout.

  1. git clone git@github.com:nagl-stripe/samples-java.git
  1. Run GreetingWorkflow.java. As checked in, it demonstrates the hanging behavior.
  1. To see the expected behavior, `docker-compose up`, change useReal to true on line 154, and re-run.

## Specifications

  - Version: 1.2.0
  - Platform: Cassandra, but probably doesn't matter.

Spikhalskiy:
Issue #628 

## What was changed
Don't buffer events and wait till the completion of the workflow task if the workflow was terminated.

## Why?
Right now if we get a termination request, the test service doesn't flush it to the history and adds to the buffered events.
Which causes #628

Closes #628

### DIFF
--- a/temporal-sdk/src/test/java/io/temporal/workflow/TerminatedWorkflowTest.java
+++ b/temporal-sdk/src/test/java/io/temporal/workflow/TerminatedWorkflowTest.java
@@ -19,9 +19,10 @@
 
 package io.temporal.workflow;
 
-import io.temporal.client.WorkflowClientOptions;
-import io.temporal.client.WorkflowFailedException;
-import io.temporal.client.WorkflowOptions;
+import static org.junit.Assert.*;
+
+import io.temporal.client.*;
+import io.temporal.failure.TerminatedFailure;
 import io.temporal.testing.internal.SDKTestOptions;
 import io.temporal.testing.internal.SDKTestWorkflowRule;
 import io.temporal.workflow.shared.TestActivities.TestActivitiesImpl;
@@ -30,18 +31,22 @@ import io.temporal.workflow.shared.TestWorkflows.TestTraceWorkflow;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import org.junit.Assert;
 import org.junit.Rule;
 import org.junit.Test;
 
-public class TerminatedWorkflowQueryTest {
-
+/**
+ * Tests verifying the correct behavior of the SDK if the workflow is in unsuccessful final states
+ */
+public class TerminatedWorkflowTest {
   private final TestActivitiesImpl activitiesImpl = new TestActivitiesImpl();
 
   @Rule
   public SDKTestWorkflowRule testWorkflowRule =
       SDKTestWorkflowRule.newBuilder()
-          .setWorkflowTypes(TestTraceWorkflowImpl.class)
+          .setWorkflowTypes(TraceTimingOutWorkflowImpl.class)
           .setActivityImplementations(activitiesImpl)
           .setWorkflowClientOptions(WorkflowClientOptions.newBuilder().build())
           .build();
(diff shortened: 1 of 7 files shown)


temporal-sdk/src/test/java/io/temporal/workflow/TerminatedWorkflowQueryTest.java
```diff
@@ -65,6 +74,58 @@ public void testShouldReturnQueryResultAfterWorkflowTimeout() {
...
 
+  @Test
+  public void testShouldReturnQueryResultAfterTerminatedWorkflow() {
+    // Get a Workflow service client which can be used to start, Signal, and Query Workflow
+    // Executions.
+    TestWorkflowEnvironment testEnv =
+        TestWorkflowEnvironment.newInstance(
+            TestEnvironmentOptions.newBuilder()
+                .setWorkerFactoryOptions(WorkerFactoryOptions.newBuilder().build())
+                .build());
```
mfateev, line 85: I think you either use testWorkflowRule or TestWorkflowEnvironment. This unit test mixes them both which is really weird.

temporal-sdk/src/test/java/io/temporal/workflow/TerminatedWorkflowQueryTest.java
```diff
@@ -65,6 +74,58 @@ public void testShouldReturnQueryResultAfterWorkflowTimeout() {
...
+    // happens.
+    workflow.terminate("testing");
+
+    try {
+      System.out.println("Calling getResult");
+      workflow.getResult(10000, TimeUnit.MILLISECONDS, String.class);
+      fail();
+    } catch (WorkflowFailedException e) {
+      // This is expected
+      System.out.println("getResult failed-fast as expected");
```
mfateev, line 122: I don't think we want to print from unit tests.

Spikhalskiy:
@mfateev not ready for review yet! I will clean it up.

temporal-sdk/src/test/java/io/temporal/workflow/TerminatedWorkflowTest.java
```diff
@@ -53,24 +59,49 @@ public void testShouldReturnQueryResultAfterWorkflowTimeout() {
...
             .build();
-    TestTraceWorkflow client =
+    TestTraceWorkflow workflow =
         testWorkflowRule.getWorkflowClient().newWorkflowStub(TestTraceWorkflow.class, options);
 
     Assert.assertThrows(
         "Workflow should throw because of timeout",
         WorkflowFailedException.class,
-        () -> client.execute(SDKTestWorkflowRule.useExternalService));
+        workflow::execute);
```
Spikhalskiy, line 68: I had to clean this up as a part of this PR, because TraceWorkflow for some reason accepting `useExternalService` as a parameter just doesn't make sense and it's a part of this test class.

vitarb (approved):

temporal-testing/src/main/java/io/temporal/internal/testservice/StateUtils.java
```diff
@@ -0,0 +1,33 @@
...
+ *  an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
+package io.temporal.internal.testservice;
+
+public class StateUtils {
+  /** @return true if the workflow was completed not by workflow task event */
+  public static boolean isWorkflowExecutionTerminatedExternally(StateMachines.State state) {
```
mfateev, line 24: I would rename the function to isWorkflowExecutionForcefullyCompleted. The current name sounds weird for timed out case.

mfateev (approved):

temporal-testing/src/main/java/io/temporal/internal/testservice/StateUtils.java
Spikhalskiy: 👍 I was thinking about naming for some time. Will go with your version.

PR #709, opened 2021-09-11
GrpcRetryer now retries underlying DEADLINE_EXCEEDED if the root gRPC context deadline is not expired

### LINKED ISSUE #708: Regression on GrpcRetryer + GrpcDeadlineInterceptor
Our GrpcDeadlineInterceptor is located on the lowest level, lower GrpcAsyncRetryer and it sets more granular deadlines then root gRPC context. And these requests were previously expected to be retried.
A recent modification of GrpcRetryer to respect DEADLINE_EXCEEDED causes a situation when an underlying request fails with DEADLINE_EXCEEDED and is not being retried anymore, while the root gRPC context is not expired yet.
This may cause 1.3.0 release of java-sdk to throw when the previous release was rightfully retrying.

Spikhalskiy:
## What was changed
`GrpcRetryer` now retries underlying `DEADLINE_EXCEEDED` if the root gRPC context deadline is not expired

## Why?
The behavior of `GrpcRetryer` has been changed in java-sdk#1.3.0 that causes potential regressions when `GrpcRetryer` is used with `DeadlineGrpcInterceptor` like described in #708

Closes #708

### DIFF
--- a/temporal-serviceclient/src/main/java/io/temporal/internal/retryer/GrpcRetryerUtils.java
+++ b/temporal-serviceclient/src/main/java/io/temporal/internal/retryer/GrpcRetryerUtils.java
@@ -62,11 +62,9 @@ class GrpcRetryerUtils {
           // we try to preserve the previous exception if it's present
           return previousException != null ? previousException : currentException;
         } else {
-          // If the deadline of the specific request has been setup on a lower level than
-          // GrpcRetryer and it's
-          // fired, but we are not run out of GrpcRetryer attempts, GrpcRetryer timeout or
-          // GrpcContext deadline,
-          // we should retry.
+          // If this specific request's deadline has expired, but the higher-level deadline
+          // that was established when GrpcRetryer was initialized not been exceeded, we
+          // should keep retrying.
           break;
         }
       default:
--- a/temporal-serviceclient/src/main/java/io/temporal/internal/retryer/GrpcRetryerUtils.java
+++ b/temporal-serviceclient/src/main/java/io/temporal/internal/retryer/GrpcRetryerUtils.java
@@ -19,6 +19,7 @@
 
 package io.temporal.internal.retryer;
 
+import io.grpc.Deadline;
 import io.grpc.Status;
 import io.grpc.StatusRuntimeException;
 import io.temporal.serviceclient.RpcRetryOptions;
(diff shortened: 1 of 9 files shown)


temporal-serviceclient/src/main/java/io/temporal/internal/retryer/GrpcRetryerUtils.java
```diff
@@ -54,7 +56,19 @@
...
+          // If our higher level GRPC context deadline is expired,
+          // the underlying DEADLINE_EXCEEDED is likely meaningless, and
+          // we try to preserve the previous exception if it's present
+          return previousException != null ? previousException : currentException;
+        } else {
+          // If the deadline of the specific request has been setup on a lower level than
+          // GrpcRetryer and it's
+          // fired, but we are not run out of GrpcRetryer attempts, GrpcRetryer timeout or
+          // GrpcContext deadline,
+          // we should retry.
```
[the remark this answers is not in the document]
Sushisource: I the attempts/other timeout part is clear and not relevant in this particular branch

temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsOptions.java
```diff
@@ -445,10 +446,19 @@ public Builder setRpcTimeout(Duration timeout) {
...
+     * seconds (server timeout + additional delay). Default is 70 seconds.
+     *
+     * @throws IllegalArgumentException if {@code timeout} is less than 70s
+     * @deprecated exposing of this option for users configuration deemed non-beneficial and
+     *     dangerous
      */
+    @Deprecated
     public Builder setRpcLongPollTimeout(Duration timeout) {
+      Preconditions.checkArgument(
+          timeout.toMillis() > 70_000, "rpcLongPollTimeout has to be longer 70s");
```
Sushisource, line 461: Since we're already deprecating here I dunno that we need to potentially cause people pain by throwing here if they're already setting it too low.

I can go either way here though so if you want to keep it go for it.

Sushisource (approved):

temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsOptions.java
Spikhalskiy: Setting less than 70s here now is just dangerous. So, I think it's the best we can do here.

mfateev (approved):

PR #719, opened 2021-09-13
User-facing TestWorkflowRule doesn't enforce test timeouts anymore

### LINKED ISSUE #634: Global timeout should be relocated from TestWorkflowRule to SDKTestWorkflowRule
`TestWorkflowRule` that we provide to our users right now enforces 10 seconds timeout by default.
There is a lot of ways provided by JUnit to enforce test timeouts and that can be used by our users if they need this and our test rule shouldn't be responsible for such basic test framework functionality.
Worse is that by providing 10 seconds by default, we create a productivity problem for our users because they need to figure out how to disable this never-asked-about functionality.

We want to have this enabled and enforced across Temporal tests though and I propose to:
- Move existing global timeout implementation into Temporal internal `SDKTestWorkflowRule`.
- Remove logic of TestWorkflowRule that enforces timeout.
- Mark timeout functionality of public TestWorkflowRule as deprecated and remove in the release after the next one.

Spikhalskiy:
## What was changed
TestWorkflowRule#setTestTimeoutSeconds is deprecated and transformed to noop.
Existing timeout logic is moved to SDKTestWorkflowRule and should be used only in internal Temporal tests.
Also moved handy   `newWorkflowStub(Class<T> workflow)` and `newUntypedWorkflowStub(String workflow)` from SDK TestWorkflowRule to TestWorkflowRule, because they can be handy for our users.

## Why?
As described in #634, Temporal Test rule shouldn't reimplement basic unit test framework functionality. And default timeout of 10 seconds could be annoying for users in some cases because users have to find a way to disable it.

1. Closes #634

### DIFF
--- a/temporal-testing-junit4/src/main/java/io/temporal/testing/TestWorkflowRule.java
+++ b/temporal-testing-junit4/src/main/java/io/temporal/testing/TestWorkflowRule.java
@@ -25,6 +25,8 @@ import io.temporal.api.workflowservice.v1.GetWorkflowExecutionHistoryRequest;
 import io.temporal.api.workflowservice.v1.WorkflowServiceGrpc;
 import io.temporal.client.WorkflowClient;
 import io.temporal.client.WorkflowClientOptions;
+import io.temporal.client.WorkflowOptions;
+import io.temporal.client.WorkflowStub;
 import io.temporal.common.interceptors.WorkerInterceptor;
 import io.temporal.internal.common.DebugModeUtils;
 import io.temporal.worker.Worker;
@@ -68,8 +70,6 @@ import org.junit.runners.model.Statement;
  */
 public class TestWorkflowRule implements TestRule {
 
-  private static final long DEFAULT_TEST_TIMEOUT_SECONDS = 10;
-
   private final String namespace;
   private final boolean useExternalService;
   private final boolean doNotStart;
@@ -112,11 +112,8 @@ public class TestWorkflowRule implements TestRule {
             ? WorkflowImplementationOptions.getDefaultInstance()
             : builder.workflowImplementationOptions;
     globalTimeout =
-        !DebugModeUtils.isTemporalDebugModeOn()
-            ? Timeout.seconds(
-                builder.testTimeoutSeconds == 0
-                    ? DEFAULT_TEST_TIMEOUT_SECONDS
-                    : builder.testTimeoutSeconds)
+        !DebugModeUtils.isTemporalDebugModeOn() && builder.testTimeoutSeconds != 0
+            ? Timeout.seconds(builder.testTimeoutSeconds)
             : null;
 
     WorkflowClientOptions clientOptions =
@@ -145,7 +142,6 @@ public class TestWorkflowRule implements TestRule {
     private String target;
     private boolean useExternalService;
     private boolean doNotStart;
-    private long testTimeoutSeconds;
     private long initialTimeMillis;
 
     private Class<?>[] workflowTypes;
@@ -154,6 +150,7 @@ public class TestWorkflowRule implements TestRule {
     private WorkflowClientOptions workflowClientOptions;
     private WorkerFactoryOptions workerFactoryOptions;
     private WorkerOptions workerOptions;
+    private long testTimeoutSeconds;
 
     protected Builder() {}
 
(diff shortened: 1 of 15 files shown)


mfateev:
Could we add a unit test that validates that JUnit way to enforce timeout works with TestWorkflowRule?

Spikhalskiy:
@mfateev wouldn't it be a unit test for junit? Ideally, junit rules should be orthogonal and independent.
Do you just offer to test that

@Test(timeout=1000)
test() {
workflowThatExecutes2000Ms
}

throws?

mfateev:
The reason TestWorkflowRule has this timeout is that it sets the default one to 10 seconds. You removed this timeout and the default. In the majority of cases, Temporal related unit tests don't simply fail but get stuck as it is a complex multithreaded application. So after this change almost any unit test that uses TestWorkflowRule is going to add the timeout to every test or test class.

Spikhalskiy:
@mfateev as they should if users want this behavior. I'm not sure if you was on a standup when I was discussing this change with folks, but the main trigger for it is the fact that users are actually annoyed by this 10 seconds, don't understand why we enforce it and they don't know that they can disable it. It’a especially annoying during debugging.
So, because it's a default functionality of any unit test framework, the idea is that we shouldn’t 1. Reimplement it 2. Providence default timeouts.

Also, from my experience, “stuck inside” is actually a rare scenario for java-sdk. I’m not sure I got it ever during App development other then “stuck inside after several hours of run”.

Spikhalskiy:
The workflow history got printed as before if we use standard JUnit ability to enforce timeout and the test fails/times out.

temporal-testing-junit4/src/main/java/io/temporal/testing/internal/SDKTestWorkflowRule.java
```diff
@@ -78,24 +82,36 @@
...
     }
-    testWorkflowRule = testWorkflowRuleBuilder.build();
+
+    globalTimeout =
+        !DebugModeUtils.isTemporalDebugModeOn()
+            ? Timeout.seconds(
+                builder.testTimeoutSeconds == 0
+                    ? DEFAULT_TEST_TIMEOUT_SECONDS
+                    : builder.testTimeoutSeconds)
+            : null;
```
Sushisource, line 104: nit: Swap the return values here so the condition isn't negative.

temporal-testing-junit4/src/main/java/io/temporal/testing/TestWorkflowRule.java
```diff
@@ -221,9 +208,12 @@ public Builder setTarget(String target) {
       return this;
     }
 
-    /** Global test timeout. Default is 10 seconds. */
+    /**
+     * @deprecated Temporal test rule shouldn't be responsible for enforcing test timeouts. Use
+     *     toolchain of your test framework to enforce timeouts.
```
Sushisource, line 213: Should mention not only is this deprecated, it now does nothing.

temporal-testing-junit4/src/main/java/io/temporal/testing/internal/SDKTestWorkflowRule.java
```diff
@@ -52,18 +53,21 @@
...
+import javax.annotation.Nullable;
+import org.junit.Test;
 import org.junit.rules.TestRule;
+import org.junit.rules.Timeout;
 import org.junit.runner.Description;
 import org.junit.runners.model.Statement;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class SDKTestWorkflowRule implements TestRule {
```
Sushisource, line 67: Docstring about why you would use this over normal TestWorkflowRule would be useful

Sushisource (approved):
This makes sense to me. Ultimately it seems reasonable to depend on existing frameworks for timeouts.

temporal-testing-junit4/src/main/java/io/temporal/testing/internal/SDKTestWorkflowRule.java
Spikhalskiy: It's in internal package, it's for us (SDK devs) only

Sushisource: Right, still useful for new devs taking a look at it I mean.

temporal-testing-junit4/src/main/java/io/temporal/testing/TestWorkflowRule.java
mfateev: What about deprecating it for now, but setting timeout only if the value is set?

Spikhalskiy: I rolled back the removal of the actual implementation. Let's clean it out in 1.5

PR #720, opened 2021-09-14
Move check for isTemporalDebugModeOn down to WorkflowThreadContext#runUntilBlocked to avoid missing checks on the calling side

### LINKED ISSUE #724: TEMPORAL_DEBUG flag is ignored by deadlock detector in eventLoop
Bug report:
https://community.temporal.io/t/testing-temporal-workflows/2918/4

`SyncWorkflow#eventLoop` lacks a check for `TEMPORAL_DEBUG` flag

Spikhalskiy:
## What was changed
Check for isTemporalDebugModeOn was moved down into `WorkflowThreadContext#runUntilBlocked` so it can't be forgotten by the calling code.

## Why?
We have a missing check for `isTemporalDebugModeOn` in `SyncWorkflow#eventLoop`, so the main event loop doesn't overwrite deadline detection timeout when it should. Instead of fixing this one usage, it's a good idea to move the check to the lower level, so the calling code doesn't have to check the flag explicitly every time.

Closes #724

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowThreadContext.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowThreadContext.java
@@ -19,9 +19,8 @@
 
 package io.temporal.internal.sync;
 
-import static io.temporal.internal.sync.DeterministicRunner.getDeadlockDetectionTimeout;
-
 import com.google.common.base.Throwables;
+import io.temporal.internal.common.DebugModeUtils;
 import io.temporal.workflow.Functions;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.Condition;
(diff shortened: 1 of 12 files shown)


temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowThreadContext.java
```diff
@@ -216,11 +215,14 @@ public String getYieldReason() {
   }
 
   /**
+   * @param deadlockDetectionTimeoutMs max duration is ms a workflow code can run without yielding
+   *                                   before it's considered deadlocked.
+   *                                   Discarded if Temporal is configured in debug mode.
```
mfateev: I would state the environment variable name in this comment instead of "in debug mode".

temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowThreadContext.java
```diff
@@ -216,11 +215,14 @@ public String getYieldReason() {
...
    * @return true if thread made some progress. Which is await was unblocked and some code after it
-   *     was executed.
-   * @param deadlockDetectionTimeout
+   *    *     was executed.
    */
-  public boolean runUntilBlocked(long deadlockDetectionTimeout) {
+  public boolean runUntilBlocked(long deadlockDetectionTimeoutMs) {
+    deadlockDetectionTimeoutMs = getDeadlockDetectionTimeoutMs(deadlockDetectionTimeoutMs);
```
mfateev, line 224: I would inline this method as it is very short and called from one place only.

mfateev (commented):
Would you add a unit test that validates that DEBUG_TIMEOUT actually works?

Spikhalskiy:
@mfateev adapted DeadlockDetectorTest to test both debug and non-debug. But it's a bit ugly, overriding of env variables, especially in static util methods/classes is usually ugly.

mfateev (approved):

PR #722, opened 2021-09-14
Removed unnecessary thread interrupt 

### LINKED ISSUE #721: CleanWorkerShutdownTest::testShutdownNow[Docker]
Fails with docker

vkoby:
## What was changed
Removed unnecessary thread interrupt that prevented the sdk from sending the activity task completed call.

## Checklist

1. Closes #721 

2. How was this tested:
Unit test


temporal-sdk/src/test/java/io/temporal/worker/CleanWorkerShutdownTest.java
```diff
@@ -214,141 +98,50 @@ public void testShutdownNow() throws ExecutionException, InterruptedException {
...
+        if (now == null) {
+          shutdownLatch.countDown();
+        } else {
+          shutdownNowLatch.countDown();
+        }
+        Thread.sleep(10000);
       } catch (InterruptedException e) {
-        Thread.currentThread().interrupt();
-        return "interrupted";
+        return INTERRUPTED;
```
Spikhalskiy, line 142: Why? "Before" we have correct handling of an interrupted exception. "After" - is not correct.
If we have to implement handling in an incorrect way to make things work properly - it's a reason to investigate deeper, not replacing correct handling with incorrect.

Spikhalskiy (changes requested):
It looks like this PR cuts off an "interrupted" version of the test instead of fixing anything.

temporal-sdk/src/test/java/io/temporal/worker/CleanWorkerShutdownTest.java
vkoby: Interrupted shutdown is tested by making sure the returned value is INTERRUPTED.
The reason we can't interrupt the activity thread is that it seems to prevent gRPC from making the RespondActivityTaskCompleted call to the external service (the RPC call is getting a CANCELLED status code back because the thread is marked as interrupted). This results in the event EVENT_TYPE_ACTIVITY_TASK_COMPLETED missing from history, thus the failing test.

Spikhalskiy: > Interrupted shutdown is tested by making sure the returned value is INTERRUPTED.

Spikhalskiy: https://github.com/temporalio/sdk-java/issues/731

temporal-sdk/src/test/java/io/temporal/worker/CleanWorkerShutdownHeartBeatingActivityTest.java
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  public static class HeartBeatingActivitiesImpl implements NoArgsReturnsStringActivity {
+    private final CompletableFuture<Boolean> started;
+
+    public HeartBeatingActivitiesImpl(CompletableFuture<Boolean> started) {
+      this.started = started;
+    }
+
+    @Override
```
Spikhalskiy, line 107: This doesn't look right and I'm not sure if it's properly testing what is intended to be tested. It looks like the right implementation will be something like...


    @Override
     public String execute() {
         started.complete(true);

       try {
         Thread.sleep(1000);
            //we ignore interrupted exception here, because the main goal of `testShutdownHeartBeatingActivity` test is to check that heartbeat returns a ActivityWorkerShutdownException
       }
       try {
           Activity.getExecutionContext().heartbeat("foo");
       } catch (ActivityWorkerShutdownException e) {
         return EXPECTED_RESULT;
       }

       return "completed"; //should never happen
     }

temporal-sdk/src/test/java/io/temporal/worker/CleanWorkerShutdownTest.java
Spikhalskiy: Please add a comment why specifically do we swallow interrupted flag here.
You can add a link to #731 explaining that otherwise the result of the activity is not reported.

temporal-sdk/src/test/java/io/temporal/worker/CleanWorkerShutdownTest.java
```diff
@@ -214,141 +98,50 @@ public void testShutdownNow() throws ExecutionException, InterruptedException {
...
     @Override
-    public String execute() {
+    public String execute(String now) {
       try {
-        started.complete(true);
-        Thread.sleep(1000);
-        Activity.getExecutionContext().heartbeat("foo");
-      } catch (ActivityWorkerShutdownException e) {
-        return "workershutdown";
+        if (now == null) {
```
Spikhalskiy, line 135: Please move these latches code block up from the `try{} catch` to improve readability.
I believe `CountDownLatch#countDown` shouldn't throw InterruptedException

vkoby: Moved, but it does not throw the InterruptedException

Spikhalskiy (approved):

Issue #696, opened 2021-09-03
Port fix for following execution chain from https://github.com/temporalio/sdk-go/pull/527


Sushisource:
See details from the Go SDK pull. Fix should be straightforward.

cretz:
Here's the Go change: https://github.com/temporalio/sdk-go/pull/791

PR #730, opened 2021-09-16
Cleanup LongPoll code and update Async version

Spikhalskiy:
## What was changed

Remove unused methods that are full duplicates of used methods
Reuse code and properties between Sync and Async versions to keep them consistent
Async version now calls the long poll with waitForNewEvents = true and updated retry options.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/client/WorkflowClientLongPollHelper.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/client/WorkflowClientLongPollHelper.java
@@ -19,7 +19,6 @@
 
 package io.temporal.internal.client;
 
-import static io.temporal.internal.common.WorkflowExecutionUtils.getCloseStatus;
 import static io.temporal.serviceclient.MetricsTag.HISTORY_LONG_POLL_CALL_OPTIONS_KEY;
 import static io.temporal.serviceclient.MetricsTag.METRICS_TAGS_CALL_OPTIONS_KEY;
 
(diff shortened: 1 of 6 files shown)


temporal-sdk/src/main/java/io/temporal/internal/client/WorkflowClientLongPollHelper.java
```diff
@@ -46,10 +45,11 @@
 import java.util.concurrent.TimeoutException;
 
 /** This class encapsulates sync long poll logic of {@link RootWorkflowClientInvoker} */
-class WorkflowClientLongPollHelper {
+final class WorkflowClientLongPollHelper {
   /**
    * Returns result of a workflow instance execution or throws an exception if workflow did not
-   * complete successfully.
+   * complete successfully. Will wait for continued-as-new generations of the original workflow
```
Sushisource, line 51:
```suggestion
   * complete successfully. Will wait for continued-as-new instances of the original workflow
```

Sushisource: Btw might be worth knocking out https://github.com/temporalio/sdk-java/issues/696 in here

Sushisource (approved):

temporal-sdk/src/main/java/io/temporal/internal/client/WorkflowClientLongPollHelper.java
mfateev: We call them executions. Each execution consists form 1 or many runs.

PR #735, opened 2021-09-17
Cleanup for using external docker in CI and a temporary fix for server namespaces publishing delay

Spikhalskiy:
## What was changed

Implement 10 seconds wait for the availability of just registered test namespace. This hack is needed to make tests work with dockerized service until the namespace caching mechanics is reworked in Temporal server: https://github.com/temporalio/temporal/issues/1941

Add various cleanups for test resources.

Skip `largeHistories` test that requires larger timeouts to work with dockerized Temporal when using dockerized service


temporal-testing/src/main/java/io/temporal/internal/docker/RegisterTestNamespace.java
```diff
@@ -64,6 +90,27 @@ public static void main(String[] args) throws InterruptedException {
...
+
+  private static void waitForNamespace(WorkflowServiceStubs service) throws InterruptedException {
+    // We have a namespace cache on the server side which is refreshed every 10seconds. All
+    // namespace centric APIs
+    // don't use the cache, so the best way is just to wait for it to expire. Another way is to call
+    // some
+    // workflow/activity tasks poll APIs and wait till they stop failing
+    // https://github.com/temporalio/temporal/issues/1941
+    System.out.println("Waiting for the namespace propagation...");
+    Thread.sleep(10000);
```
Sushisource, line 103: I think just calling `doesNamespaceExist` in a loop with a 100ms wait in between or something is a better option than a straight wait here. It can time out after however long.

temporal-testing/src/main/java/io/temporal/internal/docker/RegisterTestNamespace.java
```diff
@@ -64,6 +90,27 @@ public static void main(String[] args) throws InterruptedException {
...
+  }
+
+  private static boolean isNamespaceExist(WorkflowServiceStubs service) {
```
Sushisource, line 106:
```suggestion
  private static boolean doesNamespaceExist(WorkflowServiceStubs service) {
```

temporal-testing/src/main/java/io/temporal/internal/docker/RegisterTestNamespace.java
Spikhalskiy: It doesn't work. It was my original solution. It will return that it exists, because that code path doesn't use a cache.

Sushisource: K, for reference this is what Node does which isn't too bad https://github.com/temporalio/sdk-node/blob/main/scripts/wait-on-temporal.js#L24

Sushisource (approved):

temporal-testing/src/main/java/io/temporal/internal/docker/RegisterTestNamespace.java
Spikhalskiy: Honestly, both ways are hacks and look pretty bad. Because we are not enabling dockerized tests yet, let's leave it like this and just wait for #1941 that actually should be addressed in upcoming weeks most likely.

PR #737, opened 2021-09-17
Make rpc retry log levels configurable

### LINKED ISSUE #618: a load balancer with an idle timeout causes java sdk to log UNAVAILABLE errors
## Expected Behavior

If the Temporal front-end server is behind a load balancer, and that load balancer should close an idle connection that is polling for tasks, if anything is logged it should be at a debug or info level, and the SDK should immediately create a new connection to resume the poll.

## Actual Behavior

sdk-java logs the following exception at level ERROR:

`io.grpc.StatusRuntimeException: UNAVAILABLE: unavailable\n	at io.grpc.stub.ClientCalls.toStatusRuntimeException(ClientCalls.java:262) ~[grpc-stub-1.39.0.jar!/:1.39.0
at io.grpc.stub.ClientCalls.getUnchecked(ClientCalls.java:243) ~[grpc-stub-1.39.0.jar!/:1.39.0
at io.grpc.stub.ClientCalls.blockingUnaryCall(ClientCalls.java:156) ~[grpc-stub-1.39.0.jar!/:1.39.0`

These clutter the logs and the only way to squelch them is to basically turn off all logging from the Poller logger. I have observed no functionality problems with this application, and the Poller appears to immediately resume a poll after this happens, so I don't think there are any functional problems here beyond the log spam.

These steps correspond to the configuration in which I am observing the issue.

  1. Deploy temporal in AWS, in my case it is on ECS using the all-in-one docker image but the deployment method shouldn't matter.
  2. Configure an AWS ALB with http/2 GRPC support enabled pointing at the Temporal front-end endpoint. Set the idle-timeout of the ALB to some specific value in the range of several minutes.
  3. Deploy a java application using sdk-java using the AWS ALB as the target. Create at least one worker.
  4. Observe that polling for tasks logs these exceptions, and the frequency of the exceptions scales inversely with the value of the ALB idle-timeout.

## Specifications

  - Version: 1.1.0
  - Platform: Java 11

Note that I debugged into the exceptions and can observe them coming out of the GRPC layer and bubbling up to the Poller's error handler, which is treating all errors in the same manner. Probably this error could be recognized as harmless and logged at a lower logging level?

vitarb:
## What was changed
New RPC retry option has been added that allows setting log level for RPC errors.

## Why?
Some users may not like defaults.

## Checklist

1. Closes #618

2. How was this tested:
Manually + unit test that verifies that options can be passed without any issues.

3. Any docs updates needed?
No

### DIFF
--- a/temporal-serviceclient/src/main/java/io/temporal/internal/retryer/GrpcAsyncRetryer.java
+++ b/temporal-serviceclient/src/main/java/io/temporal/internal/retryer/GrpcAsyncRetryer.java
@@ -19,6 +19,8 @@
 
 package io.temporal.internal.retryer;
 
+import static io.temporal.internal.retryer.GrpcRetryerUtils.logWithLevel;
+
 import io.grpc.Context;
 import io.grpc.Deadline;
 import io.grpc.StatusRuntimeException;
@@ -154,7 +156,7 @@ class GrpcAsyncRetryer {
       return;
     }
 
-    log.debug("Retrying after failure", currentException);
+    logWithLevel(log, options.getLogLevel(), currentException);
 
     retry(options, function, attempt + 1, startTime, throttler, statusRuntimeException, resultCF);
   }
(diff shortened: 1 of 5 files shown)


Spikhalskiy:
Is this the best way to tackle the problem though?
We lose control of the relative importance of messages that we are logging.
There is another way how we can approach it - start making more meaningful and well-grouped loggers with names. For example "gRPCRetries" logger and users can set up which levels should actually be logged in the logger config.
I think this way will allow us to save control over which messages are considered more or less important while giving users an opportunity to influence the verbosity of the log.

PR #764, opened 2021-09-24
Fix a timeskipping bug in the test service

nagl-stripe:
## What was changed
This fixes a small bug in the test service's timeskipping logic.

## Why?
A test suite we have would reliably trigger this bug because of how the workflows under test would sleep / timeout.

## Checklist
1. Closes (no issue, but I can make one if you want)

2. How was this tested:
There's a new unit test that used to fail but passes now. Also, the test suite within stripe's codebase that lead me here now passes.

3. Any docs updates needed?
No.


Spikhalskiy (approved):

Spikhalskiy (approved):

Spikhalskiy:
This is great, thanks for the contribution!

Issue #648, opened 2021-08-18
Version and Timer State Machines problem after timer cancelation


Spikhalskiy:
## Expected Behavior

Test passes without error

## Actual Behavior

The unit test ails with 

```
Caused by: io.temporal.internal.replay.InternalWorkflowTaskException: Failure handling event 7 of 'EVENT_TYPE_TIMER_STARTED' type. IsReplaying=true, PreviousStartedEventId=3, workflowTaskStartedEventId=20, Currently Processing StartedEventId=3
	at io.temporal.internal.statemachines.WorkflowStateMachines.handleEvent(WorkflowStateMachines.java:188)
	at io.temporal.internal.replay.ReplayWorkflowRunTaskHandler.handleEvent(ReplayWorkflowRunTaskHandler.java:140)
	at io.temporal.internal.replay.ReplayWorkflowRunTaskHandler.handleWorkflowTaskImpl(ReplayWorkflowRunTaskHandler.java:180)
	at io.temporal.internal.replay.ReplayWorkflowRunTaskHandler.handleWorkflowTask(ReplayWorkflowRunTaskHandler.java:150)
	at io.temporal.internal.replay.ReplayWorkflowTaskHandler.handleWorkflowTaskWithEmbeddedQuery(ReplayWorkflowTaskHandler.java:201)
	at io.temporal.internal.replay.ReplayWorkflowTaskHandler.handleWorkflowTask(ReplayWorkflowTaskHandler.java:114)
	at io.temporal.internal.worker.WorkflowWorker$TaskHandlerImpl.handle(WorkflowWorker.java:319)
	at io.temporal.internal.worker.WorkflowWorker$TaskHandlerImpl.handle(WorkflowWorker.java:279)
	at io.temporal.internal.worker.PollTaskExecutor.lambda$process$0(PollTaskExecutor.java:73)
	... 4 common frames omitted
Caused by: java.lang.IllegalArgumentException: Timer: invalid CANCEL_TIMER_COMMAND_SENT->TIMER_STARTED, transition history is [CREATED->SCHEDULE, START_COMMAND_CREATED->START_TIMER, START_COMMAND_CREATED->TIMER_STARTED, START_COMMAND_RECORDED->CANCEL, CANCEL_TIMER_COMMAND_CREATED->CANCEL_TIMER]
	at io.temporal.internal.statemachines.StateMachine.executeTransition(StateMachine.java:130)
[...]

PR #781, opened 2021-09-29
Various debugging utils developed during state machine bug investigation

Spikhalskiy:
These changes are taken from the branch related to investigation and fixes of #648 to keep that branch and future PR clean and easy to follow.
This PR includes various helper logging and functions that could be helpful with state machine debugging.
Including:
- improved states transition logging
- included optional entityName into state machines for concise debug prints
- methods helping with getting a workflow history 
- methods helping with replaying workflow history in tests using TestRule

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/statemachines/CancellableCommand.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/statemachines/CancellableCommand.java
@@ -67,7 +67,7 @@ class CancellableCommand {
   public WorkflowStateMachines.HandleEventStatus handleEvent(
       HistoryEvent event, boolean hasNextEvent) {
     if (canceled) {
-      return WorkflowStateMachines.HandleEventStatus.NON_MATCHING_EVENT;
+      return WorkflowStateMachines.HandleEventStatus.COMMAND_CANCELLED;
     }
     return entityStateMachine.handleEvent(event, hasNextEvent);
   }
--- a/temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
@@ -42,10 +42,7 @@ import io.temporal.api.common.v1.SearchAttributes;
 import io.temporal.api.common.v1.WorkflowExecution;
 import io.temporal.api.enums.v1.EventType;
 import io.temporal.api.failure.v1.Failure;
-import io.temporal.api.history.v1.ActivityTaskScheduledEventAttributes;
-import io.temporal.api.history.v1.HistoryEvent;
-import io.temporal.api.history.v1.MarkerRecordedEventAttributes;
-import io.temporal.api.history.v1.StartChildWorkflowExecutionInitiatedEventAttributes;
+import io.temporal.api.history.v1.*;
 import io.temporal.common.converter.DataConverter;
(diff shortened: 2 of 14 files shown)


temporal-sdk/src/main/java/io/temporal/internal/statemachines/CancellableCommand.java
```diff
@@ -67,7 +67,7 @@ public void handleCommand(CommandType commandType) {
   public WorkflowStateMachines.HandleEventStatus handleEvent(
       HistoryEvent event, boolean hasNextEvent) {
     if (canceled) {
-      return WorkflowStateMachines.HandleEventStatus.NON_MATCHING_EVENT;
+      return WorkflowStateMachines.HandleEventStatus.COMMAND_CANCELLED;
     }
```
Sushisource, line 71: Should this change be in this branch?

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
```diff
@@ -801,10 +802,10 @@ private void assertMatch(Command command, HistoryEvent event, Object expected, O
     public void workflowTaskStarted(
         long startedEventId, long currentTimeMillis, boolean nonProcessedWorkflowTask) {
       setCurrentTimeMillis(currentTimeMillis);
-      // If some new commands are pending and there are no more command events.
       for (CancellableCommand cancellableCommand : commands) {
+        // TODO delete this block
         if (cancellableCommand == null) {
-          break;
+          throw new Error("What the heck");
         }
```
Sushisource, line 809: Looks like this guy should be removed :)

temporal-sdk/src/main/java/io/temporal/internal/statemachines/CancellableCommand.java
Spikhalskiy: No, unrelated, and needed more for clarity and documentation. This change shouldn't change anything.
It's just incorrect to say that the event is not matching when the command is canceled and we never even checked it.
Also, with the current code this path should never actually happen.

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowTaskStateMachine.java
```diff
@@ -52,7 +53,9 @@ void workflowTaskStarted(
   private final long workflowTaskStartedEventId;
   private final Listener listener;
 
-  private long currentTimeMillis;
+  private long eventTimeOfTheLastWorkflowStartTask;
+  // TODO write a comment describing the difference between workflowTaskStartedEventId and
+  // startedEventId
```
Sushisource, line 58: Another TODO probably worth resolving in this PR

Spikhalskiy: I don't want to tackle it right now. I don't know the difference and I don't want to dig into this right now. That why I just throw a TODO for a better time.

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
Spikhalskiy: Yeah, if all tests pass I will throw this away, this thing doesn't make much sense.

Sushisource (approved):

PR #782, opened 2021-09-29
Make TestService ignore heartbeat identity like dockerized server does

### LINKED ISSUE #765: Tests checking worker identity are failing with docker-compose Temporal server
`DescribeTest:testSuccessfulActivity`and `PendingActivityHeartbeatTest::testActivityHeartbeatHasIdentity` fail with dockerized temporal server.
Failure was introduced in #698

last_worker_identity is missing from the result of `describeWorkflow` with dockerized temporal.

Also `recordActivityTaskHeartbeatById` is missing the worker identity added in #698

vkoby:
## What was changed
Added identity to RecordActivityTaskHeartbeatById, commented out part of test that needs to be fixed server side

## Why?
Theres functionality missing on the server side

## Checklist
1. Closes #765 

2. How was this tested:
Unit test


temporal-sdk/src/test/java/io/temporal/workflow/DescribeTest.java
```diff
@@ -175,12 +175,15 @@ public void testSuccessfulActivity() throws InterruptedException {
             .toBuilder()
             .setHeartbeatDetails(DescribeWorkflowAsserter.stringsToPayloads("heartbeatDetails"))
             .setLastHeartbeatTime(actual.getLastHeartbeatTime())
-            .setLastWorkerIdentity(
-                testWorkflowRule
-                    .getTestEnvironment()
-                    .getWorkflowClient()
-                    .getOptions()
-                    .getIdentity())
+            // TODO: Uncomment when the server is fixed. We should always expect to see an identity
```
Spikhalskiy, line 178: But it needs to be higher then? On line 140 where the expected is created for the first check. Because this field should be published even before the heartbeat.
May be worth relocating if you want to keep it in a commented way.

Spikhalskiy (approved):

PR #783, opened 2021-09-29
Handle case where getMemo requests a non-existent key

jeffschoner-stripe:
## What was changed

This change causes `Workflow.getMemo` to return `null` when called with a key that is not present on the memo.

The existing `testMemoInWorkflow` test has been modified to test that a missing key passed to `getMemo` returns `null`. It also extends the test to cover when a memo is queried with a mismatched type.

A new unit test, `testNoMemoInWorkflowFailsGetMemoGracefully`, has been added to check the case where `getMemo` returns null for any key when no memo was specified upon starting a workflow.

## Why?

`Workflow.getMemo` currently throws a `NullPointerException` when an unknown memo key is requested. Because there's not another way to check for a key's existence before calling `getMemo`, the only way to handle this is to catch `NullPointerException`. Checking for null is better for flow control.

## Checklist

2. How was this tested:

`cd temporal-sdk && ../gradlew test --tests io.temporal.workflow.MemoTest`


Spikhalskiy (approved):

Spikhalskiy:
Thank you for the contribution!

PR #798, opened 2021-10-06
WorkflowClientOptions#contextPropagators is now used for Workflow stubs if not overridden on WorkflowOptions

### LINKED ISSUE #796: Context propagation is not invoked when a workflow start and is configured using WorkflowClientOptions
## Expected Behavior
When starting a workflow, context propagated configured at workflow client level must be used.

## Actual Behavior
Is not invoked and must be defined using `WorkflowOptions`

In the following link, I created an example class in which we can see that the "serializeContext" method is not invoked before the workflow is started.

https://gist.github.com/jrpedrianes/2759391fa22311edf214932eeb3fd4b3

## Specifications

  - Version: sdk-java 1.3.1

Spikhalskiy:
## What was changed
WorkflowClientOptions#contextPropagators is now used for Workflow stubs if not overridden on WorkflowOptions.
Javadocs for ContextPropagator is updated with an emphasis that ContextPropagator should be set on both stub and worker sides.
Javadoc for ActivityOptions#contextPropagators was updated to reflect the fact that this method has extremely limited usage and is not intended to be used in a way how the majority of users try to apply it.

Closes #796

How was this tested:
unit tests


temporal-sdk/src/main/java/io/temporal/common/context/ContextPropagator.java
```diff
@@ -31,71 +31,75 @@
...
  * WorkflowClient client = WorkflowClient.newInstance(service, factoryOptions);
+ *
+ * Workflow workflow = client.newWorkflowStub(Workflow.class,
+ *                WorkflowOptions.newBuilder().setTaskQueue("taskQueue").build());
+ * //or
+ * WorkerFactory workerFactory = WorkerFactory.newInstance(client);
  * }</pre>
  *
  * <br>
- * If you want to have override the {@code ContextPropagator} instances for your activities, you can
```
Spikhalskiy, line 86: this section is gone, because it actually pushes users to incorrect understanding and usage of ActivityOptions#contextPropagators

temporal-sdk/src/main/java/io/temporal/activity/ActivityOptions.java
```diff
@@ -152,7 +152,35 @@ public Builder setRetryOptions(RetryOptions retryOptions) {
...
 
-    /** ContextPropagators help propagate the context from the workflow to the activities */
+    /**
+     * Note: <br>
+     * This method has an extremely limited usage.
+     *
+     * <p>Both "client" (workflow worker) and "server" (activity worker) sides of propagation from a
+     * workflow to an activity are located on a worker (and potentially even the same one) side, so
+     * they typically already share the same worker options and specifically {@code
+     * ContextPropagator}s specified on {@link
```
Sushisource, line 162:
```suggestion
     * <p>Both "client" (workflow worker) and "server" (activity worker) sides of context propagation from a
     * workflow to an activity exist in a worker process (potentially the same one), so
     * they typically share the same worker options. Specifically, {@code
     * ContextPropagator}s specified on {@link
```

temporal-sdk/src/main/java/io/temporal/activity/ActivityOptions.java
```diff
@@ -152,7 +152,35 @@ public Builder setRetryOptions(RetryOptions retryOptions) {
...
+     * io.temporal.client.WorkflowClientOptions#getContextPropagators()}. And {@link
+     * io.temporal.client.WorkflowClientOptions#getContextPropagators()} is the right place to
+     * specify {@code ContextPropagator}s between Workflow and an Activity. <br>
```
Sushisource, line 165:
```suggestion
     * io.temporal.client.WorkflowClientOptions#getContextPropagators()}. {@link
     * io.temporal.client.WorkflowClientOptions#getContextPropagators()} is the right place to
     * specify {@code ContextPropagator}s between Workflow and an Activity. <br>
```

temporal-sdk/src/main/java/io/temporal/activity/ActivityOptions.java
```diff
@@ -152,7 +152,35 @@ public Builder setRetryOptions(RetryOptions retryOptions) {
...
+     * Specifying context propagators here overrides them only on a "client" (workflow) side and
+     * can't be automatically promoted to the "server" (activity worker), which always uses {@code
+     * ContextPropagator}s from {@link
+     * io.temporal.client.WorkflowClientOptions#getContextPropagators()} <br>
+     * The only legitimate usecase for this method is probably a situation when the specific
+     * activity is implemented in a different language and in a completely different worker codebase
+     * and in that case setting a {@code ContextPropagator} that is applied only on a "client" side
+     * could make sense. <br>
+     * This is also the reason why there is no same method on {@link LocalActivityOptions}.
```
Sushisource, line 174:
```suggestion
     * This is also why there is no equivalent method on {@link LocalActivityOptions}.
```

PR #813, opened 2021-10-12
Respect WorkflowClientInterceptors in newUntypedWorkflowStub(WorkflowExecution, ...)

nagl-stripe:
## What was changed
This PR contains a bug report in the form of a test case, and also a proposed fix.

**WARNING: This is only a suggested fix. Be paranoid.**

Given some of the `@Deprecated` warnings around these calls, I'm not sure if the fix I made is the right one or not. Maybe the problem is actually somewhere in `TestWorkflowEnvironmentInternal`? Maybe it's in `TimeLockingInterceptor` itself? (Perhaps it should overload `workflowClientCallsInterceptor`?).

## Context
I have a java user at Stripe who is writing tests for a flow that starts a long-lived workflow, stores its wf-id/run-id, and checks up on it later. I think they've found a bug around `TestWorkflowEnvironment`, where timeskipping is enabled or not based on which overload of `newUntypedWorkflowStub` you use.

## Why?
This change makes it so that all 3 overloads of `newUntypedWorkflowStub` wire in the `WorkflowClientInterceptors`. This is particularly important for the `TestWorkflowEnvironment`, which uses an interceptor to enable timeskipping.

## Checklist
1. Closes (...I didn't file a bug, but I'm happy to make one, especially if the fix is more involved)

2. How was this tested:
`./gradlew test` passes, and I added a test to `TestWorkflowEnvironmentSleepTest` that fails if you remove the fix.

3. Any docs updates needed?
No

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/DeterministicRunnerImpl.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/DeterministicRunnerImpl.java
@@ -434,7 +434,9 @@ class DeterministicRunnerImpl implements DeterministicRunner {
   public void executeInWorkflowThread(String name, Runnable runnable) {
     lock.lock();
     try {
-      checkClosed();
+      // if the execution is closed, we will just add the callbacks, but we will never create
+      // threads for them,
+      // so they will be effectively ignored
       toExecuteInWorkflowThread.add(new NamedRunnable(name, runnable));
     } finally {
       lock.unlock();
--- a/temporal-sdk/src/test/java/io/temporal/internal/testing/WorkflowTestingTest.java
+++ b/temporal-sdk/src/test/java/io/temporal/internal/testing/WorkflowTestingTest.java
@@ -29,7 +29,6 @@ import static org.mockito.Mockito.when;
 import io.temporal.activity.Activity;
 import io.temporal.activity.ActivityInterface;
 import io.temporal.activity.ActivityOptions;
-import io.temporal.api.common.v1.WorkflowExecution;
 import io.temporal.api.enums.v1.TimeoutType;
 import io.temporal.api.workflow.v1.WorkflowExecutionInfo;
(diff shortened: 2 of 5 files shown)


Spikhalskiy:
@nagl-stripe `TestWorkflowEnvironmentSleepTest::timeskippingIsAlwaysOn` is failing in buildkite. Let me know if you need details of the failure.

nmeylan:
Hello, after having upgrade from 1.1.0 to 1.6.0 and a long day of debugging, I found out, using git bisect that this PR was leading to one of my test failing.
The test is an integration test where I make an activity been retries 3 times, before succeed.
It fails because I have following code in my test to await the end of the workflow:
```
WorkflowStub workflow = workflowClient.newUntypedWorkflowStub(correlationId, Optional.empty(), Optional.empty() );
try {
	return workflow.getResult(workflowReturnType); // await for workflow to complete
} catch(WorkflowFailedException e) {
	// This exception is thrown where => e.getRetryState().equals(RetryState.RETRY_STATE_TIMEOUT)
}
```
The only way I found to make it pass again, except by removing this commit, was to do add a sleep before the "newUntypedWorkflowStub" where sleep ends after all activities run, i found that useful comment https://github.com/temporalio/sdk-java/blob/v1.5.0/temporal-sdk/src/test/java/io/temporal/workflow/cancellationTests/ActivityCancellationTest.java#L59 
But I don't like this option to put a sleep.
While I understand the goal of this PR, it seems to not work for my use case. Is there a method to remove timeskipping interceptors for 1 case?
Maybe the right question would be: How am I suppose to achieve my goal: await the completion of a workflow with running activities?

Spikhalskiy:
@nmeylan Right now time-skipping is always on by default for tests, yes. And it was this bug that it wasn't for only this one method of creating the stubs. 
Me or @nagl-stripe are going to expose methods that will allow you to disable time-skipping, stay tuned.

nagl-stripe:
See https://github.com/temporalio/sdk-java/pull/937

PR #868, opened 2021-11-12
Adds workflow id and run id into workflows method names

Spikhalskiy:
Adds workflow id and run id into workflows method names
Added logging to trace WorkflowExecutionCache state

This needs to investigate a user report of workflow thread pool getting filled with open workflows which are yielding, but not getting evicted.
Also having workflowId and runId in the workflow thread names can benefit all users who need to look into thread dumps.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/replay/WorkflowExecutorCache.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/replay/WorkflowExecutorCache.java
@@ -36,8 +36,12 @@ import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public final class WorkflowExecutorCache {
+  private final Logger log = LoggerFactory.getLogger(WorkflowExecutorCache.class);
+
   private final Scope metricsScope;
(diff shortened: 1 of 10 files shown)


temporal-sdk/src/main/java/io/temporal/internal/replay/WorkflowExecutorCache.java
```diff
@@ -101,28 +114,50 @@ private WorkflowRunTaskHandler getForProcessing(String runId, Scope metricsScope
...
       for (String key : cache.asMap().keySet()) {
-        if (!key.equals(execution.getRunId()) && !inProcessing.contains(key)) {
+        if (!key.equals(inFavorOfExecution.getRunId()) && !inProcessing.contains(key)) {
           cache.invalidate(key);
           this.metricsScope.gauge(MetricsType.STICKY_CACHE_SIZE).update(size());
           metricsScope.counter(MetricsType.STICKY_CACHE_THREAD_FORCED_EVICTION).inc(1);
+          log.debug(
+              "Workflow Execution {}-{} caused eviction of Workflow Execution with runId {}",
+              inFavorOfExecution.getWorkflowId(),
+              inFavorOfExecution.getRunId(),
```
Sushisource, line 159: Could we override `toString` or whatever is needed here so you can just pass the workflow execution as the only thing to interpolate ?

Spikhalskiy: No, it's a proto instance.

Sushisource (approved):

cretz (approved):

PR #869, opened 2021-11-14
Report WORKFLOW_TASK_FAILED_CAUSE_NON_DETERMINISTIC_ERROR status

### LINKED ISSUE #625: Use `WORKFLOW_TASK_FAILED_CAUSE_NON_DETERMINISTIC_ERROR` for non-deterministic workflow
use `WORKFLOW_TASK_FAILED_CAUSE_NON_DETERMINISTIC_ERROR` as cause for `RespondWorkflowTaskFailed` if user workflow logic is non-deterministic

Ref: https://github.com/temporalio/temporal/pull/1726

Spikhalskiy:
## What was changed
Now if we detect non-determinism during replay, we fail the WFT with `WORKFLOW_TASK_FAILED_CAUSE_NON_DETERMINISTIC_ERROR` instead of "UNKNOWN"
Add new explicit exception `io.temporal.worker.NonDeterministicException` for situations that likely mean non-deterministic change of workflow implementation

Closes #625

How was this tested:
Modified existing non-deterministic unit tests with additional checks verifying the changed behavior


temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowRunTaskHandler.java
```diff
@@ -188,11 +188,6 @@ private void handleWorkflowTaskImpl(PollWorkflowTaskQueueResponseOrBuilder workf
           implementationOptions.getFailWorkflowExceptionTypes();
       for (Class<? extends Throwable> failType : failTypes) {
         if (failType.isAssignableFrom(e.getClass())) {
-          // Wrap any failure into InternalWorkflowTaskException to support specifying them
```
Spikhalskiy, line 191: This piece doesn't make any sense to me and it makes unnecessary misleading wraps in the reported error.
@mfateev could you please check if you see big reasons for this code segment to stay?
InternalWorkflowTaskException is not any special and doesn't have special handling anywhere, so this wrapping is just wrapping.

temporal-sdk/src/main/java/io/temporal/worker/NonDeterministicException.java
```diff
@@ -0,0 +1,38 @@
...
+ *  or in the "license" file accompanying this file. This file is distributed on
+ *  an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
+package io.temporal.worker;
+
+/**
+ * Thrown if history events from the server don't match commands issued by the executing or replay
```
Spikhalskiy, line 23: @Sushisource your recommendation for the Javadoc improvement would be very welcomed ;)

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
```diff
@@ -203,22 +204,23 @@ private void handleEventsBatch(List<HistoryEvent> events, boolean hasNextEvent)
...
-            + ", workflowTaskStartedEventId="
-            + this.workflowTaskStartedEventId
-            + ", Currently Processing StartedEventId="
-            + this.currentStartedEventId,
-        unwrap(e));
+  private RuntimeException createEventProcessingException(RuntimeException e, HistoryEvent event) {
+    Throwable ex = unwrap(e);
+    if (ex instanceof NonDeterministicException) {
+      // just appending the message in front of an existing message, saving the original stacktrace
+      NonDeterministicException modifiedException =
```
cretz, line 211: What if `ex` has a cause? This swallows it.

cretz (approved):

temporal-sdk/src/main/java/io/temporal/worker/NonDeterministicException.java
```diff
@@ -0,0 +1,38 @@
...
+ * of workflow code. <br>
+ * This exception usually means that there is some form of non-determinism in workflow that leads to
+ * deviation between the initial execution history and a replay. When happens during replaying code
+ * for processing of a new Workflow Task, this exception causes the Workflow Task to fail {@link
+ * io.temporal.api.enums.v1.WorkflowTaskFailedCause#WORKFLOW_TASK_FAILED_CAUSE_NON_DETERMINISTIC_ERROR}
+ */
+public class NonDeterministicException extends RuntimeException {
```
cretz, line 30: I think extending `IllegalStateException` would be very reasonable here.

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
Spikhalskiy: It doesn't now, I create all NonDeterministicExdeption without cause, but a good catch! I will fix it.

temporal-sdk/src/main/java/io/temporal/worker/NonDeterministicException.java
```diff
@@ -0,0 +1,38 @@
...
+
+package io.temporal.worker;
+
+/**
+ * Thrown if history events from the server don't match commands issued by the executing or replay
+ * of workflow code. <br>
+ * This exception usually means that there is some form of non-determinism in workflow that leads to
+ * deviation between the initial execution history and a replay. When happens during replaying code
+ * for processing of a new Workflow Task, this exception causes the Workflow Task to fail {@link
+ * io.temporal.api.enums.v1.WorkflowTaskFailedCause#WORKFLOW_TASK_FAILED_CAUSE_NON_DETERMINISTIC_ERROR}
```
Sushisource, line 28:
```suggestion
 * Thrown if history events from the server don't match commands issued by the execution or replay
 * of workflow code. <br>
 * This exception usually means that there is some form of non-determinism in workflow code that has
 * lead to a difference in the execution path taken upon replay when compared to initial execution. 
 * That is to say the history the worker received for this workflow cannot be processed by the current workflow
 * code. If this happens during the replay of a new Workflow Task, this exception will cause the Workflow Task to fail {@link
 * io.temporal.api.enums.v1.WorkflowTaskFailedCause#WORKFLOW_TASK_FAILED_CAUSE_NON_DETERMINISTIC_ERROR}
```

Sushisource:
I think you might want to check the other state machines / base classes for additional exceptions that might need to be converted.

Spikhalskiy:
@Sushisource most state machines should be good I believe, we handle anything that is non-determinism out of them. My only suspect is `MutableSideEffectStateMachine`, but I don't want to touch it right now. I don't know that state machine enough and I expect that it needs a good revisit and more than just changing exception types (yeah, I'm about preloading). So, I will take care of it during my deep revisit/refresh of `MutableSideEffectStateMachine`

PR #877, opened 2021-11-16
Memory/resource leak improvements in TestWorkflowService

nagl-stripe:
## What was changed
This makes minor improvements to resource cleanup on close()

1. If TestWorkflowService gets instantiated with an in-process gRPC server, shut it down during close()
2. Wait for the forkJoinPool to terminate during close()
3. Add a 'no gRPC service at all' mode. This is secretly prep for resetting state via a long-lived API, which will make the test service more usable to other languages.

## Why?
This is a prereq for the external gRPC API that resets the test service state. This API will make it easier to isolate tests when one uses the test service from other languages.

## Checklist
1. Closes [none]

2. How was this tested:
Existing units pass

3. Any docs updates needed?
No.

### DIFF
--- a/temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
+++ b/temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
@@ -130,6 +130,7 @@ public final class TestWorkflowService extends WorkflowServiceGrpc.WorkflowServi
   private final ForkJoinPool forkJoinPool = new ForkJoinPool(4);
   private final Lock lock = new ReentrantLock();
   private final Server outOfProcessServer;
+  private final Server inProcessServer;
   private final TestWorkflowStore store;
   private final Client client;
 
@@ -164,8 +165,18 @@ public final class TestWorkflowService extends WorkflowServiceGrpc.WorkflowServi
     return client.stubs;
   }
 
-  // Creates an in-memory service along with client stubs for use in Java code.
-  // See also createServerOnly.
+  /*
+   * Creates an in-memory service along with client stubs for use in Java code.
+   * See also createServerOnly and createInstanceOnly.
+   */
+  public TestWorkflowService() {
+    this(0);
+  }
+
+  /*
+   * Creates an in-memory service along with client stubs for use in Java code.
+   * See also createServerOnly and createInstanceOnly.
+   */
   public TestWorkflowService(boolean lockTimeSkipping) {
     this(0);
     if (lockTimeSkipping) {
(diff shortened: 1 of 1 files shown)


temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
```diff
@@ -164,79 +165,135 @@ public WorkflowServiceStubs newClientStub() {
...
+    return new TestWorkflowService(0, 0, Mode.InstanceOnly);
+  }
+
+  private enum Mode {
+    InProcess,
+    OutOfProcess,
+    InstanceOnly;
+  }
+
+  private TestWorkflowService(long initialTimeMillis, int port, Mode mode) {
```
nagl-stripe, line 221: I'm curious what y'all think about the private uber-constructor. I didn't want to use another dummy arg (like `isOutOfProc`) to differentiate the instance-only constructor, but if you prefer that pattern, I can switch.

temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
```diff
@@ -222,21 +255,38 @@ private TestWorkflowService(boolean isOutOfProc, int port) {
...
+
+    if (outOfProcessServer != null) {
+      log.info("Shutting down out-of-process GRPC server");
+      outOfProcessServer.shutdown();
+    }
+
+    if (inProcessServer != null) {
+      log.info("Shutting down in-process GRPC server");
+      inProcessServer.shutdown();
+      client.channel.shutdown();
```
Spikhalskiy, line 267: Can we bring `client.shutdown()` back? It also shuts down stubs internally.

temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
```diff
@@ -222,21 +255,38 @@ private TestWorkflowService(boolean isOutOfProc, int port) {
...
+      forkJoinPool.awaitTermination(1, TimeUnit.SECONDS);
+
       if (outOfProcessServer != null) {
         outOfProcessServer.shutdown();
         outOfProcessServer.awaitTermination(1, TimeUnit.SECONDS);
-      } else {
+      }
+
+      if (inProcessServer != null) {
+        inProcessServer.awaitTermination(1, TimeUnit.SECONDS);
```
Spikhalskiy, line 281: Do you mind implementing `client.awaitTermination` that would await on both stubs and channel?

Spikhalskiy (approved):

temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
nagl-stripe: Oof, that was a bad merge from our local copy. Thanks for catching that.

Spikhalskiy:
Thanks for the contribution!

PR #879, opened 2021-11-16
Polish WorkflowThreadContext thread status switching

Spikhalskiy:
WorkflowThreadContext has some code paths that are hard to comprehend, especially with premature switching to DONE and unswitching from it that should never happen. 
This PR polishes the statuses behavior and also simplifies destroy thread logic removing the need to implement it two ways - through a flag check and through a "coroutine" invocation simplifying and unifying threads destruction.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowThreadContext.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowThreadContext.java
@@ -78,25 +78,30 @@ public class WorkflowThreadContext {
     // Evaluates unblockFunction out of the lock to avoid deadlocks.
     lock.lock();
     try {
+      if (destroyRequested) {
+        throw new DestroyWorkflowThreadError();
+      }
+      yieldReason = reason;
+
       // TODO: Verify that calling unblockFunction under the lock is a sane thing to do.
       while (!inRunUntilBlocked || !unblockFunction.get()) {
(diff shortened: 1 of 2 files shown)


temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowThreadContext.java
```diff
@@ -78,25 +78,32 @@ public void yield(String reason, Supplier<Boolean> unblockFunction) {
...
-        }
         status = Status.YIELDED;
         runCondition.signal();
         yieldCondition.await();
+        if (destroyRequested) {
+          throw new DestroyWorkflowThreadError();
+        }
         maybeEvaluateLocked(reason);
-        yieldReason = reason;
       }
```
Sushisource, line 95: Is it impossible for `reason` to change in the loop b/c lock is held? Curious why `yieldReason = reason;` got lifted out of the loop

Spikhalskiy: How? It's a parameter of the method and we don't change it.

Sushisource: Ah I didn't see it was even a param. Makes sense.

Sushisource (approved):

PR #881, opened 2021-11-17
Fix thread leak in legacy query path

### LINKED ISSUE #880: Legacy query path leads to thread leaks in case of state machine exceptions
The following code path leads to an unreleased thread in the case of the worker / state machine exceptions (or RejectedExecutionException):

```
io.temporal.internal.worker.WorkflowExecutionException: java.util.concurrent.RejectedExecutionException: Task java.util.concurrent.FutureTask@4176e14b[Not completed, task = java.util.concurrent.Executors$RunnableAdapter@4beb70[Wrapped task = io.temporal.internal.sync.WorkflowThreadImpl$RunnableWrapper@4747fa1]] rejected from java.util.concurrent.ThreadPoolExecutor@1faac87d[Running, pool size = 600, active threads = 600, queued tasks = 0, completed tasks = 14007]
at io.temporal.internal.sync.POJOWorkflowImplementationFactory.mapToWorkflowExecutionException(POJOWorkflowImplementationFactory.java:359)
at io.temporal.internal.sync.SyncWorkflow.mapUnexpectedException(SyncWorkflow.java:192)
at io.temporal.internal.replay.ReplayWorkflowExecutor.mapUnexpectedException(ReplayWorkflowExecutor.java:152)
at io.temporal.internal.replay.ReplayWorkflowRunTaskHandler.handleWorkflowTaskImpl(ReplayWorkflowRunTaskHandler.java:191)
at io.temporal.internal.replay.ReplayWorkflowRunTaskHandler.handleQueryWorkflowTask(ReplayWorkflowRunTaskHandler.java:246)
at io.temporal.internal.replay.ReplayWorkflowTaskHandler.handleQueryOnlyWorkflowTask(ReplayWorkflowTaskHandler.java:267)
at io.temporal.internal.replay.ReplayWorkflowTaskHandler.handleWorkflowTask(ReplayWorkflowTaskHandler.java:107)
at io.temporal.internal.worker.WorkflowWorker$TaskHandlerImpl.handle(WorkflowWorker.java:314)
at io.temporal.internal.worker.WorkflowWorker$TaskHandlerImpl.handle(WorkflowWorker.java:274)
at io.temporal.internal.worker.PollTaskExecutor.lambda$process$0(PollTaskExecutor.java:73)
at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
at java.base/java.lang.Thread.run(Thread.java:829)
```

Spikhalskiy:
Fix thread leak in legacy query path
Refactor replay functionality into a separate class to keep clear from the main code path and worker code

Closes #880

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/replay/QueryResult.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/replay/QueryResult.java
@@ -1,0 +1,41 @@
+/*
+ *  Copyright (C) 2020 Temporal Technologies, Inc. All Rights Reserved.
+ *
+ *  Copyright 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ *  Modifications copyright (C) 2017 Uber Technologies, Inc.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License"). You may not
+ *  use this file except in compliance with the License. A copy of the License is
+ *  located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ *  or in the "license" file accompanying this file. This file is distributed on
+ *  an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
+package io.temporal.internal.replay;
+
+import io.temporal.api.common.v1.Payloads;
+import java.util.Optional;
+
+public class QueryResult {
+  private final Optional<Payloads> responsePayloads;
+  private final boolean workflowMethodCompleted;
+
+  public QueryResult(Optional<Payloads> responsePayloads, boolean workflowMethodCompleted) {
+    this.responsePayloads = responsePayloads;
+    this.workflowMethodCompleted = workflowMethodCompleted;
+  }
+
+  public Optional<Payloads> getResponsePayloads() {
+    return responsePayloads;
+  }
+
+  public boolean isWorkflowMethodCompleted() {
+    return workflowMethodCompleted;
+  }
+}
--- a/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
@@ -27,7 +27,6 @@ import com.uber.m3.tally.Scope;
 import com.uber.m3.util.ImmutableMap;
 import io.temporal.api.command.v1.Command;
 import io.temporal.api.command.v1.FailWorkflowExecutionCommandAttributes;
-import io.temporal.api.common.v1.Payloads;
 import io.temporal.api.common.v1.WorkflowExecution;
 import io.temporal.api.common.v1.WorkflowType;
 import io.temporal.api.enums.v1.CommandType;
@@ -54,7 +53,6 @@ import java.io.StringWriter;
 import java.time.Duration;
 import java.util.List;
 import java.util.Objects;
-import java.util.Optional;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.BiFunction;
 import org.slf4j.Logger;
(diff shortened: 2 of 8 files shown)


temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
```diff
@@ -101,20 +101,119 @@ public ReplayWorkflowTaskHandler(
...
+          getOrCreateWorkflowExecutor(useCache, workflowTask, metricsScope, createdNew);
+      WorkflowTaskResult result = workflowRunTaskHandler.handleWorkflowTask(workflowTask);
+      if (result.isFinalCommand()) {
+        cache.invalidate(execution, metricsScope, "FinalCommand", null);
+      } else if (useCache && createdNew.get()) {
+        cache.addToCache(execution, workflowRunTaskHandler);
+      }
+      return createCompletedWFTRequest(
+          workflowTask.getWorkflowType().getName(), workflowTask, result);
     } catch (Throwable e) {
```
cretz, line 139: Can any of the things in here throw themselves?

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
```diff
@@ -101,20 +101,119 @@ public ReplayWorkflowTaskHandler(
...
+        // yet and we can avoid doing a reset.
+        if (!isFullHistory(workflowTask)) {
+          resetStickyTaskQueue(execution);
+        }
+      }
+      throw e;
+    } finally {
+      if (useCache) {
+        cache.markProcessingDone(execution);
+      } else if (workflowRunTaskHandler != null) {
```
cretz, line 159: What if I'm using cache _and_ this is not null?

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
```diff
@@ -101,20 +101,119 @@ public ReplayWorkflowTaskHandler(
...
+        workflowRunTaskHandler.close();
```
cretz, line 160: Is this safe for calling multiple times? You may call this in the catch block and again here.

Spikhalskiy: Yeah, it is. I will port a comment here.

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
Spikhalskiy: We don't close, we leave it open in cache, just marked as done.

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
Spikhalskiy: Well, resetStickyTaskQueue actually can, good thinking. I will wrap it into a catch block that ignores it and logs instead.

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
```diff
@@ -168,59 +267,32 @@ private Result failureToResult(PollWorkflowTaskQueueResponse workflowTask, Throw
...
-                return createStatefulHandler(workflowTask, metricsScope);
-              });
-    }
-    return workflowRunTaskHandler;
-  }
-
-  private Result handleWorkflowTaskWithEmbeddedQuery(
-      PollWorkflowTaskQueueResponse.Builder workflowTask, Scope metricsScope) throws Throwable {
+  /** LEGACY QUERY IMPLEMENTATION */
+  private Result handleQueryOnlyWorkflowTask(
```
Sushisource, line 271: Honestly I would change this name. It's really not a real workflow task, given we respond on a totally different API

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
```diff
@@ -230,62 +302,36 @@ private Result handleWorkflowTaskWithEmbeddedQuery(
...
-        log.trace(
-            "Closing runner for {}-{} because of non-sticky worker configuration",
-            execution.getWorkflowId(),
-            execution.getRunId());
-        workflowRunTaskHandler.close();
-      } else {
+      if (useCache) {
         cache.markProcessingDone(execution);
+      } else if (workflowRunTaskHandler != null) {
+        workflowRunTaskHandler.close();
```
Sushisource, line 308: Should this be just `if` rather than `else if`? If we're not using the cache will the handler just be closed because it's disposed of?

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
```diff
@@ -230,62 +302,36 @@ private Result handleWorkflowTaskWithEmbeddedQuery(
...
-        cache.addToCache(execution, workflowRunTaskHandler);
-      }
-      if (queryResult.isPresent()) {
-        queryCompletedRequest.setQueryResult(queryResult.get());
-      }
-      queryCompletedRequest.setCompletedType(QueryResultType.QUERY_RESULT_TYPE_ANSWERED);
-    } catch (Throwable e) {
+
+    if (e != null) {
+      queryCompletedRequest.setCompletedType(QueryResultType.QUERY_RESULT_TYPE_FAILED);
```
Sushisource, line 323: Super nitpicky feel free to ignore: I would invert the conditional here. It's nice to read happy path first.

temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflowWorker.java
```diff
@@ -99,6 +96,22 @@ public SyncWorkflowWorker(
     workflowWorker =
         new WorkflowWorker(
             service, namespace, taskQueue, singleWorkerOptions, taskHandler, stickyTaskQueueName);
+
+    // it's just to support Worker#replayWorkflowExecution functionality
+    // This handler has to be non-sticky to don't evict actual executions from the cache
```
Sushisource, line 101:
```suggestion
    // Exists to support Worker#replayWorkflowExecution functionality.
    // This handler has to be non-sticky to avoid evicting actual executions from the cache
```

temporal-sdk/src/main/java/io/temporal/internal/worker/ReplayerHelper.java
```diff
@@ -0,0 +1,108 @@
...
+import io.temporal.api.history.v1.HistoryEvent;
+import io.temporal.api.history.v1.WorkflowExecutionStartedEventAttributes;
+import io.temporal.api.query.v1.WorkflowQuery;
+import io.temporal.api.workflowservice.v1.PollWorkflowTaskQueueResponse;
+import io.temporal.api.workflowservice.v1.RespondQueryTaskCompletedRequest;
+import io.temporal.internal.common.WorkflowExecutionHistory;
+import java.util.List;
+import java.util.Optional;
+
+// it's just to support Worker#replayWorkflowExecution functionality
```
Sushisource, line 35:
```suggestion
// Supports Worker#replayWorkflowExecution functionality
```

temporal-sdk/src/main/java/io/temporal/internal/worker/ReplayerHelper.java
```diff
@@ -0,0 +1,108 @@
...
+
+/**
+ * Incapsulates query implmentation to support {@link
```
Sushisource, line 38:
```suggestion
 * Encapsulates query implementation to support {@link
```

Sushisource: If this is only in support of query functionality I might name it `QueryReplayHelper`

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
cretz: I had this same question: https://github.com/temporalio/sdk-java/pull/881#discussion_r751277467. Apparently this is by intention.

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
Spikhalskiy: While I initially agreed, after a second thought... we still trigger state machines for the workflow task. I would leave the *WorkflowTask name here for this reason.

Sushisource: Yeah, I started thinking that after I wrote it. Works for me.

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
Spikhalskiy: Actually no, it doesn't throw, we use the Future non-blocking version of the call.

temporal-sdk/src/main/java/io/temporal/internal/replay/QueryResult.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+package io.temporal.internal.replay;
+
+import io.temporal.api.common.v1.Payloads;
+import java.util.Optional;
+
+public class QueryResult {
+  private final Optional<Payloads> responsePayloads;
+  private final boolean finalCommand;
```
Sushisource, line 27: This should maybe me called `wasFinalCommand` or `finalCommandCreated` here? Because the query itself doesn't really constitute a command.

Perhaps the constructor can have a little doc explaining which situations it'd be set true in.

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
```diff
@@ -346,11 +227,114 @@ private Result createCompletedRequest(
...
+
+    return new Result(
+        workflowTask.getWorkflowType().getName(),
+        null,
+        null,
+        queryCompletedRequest.build(),
+        null,
+        false);
+  }
+  /** END OF LEGACY QUERY IMPLEMENTATION */
```
Sushisource, line 312: I think this is unmatched now

temporal-sdk/src/main/java/io/temporal/internal/replay/QueryResult.java
Spikhalskiy: Makes sense, I will rename and comment. It's true if we queried a finished execution.

Sushisource (approved):
Looks nice :+1:

PR #890, opened 2021-11-19
Fix the type of an exception from the test service

nagl-stripe:
## What was changed
This changes a GRPC::Unknown to a GRPC::NotFound, which makes the test service match the behavior of the real thing.

## Why?
The test service should behave similarly to the real thing =)

## Checklist
1. Closes (none)

2. How was this tested:
There's a new unit test that fails without the change.

3. Any docs updates needed?
No.


Spikhalskiy (approved):

Spikhalskiy:
Nice! Thank you.

PR #893, opened 2021-11-20
Fix a bug with Workflow.await against the test service

nagl-stripe:
## What was changed
This **partially** fixes a bug where, if a workflow uses `Workflow.await` against the test service, the test service will timeskip past the workflow's execution timeout.

This PR is a test + half the fix. As written, the test fails in a way that demonstrates the bug. If you uncomment the first half of the fix, it demonstrates that I need help with the second half of the fix. See `REVIEW` comments for details.

## Why?
A team within Stripe is trying to test their workflow against the test service. It uses `.await` to wait for the arrival of one of many signals (ala `select`). Their test is flaky - sometimes the workflow hits its execution timeout before the test can send the signal. The execution timeout is T+21 days, so it must be timeskipping.

## Checklist
1. Closes (none)

2. How was this tested:
There's a new unit test.

3. Any docs updates needed?
No


nagl-stripe:
I'm actually leaning towards closing this without changing behavior. I don't think there is a right answer to "Should timeskipping be on during an await?" - it depends on what the await is waiting for. If it's waiting for timers, the answer is yes. If it's waiting for signals, the answer is no. Timers + signals, the answer is... it depends.

I think the right solution here is for each test case to manually lock/unlock timeskipping per its needs, which is already possible.

Spikhalskiy:
We discussed with @nagl-stripe that we need to have a bit more granular control over what specifically is called from Workflow code. Timer, await, etc. So, this has to be implemented as a WorkflowOutboundInterceptor on a worker side to get access to this level of details. At least with the current Temporal protocol.

PR #899, opened 2021-11-24
Add validation for maximumInterval vs minimumInterval

### LINKED ISSUE #898: RetryOptions allows to be built with maximumInterval < minimumInterval and later fails on the server
## Expected Behavior

RetryOptions#validateAndBuildWithDefaults performs validation and doesn't allow to build invalid RetryOptions.

## Actual Behavior

```
{"level":"info","ts":"2021-10-29T19:20:04.407Z","msg":"Failing the workflow task.","service":"history","shard-id":2700,"address":"10.254.116.207:7234","shard-item":"0xc04c03ee00","component":"history-engine","value":"BadScheduleActivityAttributes: MaximumInterval cannot be less than InitialInterval on retry policy."
```

on the Server.

Spikhalskiy:
## What was changed

Add validation for RetryOptions#maximumInterval vs RetryOptions#minimumInterval

Closes #898

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/common/RetryOptions.java
+++ b/temporal-sdk/src/main/java/io/temporal/common/RetryOptions.java
@@ -112,6 +112,21 @@ public final class RetryOptions {
         .build();
   }
 
+  private void validate() {
+    if (initialInterval == null || initialInterval.isNegative()) {
+      throw new IllegalStateException(
+          "initialInterval has to be a positive value, an actual value " + initialInterval);
+    }
+    if (maximumInterval != null && maximumInterval.compareTo(initialInterval) < 0) {
+      throw new IllegalStateException(
+          "maximumInterval cannot be less than initialInterval if set. "
+              + "maximumInterval is "
+              + maximumInterval
+              + " while initialInterval is "
+              + initialInterval);
+    }
+  }
+
   public static final class Builder {
 
     private static final Duration DEFAULT_INITIAL_INTERVAL = Duration.ofSeconds(1);
@@ -143,7 +158,7 @@ public final class RetryOptions {
     public Builder setInitialInterval(Duration initialInterval) {
       Objects.requireNonNull(initialInterval);
       if (initialInterval.isNegative() || initialInterval.isZero()) {
-        throw new IllegalArgumentException("Invalid interval: " + initialInterval);
+        throw new IllegalArgumentException("Invalid initial interval: " + initialInterval);
       }
       this.initialInterval = initialInterval;
       return this;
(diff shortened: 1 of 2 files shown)


temporal-sdk/src/main/java/io/temporal/common/RetryOptions.java
```diff
@@ -112,6 +112,21 @@ public RetryOptions merge(RetryOptions o) {
...
   }
 
+  private void validate() {
+    if (initialInterval == null || initialInterval.isNegative()) {
+      throw new IllegalStateException(
+          "initialInterval has to be a positive value, an actual value " + initialInterval);
+    }
+    if (maximumInterval != null && maximumInterval.compareTo(initialInterval) < 0) {
+      throw new IllegalStateException(
+          "maximumInterval has to be larger than initialInterval if set. "
```
cretz, line 122: Pedantic - _technically_ "has to be larger" should be "cannot be less"

cretz (approved):

mfateev:
We had such validations in the SDK and later removed them on purpose. There are two issues with doing them in the SDK. First invalid values lead to workflow failures, but we want to fail the workflow task to support fixing bugs without failing workflows. Second, they can easily get out of sync with server side validations.

akdoan1:
In our case, workflow executions were dynamically fetching configuration and building activity stubs (as part of the execution). Because this dynamic binding of activity stubs occurred as a workflow task, rollbacks to poisoned executions were not possible.

Obviously this isn't a great model, but it was the approach used in order to provide different retry values for various environments. We have since refined our approach in a manner that makes our integration more resilient.

--- 
That being said, I completely agree with @mfateev. Having dealt with this sort of issue, the ability to deploy changes to fix in progress workflow executions is incredibly valuable. 

I'd rather challenge the current design of the WorkflowServiceClient and the temporal (history?) service themselves. If there is an invalid configuration for the ActivityTask which prevents the server from scheduling the activity, I don't believe that we should allow the client to continually hammer the server. Also, the client did so pretty silently (although I'm not sure if I'm missing some metric or log, but I haven't found any). 

This sort of integration is prone to noisy tenant issues, and may have large implications for a high number of workflow executions or workflow implementation types.

Spikhalskiy:
@mfateev

>First invalid values lead to workflow failures, but we want to fail the workflow task to support fixing bugs without failing workflows. 
This will cause a RuntimeException, which will cause a failure of a workflow task, not a workflow. So, this property stays intact. There is RuntimeException right now too (INVALID_ARGUMENT gets propagated)
WFT will be just failed by the worker instead of the server. If you think about what's going on right now, it's even trickier - the WFT gets failed by BOTH.

> Second, they can easily get out of sync with server-side validations.

Well, some tricky scenarios can be handled by the server, but 
1. Without verifying it on a client, it's much harder by users to understand the rootcause. Instead of getting a clear exception on a client-side, users have to read server logs or at best workflow history.
2. We already have plenty of validations like this all over the place. Just check RetryOptions Builder code.
3. Let's not go crazy and not try to fully copy server code, but we can safely implement validation for trivial checks. If it gets out of sync with server - what's that much of a problem? It still will be caught on the server.

PR #903, opened 2021-11-28
Expose methods for loading history from files as WorkflowExecutionHistoryLoader

Spikhalskiy:
Now the methods to read history from files exist only in internal package and can be used as a part of WorkflowReplayer only.
To provide better flexibility for our users, if they want to use `worker.replayWorkflowExecution` directly, for example, the methods are moved under `WorkflowExecutionHistoryLoader`class available for testing only.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/common/WorkflowExecutionUtils.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/common/WorkflowExecutionUtils.java
@@ -19,9 +19,6 @@
 
 package io.temporal.internal.common;
 
-import static java.nio.charset.StandardCharsets.UTF_8;
-
-import com.google.common.io.CharStreams;
 import com.google.gson.JsonElement;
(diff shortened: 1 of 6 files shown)


temporal-testing/src/main/java/io/temporal/testing/WorkflowExecutionHistoryLoader.java
```diff
@@ -0,0 +1,56 @@
...
+import java.io.IOException;
+import java.io.Reader;
+import java.net.URL;
+import java.nio.file.Files;
+
+/**
+ * Expose methods to read and deserialize workflow execution history from json.<br>
+ * To be used with {@link WorkflowReplayer}
+ */
+public class WorkflowExecutionHistoryLoader {
```
cretz, line 37: If this is all static utilities, should it have a private constructor?

cretz (approved):

PR #905, opened 2021-11-29
Rearrange WFT handle logging and metrics to accommodate all failure modes

### LINKED ISSUE #900: Worker doesn't log failure during sending workflow task completion request
## Actual Behavior

A worker already logs when Workflow Task Processing fails with an unexpected exception.
But if reporting of the workflow task result to the server fails, the worker log is silent.
At the same time, both situations reach `UncaughtExceptionHandler` in the same manner.

## Expected Behavior

If a worker logs a problem during workflow task processing, the worker should log a problem during reporting the workflow task processing result to the server. Especially because it potentially can be related to misformed commands.

Create an invalid Command (for example, schedule an activity with invalid retry parameters).
`WorkflowWorker.TaskHandlerImpl#handle` gets an `INVALID_ARGUMENT` exception from the server inside `sendReply`.
Worker log is clean and doesn't show the issue anyhow.

## Notes

There is also a related problem that because `WorkflowWorker.TaskHandlerImpl#handle/sendReply` happens outside of WorkflowTaskHandler, exception there doesn't lead to a workflow thread destroy and cache invalidation, while it should. This points out that a larger refactoring here may be needed to keep things clean and easy to maintain.

Spikhalskiy:
## What was changed
WFT handling counters gathered in `WorkflowWorker` instead of being spread between `WorkflowWorker` and `WorkflowTaskHandler`
Added logging in case of successful handling o WFT on the worker, but failing to report to the server.
Cleanup usage of tags and Scopes to remove inconsistent and duplicate tagging.  

Closes #900

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
@@ -39,7 +39,6 @@ import io.temporal.api.workflowservice.v1.*;
 import io.temporal.failure.FailureConverter;
 import io.temporal.internal.common.ProtobufTimeUtils;
 import io.temporal.internal.common.WorkflowExecutionUtils;
-import io.temporal.internal.metrics.MetricsType;
 import io.temporal.internal.worker.LocalActivityWorker;
 import io.temporal.internal.worker.SingleWorkerOptions;
 import io.temporal.internal.worker.WorkflowExecutionException;
@@ -47,7 +46,6 @@ import io.temporal.internal.worker.WorkflowTaskHandler;
 import io.temporal.serviceclient.MetricsTag;
 import io.temporal.serviceclient.WorkflowServiceStubs;
 import io.temporal.worker.NonDeterministicException;
-import io.temporal.workflow.Functions;
 import java.io.PrintWriter;
 import java.io.StringWriter;
 import java.time.Duration;
@@ -67,7 +65,6 @@ public final class ReplayWorkflowTaskHandler implements WorkflowTaskHandler {
   private final WorkflowExecutorCache cache;
   private final SingleWorkerOptions options;
   private final Duration stickyTaskQueueScheduleToStartTimeout;
-  private final Functions.Func<Boolean> shutdownFn;
   private final WorkflowServiceStubs service;
   private final String stickyTaskQueueName;
   private final BiFunction<LocalActivityWorker.Task, Duration, Boolean> localActivityTaskPoller;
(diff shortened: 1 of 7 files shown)


temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
```diff
@@ -247,27 +246,33 @@ private Result failureToWFTResult(
...
+          execution.getRunId(),
+          e);
+    }
+
     // Only fail workflow task on the first attempt, subsequent failures of the same workflow task
     // should timeout. This is to avoid spin on the failed workflow task as the service doesn't
     // yet increase the retry interval.
     if (workflowTask.getAttempt() > 1) {
+      /*
+       *TODO we probably shouldn't swallow Error even if workflowTask.getAttempt() == 1.
```
cretz, line 265: IMO we should only catch `Exception` and not `Throwable` which includes `Error`. The latter signals something is seriously wrong (JVM, disk, mem, etc) and if we really have cases where `Error` is valuable, we should localize the try/catch for them and wrap in an `Exception`. But I understand this is existing behavior.

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
```diff
@@ -247,27 +246,33 @@ private Result failureToWFTResult(
...
+       * But leaving as it is for now, because a trivial change to rethrow
+       * will leave us without reporting Errors as WorkflowTaskFailure to the server,
+       * which we probably should at least attempt to do for visibility that the Error occurs.
+       */
       if (e instanceof Error) {
         throw (Error) e;
       }
       throw (Exception) e;
```
cretz, line 273: Pardon my lack of JVM knowledge. Can you not just `throw e` instead of these 4 lines here (like you do in `catch (Throwable e)` blocks)? Why cast?

cretz (approved):

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
cretz: Nevermind, catch clauses are special: https://stackoverflow.com/questions/23581611/why-is-it-legal-to-re-throw-a-throwable-in-certain-cases-without-declaring-it

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
Spikhalskiy: It's not exactly the case in JavaSDK. We have to utilize errors for some SDK functionality. But I agree that we should cleanup how we treat Errors, that's why the comment. Not going to work on it right now though.

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
Spikhalskiy: It's not because catch are specially really, but the problem that we catch all the Throwables to catch both Errors and Exceptions here and there theoretically can be a new subclass of Throwables that are not allowed to be thrown from a method that declares "throws Exception".

PR #907, opened 2021-11-30
Improve documentation on the Signal method

### LINKED ISSUE #864: Improve documentation for Signal Methods
This request is related to a user request about a workflow execution reset.

When we reset a workflow to the first workflow task in history, signals that are received by the original run are appended right in front of the first WFT of the new run causing signal methods to be called before the workflow method.

We need to explain this in `@SignalMethod` javadoc and Signal documentation page that the workflow code, signal methods, query code should be written in a way that doesn't assume an ordering. Application developers should not assume that signals will be processed only after an initial trigger of the workflow method.

[Needs verification] Query documentation should also probably get the same clarifications because there are situations when a query has to be processed before the first successful WFT.

Related to a discussion under a user request in a community slack: https://temporalio.slack.com/archives/CTT84KXK9/p1636375128042600

Spikhalskiy:
Improve documentation on the Signal method to emphasize the possibility of getting a signal before the first workflow task.

Closes #864

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/workflow/SignalMethod.java
+++ b/temporal-sdk/src/main/java/io/temporal/workflow/SignalMethod.java
@@ -25,8 +25,28 @@ import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
 /**
- * Indicates that the method is a signal handler method. Signal method is executed when workflow
- * receives signal. This annotation applies only to workflow interface methods.
+ * Indicates that the method is a signal handler method. A signal method gets executed when a
+ * workflow receives a signal.
+ *
+ * <p>A Signal can be received and the corresponding signal method can be triggered before execution
+ * of the first workflow task. Workflow and Signal method implementations should be compliant with
+ * this possibility.<br>
+ * Importantly, this happens when a reset of history to the first workflow task (with signal
+ * preservation) is performed. This results in signals being appended to the start of the new
+ * history.<br>
+ * Techniques to consider:
+ *
+ * <ul>
+ *   <li>Workflow object constructors and initialization blocks should be used to initialize the
+ *       internal data structures if possible.
+ *   <li>In rare cases signal processing may require initialization to be performed by the workflow
+ *       code first. An example is initialization that depends on the workflow input parameters. You
+ *       may persist data from the signals received when initialization is incomplete into a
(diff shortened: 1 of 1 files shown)


temporal-sdk/src/main/java/io/temporal/workflow/SignalMethod.java
```diff
@@ -25,8 +25,27 @@
...
- * receives signal. This annotation applies only to workflow interface methods.
+ * Indicates that the method is a signal handler method. A signal method gets executed when a
+ * workflow receives a signal.
+ *
+ * <p>Signal can be received and a corresponded signal method can be triggered before execution of
+ * the first workflow task. Workflow and signal method implementations should be compliant with this
+ * possibility.<br>
+ * It particularly happens when reset of a history to the first workflow task with preserving the
+ * signals is performed. It leads to signals getting appended to the very start of the history.<br>
+ * Techniques to consider:
```
Sushisource, line 36:
```suggestion
 * <p>A Signal can be received and a the corresponding signal method can be triggered before execution of
 * the first workflow task. Workflow and Signal method implementations should be compliant with this
 * possibility.<br>
 * Importantly, this happens when a reset of history to the first workflow task (with signal
 * preservation) is performed. This results in signals being appended to the start of the new history.<br>
 * 
 * Techniques to consider:
```

temporal-sdk/src/main/java/io/temporal/workflow/SignalMethod.java
```diff
@@ -25,8 +25,27 @@
...
+ *
+ * <ul>
+ *   <li>Workflow object constructors and initialization blocks should be used to initialize the
+ *       internal data structures if possible.
+ *   <li>In rare cases signal processing requires some initialization done by the workflow code
+ *       first. An example is initialization that depends on the workflow input parameters. User may
+ *       persist incoming data from the signals received when initialization is not yet done into a
+ *       workflow field. This data can be processed in the workflow method itself after the required
+ *       initialization is performed.
```
Sushisource, line 45:
```suggestion
 *   <li>In rare cases signal processing may require initialization to be performed by the workflow code
 *       first. An example is initialization that depends on the workflow input parameters. You may
 *       persist data from the signals received when initialization is incomplete into a
 *       workflow field. This data can be processed in the workflow method itself after the required
 *       initialization is performed.
```

Sushisource (approved):

PR #911, opened 2021-12-03
[Draft] Add failing test to repro a problem when a workflow chooses to ignore a cancellation.

chronoguy:


cretz:
Saw some debug and commented out code and what not in here, so assuming that since it's just a draft you don't want this reviewed at a deeper level yet?

Spikhalskiy:
According to the current CancellationScope design, once canceled, the scope stays canceled. And the cancellation is happening before the exception is thrown. So, the only way for a workflow to ignore the cancelation is to start a new detached scope. All attempts to call to activities or blocking operations like `Workflow.sleep()` will end up with CancelationException.

PR #917, opened 2021-12-06
Doc describing building test-server into a native binary using GraalVM native-image

Spikhalskiy:
This PR provides some adjustments and documentation about how to use GraalVM native-image tool to build Temporal Java Test Service into a native binary.

### DIFF
--- a/temporal-testing/README.md
+++ b/temporal-testing/README.md
@@ -45,3 +45,24 @@ and `io.temporal.testing.TestActivityExtension` for isolated testing of activiti
 See `io.temporal.testing.TestActivityEnvironment` that provides an easy way for isolated testing of
 activity implementations without needing to provide workflows calling the activities, triggering the workflows
 and bootstrapping a Temporal server or In-memory testing service.
+
+## To build a test service using GraalVM native-image
+
+From the root of the java-sdk repo:
+```
+./gradlew :temporal-testing:copyDependencies
+```
+```
+native-image -jar "./temporal-testing/build/libs/temporal-testing-1.6.0-SNAPSHOT.jar" test-server --class-path "./temporal-testing/build/dependencies/*" \
+--allow-incomplete-classpath --no-fallback
+```
+Additional `native-image` flags to consider:
+1. `--static` to build a fully static binary. _[Not supported](https://developer.apple.com/library/archive/qa/qa1118/_index.html) for MacOS._
+
+### Notes on installing GraalVM
+1. Install [prerequisites](https://www.graalvm.org/reference-manual/native-image/#prerequisites)
+2. Install [sdkman](https://sdkman.io/install)
+3. `sdk install java 21.2.0.r11-grl` GraalVM on JDK 11
+4. `sdk use java 21.2.0.r11-grl`
+5. `gu install native-image`, where `gu` is GraalVM Updater.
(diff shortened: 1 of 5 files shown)


bergundy (approved):

temporal-testing/README.md
```diff
@@ -44,4 +44,28 @@ and `io.temporal.testing.TestActivityExtension` for isolated testing of activiti
...
+./gradlew :temporal-testing:copyDependencies
+```
+```
+native-image -jar "./temporal-testing/build/libs/temporal-testing-1.6.0-SNAPSHOT.jar" test-server --class-path "./temporal-testing/build/dependencies/*" \
+--allow-incomplete-classpath --no-fallback \
+--initialize-at-build-time=org.slf4j.impl.StaticLoggerBinder \
+--initialize-at-build-time=org.slf4j.LoggerFactory \
+--initialize-at-run-time=io.grpc.netty.shaded.io.netty.util.internal.logging.Log4JLogger \
+--initialize-at-run-time=io.grpc.netty.shaded.io.netty.internal.tcnative \
+--initialize-at-run-time=io.grpc.netty.shaded.io.netty.handler.ssl
```
cretz, line 62: Pardon my lack of knowledge of native-image. Can you help me understand the use of `native-image.properties` and manual CLI arguments? What purpose does the properties file serve if these arguments are also here?

Spikhalskiy: Well, users of your library pick them up. So, native-image will inspect the underlying jars for manifests.
I think there is a way to don't have them there and make the properties file to be used. But I wasn't able to make it work just yet.

cretz (approved):

mmcshane (approved):
Works on linux (Ubuntu Focal) but FYI I had to install `libz-dev`

temporal-testing/README.md
```diff
@@ -44,4 +44,23 @@ and `io.temporal.testing.TestActivityExtension` for isolated testing of activiti
 
 See `io.temporal.testing.TestActivityEnvironment` that provides an easy way for isolated testing of
 activity implementations without needing to provide workflows calling the activities, triggering the workflows
-and bootstrapping a Temporal server or In-memory testing service.
\ No newline at end of file
+and bootstrapping a Temporal server or In-memory testing service.
+
+## To build a test service using GraalVM native-image
```
cretz, line 49: By the way, I just found https://graalvm.github.io/native-build-tools/latest/gradle-plugin.html if you wanted to wrap it up

Spikhalskiy: Thanks! We can prettify this later from a building point of view when we are sure that it will work for us at all. For now CLI is easier to make adjustments.

temporal-testing/README.md
```diff
@@ -44,4 +44,23 @@ and `io.temporal.testing.TestActivityExtension` for isolated testing of activiti
...
+
+From the root of the java-sdk repo:
+```
+./gradlew :temporal-testing:copyDependencies
+```
+```
+native-image -jar "./temporal-testing/build/libs/temporal-testing-1.6.0-SNAPSHOT.jar" test-server --class-path "./temporal-testing/build/dependencies/*" \
```
cretz, line 56: At least for the ones we might distribute, we should use `--static` here and potentially `--libc=musl` per https://www.graalvm.org/reference-manual/native-image/StaticImages.

Issue #929, opened 2021-12-15
Can't register activity implementation that uses inheritance


aniketbhatnagar:
## Expected Behavior
Consider the following:
```
@ActivityInterface
public interface BlockListenerActivities {
     @ActivityMethod
     Optional<BlockRef> processNextBlock();
}

@Slf4j
public abstract class BaseBlockListenerActivities implements BlockListenerActivities {
    @Override
    public final Optional<BlockRef> processNextBlock() {
        return doProcessNextBlock();
    }

    abstract protected Optional<BlockRef> doProcessNextBlock();
}

@Slf4j
public class BlockListenerActivitiesImpl extends BaseBlockListenerActivities {
    @Override
    protected Optional<BlockRef> doProcessNextBlock() {
        log.info("Processing next ETH block");
        return Optional.empty();
    }
}
```

Then, registering the activity implementation (`worker.registerActivitiesImplementations(new BlockListenerActivitiesImpl());`) should work.

## Actual Behavior
Registering the activity implementation (`worker.registerActivitiesImplementations(new BlockListenerActivitiesImpl());`) causes the exception `Class doesn't implement any non empty interface annotated with @ActivityInterface: BlockListenerActivitiesImpl`.

It seems that that SDK expects `BlockListenerActivitiesImpl` to explicitly say `implements BlockListenerActivities`. 

I ran into this is because I was provided a "quarkus" injected bean that automatically extends the activity implementation class to form a proxy like `ActivityImpl_ClientProxy extends ActivityImpl` and temporal SDK rejects `ActivityImpl_ClientProxy` as a valid activity implementation.

## Steps to Reproduce the Problem
Use the code snippet above and register activity using `worker.registerActivitiesImplementations(new BlockListenerActivitiesImpl());`

## Specifications

  - Version: temporal-sdk 1.6.0
  - Platform: Reproduced on Mac with jdk 14 but should be applicable on all platforms.

tsurdilo:
Community Forum ref: https://community.temporal.io/t/cant-register-activity-implementation-that-uses-inheritance/3603

PR #930, opened 2021-12-15
Fix POJOActivityImplMetadata not being able to handle activity impl that uses inheritance

aniketbhatnagar:
## What was changed
`POJOActivityImplMetadata` now uses guava to find all activity interfaces for a given activity impl class. Inspiration - https://stackoverflow.com/questions/6616055/get-all-derived-interfaces-of-a-class

## Why?
https://github.com/temporalio/sdk-java/issues/929

## Checklist

1. Closes https://github.com/temporalio/sdk-java/issues/929

2. How was this tested?
Added a test. All existing tests pass.

3. Any docs updates needed? 
I don't think this needs any doc updates.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/POJOActivityTaskHandler.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/POJOActivityTaskHandler.java
@@ -39,7 +39,6 @@ import io.temporal.common.interceptors.ActivityInboundCallsInterceptor;
 import io.temporal.common.interceptors.Header;
 import io.temporal.common.interceptors.WorkerInterceptor;
 import io.temporal.common.metadata.POJOActivityImplMetadata;
-import io.temporal.common.metadata.POJOActivityInterfaceMetadata;
 import io.temporal.common.metadata.POJOActivityMethodMetadata;
 import io.temporal.failure.FailureConverter;
 import io.temporal.failure.SimulatedTimeoutFailure;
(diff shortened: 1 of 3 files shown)


temporal-sdk/src/main/java/io/temporal/internal/sync/POJOActivityTaskHandler.java
```diff
@@ -110,18 +109,15 @@ private void registerActivityImplementation(
...
-      for (POJOActivityMethodMetadata activityMetadata : activityInterface.getMethodsMetadata()) {
-        String typeName = activityMetadata.getActivityTypeName();
-        if (activities.containsKey(typeName)) {
-          throw new IllegalArgumentException(
-              "\"" + typeName + "\" activity type is already registered with the worker");
-        }
-        Method method = activityMetadata.getMethod();
-        ActivityTaskExecutor implementation = newTaskExecutor.apply(method, activity);
-        activities.put(typeName, implementation);
+    for (POJOActivityMethodMetadata activityMetadata : activityImplMetadata.getActivityMethods()) {
```
aniketbhatnagar, line 112: I had to make this change because otherwise if inheritance exists in activity interfaces, we end up getting duplication method metadata causing IllegalArgumentException (https://buildkite.com/temporal/java-sdk-public/builds/246#84704748-21be-4519-981e-d8cfe8c91e2e). `activityImplMetadata.getActivityMethods()` returns de-duped methods

Spikhalskiy (approved):
LGTM. Thanks for the great fix.

PR #941, opened 2021-12-28
Support for PKCS12 keys in SimpleSslContextBuilder

### LINKED ISSUE #925: SimpleSslContextBuilder works only with PKCS8 while all Temporal TSL examples and manuals are with PKCS12
`SimpleSslContextBuilder` uses and exposes working with PKCS8 only.
Meanwhile, all Temporal server docs and examples like https://github.com/temporalio/samples-server/tree/main/tls/client-only are working with PKCS12 only.
This inconsistency creates unneeded troubles for users in understanding how to put pieces together correctly.
We need:
1. Add code to `SimpleSslContextBuilder` supporting PKCS12 certificates. Users shouldn't have to convert certificates to PKCS8 for JavaSDK only.
2. We should improve javadocs with some more direct usage examples.

Spikhalskiy:
## What was changed
Added support for PKCS12 keys in `SimpleSslContextBuilder`.

## Why
Before this change, JavaSDK supports only PKCS8 keys.
Most OpenSSL examples lead to the generation of PKCS12 keys, including our own Temporal Server samples: 
https://github.com/temporalio/samples-server/tree/main/tls
This leads to an additional conversion step just for JavaSDK in the already convoluted keys generation process which makes it hard to manage for users.

How was this tested:
Unit test checking that netty's SslContext gets initialized with a sample PKCS12 archive (.pfx) can be loaded without exception

Closes #925


temporal-serviceclient/src/main/java/io/temporal/serviceclient/SimpleSslContextBuilder.java
```diff
@@ -47,21 +44,49 @@
...
-    return new SimpleSslContextBuilder(keyCertChain, key);
+    return forPKCS8(keyCertChain, key);
   }
 
-  private SimpleSslContextBuilder(InputStream keyCertChain, InputStream key) {
+  /**
+   * @param keyCertChain - an input stream for an X.509 client certificate chain in PEM format.
+   * @param key - an input stream for a PKCS#8 client private key in PEM format.
+   */
+  public static SimpleSslContextBuilder forPKCS8(InputStream keyCertChain, InputStream key) {
```
bergundy, line 73: It would have been nice if we could help users out and accept both pkcs8 and 12.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/SimpleSslContextBuilder.java
```diff
@@ -47,21 +44,49 @@
...
+  }
+
+  /**
+   * Note: <a href="https://github.com/temporalio/samples-server/tree/main/tls">Temporal examples
+   * for client certificates generation</a> lead to generation of Certificate Private Keys in PKCS12
+   * format.
+   *
+   * @param pfxKeyArchive - an input stream for .pfx or .p12 PKCS12 archive file
+   */
+  public static SimpleSslContextBuilder forPKCS12(InputStream pfxKeyArchive) {
```
bergundy, line 84: Can pkcs12 come with a keyCertChain?

bergundy (approved):
Commented, nothing blocking

temporal-serviceclient/src/main/java/io/temporal/serviceclient/SimpleSslContextBuilder.java
Spikhalskiy: It can and is, but it's packaged into PFX file in this case.
This is the tricky part and why I have to do two separate methods. PKCS12 has a packaging format - PFX. And Java supports PKCS12 only packaged into PFX. You can't pass a private key and a chain separately.
PKCS8 is supported by Netty directly and there is no packaging format - so there are just two raw PEM files. One for the private key and one for cert chain.

I can try to add a separate kinda magic method that will accept just a vararg (to accommodate formats which bundle and which is not) and makes an attempt to figure things out inside, but this is a separate thing and it will not be easy..

PR #942, opened 2021-12-30
Add metadata.messageType to protobuf payloads

lorensr:
## What was changed

In `ProtobufPayloadConverter` and `ProtobufJSONPayloadConverter`, I added a second `metadata` attribute, `mesageType`, with the full proto message name:

```
metadata: 
  encoding: binary("json/protobuf"),
  messageType: binary("foo.bar.MyMessage")
data: ...
```

## Why?

1. Support decoding protobuf payloads in languages that don't have the ability to get the types of a given function's parameters at runtime (see TS SDK [PR](https://github.com/temporalio/sdk-typescript/pull/430) and [design doc](https://github.com/lorensr/sdk-node/blob/binary/protobuf/docs/protobuf-libraries.md)).
2. Support decoding binary protobuf payloads in the UI (https://github.com/temporalio/ui-server/issues/79).

## Checklist

2. How was this tested:

Added `ProtoPayloadConverterTest.testProtoMessageType()` to check that an execution payload has the correct `messageType`.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/common/converter/AbstractProtobufPayloadConverter.java
+++ b/temporal-sdk/src/main/java/io/temporal/common/converter/AbstractProtobufPayloadConverter.java
@@ -1,0 +1,79 @@
+/*
+ *  Copyright (C) 2020 Temporal Technologies, Inc. All Rights Reserved.
+ *
+ *  Copyright 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ *  Modifications copyright (C) 2017 Uber Technologies, Inc.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License"). You may not
+ *  use this file except in compliance with the License. A copy of the License is
+ *  located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ *  or in the "license" file accompanying this file. This file is distributed on
+ *  an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
+package io.temporal.common.converter;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+import com.google.protobuf.ByteString;
+import com.google.protobuf.Descriptors;
+import com.google.protobuf.MessageOrBuilder;
+import io.temporal.api.common.v1.Payload;
+import java.lang.reflect.Method;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public abstract class AbstractProtobufPayloadConverter {
+  protected static final Logger log =
+      LoggerFactory.getLogger(AbstractProtobufPayloadConverter.class);
+
+  private final boolean excludeProtobufMessageTypes;
+
+  protected AbstractProtobufPayloadConverter() {
+    this.excludeProtobufMessageTypes = false;
+  }
+
+  protected AbstractProtobufPayloadConverter(boolean excludeProtobufMessageTypes) {
+    this.excludeProtobufMessageTypes = excludeProtobufMessageTypes;
+  }
+
+  protected void addMessageType(Payload.Builder builder, Object value) {
+    if (this.excludeProtobufMessageTypes) {
+      return;
+    }
+    String messageTypeName = ((MessageOrBuilder) value).getDescriptorForType().getFullName();
+    builder.putMetadata(
+        EncodingKeys.METADATA_MESSAGE_TYPE_KEY, ByteString.copyFrom(messageTypeName, UTF_8));
+  }
+
+  protected void checkMessageType(Payload payload, Class valueClass) throws DataConverterException {
+    ByteString messageTypeBytes =
+        payload.getMetadataMap().get(EncodingKeys.METADATA_MESSAGE_TYPE_KEY);
+    if (messageTypeBytes != null) {
+      try {
+        String messageType = messageTypeBytes.toString(UTF_8);
+        Method getDescriptor = valueClass.getMethod("getDescriptor");
+        String valueMessageType =
+            ((Descriptors.Descriptor) getDescriptor.invoke(null)).getFullName();
+        if (!messageType.equals(valueMessageType) && log.isWarnEnabled()) {
+          log.warn(
+              "Encoded protobuf message type \""
+                  + messageType
+                  + "\" does not match valueClass `"
+                  + valueClass.getName()
+                  + "`, which has message type \""
+                  + valueMessageType
+                  + '"');
+        }
+      } catch (Exception e) {
+        throw new DataConverterException(e);
+      }
+    }
+  }
+}
--- a/temporal-sdk/src/main/java/io/temporal/common/converter/AbstractProtobufPayloadConverter.java
+++ b/temporal-sdk/src/main/java/io/temporal/common/converter/AbstractProtobufPayloadConverter.java
@@ -22,10 +22,8 @@ package io.temporal.common.converter;
 import static java.nio.charset.StandardCharsets.UTF_8;
 
 import com.google.protobuf.ByteString;
-import com.google.protobuf.Descriptors;
 import com.google.protobuf.MessageOrBuilder;
 import io.temporal.api.common.v1.Payload;
-import java.lang.reflect.Method;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -52,27 +50,24 @@ public abstract class AbstractProtobufPayloadConverter {
         EncodingKeys.METADATA_MESSAGE_TYPE_KEY, ByteString.copyFrom(messageTypeName, UTF_8));
   }
 
-  protected void checkMessageType(Payload payload, Class valueClass) throws DataConverterException {
+  protected void checkMessageType(Payload payload, Object instance) {
+    if (!log.isWarnEnabled()) {
+      return;
+    }
+
     ByteString messageTypeBytes =
         payload.getMetadataMap().get(EncodingKeys.METADATA_MESSAGE_TYPE_KEY);
     if (messageTypeBytes != null) {
-      try {
-        String messageType = messageTypeBytes.toString(UTF_8);
-        Method getDescriptor = valueClass.getMethod("getDescriptor");
-        String valueMessageType =
-            ((Descriptors.Descriptor) getDescriptor.invoke(null)).getFullName();
-        if (!messageType.equals(valueMessageType) && log.isWarnEnabled()) {
-          log.warn(
-              "Encoded protobuf message type \""
-                  + messageType
-                  + "\" does not match valueClass `"
-                  + valueClass.getName()
-                  + "`, which has message type \""
-                  + valueMessageType
-                  + '"');
-        }
-      } catch (Exception e) {
-        throw new DataConverterException(e);
+      String messageType = messageTypeBytes.toString(UTF_8);
+      String instanceType = ((MessageOrBuilder) instance).getDescriptorForType().getFullName();
+
+      if (!messageType.equals(instanceType)) {
+        log.warn(
+            "Encoded protobuf message type \""
+                + messageType
+                + "\" does not match value type \""
+                + instanceType
+                + '"');
       }
     }
   }
--- a/temporal-sdk/src/main/java/io/temporal/common/converter/DefaultDataConverter.java
+++ b/temporal-sdk/src/main/java/io/temporal/common/converter/DefaultDataConverter.java
@@ -40,26 +40,23 @@ import java.util.concurrent.atomic.AtomicReference;
  */
 public class DefaultDataConverter implements DataConverter {
 
+  // Order is important as the first converter that can convert the payload is used. Needs to match
+  // the other SDKs. Go SDK:
+  // https://github.com/temporalio/sdk-go/blob/5e5645f0c550dcf717c095ae32c76a7087d2e985/converter/default_data_converter.go#L28
+  private static final PayloadConverter[] DEFAULT_PAYLOAD_CONVERTERS = {
+    new NullPayloadConverter(),
+    new ByteArrayPayloadConverter(),
+    new ProtobufJsonPayloadConverter(),
+    new ProtobufPayloadConverter(),
+    new JacksonJsonPayloadConverter()
+  };
+
   private static final AtomicReference<DataConverter> defaultDataConverterInstance =
       new AtomicReference<>(newDefaultInstance());
 
   private final Map<String, PayloadConverter> converterMap = new ConcurrentHashMap<>();
 
   private final List<PayloadConverter> converters = new ArrayList<>();
-
-  // Order is important as the first converter that can convert the payload is used. Needs to match
-  // the other SDKs. Go SDK:
-  // https://github.com/temporalio/sdk-go/blob/5e5645f0c550dcf717c095ae32c76a7087d2e985/converter/default_data_converter.go#L28
-  private static final PayloadConverter[] getDefaultPayloadConverters(
-      boolean excludeProtobufMessageTypes) {
-    return new PayloadConverter[] {
-      new NullPayloadConverter(),
-      new ByteArrayPayloadConverter(),
-      new ProtobufJsonPayloadConverter(excludeProtobufMessageTypes),
-      new ProtobufPayloadConverter(excludeProtobufMessageTypes),
-      new JacksonJsonPayloadConverter()
-    };
-  }
 
   static DataConverter getDefaultInstance() {
     return defaultDataConverterInstance.get();
@@ -80,11 +77,7 @@ public class DefaultDataConverter implements DataConverter {
    * payload converters.
    */
   public static DefaultDataConverter newDefaultInstance() {
-    return new DefaultDataConverter(getDefaultPayloadConverters(false));
-  }
-
-  public static DefaultDataConverter newDefaultInstance(boolean excludeProtobufMessageTypes) {
-    return new DefaultDataConverter(getDefaultPayloadConverters(excludeProtobufMessageTypes));
+    return new DefaultDataConverter(DEFAULT_PAYLOAD_CONVERTERS);
   }
 
   /**
--- a/temporal-sdk/src/main/java/io/temporal/common/converter/DefaultDataConverter.java
+++ b/temporal-sdk/src/main/java/io/temporal/common/converter/DefaultDataConverter.java
@@ -40,20 +40,26 @@ import java.util.concurrent.atomic.AtomicReference;
  */
 public class DefaultDataConverter implements DataConverter {
 
-  // Order is important as the first converter that can convert the payload is used
-  private static final PayloadConverter[] DEFAULT_PAYLOAD_CONVERTERS = {
-    new NullPayloadConverter(),
-    new ByteArrayPayloadConverter(),
-    new ProtobufJsonPayloadConverter(),
-    new JacksonJsonPayloadConverter()
-  };
-
   private static final AtomicReference<DataConverter> defaultDataConverterInstance =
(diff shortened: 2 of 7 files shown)


temporal-sdk/src/test/java/io/temporal/common/converter/ProtoPayloadConverterTest.java
```diff
@@ -96,6 +102,34 @@ public void testCustomProto() {
...
+      if (metadata instanceof List && ((List) metadata).get(1) instanceof MapEntry) {
+        MapEntry<?, ?> secondMetadata = (MapEntry<?, ?>) ((List<?>) metadata).get(1);
+
+        assertEquals("messageType", secondMetadata.getKey());
+        assertEquals(
+            "temporal.api.common.v1.WorkflowExecution",
+            ((ByteString) secondMetadata.getValue()).toString(StandardCharsets.UTF_8));
+        return;
+      }
+    }
```
lorensr, line 129: This would be shorter with `@SuppressWarnings("unchecked")` instead of checking `instanceof`—not sure which is preferred.

Spikhalskiy:
Thanks for your submission, Loren!

Right now at a first glance, I'm kinda against this design. But keep in mind that I didn't participate in the initial discussions and may be out of the loop.

If TS can't infer a type of parameters in runtime, TS should have a way to explicitly pass the type as an additional parameter during the deserialization and not require the whole stack and other languages to pass additional metadata fields for TS deserialization only. So, if in Java we infer `Class<T> valueClass` from the parameter, TS should have a way for a user to explicitly pass the parameter. 
This way we bring TS in line with other languages by making it know the type in runtime. Which brings TS semantically closer to the behavior of other SDKs. Workflow/activity code should dectate what type to deserialize, not the data or metadata in the payload.
This is also the conventional way of dealing with such issues from my experience. For example, how Jackson knows what type of object to restore from the Json? It accepts it as a parameter. How do Java devs deal with type erasure in generics? But explicitly supplying the type of information as an input where it's needed. 

I will catch up with @bergundy on the design of this. Right now it feels off for me.

Even if we go with a solution like this, it feels like it should be an optional and configurable behavior that is not enabled by default. So, I would put it under a flag and would allow the users to turn it on on some *Options if they want it for compatibility with TS. I don't think it's a good idea to make our users to store this additional metadata for each payload, it's just expensive to store an additional piece of metadata for each integer input parameter of an activity.

Spikhalskiy (commented):
Thanks for your submission, Loren!

Some thoughts on the approach:

1. Expensive. Each tiny proto object with one int field will now have significant additional metadata to store next to it.
2. Inverts the flow of control and source of the truth for the data type (from the code implementation to the stored metadata) with side effects that are unclear for me right now. One of the immediately clear side-effects is (3), but there could be more problems down the line.
3. Renaming of the protobuf type is forward and backward compatible change. The approach in this PR makes it not a compatible change. We start to impose some Temporal-specific limitations on protobuf, which is not good, especially when we state that Temporal doesn't handle serialization and try to stick to this paradigm. 

I'm not sure it's worth doing just because TS doesn't have runtime information about the type of the input parameters.
If TS can't infer a type of parameters in runtime, TS SDK should handle it somehow and probably have a way to explicitly pass the type information as an additional parameter for the deserialization and not require the whole stack and other languages to pass additional metadata fields for TS deserialization only. So, if in Java we infer `Class<T> valueClass` in 'ProtobufPayloadConverter#fromData' from the parameter, TS should have a way for a user to explicitly pass it as an input somehow.
This way TS will know the expected type in the runtime and it will be coming from the implementation and it brings TS in line with other SDKs for this matter. Workflow/Activity implementation dictates what type to deserialize, not the data or metadata in the payload. This fixes the actual problem - lack of runtime info in TS and it's done without inversion of the source of truth that creates unintended consequences for other SDKs.

This is also the conventional way of dealing with such issues from my experience.
For example, how Jackson knows what type of object to restore from the JSON? It accepts it as a parameter.
How do Java devs deal with type erasure in generics? By explicitly supplying the type of information as an input where it's needed.

Even if we go with a solution close to what is implemented in this PR, it feels like it should be an optional and configurable behavior that is not enabled by default. So, I would put it under a flag and would allow the users to turn it on on some *Options if they want it for compatibility with TS.

Right now at a first glance, I'm against this design. But keep in mind that I didn't participate in the initial discussions and may be out of the loop.

temporal-sdk/src/main/java/io/temporal/common/converter/ProtobufJsonPayloadConverter.java
```diff
@@ -59,11 +59,15 @@ public String getEncodingType() {
     }
 
     try {
+      String messageTypeName = ((MessageOrBuilder) value).getDescriptorForType().getFullName();
       String data = printer.print((MessageOrBuilder) value);
       return Optional.of(
           Payload.newBuilder()
               .putMetadata(
                   EncodingKeys.METADATA_ENCODING_KEY, EncodingKeys.METADATA_ENCODING_PROTOBUF_JSON)
+              .putMetadata(
```
cretz, line 68: Should we also require the message type to be accurate in `fromData` if it's present?

temporal-sdk/src/test/java/io/temporal/common/converter/ProtoPayloadConverterTest.java
cretz: In this test, why not just cast and have permanent expectations instead of put this behind a conditional? Are there cases where this tests follows different code paths in different situations?

bergundy:
@Spikhalskiy We've considered your proposed alternative but voted against it because having the type information in the metadata is both simpler on the TS side and allows us to show binary protobuf payloads in the UI.

We can make the additional header optional by configuration but I'd rather have this on by default, it's not *that* expensive and the added value feels like it's worth it especially now that we can put binary serialization in the default data converter and cut the overhead of JSON serialization.

> 2. It inverts the flow of control and source of truth for the data type (from the code implementation to the metadata) with side effects that not be immediately clear. One of them is (3), but there could be more problems down the line.
> 3. It brings Temporal SDK one step into the serialization business. Renaming the protobuf type is a forward and backward compatible change. The approach in this PR breaks this nice property. We start to impose some Temporal-specific limitations on protobuf serialization, which is problematic, especially when we state that Temporal doesn't handle serialization and we want to keep it this way.

I agree about this point but in practice I don't think people rely on it too much, in languages where we can infer the type at runtime we keep the old behavior.

Overall this change opens up possibilities at a seemingly low cost so I vote we keep this design.

cretz:
@Spikhalskiy - I initially agreed with you. All the TS SDK had to do was accept the type it wanted to deserialize into, but alas it doesn't. I also disagreed with forcing users to pre-register their proto types that work, but alas it is forced on the user. A much cleaner solution is to take the type inside "fromPayload" (e.g. `fromPayload(v: { decode(arr: Uint8Array) })`).

Having said that, the only thing that changed my mind for including message type is that it has value in other domains. Other dynamic libraries or reflection-based history readers or whatever can now use this information to reconstruct data which I think is helpful at no additional cost.

So, IMO the TS SDK doesn't have the best API for this, but having a message type has value elsewhere.

Sushisource:
From our meeting:
* We should add this regardless for UI reasons
* Semi-open question is (we're leaning towards yes, but didn't quite finalize) should this be on by default, disable-able as needed to save on storage?

Something that would help resolve this: What exactly does it look like to include the necessary type info / constructors / whatever to get this all working in JS/TS land. Is it truly that ugly? If so, then we should default to including the metadata. If not, maybe we won't.  @bergundy could you show a brief little example of what that'd look like?

cretz:
@Sushisource - https://github.com/temporalio/sdk-typescript/pull/430 was the original PR which is reverted, but you can see in v1.17.1 tag. I think based on https://github.com/temporalio/sdk-typescript/blob/v0.17.1/docs/protobuf-libraries.md#current-solution they landed on something like https://github.com/temporalio/sdk-typescript/blob/v0.17.1/packages/test/protos/root.js where you have to use a special protobuf call that puts all the proto types as JSON which are then registered on a common root object.

This is unfortunate that protobufjs doesn't have a global registry for access like other langs that have easier-to-use `google.protobuf.Any` support.

temporal-sdk/src/test/java/io/temporal/common/converter/ProtoPayloadConverterTest.java
lorensr: > Are there cases where this tests follows different code paths in different situations?

No

> In this test, why not just cast and have permanent expectations instead of put this behind a conditional?

When I did that, I got the unchecked casting linting warning. Happy to cast and suppress.

temporal-sdk/src/main/java/io/temporal/common/converter/ProtobufJsonPayloadConverter.java
lorensr: Pro: that could help catch bugs, if the wrong message type is accidentally used
Con: less flexibility, if users intend to use a different message type

I'm inclined toward flexibility

lorensr:
> Is it truly that ugly? 

Yep 😄 Here's the Slack discussion with an example of what it could look like: https://temporaltechnologies.slack.com/archives/C01FT8U10GK/p1641276095140200

Seems like current sentiment is pro making it the default:

image

Curious how you're feeling about it @Spikhalskiy ?

If we have consensus on default, then this PR is missing the config for turning it off, perhaps adding an `options` object to the constructors of: 

```
ProtobufJsonPayloadConverter
ProtobufPayloadConverter
DefaultDataConverter 
```

And not including `metadata.messageType` if `options.includeProtobufMessageTypes` is `false`.

One option for slightly reducing payload size is renaming `messageType` to `message` or `type`.

temporal-sdk/src/test/java/io/temporal/common/converter/ProtoPayloadConverterTest.java
cretz: I think worth suppressing in the test. What if metadata came back with a different type? You'd succeed the test.

temporal-sdk/src/main/java/io/temporal/common/converter/ProtobufJsonPayloadConverter.java
cretz: The problem is wrongness. Assert the correctness of metadata, be it message type, encoding, or whatever seems very reasonable. At the least, maybe a warning could be issued.

cretz:
> And not including metadata.messageType if options.includeProtobufMessageTypes is false.

I would prefer the inverse, and have `excludeProtobufMessageTypes` which defaults to false (as booleans do). Same for Go.

Spikhalskiy:
@lorensr Hey Loren! Yeah, the decision is to add it with a flag allowing to disable this feature for people who are optimizing their storage and don't need it for UI and typescript.

temporal-sdk/src/test/java/io/temporal/common/converter/ProtoPayloadConverterTest.java
```diff
@@ -96,6 +101,42 @@ public void testCustomProto() {
...
+            .setRunId(UUID.randomUUID().toString())
+            .build();
+    Optional<Payloads> data = converter.toPayloads(execution);
+    Payloads payloads = data.get();
+    Object field = payloads.getField(payloads.getDescriptorForType().findFieldByName("payloads"));
+    Payload payload = (Payload) ((List<?>) field).get(0);
+    Object metadata = payload.getField(payload.getDescriptorForType().findFieldByName("metadata"));
+    assertEquals(1, ((List<?>) metadata).size());
+  }
+
```
lorensr, line 139: I could also test the warning message if there's a way to mock/spy on `log.warn`?

temporal-sdk/src/main/java/io/temporal/common/converter/AbstractProtobufPayloadConverter.java
```diff
@@ -0,0 +1,79 @@
...
+
+import com.google.protobuf.ByteString;
+import com.google.protobuf.Descriptors;
+import com.google.protobuf.MessageOrBuilder;
+import io.temporal.api.common.v1.Payload;
+import java.lang.reflect.Method;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public abstract class AbstractProtobufPayloadConverter {
```
lorensr, line 32: I'd name this `ProtobufPayloadConverter` and rename `ProtobufPayloadConverter -> ProtobufBinaryPayloadConverter` but that would be a breaking change.

temporal-sdk/src/test/java/io/temporal/common/converter/ProtoPayloadConverterTest.java
Spikhalskiy: No need really, this is waaaay too deep.

temporal-sdk/src/main/java/io/temporal/common/converter/AbstractProtobufPayloadConverter.java
Spikhalskiy: Honestly, I think that `ProtobufPayloadConverter` is a better name than `ProtobufBinaryPayloadConverter `. Because for me protobuf is a binary format and this doesn't really need Binary here. Also, it will make it clear that this is "the main one" for people who want proto serialization. ProtoJson is some strange thing, that hopefully will not be used much by anybody, because it's a mix of the worst things of both worlds at the end.

temporal-sdk/src/main/java/io/temporal/common/converter/AbstractProtobufPayloadConverter.java
```diff
@@ -0,0 +1,79 @@
...
+  protected void addMessageType(Payload.Builder builder, Object value) {
+    if (this.excludeProtobufMessageTypes) {
+      return;
+    }
+    String messageTypeName = ((MessageOrBuilder) value).getDescriptorForType().getFullName();
+    builder.putMetadata(
+        EncodingKeys.METADATA_MESSAGE_TYPE_KEY, ByteString.copyFrom(messageTypeName, UTF_8));
+  }
+
+  protected void checkMessageType(Payload payload, Class valueClass) throws DataConverterException {
```
Spikhalskiy, line 55: ⚠️ Java and Go SDK shouldn't use the encoded type in any way.

As we discussed in Cons, this serialization of message type adds additional limitations to Protobuf compatibility.
If TS decided to go this way and dodge the bullet and prefer it to other solutions, it's TS SDK choice.
But Java and Go shouldn't be using this piece of information in any way, they don't need it.

cretz: This was my suggestion from https://github.com/temporalio/sdk-java/pull/942#discussion_r777509609. If the message type is wrong I think a warning is reasonable.

Spikhalskiy: Alright, if it's just a log.

temporal-sdk/src/main/java/io/temporal/common/converter/AbstractProtobufPayloadConverter.java
```diff
@@ -0,0 +1,79 @@
...
+              "Encoded protobuf message type \""
+                  + messageType
+                  + "\" does not match valueClass `"
+                  + valueClass.getName()
+                  + "`, which has message type \""
+                  + valueMessageType
+                  + '"');
+        }
+      } catch (Exception e) {
+        throw new DataConverterException(e);
```
Spikhalskiy, line 75: If it's just a log thing, let's never rethrow anything here. If an exception happens, let's swallow it and just print it in a log.

temporal-sdk/src/main/java/io/temporal/common/converter/AbstractProtobufPayloadConverter.java
Spikhalskiy: I really don't like the fact that we do a reflection call now on each proto deserialization here. We make things expensive without clear benefits.

temporal-sdk/src/main/java/io/temporal/common/converter/AbstractProtobufPayloadConverter.java
```diff
@@ -0,0 +1,79 @@
...
+  protected void checkMessageType(Payload payload, Class valueClass) throws DataConverterException {
+    ByteString messageTypeBytes =
+        payload.getMetadataMap().get(EncodingKeys.METADATA_MESSAGE_TYPE_KEY);
+    if (messageTypeBytes != null) {
+      try {
+        String messageType = messageTypeBytes.toString(UTF_8);
+        Method getDescriptor = valueClass.getMethod("getDescriptor");
+        String valueMessageType =
+            ((Descriptors.Descriptor) getDescriptor.invoke(null)).getFullName();
+        if (!messageType.equals(valueMessageType) && log.isWarnEnabled()) {
```
Spikhalskiy, line 64: isWarnEnabled check should be much earlier. If it's false, all this logic above is not needed.

temporal-sdk/src/main/java/io/temporal/common/converter/AbstractProtobufPayloadConverter.java
```diff
@@ -0,0 +1,79 @@
...
+        EncodingKeys.METADATA_MESSAGE_TYPE_KEY, ByteString.copyFrom(messageTypeName, UTF_8));
+  }
+
+  protected void checkMessageType(Payload payload, Class valueClass) throws DataConverterException {
+    ByteString messageTypeBytes =
+        payload.getMetadataMap().get(EncodingKeys.METADATA_MESSAGE_TYPE_KEY);
+    if (messageTypeBytes != null) {
+      try {
+        String messageType = messageTypeBytes.toString(UTF_8);
+        Method getDescriptor = valueClass.getMethod("getDescriptor");
```
Spikhalskiy, line 61: Let's move `checkMessageType` to AFTER the deserialization is done. This way it can accept not a class, but a specific deserialized instance. And instead of an expensive reflection we can do `((MessageOrBuilder)instance).getDescriptorForType().getFullName()`.

cretz: At this point, I'd be ok leaving the check off and taking back my request to warn. We need to make it clear that the metadata is only for dynamic languages and not for intra-lang use.

temporal-sdk/src/main/java/io/temporal/common/converter/AbstractProtobufPayloadConverter.java
lorensr: Haha, makes sense 😄

temporal-sdk/src/main/java/io/temporal/common/converter/AbstractProtobufPayloadConverter.java
lorensr: Makes sense, will do

temporal-sdk/src/main/java/io/temporal/common/converter/DefaultDataConverter.java
```diff
@@ -40,21 +40,27 @@
...
       new AtomicReference<>(newDefaultInstance());
 
   private final Map<String, PayloadConverter> converterMap = new ConcurrentHashMap<>();
 
   private final List<PayloadConverter> converters = new ArrayList<>();
 
+  // Order is important as the first converter that can convert the payload is used. Needs to match
+  // the other SDKs. Go SDK:
+  // https://github.com/temporalio/sdk-go/blob/5e5645f0c550dcf717c095ae32c76a7087d2e985/converter/default_data_converter.go#L28
+  private static final PayloadConverter[] getDefaultPayloadConverters(
```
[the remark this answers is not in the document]
lorensr: Removing the boolean, keeping the comment and addition of `ProtobufPayloadConverter`.

cretz (approved):

Spikhalskiy (approved):
Thank you for patiently addressing all the feedback, Loren! Great job.

Issue #752, opened 2021-09-22
Revisit ActivityExecutionContext and ActivityCompletion clients


Spikhalskiy:
Code quality and readability around `ActivityExecutionContext`, `ActivityCompletionClient`, `ManualActivityCompletionClient` requires refactoring to make code more maintainable. 

- [x] #943 
- [x] Manual Completion Client is a public user-facing interface that is undocumented. #1181
- [ ] Errors handling around heartbeating in ActivityExecutionContext and ManualActivityCompletionClient are different without apparent reasons.
- [x] #1233
- [ ] In case of some exceptions, ActivityExecutionContext stops heartbeating.
- [ ] ActivityExecutionContext#heartbeat and other methods like that should have a short explanation about the heartbeat throttling
- [ ] #1354

Spikhalskiy:
`ActivityExecutionContext` is done

Quinn-With-Two-Ns:
Closing since we no longer want to use GH issues as epics

PR #943, opened 2022-01-03
Heartbeat throttling is revisited for a new design, gets a default if heartbeatTimeout is not specified

### LINKED ISSUE #906: Heartbeat throttling interval should be configurable and default interval should be greater than 0
The Java SDK should implement the new heartbeat throttling behavior:

Add 2 new `WorkerOption`s:

- `maxHeartbeatThrottleInterval` - default 60s
- `defaultHeartbeatThrottleInterval` - default 30s

`throttleInterval` = `min`(

`heartbeat_timeout * .8` if `heartbeat_timeout` is provided

else `defaultHeartbeatThrottleInterval`

, `maxHeartbeatThrottleInterval`)

## Rationale

- Make the behavior explicit and configurable
- Keep the behavior backwards compatible
    - Java currently has an implicit maximum (30s) and default (0)
    - Go has an implicit default (10m) and no maximum
    - Core is not yet stable, from user feedback the behavior is confusing
- Not heartbeating means no cancellations, need a default value
- Consistency between the SDKs 

GoSDK equivalent task: https://github.com/temporalio/sdk-go/issues/656

Spikhalskiy:
## What was changed

`WorkerOptions#maxHeartbeatThrottleInterval` and `WorkerOptions#defaultHeartbeatThrottleInterval` has been added with reasonable defaults.
The behavior of JavaSDK in case of an absent `ActivityOptions#heartbeatTimeout` is brought on par with a new universal across all SDKs design in #906.
Refactoring of the code around heartbeats to make it easier to maintain. Partially addresses #752

### How was this tested
More tests to be added

Closes #906


temporal-sdk/src/main/java/io/temporal/worker/WorkerOptions.java
```diff
@@ -214,6 +222,33 @@ public Builder setDefaultDeadlockDetectionTimeout(long defaultDeadlockDetectionT
...
 
+    /**
+     * @param interval the maximum amount of time between sending each pending heartbeat to the
+     *     server. Regardless of heartbeat timeout, no pending heartbeat will wait longer than this
+     *     amount of time to send. Default is 60s, which is chosen if set to null or 0.
+     * @return {@code this}
+     */
+    public Builder setMaxHeartbeatThrottleInterval(Duration interval) {
+      Preconditions.checkArgument(
+          !interval.isNegative(), "Negative maxHeartbeatThrottleInterval value: " + interval);
```
cretz, line 233: Docs say "if set to null". If you're implying that null can be passed as a parameter, this will throw an NPE.

cretz (approved):

PR #946, opened 2022-01-03
Interrupt pollers when client disconnects

mmcshane:
## What was changed

Connects gRPC context cancellation up to the threads performing
long-polls on task queues so that when a client exits/dies and the
attendant connections go away, the polling threads are broken out of
their blocking polls.

Also adds queue closing into the mix so that closing a workflow store
will reconfigure the queues such that they no longer block and instead
always return an empty Optional.

## Why?

Allows an out-of-process test server to be reused by multiple test cases without needing to wait the ~60s for queue polling to timeout.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
+++ b/temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
@@ -109,10 +109,13 @@ import java.util.OptionalLong;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Executors;
 import java.util.concurrent.ForkJoinPool;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.Supplier;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -133,6 +136,8 @@ public final class TestWorkflowService extends WorkflowServiceGrpc.WorkflowServi
   private final Server inProcessServer;
   private final TestWorkflowStore store;
   private final Client client;
+  private final ScheduledExecutorService backgroundScheduler =
+      Executors.newSingleThreadScheduledExecutor();
 
   private static class Client implements Closeable {
     public Client() {
@@ -260,6 +265,9 @@ public final class TestWorkflowService extends WorkflowServiceGrpc.WorkflowServi
   @Override
   public void close() {
     log.debug("Shutting down TestWorkflowService");
+
+    log.info("Shutting down background scheduler");
+    backgroundScheduler.shutdown();
 
     if (outOfProcessServer != null) {
       log.info("Shutting down out-of-process GRPC server");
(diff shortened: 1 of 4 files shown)


temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
```diff
@@ -133,6 +136,7 @@
   private final Server inProcessServer;
   private final TestWorkflowStore store;
   private final Client client;
+  private final ScheduledExecutorService bgSched = Executors.newSingleThreadScheduledExecutor();
```
Spikhalskiy, line 139: What is bg? Could we rename without the abbreviation?

temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
```diff
@@ -572,20 +581,46 @@ public void respondWorkflowTaskFailed(
     }
   }
 
+  private Context.CancellableContext deadlineCtx(final Duration dur) {
```
Spikhalskiy, line 584: We don't use `final` for method parameters if it's not needed (for example in case of using it in closure)

temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
Spikhalskiy, line 139: Please terminate in TestWorkflowService#close so it's not leaking

temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowStoreImpl.java
```diff
@@ -328,57 +323,32 @@ public void registerDelayedCallback(Duration delay, Runnable r) {
 
   @Override
   public Optional<PollWorkflowTaskQueueResponse.Builder> pollWorkflowTaskQueue(
-      PollWorkflowTaskQueueRequest pollRequest, Deadline deadline) {
-    TaskQueueId taskQueueId =
+      PollWorkflowTaskQueueRequest pollRequest) {
+    final TaskQueueId taskQueueId =
```
Spikhalskiy, line 327: Same here, this final doesn't do anything useful.

temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
```diff
@@ -572,20 +581,46 @@ public void respondWorkflowTaskFailed(
...
+   * @param <T> the result of calling the provided Supplier
+   * @return the result of calling the provided Supplier.
+   */
+  private <T> T withContextInterruption(final Context ctx, final Supplier<T> work) {
+    final Thread thread = Thread.currentThread();
+    final Context.CancellationListener interrupter = context -> thread.interrupt();
+    ctx.addListener(interrupter, this.bgSched);
+    final T out = work.get();
+    ctx.removeListener(interrupter);
+    Thread.interrupted(); // clears interrupted flag
```
Spikhalskiy, line 604: What if this thread was interrupted not by `interrupter`, but for some other reason? 
Or not only by `interrupter`, but for some other reason? (external shutdownNow on the corresponded thread pool)
We would lose the interrupted flag in this case.

Instead of this "interrupted hack", I would better make an explicit `AtomicBoolean cancelled` that will be set inside `interrupter`. 
And replace untimed .get() call with 
1. Timed get calls to `work`.
2. Checking canceling flag
3. Repeat until timeout is exeeded.

temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowStoreImpl.java
```diff
@@ -556,7 +526,17 @@ private WorkflowExecutionInfo constructWorkflowExecutionInfo(
...
   public void close() {
-    isShuttingDown = true;
-    timerService.shutdown();
+    lock.lock();
+    try {
+      if (isShuttingDown) {
+        return;
+      }
+      isShuttingDown = true;
+      activityTaskQueues.values().forEach(TaskQueue::close);
```
Spikhalskiy, line 535: I'm not sure what problem does the closing of TaskQueue solves (and the existence of `TaskQueue`).
It looks like we close them when we close the whole TestService only. The first thing we do when we close 


TestService actually stops GRPC's Server and corresponding Client. This should trigger interruption of the thread waiting on the long polls and prevention of anything new to be added to the queues too. 

Is this broken somehow? Was it maybe broken by this newly added `withContextInterruption` that swallows interrupted flag that is installed not by `interrupter`?
Let's maybe fix `withContextInterruption` and reevaluate the need for TaskQueue after that?

Spikhalskiy (changes requested):
I feel that the added `withContextInterruption` swallows interruptions caused by TestService closing thinking that it's an interruption from context cancelation. Which leads to the addition of TaskQueue.
I do understand the purpose of `withContextInterruption`, but the way how it's implemented caused a bug which in turn is addressed by `TaskQueue` bandaid. 
I think if `withContextInterruption` is reworked to another way of passing cancellations using a timed spin loop with a condition check, there will be no need in TaskQueue, because it will stop swallowing interruptions from the shutdown.
It's a good idea anyway because usage of thread interruptions as a signal is not intended usage of interruptions.

temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
```diff
@@ -572,20 +581,46 @@ public void respondWorkflowTaskFailed(
...
+  /**
+   * Execute blocking function but use the supplied Context to trigger an interrupt of the current
+   * thread in the case that the Context is cancelled.
+   *
+   * @param ctx the context that may be canceled while this thread is blocked in work
+   * @param work the blocking work to perform
+   * @param <T> the result of calling the provided Supplier
+   * @return the result of calling the provided Supplier.
+   */
+  private <T> T withContextInterruption(final Context ctx, final Supplier<T> work) {
```
Spikhalskiy, line 598: It looks like it should be "withContextCancellation"? because we handle context cancellation here. 
Or "withThreadInterruption"? But this looks wrong. It looks like thread interruption is an implementation detail.

temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
mmcshane: That's more or less what the original code did (but it only did it for a duration timeout, not for context cancellation). The reason I didn't double-down on this approach is that the timing and timeout management gets messy. It's difficult (impossible?) to pick a poll time that won't just spin the CPUs but also breaks out of the poll fast enough that the next test execution doesn't fail. All of this is really about getting this poll thread out of the way in time for the next test to execute.

Having said all that, interruption is far from perfect. You're right that this call could clear an interrupt state that wasn't actually created by the Interrupter context listener and for that reason I'd be fine with removing this line altogether. I don't _think_ there's a way to guarantee some sort of interrupt-if-not-already-interrupted call or atomically interrupt the thread _and_ flip a bit in the Interrupter such that we could tell that it was indeed the cause of the interrupt (but if anyone can think of anything ...).

Given the drawbacks, I submit that interruption is the better of two flawed approaches but I think I can be convinced the other way

temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowStoreImpl.java
mmcshane: Well this may be rather more than is needed but the thinking here is that there are 3 ways to stop a grpc-java server
1. `shutdown()` which does _not_ interrupt in-flight RPCs
2. `shutdownNow()` which _does_ interrupt in-flight RPCs
3. Just exit the process

Right now we appear to do number three when running out-of-proc. But I take your point - it would be preferable to have appropriate startup/shutdown code.

temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
mmcshane: Just though of a way to achieve a preemptible queue poll without using `Thread.interrupt()` or spinning. Working through it now ... ~50% confident it will work :crossed_fingers:

PR #949, opened 2022-01-04
Refactor a local/regular activities difference out from POJOActivityTaskHandler

### LINKED ISSUE #948: Local DynamicActivity has a regular activity context
## Expected Behavior

DynamicActivity should behave as a normal regular or local activity in respect to heart beating and other concepts. 

## Actual Behavior

DynamicActivity has a regular activity context even if called as a local activity.

Spikhalskiy:
## What was changed

Code duplication between `POJOActivityImplementation` and `POJOLocalActivityImplementation` is cleaned up by unifying the code base and moving the difference out as a `ActivityExecutionContextFactory executionContextFactory`.
Abstracted `executionContextFactory` is reused in `DynamicActivityImplementation` to correctly supporting local mode.

## Why?
`POJOActivityTaskHandler` before contained a large code duplicate in `POJOActivityImplementation` and `POJOLocalActivityImplementation` classes implemented a difference between local and regular activities. This also led to the fact that `DynamicActivityImplementation` was not handling local/regular mode correctly because there was only one version.

Closes #948

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/activity/Activity.java
+++ b/temporal-sdk/src/main/java/io/temporal/activity/Activity.java
@@ -19,7 +19,7 @@
 
 package io.temporal.activity;
 
-import io.temporal.internal.sync.ActivityInternal;
+import io.temporal.internal.activity.ActivityInternal;
 import io.temporal.internal.sync.WorkflowInternal;
 
 /**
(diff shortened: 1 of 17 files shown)


cretz (approved):
Confirmed metrics and log levels/statements and such remained the same and just moved

Issue #922, opened 2021-12-07
OpenTracing context is not propagated in some cases


Spikhalskiy:
- [x] Child workflows as reported in comments in #537
- [x] Async invocation of a function or procedure as lambda (works if we invoke a stub method reference, like `Async.function(activityStub:activityMethod)`), see #537 too
- [x] `continueAsNew` needs a proper wiring of headers and OT implementation #844
- [ ] signal methods (headers on these methods are recently implemented) #1306

PR #951, opened 2022-01-05
Fix opentracing context propagation to async executed function

### LINKED ISSUE #537: Open tracing span context not propagated when activity or child workflow invoked asynchronously

Spikhalskiy:
## What was changed
Add propagation of opentracing span into child workflow threads, including asynchronous lambda calls
Added a test verifying that opentracing propagation works for child workflows

How was this tested: Unit test for lambda baggage item propagation and span structure

Closes #537
Partially addresses #922

### DIFF
--- a/temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowOutboundCallsInterceptor.java
+++ b/temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowOutboundCallsInterceptor.java
@@ -90,6 +90,21 @@ public class OpenTracingWorkflowOutboundCallsInterceptor
     }
   }
 
+  @Override
+  public Object newChildThread(Runnable runnable, boolean detached, String name) {
+    Span activeSpan = tracer.scopeManager().activeSpan();
+    Runnable wrappedRunnable =
+        activeSpan != null
+            ? () -> {
+              // transfer the existing active span into another thread
+              try (Scope scope = tracer.scopeManager().activate(activeSpan)) {
+                runnable.run();
+              }
+            }
+            : runnable;
+    return super.newChildThread(wrappedRunnable, detached, name);
+  }
+
   private Span createAndPassActivityStartSpan(String activityName, Header header) {
     Span span = createActivityStartSpanBuilder(activityName).start();
     contextAccessor.writeSpanContextToHeader(span.context(), header, tracer);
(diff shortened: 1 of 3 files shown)


bergundy (approved):
Too bad Java (and go) is still limited by not being able to create spans inside the workflow for lack of deterministic span id generation, can we maybe use a side effect to create spans?

Spikhalskiy:
@bergundy Actually, they don't. We can implement it exactly the same way we do with slf4j loggers - give a very thin
Workflow.getTracer() that returns tracer ignoring span creation during replays.
Side effects solution looks quite expensive from the storage perspective.

bergundy:
@Spikhalskiy that won't work.
Imagine this scenario:

```java
void execute() {
  withSpan(() -> {
    runActivity();
  })
}
```

#### First run:
- start workflow
  - create span
    - run activity (parent is the span created in WF)

#### Replay - activity resolves:
- start workflow
  - create span (span is not recorded)
    - run activity (span has no parent)

Spikhalskiy:
@bergundy 
Overseeing this exact issue, JavaSDK opentracing span structure was initially designed in a way that each run creates its own span, so it should be all good.
The exact situation that you described is already tested here:
https://github.com/temporalio/sdk-java/blob/a2bc9373201dd58a607197ac39b7f87c1c0059c2/temporal-opentracing/src/test/java/io/temporal/opentracing/WorkflowReplayTest.java#L152
Check the javadoc for this test method, it contains a created span structure - spans will be connected.

Spikhalskiy:
@bergundy 
The situation that you described is already tested here:
https://github.com/temporalio/sdk-java/blob/a2bc9373201dd58a607197ac39b7f87c1c0059c2/temporal-opentracing/src/test/java/io/temporal/opentracing/WorkflowReplayTest.java#L152
Check the javadoc for this test method, it contains a created span structure - spans will be connected in JavaSDK.

But what is true, user-created span may be ignored during replay, true. And in your sample the activity will be directly under the WorkflowRun span. But this is a smaller issue, let me think about how we can resolve it... 
I really believe that we can make it resolved just as a tracer wrapper that can make some smart "buffering" inside and track when isReplaying switches from true to false.

Spikhalskiy:
I created a follow-up to this discussion as a separate issue: https://github.com/temporalio/sdk-java/issues/952
We can continue the discussion there.

PR #953, opened 2022-01-05
Cancel pollers when client disconnects

mmcshane:
## What was changed

Connects gRPC context cancellation up to the threads performing
long-polls on task queues so that when a client exits/dies and the
attendant connections go away, the polling threads are broken out of
their blocking polls.

## Why?

To reuse a test server process across multiple test cases without having 
to wait for the 60s long-poll timeout.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
+++ b/temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
@@ -107,9 +107,13 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.OptionalLong;
 import java.util.UUID;
+import java.util.concurrent.CancellationException;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Executors;
 import java.util.concurrent.ForkJoinPool;
+import java.util.concurrent.Future;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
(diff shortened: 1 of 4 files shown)


temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
```diff
@@ -261,6 +268,9 @@ private TestWorkflowService(boolean isOutOfProc, int port) {
   public void close() {
     log.debug("Shutting down TestWorkflowService");
 
+    log.info("Shutting down background scheduler");
```
Spikhalskiy, line 271: Let's maybe make it debug? It looks pretty minor.

temporal-testing/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
```diff
@@ -492,46 +502,65 @@ public void getWorkflowExecutionHistory(
         });
   }
 
+  private <T> T pollTaskQueue(Context ctx, Future<T> futureValue)
+      throws ExecutionException, InterruptedException {
+    final Context.CancellationListener canceler = context -> futureValue.cancel(true);
```
Spikhalskiy, line 507: 👍  great solution, really like it

Spikhalskiy (approved):

Spikhalskiy:
Do we really still need this TaskQueue after implementing cancelation this new way?

mmcshane:
> Do we really still need this TaskQueue after implementing cancelation this new way?

What is needed to avoid spinning is to be able to wake up the condition that is being waited on when the queue read is blocked due to the queue being empty. This code achieves that end by introducing an object (the Future) that logically represents the pop operation and then funnels all the concurrency into that object - the write, the wait, and the cancel all end up going to the Future. With any of the various BlockingQueue impls in the jdk, there's no way to access the internal concurrency control structures - the thread that calls BlockingQueue.take waits on some condition variable (or monitor) _internal to that type_ and there's no way to wake it up (apart from interruption, which I think we all agree is kind of gross).

So that's a long way of saying I don't think we can get away with using any of the off-the-shelf queues because they don't give us access to the concurrency primitives in a way that's useful for cancellation.

bergundy:
🙌

PR #959, opened 2022-01-09
WorkflowUnsafe#isWorkflowThread

### LINKED ISSUE #955: Expose a method allowing user code to defined if they are inside workflow thread
Users want to reuse pieces of code in workflow and outside of the workflow context and use deterministic `Workflow` class methods if they are inside a workflow context and non-deterministic versions if they are outside.

Probably a method on the `Workflow` class returning true if we are executing inside a workflow thread

- Not exposing access to this info, maybe enforcing users to don't share code in such a manner.
- Give an API, but not on `Workflow`, but somewhere deeper, which will highlight the fact that it may be not a good practice.

**Additional context**
Related discussion in a Temporal Support Slack: https://temporalio.slack.com/archives/CTT84KXK9/p1641489380001200

Spikhalskiy:
Added WorkflowUnsafe#isWorkflowThread
Relocated general Workflows javadoc from Workflow class onto `io.temporal.workflow.package-info`

Closes #955

### DIFF
--- a/temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowOutboundCallsInterceptor.java
+++ b/temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowOutboundCallsInterceptor.java
@@ -28,6 +28,7 @@ import io.temporal.common.interceptors.WorkflowOutboundCallsInterceptorBase;
 import io.temporal.opentracing.OpenTracingOptions;
 import io.temporal.workflow.Workflow;
 import io.temporal.workflow.WorkflowInfo;
+import io.temporal.workflow.unsafe.WorkflowUnsafe;
 
 public class OpenTracingWorkflowOutboundCallsInterceptor
     extends WorkflowOutboundCallsInterceptorBase {
@@ -48,7 +49,7 @@ public class OpenTracingWorkflowOutboundCallsInterceptor
 
   @Override
   public <R> ActivityOutput<R> executeActivity(ActivityInput<R> input) {
-    if (!Workflow.isReplaying()) {
+    if (!WorkflowUnsafe.isReplaying()) {
       Span activityStartSpan =
           contextAccessor.writeSpanContextToHeader(
               () -> createActivityStartSpanBuilder(input.getActivityName()).start(),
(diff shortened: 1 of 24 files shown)


temporal-sdk/src/main/java/io/temporal/workflow/unsafe/WorkflowUnsafe.java
```diff
@@ -0,0 +1,70 @@
...
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ *  or in the "license" file accompanying this file. This file is distributed on
+ *  an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
+package io.temporal.workflow.unsafe;
```
cretz, line 20: Not sure it deserves its own package, but no biggie

cretz (approved):

temporal-sdk/src/main/java/io/temporal/workflow/unsafe/WorkflowUnsafe.java
Spikhalskiy: It doesn't. I just want to dig it deeper :)

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowOutboundCallsInterceptor.java
```diff
@@ -48,7 +49,7 @@ public OpenTracingWorkflowOutboundCallsInterceptor(
 
   @Override
   public <R> ActivityOutput<R> executeActivity(ActivityInput<R> input) {
-    if (!Workflow.isReplaying()) {
+    if (!WorkflowUnsafe.isReplaying()) {
```
bergundy, line 52: Should we expose this as `Workflow.unsafe.isReplaying()`?

bergundy (approved):

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowOutboundCallsInterceptor.java
cretz: I think the class level is common in Java.

Spikhalskiy: Do you mean like `Workflow.Unsafe.isReplaying` where `Unsafe` is an inner class of `Workflow`?
We can, but it will actually increase the discoverability (people will see it autocomplete for `Workflow.`), so I'm in favor of a separate class.

PR #969, opened 2022-01-12
Fix rpcRetryOptions not being set in WorkflowServiceStubsOptions

cliffsun91:
It wasn't being set in the private constructor `WorkflowServiceStubsOptions(Builder builder, boolean ignore)`.
It also seems like the parameter `ignore` isn't being used?

There also doesn't seem to be any existing tests for this class / builder...

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


Spikhalskiy:
Closing in favor of https://github.com/temporalio/sdk-java/pull/972

PR #974, opened 2022-01-13
Make GraalVM exe generation part of the build

mmcshane:
## What was changed

The Gradle task `nativeImage` will generate a
build/graal/temporal-test-server executable and the task `docker` will
wrap the same into a Docker image.

## Why?

Encapsulate all the fiddly bits of native image generation behind gradle tasks.

## Checklist

1. Closes 

2. How was this tested:

This is works-on-my-machine level on linux. I need to provision an EC2 Mac instance next to test out the OSX compatibility.

3. Any docs updates needed?


cretz:
Doesn't have to be part of this issue, but knowing all the specific Graal config you had to do, I wonder if we can have CI build the server via gradle and try to run at least its exe form and make a call against it just to confirm it works (don't even need multi-platform or docker, just a sanity check would be good enough I think).

I fear the settings may be brittle and a gRPC/netty upgrade may break things. Again, doesn't have to be in this issue of course.

mmcshane:
> I wonder if we can have CI build ...

Once I'm happy with OSX/arm I can make sure there are some appropriate sanity tests.

> I fear the settings may be brittle

Goodness yes. This support is highly coupled to the implementation internals of a complex third party library. But if we _must_ have a native exe and we _must not_ have yet another partial implementation of the server api then I don't know what else to try.

cretz:
@mmcshane - :+1: My concern w/ brittleness is allayed by having a CI check, so we're good.

temporal-testing/build.gradle
```diff
@@ -53,6 +55,23 @@ task testServiceServer(type: CreateStartScripts) {
     dependsOn 'sourcesJar'
 }
 
+graal {
```
Spikhalskiy, line 58: temporal-testing now is a mix of the testing server with some java-specific unit test helpers.
Do you think it's time to move test server into temporal-test-server and make temporal-testing depend on it?
So all this graal config can be in a more appropriate separate module than temporal-testing.

temporal-testing/build.gradle
```diff
@@ -53,6 +55,23 @@ task testServiceServer(type: CreateStartScripts) {
...
+  outputName "temporal-test-server"
+  mainClass application.mainClassName
+  javaVersion '17'
+  graalVersion '21.3.0'
+  option "--class-path=${buildDir}/dependencies/*"
+  option "--allow-incomplete-classpath"
+  option "--no-fallback"
+  option "--install-exit-handlers"
```
Spikhalskiy, line 66: It would be great if we can provide some commentary for each of the options that you still remember why were added :)

temporal-testing/src/main/resources/META-INF/native-image/io.temporal/temporal-testing/jni-config.json
```diff
@@ -0,0 +1,29 @@
+[
+{
+  "name":"java.lang.ClassLoader",
```
Spikhalskiy, line 3: It this file generated automatically using one of the Graal agents?

If so, it would be nice to get a Readme.md containing steps to generate/regenerate it.

Spikhalskiy (approved):
LGTM. All comments are advisory.
It would be really great to get more documentation with this PR about how this and that files were generated if they were generated or what's the logic behind them if they were manually crafted.

Also, I feel that the original documentation that I added about building the native distribution should likely be updated: https://github.com/temporalio/sdk-java/blob/8998879964e38d44bdd0aeb90e35141569f7bf55/temporal-testing/README.md#L50-L49

temporal-testing/build.gradle
bergundy: I would do it but doesn't have to be in this PR

temporal-testing/build.gradle
```diff
@@ -53,6 +55,23 @@ task testServiceServer(type: CreateStartScripts) {
...
+  option "-H:+StaticExecutableWithDynamicLibC"
+}
+
+docker {
```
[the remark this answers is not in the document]
bergundy: Yeah, you'd need to compile for linux on the host and copy it to the docker image

bergundy (approved):
LGTM.

The next steps should be to build and test the executable for different targets.
You will have to use dedicated builders for that so it's a considerable effort.
For arm64 linux and mac you can use EC2 (https://aws.amazon.com/ec2/instance-types/mac/).
For the x86 windows, linux and mac we can use Github Actions free builders.

I would use Github Actions for this.

mmcshane:
Landing this now but TODO list is

1. break out a separate module (same repo) for the server
2. document the steps to regenerate/update the native-image json config files
3. Only use static linking on linux
4. get GHA matrix builds going

PR #987, opened 2022-01-19
Rework and cleanup Worker metrics and contexts

### LINKED ISSUE #986: Add metrics for active executors threads

Right now metrics that are provided by JavaSDK don't give a direct answer to a question of what percentage of WTF, Activity, or Local Activity executors threads are actually used. 
We have a metric that gives us time between SCHEDULE and PULL of the task (SCHEDULE_TO_START) that can be judged that there are not enough Pollers.
We have a metric that gives us information on how long does it take to process a task (execution_latency)
But we don't have any direct stats helping to understand that we don't have enough executors. We can theoretically have an underloaded host with not enough executor threads and no metric to understand that we utilize the executor pool fully and we should increase its size.

Report a gauge metric with a number of available threads in Executor ThreadPools.

Spikhalskiy:
This PR doesn't have functional changes. It's a unification, cleanup, and refactoring of metrics and context between different types of workers.


Sushisource (approved):

cretz:
In Go, we recently added `temporal_worker_task_slots_available` which we found is clearer for determining utilization (see https://github.com/temporalio/sdk-go/pull/637). Might we want something similar here?

Sushisource:
Yeah actually I had forgotten about that. I originally advocated for available slots rather than used b/c it's generally more what you care about without needing to know how many total slots there are. Probably make sense to do those instead.

PR #996, opened 2022-01-22
Expose WorkflowInfo.getCronSchedule

tsurdilo:

exposes WorkflowInfo.getCronSchedule so can get the defined cron schedule inside workflow code

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowContext.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowContext.java
@@ -261,7 +261,7 @@ public interface ReplayWorkflowContext extends ReplayAware {
   /** Updates or inserts search attributes used to index workflows. */
   void upsertSearchAttributes(SearchAttributes searchAttributes);
 
-  /** @return workflow retry attempt */
+  /** @return workflow retry attempt. default is 1 */
   int getAttempt();
 
   /** @return workflow cron schedule */
--- a/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowContext.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowContext.java
@@ -261,5 +261,9 @@ public interface ReplayWorkflowContext extends ReplayAware {
   /** Updates or inserts search attributes used to index workflows. */
   void upsertSearchAttributes(SearchAttributes searchAttributes);
 
+  /** @return workflow retry attempt */
   int getAttempt();
+
+  /** @return workflow cron schedule */
+  String getCronSchedule();
 }
--- a/temporal-sdk/src/test/java/io/temporal/workflow/GetCronScheduleFromWorkflowInfoTest.java
+++ b/temporal-sdk/src/test/java/io/temporal/workflow/GetCronScheduleFromWorkflowInfoTest.java
@@ -1,0 +1,105 @@
+/*
+ *  Copyright (C) 2020 Temporal Technologies, Inc. All Rights Reserved.
+ *
+ *  Copyright 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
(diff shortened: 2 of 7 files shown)


bergundy (approved):

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowContext.java
```diff
@@ -261,5 +261,9 @@ void getVersion(
   /** Updates or inserts search attributes used to index workflows. */
   void upsertSearchAttributes(SearchAttributes searchAttributes);
 
+  /** @return workflow retry attempt */
```
Spikhalskiy, line 264: If we add this javadoc, it may be good to specify what is the stat value (0 or 1).

temporal-sdk/src/test/java/io/temporal/workflow/GetCronScheduleFromWorkflowInfoTest.java
```diff
@@ -0,0 +1,107 @@
...
+    WorkflowStub workflowStub =
+        testWorkflowRule
+            .getWorkflowClient()
+            .newUntypedWorkflowStub(
+                "TestGetCronScheduleWorkflowsFunc",
+                newWorkflowOptionsWithTimeouts(testWorkflowRule.getTaskQueue())
+                    .toBuilder()
+                    .setCronSchedule("0 0 * * *")
+                    .build());
+    testWorkflowRule.registerDelayedCallback(Duration.ofDays(1), workflowStub::cancel);
```
Spikhalskiy, line 64: Why do we need it here?

temporal-sdk/src/test/java/io/temporal/workflow/GetCronScheduleFromWorkflowInfoTest.java
```diff
@@ -0,0 +1,107 @@
...
+
+  @Rule
+  public SDKTestWorkflowRule testWorkflowRule =
+      SDKTestWorkflowRule.newBuilder()
+          .setWorkflowTypes(TestGetCronScheduleWorkflowsFuncImpl.class)
+          .build();
+
+  @Test
+  public void testGetCronScheduleFromWorkflowInfo() throws InterruptedException {
+    Assume.assumeFalse("skipping for docker tests", testWorkflowRule.isUseExternalService());
```
Spikhalskiy, line 53: Why can't we make it work with the external service?

Spikhalskiy (approved):

temporal-sdk/src/test/java/io/temporal/workflow/GetCronScheduleFromWorkflowInfoTest.java
tsurdilo: to hit the cron at least once, and then cancel it after a day (to stop it)

tsurdilo: the cron is set for "At 00:00" and we cancel it after a day to make sure it invokes the workflow method at least once when its tested. there seems to be some issues with testing cron workflows in general that will take a look at next.

cretz (approved):

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowContext.java
tsurdilo: added (default is 1)

temporal-sdk/src/test/java/io/temporal/workflow/GetCronScheduleFromWorkflowInfoTest.java
Spikhalskiy: Why do we need to cancel it at all after that day? I'm just not sure how it affects the test and why it's important for it. If the test can be implemented without this cancellation - it should (for example, by canceling from inside the worklfow?). Because you likely will face troubles adapting this test for a dockerized temporal if this is an important piece for the test.

PR #1000, opened 2022-01-24
Introduce GH actions for native-image builds

mmcshane:
## What was changed

A GH action that will build the test server native image on linux/amd64,
macOS/amd64, and windows/amd64. Build outputs are stored with the
workflow execution for 1 day and optionally attached to a user-supplied
GH release.

Missing from this work is a matrix build entry for macOS/aarch64 as
we don't have build machines for that platform yet.

The tgz and zip files that are ultimately released are modeled on
https://github.com/cli/cli/releases/tag/v2.4.0

## Why?

Cross-platform native builds of the test server

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/.github/workflows/native-images.yml
+++ b/.github/workflows/native-images.yml
@@ -1,0 +1,115 @@
+name: Build Native Images
+defaults:
+  run:
+    shell: bash
+on:
+  workflow_dispatch:
+    inputs:
+      release:
+        type: string
+        description: 'Release version (e.g. v1.2.3); omit to skip attaching artifacts to a release'
+        required: false
+      release_repo:
+        type: string
+        description: 'The repository hosting the github release; used only if the release version parameter is set'
+        required: false
+        default: temporalio/sdk-java
+      clobber:
+        type: boolean
+        description: 'Overwrite existing release artifacts'
+        required: false
+        default: no
+
+jobs:
+  build:
+    name: Build
+    strategy:
+      matrix:
+        include:
+          - dist: ubuntu-latest
+            os_family: linux
+            arch: amd64
+          - dist: macos-latest
+            os_family: macOS
+            arch: amd64
+          - dist: windows-latest
+            os_family: windows
+            arch: amd64
+    runs-on: ${{ matrix.dist }}
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+        
+      - name: Set up Java
+        uses: actions/setup-java@v2
+        with:
+          java-version: '11'
+          distribution: 'temurin'
+          
+      - name: Setup Gradle
+        uses: gradle/gradle-build-action@v2
+        
+      - name: Build Native Image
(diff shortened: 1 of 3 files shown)


cretz (approved):

.github/workflows/native-images.yml
```diff
@@ -0,0 +1,115 @@
...
+      # example: linux_amd64/ -> temporal-test-server_1.2.3_linux_amd64
+      # the name of the directory created becomes the basename of the archive (*.tar.gz or *.zip) and 
+      # the root directory of the contents of the archive.
+      - name: Rename dirs
+        run: |
+          shopt -s nullglob
+          version="$(sed 's/v//'<<<'${{ github.event.inputs.release }}')"
+          for dir in *; do
+            mv "$dir" "temporal-test-server_$version_$dir"
+          done
```
Sushisource, line 89: These scripts are short but IMO we should be putting all of them in external files rather than inline. Just dramatically easier to work with in IDEs etc

Also PR 1k :tada:

bergundy: I agree, I'd ideally want scripts that are more than a line long to not be shell scripts but I won't block the PR on it.
We should also verify that github sets `-euo pipefail` for these scripts if we plan to leave them here.

bergundy (approved):

bergundy:
How did you test this? (mostly asking about how to test releases)

It'd be nice to add some notes somewhere to explain how to test in case we want to modify the process.

.github/workflows/native-images.yml
mmcshane: I did check - bash defaults to  `set -eo pipefail` but not with `-u`. I can add `-u` in defaults.run.bash at the top of this file (and `-O nullglob` for that matter)

mmcshane: I can convert these stanzas to files but there's one good and one half-good reason not to do so - LMK if this changes your opinion...

The good reason is that as inline snippets, GH will include the code verbatim in the output of the action such that when it fails you get the code, the output, and the error in a little web UI that shows all the jobs and steps. If the code is in a separate file you just see the filename rather than the code itself. Debugging build systems tends to be impoverished and I'd lean toward making as much information as possible available wherever the error is indicated.

The half-good reason is that these snippets are not really bash. The `${{ foo }}` syntax is effectively a macro language (including its own expression language) that gets expanded prior to execution by the GH action runner so something like shellcheck is going to get false positives. Even syntax coloring gets confused.

bergundy: I get it, technically you can have a 1 liner in the yml command that calls a script with provided variables and for the script itself you can `set -x` to show which line is executed but it's not critical.

PR `0x3e8`! 🎉

mmcshane:
> It'd be nice to add some notes somewhere to explain how to test in case we want to modify the process.

I added a README with instructions on how to invoke this workflow in a fork for testing purposes.

PR #1006, opened 2022-01-26
Retain strong references to all Gauges

### LINKED ISSUE #979: Gauge metrics disappear after ~60 seconds
## Expected Behavior

After configuring a metrics scope (similar to what is described here: https://community.temporal.io/t/how-to-enable-metrics-support-using-java-sdk/366/2), I am having issues getting "gauge" metrics to register properly.

It seems that I can report metrics directly to the `MicrometerClientStatsReporter`, like this:

```
reporter.reportGauge("my_gauge", new HashMap<>(), 123.0);
```

Or, if I am inside the context of a Workflow, like this:

```
Workflow.getMetricsScope().gauge("my_gauge").update(123.0);
```

In either example, I would expect to see "my_gauge" in my metrics backend, and I would expect to see it updated whenever this code runs.

## Actual Behavior

In actuality, I am seeing the gauge value being reported, but it disappears approximately 60 seconds later. If I restart my program, the gauge starts reporting again, but it stops again after 60 seconds. If I try using a counter or a timer, then those seem to work fine. But there seems to be a problem that is specific to gauges.

Digging a bit deeper... It seems that both code examples above make use of the `reportGauge` method in the `MicrometerClientStatsReporter`. This method uses the `io.micrometer.core.instrument.MeterRegistry` to create a new gauge each time it is called. Once created, it updates the gauge using a value of type `double`.

This appears to violate 2 warnings listed in the [Gauges section](https://micrometer.io/docs/concepts#_gauges) of the Micrometer documentation. To summarize from section 9.1:

> Attempting to construct a gauge with a primitive number or one of its `java.lang` object forms is always incorrect.

From section 9.3, titled "Why is My Gauge Reporting NaN or Disappearing?":

> It is your responsibility to hold a strong reference to the state object that you are measuring with a  `Gauge`.

One more note is that if I create a `Gauge` through the `io.micrometer.core.instrument.MeterRegistry` directly (as well as using an `AtomicLong` and ensuring that it's not garbage collected), then the bug does not appear. In other words, I don't believe that my metrics backend is where the values are getting dropped.

  1. Enable metrics as described here: https://community.temporal.io/t/how-to-enable-metrics-support-using-java-sdk/366/2
  2. Create a new gauge using one of the examples above
  3. Verify that the gauge is being reported to your metrics backend
  4. Wait for approximately 60 seconds and the gauge _should_ stop being reported

## Specifications

  - Version: tested with Java SDK v1.5.0 and v1.7.0; Temporal is v1.14.0
  - Platform: Docker on Linux
  - Metrics backend: Prometheus

Spikhalskiy:
## What was changed
MicrometerReporter now retains references to all created guages to comply with Micrometer API requirements.

Closes #979

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/common/reporter/MicrometerClientStatsReporter.java
+++ b/temporal-sdk/src/main/java/io/temporal/common/reporter/MicrometerClientStatsReporter.java
@@ -19,6 +19,8 @@
 
 package io.temporal.common.reporter;
 
+import com.google.common.base.Preconditions;
+import com.google.common.util.concurrent.AtomicDouble;
 import com.uber.m3.tally.Capabilities;
 import com.uber.m3.tally.CapableOf;
 import com.uber.m3.tally.StatsReporter;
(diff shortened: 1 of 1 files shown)


temporal-sdk/src/main/java/io/temporal/common/reporter/MicrometerClientStatsReporter.java
```diff
@@ -19,22 +19,44 @@
...
+   * doesn't expose any way to retain or not retain a gauge) and Micrometer/Prometheus Gauge API.
+   *
+   * <p>This theoretically may cause a memory leak if users create gauges with names or tags of very
+   * large cardinalities or if the names of gauges or tags have unique values.
+   *
+   * <p>The design to retain gauges is based on <a
+   * href="https://github.com/uber-java/tally/blob/master/prometheus/src/main/java/com/uber/m3/tally/experimental/prometheus/PrometheusReporter.java">Tally's
+   * PrometheusReporter<a/>
+   */
+  private final ConcurrentMap<MetricID, AtomicDouble> registeredGauges = new ConcurrentHashMap<>();
```
Sushisource, line 58: The gauge itself is just an atomic? Odd

Sushisource (approved):

temporal-sdk/src/main/java/io/temporal/common/reporter/MicrometerClientStatsReporter.java
Spikhalskiy: Well, a Micrometer has a concept of "value measured by a Gauge". You rarely need to work with the Gauge object itself (while you can, using Gauge.builder()). To retain a Gauge, you need to actually retain this "measured value", not the Gauge itself, because by default Gauges have weak references to the "measured value" (which can be changed by passing strongReference=true to the Gauge Builder). Solution with retaining the Gauge itself ends up more complicated and less idiomatic Micrometer.

Sushisource: Ahhhh, that makes sense in a roundabout way

PR #1029, opened 2022-02-03
Improve Promise and io.temporal.workflow javadocs

Spikhalskiy:
## What was changed
Improve Promise and io.temporal.workflow javadocs.
Add a test for Promise handler calling a sleeping activity.

## Why?
Currently Promise handlers doc states that it can't contain blocking calls, while in reality, this code has all the same limitations as a regular workflow code.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/workflow/package-info.java
+++ b/temporal-sdk/src/main/java/io/temporal/workflow/package-info.java
@@ -319,32 +319,42 @@
  * <ul>
  *   <li>Do not use any mutable global variables because multiple instances of workflows are
  *       executed in parallel.
- *   <li>Do not call any non-deterministic functions like non seeded random, {@link java.lang.System
- *       .currentTimeMillis()} or {@link java.util.UUID#randomUUID()} directly form the workflow
- *       code. Always do this in activities.
- *   <li>Don’t perform any IO or service calls as they are not usually deterministic. Use activities
- *       for this.
- *   <li>Use methods on {@link io.temporal.workflow.Workflow} as safe deterministic alternatives to
- *       non-deterministic methods. For example, only use {@link
- *       io.temporal.workflow.Workflow#currentTimeMillis()} to get the current time inside a
- *       workflow.
+ *   <li>Do not call any non-deterministic functions, like non-seeded random, directly form the
+ *       workflow code. Always use safe deterministic alternatives provided by Temporal SDK on
+ *       {@link io.temporal.workflow.Workflow} or perform such calls in activities if have to. For
+ *       example:
+ *       <ul>
+ *         <li>Use {@link io.temporal.workflow.Workflow#currentTimeMillis()} instead of {@link
+ *             java.lang.System#currentTimeMillis()} to get the current time inside a workflow
+ *         <li>Use {@link io.temporal.workflow.Workflow#randomUUID()} instead of {@link
+ *             java.util.UUID#randomUUID()}
+ *       </ul>
+ *   <li>Don't perform long blocking operations other than calls that block inside Temporal SDK
+ *       (like Activity invocations or {@link io.temporal.workflow.Workflow} APIs). Use activities
+ *       for this. For example:
+ *       <ul>
+ *         <li>Call {@link io.temporal.workflow.Workflow#sleep(Duration)} instead of {@link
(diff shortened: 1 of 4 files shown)


mschielmann:
@Spikhalskiy  please also see the java doc of the Promise class itself. The [last point](https://github.com/temporalio/sdk-java/blob/master/temporal-sdk/src/main/java/io/temporal/workflow/Promise.java#L43) contains the same statement regarding the blocking calls.

tsurdilo (approved):

tsurdilo:
@Spikhalskiy LGTM, thanks for fixing this so fast! Only thing we could change is 
swap "limitations" with "constraints" imo if you agree.

cretz (approved):

temporal-sdk/src/main/java/io/temporal/workflow/package-info.java
```diff
@@ -319,32 +319,42 @@
...
- *       .currentTimeMillis()} or {@link java.util.UUID#randomUUID()} directly form the workflow
- *       code. Always do this in activities.
- *   <li>Don’t perform any IO or service calls as they are not usually deterministic. Use activities
- *       for this.
- *   <li>Use methods on {@link io.temporal.workflow.Workflow} as safe deterministic alternatives to
- *       non-deterministic methods. For example, only use {@link
- *       io.temporal.workflow.Workflow#currentTimeMillis()} to get the current time inside a
- *       workflow.
+ *   <li>Do not call any non-deterministic functions, like non-seeded random, directly form the
+ *       workflow code. Always use safe deterministic alternatives provided by Temporal SDK on
```
Sushisource, line 323:
```suggestion
 *   <li>Do not call non-deterministic functions, like non-seeded random, directly from the
 *       workflow code. Always use safe deterministic alternatives provided by the Temporal SDK on
```

temporal-sdk/src/main/java/io/temporal/workflow/package-info.java
```diff
@@ -319,32 +319,42 @@
...
+ *       {@link io.temporal.workflow.Workflow} or perform such calls in activities if have to. For
```
Sushisource, line 324:
```suggestion
 *       {@link io.temporal.workflow.Workflow} or perform such calls in activities when required. For
```

temporal-sdk/src/main/java/io/temporal/workflow/package-info.java
```diff
@@ -319,32 +319,42 @@
...
+ *       example:
+ *       <ul>
+ *         <li>Use {@link io.temporal.workflow.Workflow#currentTimeMillis()} instead of {@link
+ *             java.lang.System#currentTimeMillis()} to get the current time inside a workflow
+ *         <li>Use {@link io.temporal.workflow.Workflow#randomUUID()} instead of {@link
+ *             java.util.UUID#randomUUID()}
+ *       </ul>
+ *   <li>Don't perform long blocking operations other than calls that block inside Temporal SDK
+ *       (like Activity invocations or {@link io.temporal.workflow.Workflow} APIs). Use activities
+ *       for this. For example:
```
Sushisource, line 334:
```suggestion
 *   <li>Don't perform long (more than a few ms) blocking operations other than Temporal SDK-provided operations
 *       (like Activity invocations or {@link io.temporal.workflow.Workflow} APIs). Use activities
 *       for this. For example:
```

temporal-sdk/src/main/java/io/temporal/workflow/package-info.java
```diff
@@ -319,32 +319,42 @@
...
- *       java.lang.Thread#sleep(long)}.
- *   <li>Use {@link io.temporal.workflow.Promise} and {@link
- *       io.temporal.workflow.CompletablePromise} instead of {@link java.util.concurrent.Future} and
- *       {@link java.util.concurrent.CompletableFuture}.
- *   <li>Use {@link io.temporal.workflow.WorkflowQueue} instead of {@link
- *       java.util.concurrent.BlockingQueue}.
- *   <li>Don't change workflow code when there are open workflows. The ability to do updates through
- *       visioning is TBD.
+ *       synchronization because multi-threaded code inside a single workflow execution is executed
+ *       one thread at a time and under a global lock. For example:
```
Sushisource, line 351: GH won't let me suggest here, but:

"...no need **for** explicit..."

PR #1031, opened 2022-02-04
Unify worker pollers options naming between GoSDK and JavaSDK

Spikhalskiy:
Synchronization of similar option names helps to keep guides and docs less cluttered with references between different SDKs.


bergundy (approved):
LGTM

I wish we'd just autoscale the pollers and get rid of these knobs altogether.

Spikhalskiy:
@bergundy agree. But until we have autoscaling, let's just align the names to reduce documentation clutter.

PR #1045, opened 2022-02-19
Allow OpenTracing interceptor to ignore certain exceptions

### LINKED ISSUE #1012: OpenTracing interceptor should be able to ignore certain failures
Our workflow involves a lot of polling external services, following the pattern whereby we check for a result and throw an exception if it's not ready yet. As a result, we produce a lot of OpenTracing spans with errors that are actually normal, expected behavior. We also show an artificially inflated error % as a result.

I'd like to be able to exclude certain exception types from being logged as errors in the OpenTracing interceptor.

I'm not sure if there's another use case for this beside the specific one I'm dealing with, so if it doesn't seem generally-applicable enough I could always just implement a new interceptor for this project's use. And of course the optimal way would be to simply not have to poll (but setting up the external services to send signals is outside of our current scope).

tylercunnion:
## What was changed
Adds a Predicate as a new option on the OpenTracingOptions, which controls whether or not the tested exception should be considered an error for the purposes of the span. If the predicate returns false, the exception will not be attached to the span; otherwise it will. By default, the predicate simply returns true.

## Why?
Specifically, the use case that I'm solving is this: for Activities which have to periodically poll an external service, the general best practice is to throw an exception if the polled service isn't ready to respond yet (a request isn't finished for instance). In this case, the normal retry logic will kick in and periodically try again. There are surely other instances where one might throw an exception in order to repeat an Activity without it being considered abnormal.
However, when this happens, the OpenTracing integration will attach it to the Activity span as an error, even though it's healthy, expected behavior; the exception is more of an implementation detail. In this case, the workflow has an artificially inflated error rate and your Quality team starts asking you why your workflow is failing so much.

## Checklist

1. Closes #1012 

2. How was this tested:
New unit test created which throws two exceptions from an activity and tests that only one was logged.

3. Any docs updates needed?
Javadoc is updated, not sure if anything else is necessary.

### DIFF
--- a/temporal-opentracing/src/main/java/io/temporal/opentracing/OpenTracingOptions.java
+++ b/temporal-opentracing/src/main/java/io/temporal/opentracing/OpenTracingOptions.java
@@ -119,7 +119,8 @@ public class OpenTracingOptions {
      *     regardless, in order to provide a complete picture of the execution outcome. The "error"
      *     tag on the span will be set according to the value returned by this Predicate. By
      *     default, all exceptions will be considered errors.
-     * @see <a href="https://github.com/opentracing/specification/blob/master/semantic_conventions.md#span-and-log-errors">
+     * @see <a
+     *     href="https://github.com/opentracing/specification/blob/master/semantic_conventions.md#span-and-log-errors">
      *     OpenTracing Documentation</a> regarding error spans and logging exceptions.
      * @return this
      */
--- a/temporal-opentracing/src/main/java/io/temporal/opentracing/OpenTracingOptions.java
+++ b/temporal-opentracing/src/main/java/io/temporal/opentracing/OpenTracingOptions.java
@@ -114,9 +114,13 @@ public class OpenTracingOptions {
     }
 
     /**
-     * @param isErrorPredicate indicates whether an exception should or should not be considered an
-     *     error for the purpose of the OpenTracing span. By default, all exceptions are considered
-     *     errors.
+     * @param isErrorPredicate indicates whether the received exception should cause the OpenTracing
+     *     span to finish in an error state or not. All exceptions will be logged to the span
+     *     regardless, in order to provide a complete picture of the execution outcome. The "error"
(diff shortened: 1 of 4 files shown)


Spikhalskiy:
Thanks a lot for your contribution! 

This looks great, but I have one concern. 

If we merge this, the spans with failures will be indistinguishable from the successful executions and this may get tricky for an operator.
I think we should remove the deprecation from `io.temporal.opentracing.StandardTagNames.FAILED` and log this flag on exception irrespectively of `isErrorPredicate` result to show that it wasn't a successful run/execution.
WDYT?

tylercunnion:
Yeah, that's a really good point. I took a look over at the [OpenTracing docs](https://github.com/opentracing/specification/blob/master/semantic_conventions.md#logged-application-level-errors) for error logging; based on this what I'm thinking is 
a) The Predicate only ought to control whether or not we set the "error" tag on the _span_ (since this is what analytics tools use to determine error state or not)
b) Regardless, we should still add an error _log_ containing the thrown exception/stack trace/etc. (Same payload that's already in there)

In this case we might not even need to reintroduce the old "failed" tag since the combination of the exception log plus the presence or absence of the "error" tag should be enough for observers to determine what happened. 

Does that make sense? I think such a span would get across all the pertinent information while still signaling the correct semantics, which I think is exactly what we want.

Spikhalskiy:
This looks reasonable. It looks like it conveys enough (if we describe this idea in the predicate javadoc). We can always introduce the additional flag later if there is a corner case we didn't think about.

tylercunnion:
OK cool - I went ahead and moved the use of the Predicate to the SpanFactory class; so the only thing it affects now is simply whether the error tag is set to true or false. Hopefully the Javadoc I wrote is clear enough.
The other thing is that because of this, it would now affect both Activity and Workflow spans, which might be just fine but I figured I should call it out.

Spikhalskiy (approved):

Spikhalskiy:
Great job. Thanks for your contribution, @tylercunnion!

PR #1057, opened 2022-02-28
Expose OrchestratorServiceStubs

### LINKED ISSUE #882: Allow to register new search attributes
Temporal SDK allows to communicate with temporal server through GRPC calls. Currently you can create workflows with custom search attributes but those attributes must be already registered on Elasticsearch through the `add-search-attributes` sub-command on `tctl`

Allow to register new search attributes through the SDK

The only alternative by now is to run tctl 

**Additional context**
Nope

Spikhalskiy:
## What was changed
WorkflowServiceStubsImpl was refactored to reduce code duplication

## Why?
Allow users to register search attributes using sdk
Refactoring of WorkflowServiceStubs to reduce code duplication

Closes #882

### DIFF
--- a/temporal-serviceclient/src/main/java/io/temporal/serviceclient/ChannelManager.java
+++ b/temporal-serviceclient/src/main/java/io/temporal/serviceclient/ChannelManager.java
@@ -1,0 +1,350 @@
+/*
+ *  Copyright (C) 2020 Temporal Technologies, Inc. All Rights Reserved.
+ *
+ *  Copyright 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ *  Modifications copyright (C) 2017 Uber Technologies, Inc.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License"). You may not
+ *  use this file except in compliance with the License. A copy of the License is
+ *  located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ *  or in the "license" file accompanying this file. This file is distributed on
+ *  an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
+package io.temporal.serviceclient;
+
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
+import io.grpc.*;
+import io.grpc.health.v1.HealthCheckRequest;
+import io.grpc.health.v1.HealthCheckResponse;
+import io.grpc.health.v1.HealthGrpc;
+import io.grpc.netty.shaded.io.grpc.netty.NettyChannelBuilder;
+import io.grpc.stub.MetadataUtils;
+import io.temporal.internal.retryer.GrpcRetryer;
+import java.util.Collection;
+import java.util.List;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+final class ChannelManager {
+  private static final Logger log = LoggerFactory.getLogger(ChannelManager.class);
+
+  private static final int MAX_INBOUND_MESSAGE_SIZE = 25_000_000;
+
+  /** refers to the name of the gRPC header that contains the client library version */
+  private static final Metadata.Key<String> LIBRARY_VERSION_HEADER_KEY =
+      Metadata.Key.of("client-version", Metadata.ASCII_STRING_MARSHALLER);
+
+  /** refers to the name of the gRPC header that contains supported server versions */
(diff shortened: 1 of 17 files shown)


temporal-serviceclient/src/main/java/io/temporal/serviceclient/OperatorServiceStubs.java
```diff
@@ -0,0 +1,48 @@
...
+
+import io.temporal.api.operatorservice.v1.OperatorServiceGrpc;
+import io.temporal.internal.WorkflowThreadMarker;
+
+/** Initializes and holds gRPC blocking and future stubs. */
+public interface OperatorServiceStubs
+    extends ServiceStubs<
+        OperatorServiceGrpc.OperatorServiceBlockingStub,
+        OperatorServiceGrpc.OperatorServiceFutureStub> {
+  String HEALTH_CHECK_SERVICE_NAME = "temporal.api.workflowservice.v1.WorkflowService";
```
cretz, line 32: Should we change this service name?

temporal-serviceclient/src/main/java/io/temporal/serviceclient/ChannelManager.java
```diff
@@ -0,0 +1,341 @@
...
+  private void initConnectionManagement() {
+    // Currently, it is impossible to modify backoff policy on NettyChannelBuilder.
+    // For this reason we reset connection backoff every few seconds in order to limit maximum
+    // retry interval, which by default equals to 2 minutes.
+    // Once https://github.com/grpc/grpc-java/issues/7456 is done we should be able to define
+    // custom policy during channel creation and get rid of the code below.
+    if (options.getConnectionBackoffResetFrequency() != null) {
+      grpcConnectionManager.scheduleWithFixedDelay(
+          resetGrpcConnectionBackoffTask(),
+          options.getConnectionBackoffResetFrequency().getSeconds(),
```
cretz, line 195: There is a lot of `getSeconds()` here and health check timeout and such. Can we use the more granular unit, e.g. milliseconds?

temporal-serviceclient/src/main/java/io/temporal/serviceclient/ChannelManager.java
```diff
@@ -0,0 +1,341 @@
...
+      ServiceStubsOptions options, List<ClientInterceptor> additionalHeadInterceptors) {
+    // Do not shutdown a channel passed to the constructor from outside
+    this.channelNeedsShutdown = options.getChannel() == null;
+
+    this.options = options;
+    if (options.getChannel() != null) {
+      this.channel = options.getChannel();
+      this.grpcConnectionManager = null;
+    } else {
+      this.channel = prepareChannel();
```
cretz, line 79: To confirm, by default we are intentionally creating separate channels for different gRPC services on the same server? I haven't put in thought to how the API would look if you wanted to share them, but I just want to make sure that is intentional.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/OperatorServiceStubsImpl.java
```diff
@@ -0,0 +1,102 @@
...
+        new ChannelManager(this.options, Collections.singletonList(deadlineInterceptor));
+
+    HealthCheckResponse healthCheckResponse =
+        this.channelManager.waitForServer(HEALTH_CHECK_SERVICE_NAME);
+    if (!HealthCheckResponse.ServingStatus.SERVING.equals(healthCheckResponse.getStatus())) {
+      throw new RuntimeException(
+          "Health check returned unhealthy status: " + healthCheckResponse.getStatus());
+    }
+
+    log.info(String.format("Created WorkflowServiceStubs for channel: %s", channelManager.channel));
```
cretz, line 60: Nit - I am used to seeing something like `log.info("Created WorkflowServiceStubs for channel: {}", channelManager.channel);` so the channel's `toString` is only invoked if that log level is enabled.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/OperatorServiceStubs.java
Spikhalskiy: No, there is nothing responding to the OperatorService yet. This is as good as it gets for right now.
https://github.com/temporalio/temporal/pull/2579

temporal-serviceclient/src/main/java/io/temporal/serviceclient/OperatorServiceStubsImpl.java
Spikhalskiy: It's a relocated code, but you are right, it's better to rework.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/ChannelManager.java
Spikhalskiy: Yes. By default, every stub will have its own channel. It happens now if we create several WorkflowServiceStubs too. At the same time, if the users really want, they can reuse the channel. I exposed ServiceStubs#getRawChannel and the channel can be taken from the existing stubs and reused in the Options for a new stub.

Also, keep in mind that a regular App shouldn't be creating OperatorServiceStubs at all.

cretz (approved):

PR #1065, opened 2022-03-08
Synchronize GrpcSyncRetryer and GrpcAsyncRetryer behavior, make non-final exceptions INFO level

Spikhalskiy:
This is mostly a refactoring of GrpcSyncRetryer and GrpcAsyncRetryer that keeps the codebase unified (especially in relation to preserving "last meaningful exception"). 
LOG level for non-terminal exceptions is INFO now for both. Previously WARN in Sync and DEBUG in Async Retryer, which doesn't make much sense.

### DIFF
--- a/temporal-serviceclient/src/main/java/io/temporal/internal/retryer/GrpcAsyncRetryer.java
+++ b/temporal-serviceclient/src/main/java/io/temporal/internal/retryer/GrpcAsyncRetryer.java
@@ -144,7 +144,7 @@ class GrpcAsyncRetryer {
         GrpcRetryerUtils.createFinalExceptionIfNotRetryable(
             statusRuntimeException, previousException, options, grpcContextDeadline);
     if (finalException != null) {
-      log.warn("Not retryable failure", finalException);
+      log.warn("Non retryable failure", finalException);
       resultCF.completeExceptionally(finalException);
       return;
     }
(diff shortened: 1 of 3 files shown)


temporal-serviceclient/src/main/java/io/temporal/internal/retryer/GrpcAsyncRetryer.java
```diff
@@ -144,19 +144,22 @@ public GrpcAsyncRetryer(Clock clock) {
         GrpcRetryerUtils.createFinalExceptionIfNotRetryable(
             statusRuntimeException, previousException, options, grpcContextDeadline);
     if (finalException != null) {
+      log.warn("Not retryable failure", finalException);
```
Sushisource, line 147:
```suggestion
      log.warn("Non retryable failure", finalException);
```

temporal-serviceclient/src/main/java/io/temporal/internal/retryer/GrpcSyncRetryer.java
```diff
@@ -67,11 +67,13 @@ public GrpcSyncRetryer(Clock clock) {
       } catch (StatusRuntimeException e) {
         RuntimeException finalException =
             GrpcRetryerUtils.createFinalExceptionIfNotRetryable(
-                e, lastException, options, grpcContextDeadline);
+                e, lastMeaningfulException, options, grpcContextDeadline);
         if (finalException != null) {
+          log.warn("Not retryable failure", finalException);
```
Sushisource, line 72:
```suggestion
          log.warn("Non retryable failure", finalException);
```

Sushisource (approved):

PR #1067, opened 2022-03-08
Deserialized Search Attributes API and full Test Server implementation for Search Attributes

### LINKED ISSUE #617: Deserialize Search Attributes

### LINKED ISSUE #673: SearchAttributesTest fails against docker due to LocalDateTime serialization problem
## Expected Behavior
Not sure actually... either [this javadoc](https://github.com/temporalio/sdk-java/blob/master/temporal-sdk/src/main/java/io/temporal/client/WorkflowOptions.java#L212-L216) needs to change, or there's a bug around how the server `json.Unmarshal`s strings like `2021-08-26T13:21:52.059738` as a `time.Time`.

## Actual Behavior
Using a LocalDateTime as a SearchAttribute value fails with a message like `INVALID_ARGUMENT: 2021-08-26T13:21:52.059738 is not a valid value for search attribute CustomDatetimeField of type Datetime`

`export USE_DOCKER_SERVICE=true`, remove [these lines](https://github.com/temporalio/sdk-java/blob/5d8734be38648c9c1837832fb4fc424f26136ecb/temporal-sdk/src/test/java/io/temporal/workflow/SearchAttributesTest.java#L82-L84), and run SearchAttributesTest.testSearchAttributes().

## Specifications

  - Version: [1.11.4](https://github.com/temporalio/docker-compose/commit/ff95bcdf8371fbc10b5f4230fd4c0f46a450cfc8) in docker-compose.
  - Platform: cassandra

Spikhalskiy:
## What was changed

Test server now implements and verifies search attributes upserts and in start call on par with the real server
New methods are added that deserialize search attributes from Payload. 

## How was this tested

Unit tests covering new methods and test server functionality

Closes #673
Closes #617

Blocked by:
- [x] #1124
- [x] #1115
- [x] #1121


temporal-sdk/src/main/java/io/temporal/internal/common/HistoryJsonUtils.java
```diff
@@ -83,26 +78,4 @@ private static String convertEnumValues(
     }
     return parsed.jsonString();
   }
-
```
Spikhalskiy, line 86: This code was generalized and moved to ProtobufEnumNameUtils to be reused in Search Attributes code

temporal-sdk/src/main/java/io/temporal/internal/common/SearchAttributePayloadConverter.java
```diff
@@ -0,0 +1,254 @@
...
+    } else if (Long.class.equals(type)
+        || Integer.class.equals(type)
+        || Short.class.equals(type)
+        || Byte.class.equals(type)) {
+      return IndexedValueType.INDEXED_VALUE_TYPE_INT;
+    } else if (Double.class.equals(type) || Float.class.equals(type)) {
+      return IndexedValueType.INDEXED_VALUE_TYPE_DOUBLE;
+    } else if (Boolean.class.equals(type)) {
+      return IndexedValueType.INDEXED_VALUE_TYPE_BOOL;
+    } else if (OffsetDateTime.class.equals(type)) {
```
cretz, line 196: Hrmm, I wonder if `Instant` or even `java.util.Date` should be supported? Probably not if we expect to be able to decode to the same type we receive.

temporal-sdk/src/main/java/io/temporal/internal/common/SearchAttributePayloadConverter.java
```diff
@@ -0,0 +1,254 @@
...
+   * metadata that the SDK sends. When the attribute is registered with the service, so is its
+   * intended type which is used to validate the data. However, we do include type metadata in
+   * payload here for compatibility with older versions of the server. Earlier version of Temporal
+   * save the type metadata and return exactly the same payload back to the SDK, which will be
+   * needed to deserialize the attribute into it's initial type.
+   */
+  public Payload encode(@Nonnull Object obj) {
+    if (obj instanceof Collection && ((Collection<?>) obj).size() == 1) {
+      // always serialize an empty collection as one value
+      obj = ((Collection<?>) obj).iterator().next();
```
cretz, line 62: Hrmm, it is a bit confusing that if I encode a string list of size 1 and decode it I get a string but if I encode a string list of size 2 and decode it I get a list. If we did encode this as a single-item list, would the server just send us back a single item anyways?

temporal-sdk/src/main/java/io/temporal/internal/common/SearchAttributesUtil.java
```diff
@@ -0,0 +1,117 @@
...
+
+    return builder.build();
+  }
+
+  @Nonnull
+  public static Map<String, List<?>> decode(@Nonnull SearchAttributes searchAttributes) {
+    if (isEmpty(searchAttributes)) {
+      return Collections.emptyMap();
+    }
+    Map<String, List<?>> deserializedMap = new HashMap<>();
```
cretz, line 59: Since the empty-collection type is rare, should probably give an initial capacity here

temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowInternal.java
```diff
@@ -539,7 +541,41 @@ public static <R> R getLastCompletionResult(Class<R> resultClass, Type resultTyp
...
+        name,
+        list);
+    return list.get(0);
+  }
+
+  @Nullable
+  public static <T> List<T> getSearchAttributeValues(String name) {
+    SearchAttributes searchAttributes = getRootWorkflowContext().getContext().getSearchAttributes();
+    if (searchAttributes == null) {
+      return null;
```
cretz, line 563: I wonder if this should return an empty list for no value. Won't `getSearchAttributes` map values be empty lists instead of null?

It seems we return null in all cases of no attributes, no attribute, and no attribute value. So they can't use nullability to differentiate set or not anyways. If not, I haven't checked, but we probably need to state in the javadoc that this can never return an empty list then.

cretz:
Per off-PR discussion, marking approve with more off-PR discussion to come.

Specifically, here is the scenario in Java with this PR:

* Encode
  * string -> SA string
  * list of one string -> SA string
  * list of multiple strings -> SA list of multiple strings
* Decode
  * SA string -> list of one string (can be read as just string w/ separate getter though)
  * SA list of multiple strings -> list of multiple strings (fails if trying to use that single-item getter though)

While this makes sense for Java, in Go I think we might not unwrap a list of one string as just a string but rather leave as list. Unsure about TypeScript.

cretz (approved):

temporal-sdk/src/main/java/io/temporal/internal/common/ProtoEnumNameUtils.java
```diff
@@ -0,0 +1,80 @@
...
+
+import com.google.common.base.CaseFormat;
+import com.google.common.collect.ImmutableMap;
+import io.temporal.api.enums.v1.IndexedValueType;
+import io.temporal.api.enums.v1.WorkflowTaskFailedCause;
+import java.util.Map;
+
+/**
+ * Because in Go enums are in a shared namespace, protobuf best practices is to prefix the enums
+ * with enum type prefix to create unique names. The final result is verbose and when we use a
```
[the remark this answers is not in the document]
Spikhalskiy: Well, in this AIP:
"The other values should not be prefixed by the name of the enum itself. This generally requires users to write MyState.MYSTATE_ACTIVE in their code, which is unnecessarily verbose."
But this is also right under:
"To avoid sharing values, APIs may prefix enum values with the name of the enum. In this case, they must do so consistently within the enum."

temporal-sdk/src/main/java/io/temporal/internal/common/ProtoEnumNameUtils.java
```diff
@@ -0,0 +1,80 @@
...
+
+  public static final Map<Class<?>, String> ENUM_CLASS_TO_PREFIX =
+      ImmutableMap.of(
+          WorkflowTaskFailedCause.class, WORKFLOW_TASK_FAILED_CAUSE_PREFIX,
+          IndexedValueType.class, INDEXED_VALUE_TYPE_PREFIX);
+
+  public static String uniqueToSimplifiedName(Enum<?> enumm) {
+    String protoEnumName = enumm.name();
+    String prefix = ENUM_CLASS_TO_PREFIX.get(enumm.getClass());
+    if (!protoEnumName.startsWith(prefix)) {
```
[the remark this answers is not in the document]
Spikhalskiy: I will add an explicit exception, thanks for pointing out. It's not always safe, the enums have to be added.

temporal-sdk/src/main/java/io/temporal/internal/common/SearchAttributePayloadConverter.java
Spikhalskiy: Date is, probably, not doable looking at the underlying ES Type.
 
Instant is a really good thought though, but yes, to your point, if we do Instant, we will have to return Instant ideally. And it will restrict us because OffsetDateTime is a more powerful class.

bergundy:
@cretz We haven't implemented search attributes in TS yet.
We will do that soon and always use `Array`s when reading attributes and collections for writing.
We will also verify that the values are valid ElasticSearch types.

temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowInternal.java
Spikhalskiy: Well, we use an empty list as a "delete" token, so I would day it's safer to deserialize it always to null to users.

> If not, I haven't checked, but we probably need to state in the javadoc that this can never return an empty list then.
👍

PR #1068, opened 2022-03-10
Add port callback URL for test server

bergundy:
To avoid port conflicts, I've added a way for the test server executable to accept a URL as a second argument, it can use this URL to report which port it is bound to (assuming the requested bind port is 0 in which case an unused port is assigned).

### DIFF
--- a/temporal-test-server/build.gradle
+++ b/temporal-test-server/build.gradle
@@ -39,6 +39,9 @@ graal {
   // Signal handling so that ^C actually stops the process
   option "--install-exit-handlers"
 
+  // Enable calling back an HTTP URL with the listening port
+  option "--enable-url-protocols=http"
+
   // If we're on linux, static link everything but libc. Otherwise link
   // everything dynamically (note the '-' rather than '+' in fromt of
   // StaticExecutable)
--- a/temporal-test-server/src/main/java/io/temporal/internal/testservice/TestServiceServer.java
+++ b/temporal-test-server/src/main/java/io/temporal/internal/testservice/TestServiceServer.java
@@ -19,14 +19,29 @@
 
 package io.temporal.internal.testservice;
 
+import java.io.IOException;
+import java.net.HttpURLConnection;
+import java.net.URL;
+
 public class TestServiceServer {
 
-  public static void main(String[] args) {
-    if (args.length != 1) {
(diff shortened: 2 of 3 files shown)


cretz:
Is this the only time you need to communicate with the test server? I wonder if the test server should instead surface an HTTP endpoint (or even gRPC) that the starting process calls? Yes it would be kinda annoying to continuously poll until online, but said RPC could also be used for other communication needs.

Spikhalskiy:
I don't really understand the usecase. 

Right now you already can pass a specific port to listen on, right? If you need to start several servers, why the caller can't check that the port is free and trigger the creation of the test server on the free port?

cretz:
Having the system choose a free port is reasonable, but I am not sure an HTTP callback is the best approach, and definitely not one just for port that has no other reuse potential. I think a "test server admin API" may be reasonable and it can provide the port.

Spikhalskiy:
We need test server api anyway, because we need time forwarding functionality to be provided to other languages.
But for that please halt till my refactoring is done that will make it easy.

cretz:
> it first needs to be communicated back to the calling process before that process is able to call gRPC endpoints.

True, I was thinking the admin API does not need to be the same gRPC. But then you're having to provide a fixed port anyways.

So yeah, maybe we should just make it where the caller has to pick a free port. Temporalite does this (see [here](https://github.com/DataDog/temporalite/blob/main/internal/liteconfig/freeport.go)), it's not like we let our real server communicate back a dynamically chosen port. If they do it we can do it. I can show how in various languages if necessary.

> And instead of doing that, better to let the system choose a free port outside, free the port if needed (if the choosing was implemented by creating a socket) and pass it inside the test server process and avoid the whole callback loop.

I now agree with this, :+1:

bergundy:
The temporalite approach might still fail AFAIK, some other process could take up this generated port between the resolve call and the bind call.
The problem gets worse when you find the port in one process and bind to it in another, there might be a significant delay between the 2 calls.

bergundy:
I can instead of this approach, try the temporalite approach in a loop until it succeeds.

bergundy:
Decided to go with the other approach where the SDK test framework gets a free port and passes that on to the test server via command line arg.

If we see that users complain about port conflicts with this approach we'll revisit the callback solution.

PR #1070, opened 2022-03-11
Code cleanup around Pollers

Spikhalskiy:
Fix for potential pollers deadlock if suspendPolling called twice.
Refactorings that make the code around Polling easier to follow.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/worker/ActivityPollTask.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/worker/ActivityPollTask.java
@@ -110,6 +110,7 @@ final class ActivityPollTask implements Poller.PollTask<ActivityTask> {
               ProtobufTimeUtils.toM3Duration(
                   response.getStartedTime(), response.getCurrentAttemptScheduledTime()));
       isSuccessful = true;
+      return new ActivityTask(response, pollSemaphore::release);
     } catch (StatusRuntimeException e) {
       if (e.getStatus().getCode() == Status.Code.UNAVAILABLE
           && e.getMessage().startsWith("UNAVAILABLE: Channel shutdown")) {
@@ -119,6 +120,5 @@ final class ActivityPollTask implements Poller.PollTask<ActivityTask> {
     } finally {
       if (!isSuccessful) pollSemaphore.release();
     }
-    return new ActivityTask(response, pollSemaphore::release);
   }
 }
--- a/temporal-sdk/src/main/java/io/temporal/internal/worker/Poller.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/worker/Poller.java
@@ -168,15 +168,18 @@ public final class Poller<T> implements SuspendableWorker {
 
   @Override
   public void suspendPolling() {
-    log.info("suspendPolling: {}", this);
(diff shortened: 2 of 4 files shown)


temporal-sdk/src/main/java/io/temporal/internal/worker/ActivityPollTask.java
```diff
@@ -110,6 +110,7 @@ public ActivityTask poll() {
               ProtobufTimeUtils.toM3Duration(
                   response.getStartedTime(), response.getCurrentAttemptScheduledTime()));
       isSuccessful = true;
+      return new ActivityTask(response, pollSemaphore::release);
```
Spikhalskiy, line 113: Paired with 
```
if (!isSuccessful) pollSemaphore.release();
```
in finally, this location of `return` is cleaner for the reader.

temporal-sdk/src/main/java/io/temporal/internal/worker/Poller.java
```diff
@@ -168,15 +168,18 @@ public void awaitTermination(long timeout, TimeUnit unit) {
 
   @Override
   public void suspendPolling() {
-    log.info("suspendPolling: {}", this);
-    suspendLatch.set(new CountDownLatch(1));
```
Spikhalskiy, line 172: Absense of CAS here may replace and lose the latch which is already used by Pollers to block on. This way they will never be unblocked if suspentPolling was called twice.

temporal-sdk/src/main/java/io/temporal/internal/worker/SingleWorkerOptions.java
```diff
@@ -126,35 +126,33 @@ public Builder setDefaultHeartbeatThrottleInterval(Duration defaultHeartbeatThro
     }
 
     public SingleWorkerOptions build() {
+      PollerOptions pollerOptions = this.pollerOptions;
       if (pollerOptions == null) {
-        pollerOptions =
-            PollerOptions.newBuilder()
-                .setPollBackoffInitialInterval(Duration.ofMillis(200))
```
Spikhalskiy, line 132: This code path is dead. We come here only in some tests where we do
```
        SingleWorkerOptions.newBuilder().build()
```

and these options and not important for these tests. As a result, these magic values 200 and 20 here are never used and beter be removed from the production code paths.

temporal-sdk/src/main/java/io/temporal/worker/Worker.java
```diff
@@ -181,7 +181,7 @@ private static SingleWorkerOptions toLocalActivityOptions(
       List<ContextPropagator> contextPropagators,
       Scope metricsScope) {
     return toSingleWorkerOptions(factoryOptions, options, clientOptions, contextPropagators)
-        .setPollerOptions(PollerOptions.newBuilder().build())
```
Spikhalskiy, line 184: ActivityWorker and WorkflowWorker PollerOptions specify this parameter explicitly. 
To simplify the untangling for the reader, let's specify it for LocalActivityWorker explicitly too. 1 is the default value for pollThreadCount used inside PollerOptions.

Sushisource (approved):

Issue #392, opened 2021-03-20
TimeLockingFuture implementation is incorrect and easy to break 


Spikhalskiy:
## Expected Behavior

CompletableFuture returned from ```WorkflowStub#getResult``` and ```WorkflowStub#getResultAsync``` can correctly work with all CompletableFuture methods.

## Actual Behavior

TimeLockingFuture is returned from getResultAsync and it overrides CompletableFuture ```#get``` and ```#join```.
This implementation is extremely brittle because any chaining of this CompletableFuture will not call the overridden ```#get``` and ```#join```. So, even trivial chaining completely breaks time skipping in unit tests.

## Steps to Reproduce the Problem

  1.
  1.
  1.

## Specifications

  - Version: Current master (5f2ac6133c6560984eff35cc85faafb85cbf2c94)

PR #1076, opened 2022-03-15
Time Skipping Service

Spikhalskiy:
## What was changed

Test Service gRPC service is added with time-skipping methods.
JavaSDK is fully reworked to use Time Skipping gRPC Stubs instead of accessing the in-memory TestService internal instances and classes directly when time skipping functionality is needed, reducing coupling between SDK test classes and test-server implementation.
Native Test Service default starting behavior changed to start with locked (disabled) time-skipping to behave closer to a real Temporal Server by default.

### DIFF
--- a/temporal-test-server/src/main/java/io/temporal/internal/testservice/TestService.java
+++ b/temporal-test-server/src/main/java/io/temporal/internal/testservice/TestService.java
@@ -1,0 +1,121 @@
+/*
+ *  Copyright (C) 2020 Temporal Technologies, Inc. All Rights Reserved.
+ *
+ *  Copyright 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ *  Modifications copyright (C) 2017 Uber Technologies, Inc.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License"). You may not
+ *  use this file except in compliance with the License. A copy of the License is
+ *  located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ *  or in the "license" file accompanying this file. This file is distributed on
+ *  an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
+package io.temporal.internal.testservice;
+
+import com.google.protobuf.Empty;
+import com.google.protobuf.Timestamp;
+import io.grpc.stub.StreamObserver;
+import io.temporal.api.testservice.v1.*;
+import io.temporal.internal.common.ProtobufTimeUtils;
+import java.io.Closeable;
+import java.time.Duration;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+
+/**
+ * In memory implementation of the Operator Service. To be used for testing purposes only.
+ *
+ * <p>Do not use directly, instead use {@link io.temporal.testing.TestWorkflowEnvironment}.
+ */
+public final class TestService extends TestServiceGrpc.TestServiceImplBase implements Closeable {
+
+  private final TestWorkflowStore workflowStore;
+
+  public TestService(TestWorkflowStore workflowStore) {
+    this.workflowStore = workflowStore;
+  }
+
+  @Override
+  public void lockTimeSkipping(
+      LockTimeSkippingRequest request, StreamObserver<LockTimeSkippingResponse> responseObserver) {
+    lockTimeSkipping(request.getCallerId());
+    responseObserver.onNext(LockTimeSkippingResponse.newBuilder().build());
+    responseObserver.onCompleted();
+  }
+
+  @Override
+  public void unlockTimeSkipping(
+      UnlockTimeSkippingRequest request,
+      StreamObserver<UnlockTimeSkippingResponse> responseObserver) {
+    unlockTimeSkipping(request.getCallerId());
+    responseObserver.onNext(UnlockTimeSkippingResponse.newBuilder().build());
+    responseObserver.onCompleted();
+  }
+
+  @Override
+  public void sleep(SleepRequest request, StreamObserver<SleepResponse> responseObserver) {
+    sleep(ProtobufTimeUtils.toJavaDuration(request.getDuration()));
+    responseObserver.onNext(SleepResponse.newBuilder().build());
+    responseObserver.onCompleted();
+  }
+
+  @Override
+  public void getCurrentTime(
+      Empty request, StreamObserver<GetCurrentTimeResponse> responseObserver) {
+    Timestamp timestamp = workflowStore.currentTime();
+    responseObserver.onNext(GetCurrentTimeResponse.newBuilder().setTime(timestamp).build());
+    responseObserver.onCompleted();
+  }
+
+  /**
+   * Disables time skipping. To re-enable call {@link #unlockTimeSkipping(String)}. These calls are
+   * counted, so calling unlock does not guarantee that time is going to be skipped immediately as
+   * another lock can be holding it.
+   */
+  public void lockTimeSkipping(String caller) {
+    workflowStore.getTimer().lockTimeSkipping(caller);
+  }
+
+  public void unlockTimeSkipping(String caller) {
+    workflowStore.getTimer().unlockTimeSkipping(caller);
+  }
+
+  /**
+   * Unlocks time skipping and blocks the calling thread until internal clock passes the current +
+   * duration time.<br>
+   * When the time is reached, locks time skipping and returns.<br>
+   * Might not block at all due to time skipping. Or might block if the time skipping lock counter
+   * was more than 1.
+   */
+  public void sleep(Duration duration) {
+    CompletableFuture<Void> result = new CompletableFuture<>();
+    workflowStore
+        .getTimer()
+        .schedule(
+            duration,
+            () -> {
+              workflowStore.getTimer().lockTimeSkipping("TestWorkflowService sleep");
+              result.complete(null);
+            },
+            "workflow sleep");
+    workflowStore.getTimer().unlockTimeSkipping("TestWorkflowService sleep");
+    try {
+      result.get();
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+      throw new RuntimeException(e);
+    } catch (ExecutionException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  @Override
+  public void close() {}
+}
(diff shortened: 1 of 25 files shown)


Spikhalskiy:
@bergundy This is the most important for you for time-skipping in typescript: https://github.com/temporalio/sdk-java/pull/1076/files#diff-d7a107df90fe446efc696d828e3df42e829262e802060ade229d4cdddbb6fb26R41

temporal-testing/src/main/java/io/temporal/testing/IdempotentTimeLocker.java
```diff
@@ -0,0 +1,70 @@
...
+ */
+class IdempotentTimeLocker {
+  private final TestServiceStubs testServiceStubs;
+  private final AtomicInteger count = new AtomicInteger(1);
+
+  IdempotentTimeLocker(TestServiceStubs testServiceStubs) {
+    this.testServiceStubs = testServiceStubs;
+  }
+
+  public void lockTimeSkipping(String caller) {
```
mmcshane, line 40: There may be a problem with this approach - when a given call to `lockTimeSkipping` returns, it is _not_ guaranteed that time skipping has been locked. Any call subsequent to the first call might return early _while that first call is still executing the RPC call_. This leaves a window where this method has returned but time skipping is not (yet) locked.

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
```diff
@@ -1027,6 +1031,11 @@ public void getDiagnostics(StringBuilder result) {
     store.getDiagnostics(result);
   }
 
+  /**
+   * @deprecated use {@link io.temporal.serviceclient.TestServiceStubs} and {@link
+   *     io.temporal.api.testservice.v1.TestServiceGrpc.TestServiceBlockingStub#getCurrentTime(Empty)}
+   */
+  @Deprecated
```
Sushisource, line 1038: Are users using these directly and that's why we need to deprecate them?

temporal-test-server/src/main/java/io/temporal/serviceclient/TestServiceStubsImpl.java
```diff
@@ -0,0 +1,106 @@
...
+public class TestServiceStubsImpl implements TestServiceStubs {
+  private static final Logger log = LoggerFactory.getLogger(TestServiceStubsImpl.class);
+
+  private final ChannelManager channelManager;
+
+  private final TestServiceGrpc.TestServiceBlockingStub blockingStub;
+  private final TestServiceGrpc.TestServiceFutureStub futureStub;
+
+  /**
+   * Creates a factory that connects to the Temporal according to the specified options. When
```
Sushisource, line 41: To the Temporal....

temporal-test-server/src/main/proto/temporal/api/testservice/v1/service.proto
```diff
@@ -0,0 +1,63 @@
...
+option ruby_package = "Temporal::Api::TestService::V1";
+option csharp_namespace = "Temporal.Api.TestService.V1";
+
+import "temporal/api/testservice/v1/request_response.proto";
+import "google/protobuf/empty.proto";
+
+// OperatorService API defines how Temporal SDKs and other clients interact with the Temporal server
+// to perform administrative functions like registering a search attribute or a namespace.
+// APIs in this file could be not compatible with Temporal Cloud, hence it's usage in SDKs should be limited by
+// designated APIs that clearly state that they shouldn't be used by the main Application (Workflows & Activities) framework.
```
Sushisource, line 40: Copy-pasted and not-updated comment

temporal-test-server/src/main/proto/temporal/api/testservice/v1/service.proto
```diff
@@ -0,0 +1,63 @@
...
+service TestService {
+    // LockTimeSkipping disables time skipping.
+    //
+    // To enable back call UnlockTimeSkipping. These calls are counted,
```
Sushisource, line 44:
```suggestion
    // To re-enable call UnlockTimeSkipping. These calls are counted,
```

temporal-test-server/src/main/proto/temporal/api/testservice/v1/service.proto
```diff
@@ -0,0 +1,63 @@
...
+    // so calling unlock does not guarantee that time is going to be skipped immediately
+    // as another lock can be holding it.
+    rpc LockTimeSkipping (LockTimeSkippingRequest) returns (LockTimeSkippingResponse) {
+    }
+
+    // UnlockTimeSkipping
+    rpc UnlockTimeSkipping (UnlockTimeSkippingRequest) returns (UnlockTimeSkippingResponse) {
+    }
+
+    // SkipTime fasts forward the server time by the duration specified in the request
```
Sushisource, line 54:
```suggestion
    // SkipTime fast forwards the server time by the duration specified in the request
```

But only if time skipping is currently unlocked, correct? If it is currently locked, is the request "queued" or is it simply dropped? Would be good to elaborate here.

temporal-testing/src/main/java/io/temporal/testing/IdempotentTimeLocker.java
```diff
@@ -0,0 +1,70 @@
...
+            // 1. make sure that this operation is actually performed and
+            // 2. more importantly, don't override and hide any underlying exceptions
+            testServiceStubs
+                .blockingStub()
+                .lockTimeSkipping(LockTimeSkippingRequest.newBuilder().setCallerId(caller).build());
+          });
+    }
+  }
+
+  public void unlockTimeSkipping(String caller) {
```
Sushisource, line 55: If there are other clients not using this idempotent locker, when this returns time also might not be un-locked. Is that a potential issue? Is the expectation that, if you're using this class, every lock/unlock goes through this class?

temporal-test-server/src/main/proto/temporal/api/testservice/v1/service.proto
Spikhalskiy: Nope. Actually, when it's locked. When it's unlocked - it fast forwards how much it can.

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestService.java
```diff
@@ -0,0 +1,118 @@
...
+    workflowStore
+        .getTimer()
+        .schedule(
+            duration,
+            () -> {
+              workflowStore.getTimer().lockTimeSkipping("TestWorkflowService sleep");
+              result.complete(null);
+            },
+            "workflow sleep");
+    workflowStore.getTimer().unlockTimeSkipping("TestWorkflowService sleep");
```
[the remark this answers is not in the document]
Spikhalskiy: The only situation when we need to use sleep is when the time skipping is locked. Otherwise time fast forwards without limits, no sleeps is needed.
What happens here, the timer gets scheduled. After that, the time gets unlocked. After that time gets locked by the fired timer that we just scheduled.

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestService.java
```diff
@@ -0,0 +1,118 @@
...
+  public void lockTimeSkipping(String caller) {
+    workflowStore.getTimer().lockTimeSkipping(caller);
+  }
+
+  public void unlockTimeSkipping(String caller) {
+    workflowStore.getTimer().unlockTimeSkipping(caller);
+  }
+
+  /**
+   * Blocks calling thread until internal clock doesn't pass the current + duration time. Might not
```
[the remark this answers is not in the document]
Spikhalskiy: It's a copy-paste, but you are correct, will fix the wording.

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
Spikhalskiy: They shouldn't, because it's in `internal` package and has comments to don't use it directly. But I know some users who were playing with the test server quite a bit maybe using and exposing them in their own wrappers, yes.
I'm going to clean this up next time I touch code around it (even if it's the next release)

temporal-test-server/src/main/java/io/temporal/serviceclient/TestServiceStubsOptions.java
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  public static Builder newBuilder(TestServiceStubsOptions options) {
+    return new Builder(options);
+  }
+
+  public static Builder newBuilder(ServiceStubsOptions options) {
+    return options instanceof TestServiceStubsOptions
+        ? newBuilder((TestServiceStubsOptions) options)
+        : new Builder(options);
```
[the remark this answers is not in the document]
Spikhalskiy: Well, this structure was copied from `WorkflowServiceStubsOptions` and it makes sense there. Here is doesn't, but I don't want to remove it. The only reason it doesn't make sense right now is because the TestServiceStubsOptions don't add anything to ServiceStubsOptions.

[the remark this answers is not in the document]
Spikhalskiy: On the second thought... let me restructure the code to both leave the framework in place and don't leave such a strange-looking code.

temporal-testing/src/main/java/io/temporal/testing/IdempotentTimeLocker.java
Spikhalskiy: This code is a refactoring of the existing one, not a new code, so I'm not gonna tackle this in this PR.
But this (return without lock/unlock actually happening) actually shouldn't be an issue for the caller code. The only thing this wrapper is checking is that one WorkflowStub doesn't cause several locks/unlocks.
There is another problem with this code though: https://github.com/temporalio/sdk-java/issues/392 But again, not something I'm gonna tackle here.

temporal-testing/src/main/java/io/temporal/testing/IdempotentTimeLocker.java
Spikhalskiy: No, it's not a potential issue. Time is not guaranteed to be unlocked when this method returns. 

This IdempotentLocker only guarantees that time will be locked/unlocked once for a WorkflowStub (see TimeLockingInterceptor where it's used), it doesn't guarantee that the whole test process locks/unlocks only once. Quite opposite, time locking feature is counting, so it's explicitly supported that it can be locked several times. And we actually use it. For example, to disable time locking by default, we just call lockTimeSkipping WITHOUT a correspondent unlock (see https://github.com/temporalio/sdk-java/pull/1076/files#diff-ce2ffc1d1e2ae7b11577efed9e362e7148f6304a85da5e78ba6b3b45a896f15eR101)

Spikhalskiy:
@Sushisource @mmcshane completely reformulated docs on TestService (https://github.com/temporalio/sdk-java/pull/1076/files#diff-d7a107df90fe446efc696d828e3df42e829262e802060ade229d4cdddbb6fb26R39)
Please check if this more formal definition makes more sense.

PR #1095, opened 2022-03-23
Improve client WorkflowFailedException message with the reason

Spikhalskiy:
Improves top client `WorkflowFailedException` message by making it more human-readable and enriches it with the reason (terminated, canceled, timed_out, failed) to make what happened with the workflow immediately clear without looking at the root exception down the stack.

Simplifies exception wrapping and handling to avoid non-needed intermediate representation. 

Before: `WorkflowExecutionUtils.getResultFromCloseEvent` throws internal `WorkflowExecutionFailedException` and client facing `WorkflowFailedException`. `WorkflowExecutionFailedException` gets converted to `WorkflowFailedException` higher in the chain (in the stubs themselves).

After: `WorkflowExecutionUtils.getResultFromCloseEvent` throws just `WorkflowFailedException`.


temporal-sdk/src/main/java/io/temporal/client/WorkflowFailedException.java
```diff
@@ -50,23 +59,72 @@ public RetryState getRetryState() {
...
+  }
+
+  private static String getMessage(
       WorkflowExecution workflowExecution,
       String workflowType,
+      EventType closeEventType,
       long workflowTaskCompletedEventId,
       RetryState retryState) {
-    return "workflowId='"
+    return "Workflow execution "
```
cretz, line 95: This would look much nicer with Guava's `MoreObjects.ToStringHelper` :-)

cretz (approved):

temporal-sdk/src/main/java/io/temporal/client/WorkflowFailedException.java
Spikhalskiy: We try our best to don't use Guava where it's not absolutely needed :) Also, this one is not a standard toString, it's a message and it's customized a little.

cretz: I find even a string builder is easier to read than all these concatenations (granted that's what the JVM compiles this down to anyways)

PR #1099, opened 2022-03-27
Cleanup test timeskipping API before the release

Spikhalskiy:
Cleanup test time-skipping API before the release. Removing the experimental methods that are considered dead-end for now to avoid users referencing them.

A trivial framework for timeskipping added
A bunch of miscellaneous improvements helping in timeskipping investigations and discussion
A simple dome showing a dilemma with timeskipping and abandoning activities for further discussion


temporal-test-server/src/main/java/io/temporal/internal/testservice/SelfAdvancingTimer.java
```diff
@@ -94,6 +94,10 @@
 }
 
 interface LockHandle {
+  /**
+   * Only call it if the same caller that takes the lock releases it, otherwise call {@link
+   * #unlock(String)}
```
Sushisource, line 99:
```suggestion
   * Only call this if the same caller that took the lock will release it, otherwise call {@link
   * #unlock(String)}
```

temporal-test-server/src/main/java/io/temporal/internal/testservice/SelfAdvancingTimerImpl.java
```diff
@@ -153,10 +152,10 @@ private void runLocked() {
         if (peekedTask != null) {
           timeToAwait = peekedTask.getExecutionTime() - currentTimeMs;
           if (log.isTraceEnabled()) {
-            log.trace("Waiting for {}", Duration.ofMillis(timeToAwait));
+            log.trace("Waiting for {} with outstanding tasks:", Duration.ofMillis(timeToAwait));
             for (TimerTask task : tasks) {
               log.trace(
-                  "Outstanding task: +{} {} {}",
+                  "    +{} <{}>, cancelled: {}",
```
Sushisource, line 158: This isn't really all that important, but I tend to prefer combining log lines like this that go together into one big one (possibly still with line breaks within it) so that interleaving doesn't lose context.

Given this is test server though, not all that important.

temporal-test-server/src/main/java/io/temporal/internal/testservice/SelfAdvancingTimerImpl.java
```diff
@@ -207,27 +206,39 @@ public TimerLockHandle(LockEvent event) {
...
+            "{-"
+                + caller
+                + "}"
+                + " {+"
+                + event.caller
+                + "}"); // the caller releases lock taken by another caller using the same handle
         condition.signal();
       } finally {
         lock.unlock();
       }
```
Sushisource, line 232: It looks like both of these `unlock` overloads could share one impl just passing the different string in.

A comment on it would be nice too about why it locks/unlocks in the same method

temporal-test-server/src/main/java/io/temporal/internal/testservice/SelfAdvancingTimerImpl.java
```diff
@@ -207,27 +206,39 @@ public TimerLockHandle(LockEvent event) {
 
     @Override
     public void unlock() {
-      unlock(null);
+      lock.lock();
+      try {
+        unlockFromHandleLocked(event.caller + "[⎌]"); // caller releases its own lock
```
Sushisource, line 211: The symbol is cute but hard to search for in logs. I'd just use some bit of text

temporal-test-server/src/main/proto/temporal/api/testservice/v1/service.proto
```diff
@@ -70,17 +70,16 @@ service TestService {
...
-    rpc Skip (SkipRequest) returns (SkipResponse) {
-    }
-
-    // This call adjusts time to the specified timestamp irrespectively if the time is locked.
-    // If the current Test Server Time is beyond the specified timestamp, does nothing.
-    // Returns immediately.
-    // This is an EXPERIMENTAL API.
-    rpc SkipTo (SkipToRequest) returns (SkipToResponse) {
+    // UnlockTimeSkippingWhileSleep decreases time locking counter by one and increases it back
+    // once the Test Server Time advances by the duration specified the the request.
```
Sushisource, line 74:
```suggestion
    // once the Test Server Time advances by the duration specified in the request.
```

temporal-test-server/src/main/proto/temporal/api/testservice/v1/service.proto
```diff
@@ -70,17 +70,16 @@ service TestService {
...
+    //
+    // This call returns only when the Test Server Time advances by the specified duration.
+    //
+    // If it is called when Time Locking Counter is
+    //   - more than 1 and no other unlocks are coming in, rpc call will block for the specified duration, time will not be fast forwarded.
+    //   - 1, it will lead to fast forwarding of the time by the duration specified in the request and quick return of this rpc call.
+    //   - 0 will lead to rpc call failure same way as an unbalanced UnlockTimeSkipping.
+    rpc UnlockTimeSkippingWhileSleep (SleepRequest) returns (SleepResponse) {
```
Sushisource, line 82:
```suggestion
    rpc UnlockTimeSkippingWhileSleeping (SleepRequest) returns (SleepResponse) {
```

Can be applied to the other method too

Sushisource (approved):
Minor stuff

temporal-test-server/src/main/java/io/temporal/internal/testservice/SelfAdvancingTimerImpl.java
Spikhalskiy: I used text before this symbol, this makes it easier to understand what's going on, especially when you look through longish lock histories. Mostly becuase lines can be logins sometimes to clearly identify the caller. Also, it makes it easier to search because it's a very specific symbol.

PR #1109, opened 2022-04-01
Added local whitesource config files

mcbryde:
## What was changed
Added a local .whitesource and whitesource.config files, to exclude Gradle test scopes.

## Why?
Test scopes are:
1. Not expected to be in the classpath at runtime
2. Not included into dependencies descriptor of the jar publishing into maven central repos
3. Not used for compilation of the sources that end up in the target jar.


Spikhalskiy (approved):

Spikhalskiy:
Awesome, thanks for doing it!

PR #1122, opened 2022-04-10
Improve PotentialDeadlockException message with timestamps

Spikhalskiy:
- Potential Deadlock detection firing time
- The time when SDK was able to obtain a stacktrace of a potentially deadlocked thread
- A timestamp when a thread dump of other threads was taken 
could be different because of JVM safepoints. To make it more clear to the investigator, timestamps are now included in the Potential Deadlock message.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowThreadContext.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowThreadContext.java
@@ -114,7 +114,7 @@ public class WorkflowThreadContext {
    *
    * <p>Should be called under the lock
    *
-   * @param reason human readable reason for current thread blockage passed to await call.
+   * @param reason human-readable reason for current thread blockage passed to await call.
    */
   private void maybeEvaluateLocked(String reason) {
     if (status == Status.EVALUATING) {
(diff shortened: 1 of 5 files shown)


temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowThreadContext.java
```diff
@@ -123,17 +123,17 @@ private void maybeEvaluateLocked(String reason) {
       } catch (Exception e) {
         evaluationFunction.apply(Throwables.getStackTraceAsString(e));
       } finally {
-        status = Status.YIELDED;
+        status = Status.RUNNING;
```
Spikhalskiy, line 126: It's a more correct state to switch after evaluation. If running conditions are not met, we will go into YIELDED anyway after the check.

Before this change the following chains of states were possible:
EVALUATING -> YIELDED
EVALUATING -> YIELDED (shortly) -> RUNNING and YIELDED will be set and observable without actually yielding. Ideally, we want to switch to and publish YIELDED state only when we ACTUALLY yield.

After the change, we can get
EVALUATING -> RUNNING (shortly) -> YIELDED
or EVALUATING -> RUNNING
and we go into YIELDED ONLY if we actually yield. This is a much more consistent state model. Publishing of RUNNING in a transitional state is much safer than YIELDED.

cretz (approved):

PR #1123, opened 2022-04-10
Fix for potential spurious wake-ups in WorkflowThreadContext

Spikhalskiy:
Some awaits in WorkflowThreadContext were not shielded against spurious wake-ups with condition checks or were shielded with condition checks that were not safe.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowThreadContext.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowThreadContext.java
@@ -19,6 +19,7 @@
 
 package io.temporal.internal.sync;
 
+import com.google.common.base.Preconditions;
 import com.google.common.base.Throwables;
 import io.temporal.internal.common.DebugModeUtils;
 import io.temporal.workflow.Functions;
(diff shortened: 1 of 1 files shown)


bergundy (commented):
Are there any concerns of breaking history compatibility with this change?

Spikhalskiy:
@bergundy nope, it's to eliminate potential rare spurious wake-ups and race conditions. Normal executions will not be affected by this change.

temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowThreadContext.java
```diff
@@ -123,33 +130,35 @@ private void maybeEvaluateLocked(String reason) {
       } catch (Exception e) {
         evaluationFunction.apply(Throwables.getStackTraceAsString(e));
       } finally {
-        status = Status.YIELDED;
+        status = Status.RUNNING;
```
Spikhalskiy, line 133: It's a more correct state to switch after evaluation. If running conditions are not met, we will go into YIELDED anyway after the check.

Before this change the following chains of states were possible: 
EVALUATING -> YIELDED
EVALUATING -> YIELDED (shortly) -> RUNNING and YIELDED will be set and observable without actually yielding. Ideally, we want to switch to and publish YIELDED state only when we ACTUALLY yield.

After the change, we can get 
EVALUATING -> RUNNING (shortly) -> YIELDED 
or EVALUATING -> RUNNING
and we go into YIELDED ONLY if we actually yield. This is a much more consistent state model.

Spikhalskiy:
After thinking more, this is reduced to a small change that is not worth a separate pr, will just include it into another PR touching this class for simplicity.

PR #1134, opened 2022-04-12
State Machines now don't try to handle already handled events a second time

### LINKED ISSUE #643: Unexpected WFT Completed events in java bench runs
The Java benches have consistently been having issues with the `reactor_stress` test, resulting in worker logs that look like the below. Eventually this causes the test to time out. I would not discount the possibility that the bencher itself is actually the problem here and not something in the SDK

```
16:54:23.234 [Workflow Executor taskQueue="temporal-bench", namespace="java-bench": 13767] ERROR i.t.i.r.ReplayWorkflowTaskHandler - Workflow task failure. startedEventId=73, WorkflowId=config-driven-throughputstress-re │
│ actor_stress_i9wr-0-585, RunId=fc0f95b5-58c6-48d4-b069-cf088d2be395. If see continuously the workflow might be stuck.                                                                                                       │
│ io.temporal.internal.replay.InternalWorkflowTaskException: Failure handling event 67 of 'EVENT_TYPE_WORKFLOW_TASK_COMPLETED' type. IsReplaying=true, PreviousStartedEventId=66, workflowTaskStartedEventId=73, Currently Pr │
│ ocessing StartedEventId=66                                                                                                                                                                                                  │
│     at io.temporal.internal.statemachines.WorkflowStateMachines.handleEvent(WorkflowStateMachines.java:193)                                                                                                                 │
│     at io.temporal.internal.replay.ReplayWorkflowRunTaskHandler.handleEvent(ReplayWorkflowRunTaskHandler.java:140)                                                                                                          │
│     at io.temporal.internal.replay.ReplayWorkflowRunTaskHandler.handleWorkflowTaskImpl(ReplayWorkflowRunTaskHandler.java:180)                                                                                               │
│     at io.temporal.internal.replay.ReplayWorkflowRunTaskHandler.handleWorkflowTask(ReplayWorkflowRunTaskHandler.java:150)                                                                                                   │
│     at io.temporal.internal.replay.ReplayWorkflowTaskHandler.handleWorkflowTaskWithEmbeddedQuery(ReplayWorkflowTaskHandler.java:201)                                                                                        │
│     at io.temporal.internal.replay.ReplayWorkflowTaskHandler.handleWorkflowTask(ReplayWorkflowTaskHandler.java:114)                                                                                                         │
│     at io.temporal.internal.worker.WorkflowWorker$TaskHandlerImpl.handle(WorkflowWorker.java:319)                                                                                                                           │
│     at io.temporal.internal.worker.WorkflowWorker$TaskHandlerImpl.handle(WorkflowWorker.java:279)                                                                                                                           │
│     at io.temporal.internal.worker.PollTaskExecutor.lambda$process$0(PollTaskExecutor.java:73)                                                                                                                              │
│     at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)                                                                                                                                      │
│     at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)                                                                                                                                      │
│     at java.lang.Thread.run(Thread.java:748)                                                                                                                                                                                │
│ Caused by: java.lang.IllegalArgumentException: Unexpected event:event_id: 67                                                                                                                                                │
│ event_time {                                                                                                                                                                                                                │
│   seconds: 1629219263                                                                                                                                                                                                       │
│   nanos: 10300937                                                                                                                                                                                                           │
│ }                                                                                                                                                                                                                           │
│ event_type: EVENT_TYPE_WORKFLOW_TASK_COMPLETED                                                                                                                                                                              │
│ task_id: 2107374                                                                                                                                                                                                            │
│ workflow_task_completed_event_attributes {                                                                                                                                                                                  │
│   scheduled_event_id: 65                                                                                                                                                                                                    │
│   started_event_id: 66                                                                                                                                                                                                      │
│   identity: "1@java-bench-temporal-bench-6f6f76fccf-hpxgm"                                                                                                                                                                  │
│ }                                                                                                                                                                                                                           │
│     at io.temporal.internal.statemachines.WorkflowStateMachines.handleNonStatefulEvent(WorkflowStateMachines.java:419)                                                                                                      │
│     at io.temporal.internal.statemachines.WorkflowStateMachines.handleEventImpl(WorkflowStateMachines.java:215)                                                                                                             │
│     at io.temporal.internal.statemachines.WorkflowStateMachines.handleEvent(WorkflowStateMachines.java:178)                                                                                                                 │
│     ... 11 common frames omitted 
```

Spikhalskiy:
## What was changed
State Machines now don't try to handle already handled events a second time

## Why?
There is a tricky situation where a legacy query path can restore cached workflow in a more advanced state that the sticky queue remembered it which leads to subsequent WFT with partial history to have duplicated already processed event.
This fix shields against partial histories containing already handled events.

## How was this tested
Unit test emulating the specific scenario described in https://github.com/temporalio/sdk-java/issues/643#issuecomment-1096030871 and verifying that there is no WFT failure.

Closes #643


cretz:
I wonder if this is similar to https://github.com/temporalio/sdk-go/issues/755?

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
```diff
@@ -171,6 +177,12 @@ public void setStartedIds(long previousStartedEventId, long workflowTaskStartedE
    * @param hasNextEvent false if this is the last event in the history.
    */
   public void handleEvent(HistoryEvent event, boolean hasNextEvent) {
+    long eventId = event.getEventId();
+    if (eventId <= lastHandledEventId) {
```
cretz, line 181: Is it possible that this could adversely affect code paths in any backwards compatible way with existing history? I wouldn't  think so since it's fixing a cache expiration edge case, I just want to confirm from your POV.

temporal-sdk/src/test/java/io/temporal/worker/StickyWorkerTest.java
```diff
@@ -471,74 +462,94 @@ public void workflowsCanBeQueriedAfterEviction() throws Exception {
...
     assertEquals(workflow.getProgress(), GreetingSignalWorkflow.Status.WAITING_FOR_NAME);
+    SDKTestWorkflowRule.assertNoHistoryEvent(
+        wrapper.testEnv.getWorkflowExecutionHistory(execution).getHistory(),
+        EventType.EVENT_TYPE_WORKFLOW_TASK_FAILED);
 
+    wrapper.close();
+  }
+
+  @Test
+  public void workflowCanContinueExecutionAfterBeingEvictedAndQueried() throws Exception {
```
cretz, line 473: Can you confirm that this test would fail if this fix wasn't in place?

temporal-sdk/src/test/java/io/temporal/worker/StickyWorkerTest.java
```diff
@@ -471,74 +462,94 @@ public void workflowsCanBeQueriedAfterEviction() throws Exception {
...
+            .setWorkflowIdReusePolicy(
+                WorkflowIdReusePolicy.WORKFLOW_ID_REUSE_POLICY_REJECT_DUPLICATE)
+            .build();
+    GreetingSignalWorkflow workflow =
+        wrapper.getWorkflowClient().newWorkflowStub(GreetingSignalWorkflow.class, workflowOptions);
+
+    // Act
+    WorkflowExecution execution = WorkflowClient.start(workflow::getGreeting);
+
+    Thread.sleep(200); // Wait for workflow to start
```
cretz, line 497: In order to avoid flakiness, in Go SDKs and others we often have a query setup on the workflow and poll until it returns instead of relying on time.

cretz (approved):

temporal-sdk/src/test/java/io/temporal/worker/StickyWorkerTest.java
Spikhalskiy: Yep, I checked it.

temporal-sdk/src/test/java/io/temporal/worker/StickyWorkerTest.java
Spikhalskiy: Yeah, same in Java, this class needs a little bit more love that I'm not ready to give now to do that I believe. It's a part of Test Rule in Java and this class doesn't use the test rule. But let me double check if I can hook it in anyway.

cretz:
Per off-PR discussion, marking approve with more off-PR discussion to come.

Specifically, here is the scenario in Java today:

* Encode
  * string -> SA string
  * list of one string -> SA string
  * list of multiple strings -> SA list of multiple strings
* Decode
  * SA string -> list of one string (can be read as just string w/ separate getter though)
  * SA list of multiple strings -> list of multiple strings (fails if trying to use that single-item getter though)

While this makes sense for Java, in Go I think we might not unwrap a list of one string as just a string but rather leave as list. Unsure about TypeScript.

Sushisource (approved):
I can confirm I ended up doing this same thing in Core.

PR #1144, opened 2022-04-14
Fix workflow thread name strategy to append a clear marker of truncation

Spikhalskiy:
Now if we truncate workflowId before we append it to the workflow thread name, we truncate it and it may be unclear and misleading in some cases. Add a clear marker at the end of workflowId if we indeed truncate.


cretz (approved):

temporal-sdk/src/main/java/io/temporal/internal/worker/workflow/ExecutionInfoStrategy.java
```diff
@@ -30,8 +30,14 @@ private ExecutionInfoStrategy() {}
   @Override
   public String createThreadName(WorkflowExecution workflowExecution) {
     String workflowId = workflowExecution.getWorkflowId();
+
     String trimmedWorkflowId =
-        workflowId.substring(0, Math.min(WORKFLOW_ID_TRIM_LENGTH, workflowId.length()) - 1);
+        workflowId.length() > WORKFLOW_ID_TRIM_LENGTH
+            ?
+            // add a ' at the end to explicitly show that the id was trimmed
```
cretz, line 37: I prefer `...`, but this works too

PR #1147, opened 2022-04-17
Reimplemented WorkflowRunLockManager to fix design flaw with an unsafe unlock

### LINKED ISSUE #1146: Code using WorkflowRunLockManager may release locks that are owned by other threads
WorkflowWorker.TaskHandlerImpl#handle:

```
if (...) {
        runLock = runLocks.getLockForLocking(task.getWorkflowExecution().getRunId());
        // Acquiring a lock with a timeout to avoid having lots of workflow tasks for the same run
        // id waiting for a lock and consuming threads in case if lock is unavailable.
        if (!runLock.tryLock(1, TimeUnit.SECONDS)) {
          throw new UnableToAcquireLockException(
              "Workflow lock for the run id hasn't been released by one of previous execution attempts, "
                  + "consider increasing workflow task timeout.");
        }
      }
      ...
      } finally {
        if (runLock != null) {
          System.out.println("released a lock for: " + task.getWorkflowExecution().getRunId());
          runLocks.unlock(task.getWorkflowExecution().getRunId());
        }
      }
```

If the exception is thrown and the lock is not taken
- it means that another thread acquired the lock and still processing this runId
- the `finally` block will actually release the lock while another thread is still working on the task
If the task gets distributed to the same worker after that, two threads will start working on the same instance of the workflow, which will lead to all kinds of problems, because workflow state machines are expected to be managed by one thread only.

This scenario is not happening on a normal happy path. This scenario usually doesn't happen on a bad path too, because if JavaSDL detects blocking of workflow task, it kills it with DeadlockDetector. Situations where local activities keep Workflow Task open also have special handling. But it may happen if something causes blocking of workflow tasks that are not getting detected by the Deadlock Detector.

Spikhalskiy:
## What was changed

WorkflowRunLockManager was reimplemented in a safe way. Now unlock for a runId can be called only from a thread that acquired the lock.

## Why?
See #1146. The current API and implementation of WorkflowRunLockManager may lead to a lock for a runId being released by a thread that is not eligible for that.

Closes #1146

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowRunLockManager.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowRunLockManager.java
@@ -20,99 +20,65 @@
 package io.temporal.internal.worker;
 
 import com.google.common.annotations.VisibleForTesting;
-import java.util.HashMap;
+import com.google.common.base.Preconditions;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.locks.Condition;
 import java.util.concurrent.locks.ReentrantLock;
 
 final class WorkflowRunLockManager {
-  // This is a single lock for the whole worker.
-  // It's ok, because it's acquired for a very short time.
-  // If this ever becomes a bottleneck, consider:
-  //   - rework this lock to use com.google.common.util.concurrent.Striped and
-  //   - make `perRunLock` a ConcurrentHashMap to increase parallelism of this code
-  private final ReentrantLock mapLock = new ReentrantLock();
-  private final Map<String, LockData> perRunLock = new HashMap<>();
+  private final Map<String, RefCountedLock> runIdLock = new ConcurrentHashMap<>();
 
-  /**
-   * @param runId to take a lock for
-   * @param time the maximum time to wait for the lock
-   * @param unit the time unit of the time argument
-   * @return true if the lock is taken
-   */
-  boolean tryLock(String runId, long time, TimeUnit unit) throws InterruptedException {
-    mapLock.lock();
+  public boolean tryLock(String runId, long timeout, TimeUnit unit) throws InterruptedException {
+    RefCountedLock runLock =
+        runIdLock.compute(
+            runId,
+            (id, lock) -> {
+              if (lock == null) {
+                lock = new RefCountedLock();
+              }
+              lock.refCount++;
+              return lock;
+            });
+
+    boolean obtained = false;
     try {
-      long timeoutNanos = unit.toNanos(time);
-      LockData lockData = tryPersistLockByCurrentThreadLocked(runId);
-      while (Thread.currentThread() != lockData.thread) {
-        if (timeoutNanos <= 0) {
-          return false;
-        }
-        timeoutNanos = lockData.condition.awaitNanos(timeoutNanos);
-        lockData = tryPersistLockByCurrentThreadLocked(runId);
-      }
-
-      lockData.count++;
-      return true;
+      obtained = runLock.lock.tryLock(timeout, unit);
+      return obtained;
     } finally {
-      mapLock.unlock();
+      if (!obtained) {
+        derefAndUnlock(runId, false);
+      }
     }
   }
 
-  private LockData tryPersistLockByCurrentThreadLocked(String runId) {
-    return perRunLock.computeIfAbsent(
-        runId, id -> new LockData(Thread.currentThread(), mapLock.newCondition()));
+  public void unlock(String runId) {
+    derefAndUnlock(runId, true);
   }
 
-  void unlock(String runId) {
-    mapLock.lock();
-    try {
-      LockData lockData = perRunLock.get(runId);
-      if (lockData == null) {
-        throw new IllegalStateException("Lock for " + runId + " is not taken");
-      }
-      if (Thread.currentThread() == lockData.thread) {
-        lockData.count--;
-        if (lockData.count == 0) {
-          perRunLock.remove(runId);
-          // it's important to signal all threads,
-          // otherwise n-1 of them will be stuck waiting on a condition that is not in the map
-          // already
-          lockData.condition.signalAll();
-        }
-      } else {
-        throw new IllegalStateException(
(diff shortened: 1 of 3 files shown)


temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
```diff
@@ -211,16 +210,37 @@ public void handle(PollWorkflowTaskQueueResponse task) throws Exception {
...
+        // TODO 2: Does "consider increasing workflow task timeout" advice in this exception makes
+        //   any sense?
+        //   This MAYBE makes sense only if a previous workflow task timed out, it's still in
+        //   progress on the worker and the next workflow task got picked up by the same exact
+        //   worker from the general non-sticky task queue.
+        //   Even in this case, this advice looks misleading, something else is going on
+        //   (like an extreme network latency).
+        locked = runLocks.tryLock(runId, 1, TimeUnit.SECONDS);
+
+        if (!locked) {
```
Spikhalskiy, line 243: The change behind this line in this class is refactoring which makes the logic easier to follow, and may be ignored during the review.

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowRunLockManager.java
```diff
@@ -19,89 +19,101 @@
...
-
-      cl.decrementCount();
-      if (cl.getCount() == 0) {
-        perRunLock.remove(runId);
+      if (Thread.currentThread() == lockData.thread) {
+        lockData.count--;
+        if (lockData.count == 0) {
+          perRunLock.remove(runId);
+          // it's important to signal all threads,
+          // otherwise n-1 of them will stuck waiting on a condition that is not in the map already
```
bergundy, line 83:
```suggestion
          // otherwise n-1 of them will be stuck waiting on a condition that is not in the map already
```

bergundy (commented):
Hard to review this without more knowledge of Java SDK internals.

If 1 second refers to net workflow code processing time it sounds like enough but if that includes data converters and local activities, either of which could do I/O then that is definitely insufficient.

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowRunLockManager.java
```diff
@@ -19,89 +19,101 @@
...
-      } else {
-        cl.incrementCount();
+      Date deadline = new Date(System.currentTimeMillis() + unit.toMillis(time));
+      boolean stillWaiting = true;
+      LockData lockData = tryPersistLockByCurrentThreadLocked(runId);
+      while (Thread.currentThread() != lockData.thread) {
+        if (!stillWaiting) {
+          return false;
+        }
+        stillWaiting = lockData.condition.awaitUntil(deadline);
```
cretz, line 55: Can you use `awaitNanos` here and use the return value for the next loop? `System.currentTimeMillis()` is not technically monotonic so these uses scare me.

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowRunLockManager.java
```diff
@@ -19,89 +19,101 @@
...
-      CountableLock cl = perRunLock.get(runId);
-      if (cl == null) {
-        cl = new CountableLock();
-        perRunLock.put(runId, cl);
-      } else {
-        cl.incrementCount();
+      Date deadline = new Date(System.currentTimeMillis() + unit.toMillis(time));
+      boolean stillWaiting = true;
+      LockData lockData = tryPersistLockByCurrentThreadLocked(runId);
+      while (Thread.currentThread() != lockData.thread) {
```
cretz, line 51: From what I'm reading, this will never evaluate to true on successive iterations. How can the map remove an old lock data (i.e. invoking/persisting `computeIfAbsent` result) if you've locked the entire map here with `mapLock`? If this is just for the first invocation, an `if` is clearer.

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowRunLockManager.java
```diff
@@ -19,89 +19,101 @@
```
cretz, line 19: I think the approach from https://stackoverflow.com/a/23920910 is easier to follow. `ConcurrentHashMap` + `BlockingQueue` since the latter has `offer` and `poll` with built in timeouts.

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowRunLockManager.java
Spikhalskiy: Spurious wake-ups

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowRunLockManager.java
Spikhalskiy: The approach from this link will not work because it will leak. If the approach by this link would be acceptable, I would just make a map from keys to lock and be done with it.

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowRunLockManager.java
Spikhalskiy: That's a great advice, reworked. Thanks!

cretz (approved):

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
mfateev: I think it should be changed to the workflow task timeout.

Spikhalskiy: Thanks for confirming that this 1 sec looks off! Let's discuss more and change it. While important to do, unrelated to the scope of this PR, we can do it separately.

Spikhalskiy: It's more to actually let the workflow task in flight to finish. Workflow task timeout is about 10s, deadlock detector is 1 second by default, seconds typically, there is more room to wait more than 1 second typically.

PR #1151, opened 2022-04-18
Make WorkflowThreadContext#destroy non-blocking and remove deadlock detector from the call

Spikhalskiy:


temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowThreadContext.java
```diff
@@ -381,4 +383,14 @@ public Thread getCurrentThread() {
       lock.unlock();
     }
   }
+
+  /** Never throws, prerequisites should be checked by the calling code */
+  private void runLocked() {
+    inRunUntilBlocked = true;
```
Sushisource, line 389: This name is now confusing or this assignment is a bit of a lie, as this could be from `destroy` now

temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowThreadContext.java
```diff
@@ -318,7 +312,11 @@ public boolean isDestroyRequested() {
     }
   }
 
-  public void destroy() {
+  /**
+   * Non-blocking call, never throws.<br>
+   * There is no guarantee that the thread is destroyed at the end of this call.
+   */
+  void destroy() {
```
Sushisource, line 319: This should maybe be called tryDestroy or something since it's not guaranteed to actually destroy

Spikhalskiy: It does guarantee, it just doesn't block on it. I will rename to `initiateDestroy`.

Sushisource (approved):

PR #1159, opened 2022-04-21
Fix sticky_cache_size gauge behavior

Spikhalskiy:
sticky_cache_size is broken for some time and is published only during eviction from the sticky cache.
This PR fixes it and covers with a test.


bergundy (approved):
Is there any way we can centralize all of these gauge updates?

Spikhalskiy:
@bergundy not with the current tally interface. We should get rid of tally at some moment and get direct access to micrometer interfaces which are very nice. For example, you can just register something on a gauge and it will just periodically calculate a new value.

PR #1168, opened 2022-04-26
Explicit methods for lazy and eager connecting client, new #healthCheck method on stubs

### LINKED ISSUE #828: Expose health check on WorkflowServiceStubs or WorkflowClient
Temporal SDK should expose a method like `WorkflowServiceStubs#healthCheck` or `WorkflowServiceStubs#isHealthy` that allows users to check if the WorkflowClient/WorkflowServiceStubs underlying channel / connection to Temporal server is healthy.
There were quite a bit of user requests for such a method from users for all SDKs.
Having this method on Stubs is preferable over WorkflowClient, because it's the lowest level abstraction on top of the channel and WorkflowClient can just fall back on the WorkflowServiceStubs method. But if the language doesn't have a wrapper for stubs (like Java does), this method may be on WorkflowClient.

Issues for other SDKs:
https://github.com/temporalio/sdk-go/issues/595
https://github.com/temporalio/sdk-node/issues/312

Spikhalskiy:
Add and Implemented `ServiceStubs#healthCheck()`, make eager health check during client creation disabled by default.

Closes #828
Corresponded GoSDK change: https://github.com/temporalio/sdk-go/pull/795

### DIFF
--- a/temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsImpl.java
+++ b/temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsImpl.java
@@ -24,8 +24,10 @@ import io.grpc.*;
 import io.grpc.health.v1.HealthCheckResponse;
 import io.temporal.api.workflowservice.v1.WorkflowServiceGrpc;
 import io.temporal.internal.testservice.InProcessGRPCServer;
+import java.time.Duration;
 import java.util.Collections;
 import java.util.concurrent.TimeUnit;
+import javax.annotation.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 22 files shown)


bergundy:
@Spikhalskiy looks like we need to discuss this behavior again, waiting with this review.

Spikhalskiy:
@bergundy that's why it's a draft ;) I will finalize it after the discussion of our contract.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubsImpl.java
```diff
@@ -159,6 +150,17 @@ public boolean awaitTermination(long timeout, TimeUnit unit) {
     return true;
   }
 
+  @Override
+  public void connect(@Nullable Duration timeout) {
+    channelManager.connect(HEALTH_CHECK_SERVICE_NAME, timeout);
```
cretz, line 155: I think "get system info" can replace health check for this use case (and therefore you may want to change `disableHealthCheck` option to `disableEagerConnection`). I don't think there's value in doing both for workflow service.

Spikhalskiy: It will, yeah, I just didn't implement it yet. It will be in a separate PR. Don't want to mix concerns here.

cretz: :+1:

PR #1172, opened 2022-04-27
Add per-module Jacoco config

Spikhalskiy:
Add Jacoco configuration that produces reports for each module.
Some tests refactoring to remove flakes caused by increased execution times with instrumentation for code coverage.

### DIFF
--- a/temporal-sdk/src/test/java/io/temporal/workflow/StartTest.java
+++ b/temporal-sdk/src/test/java/io/temporal/workflow/StartTest.java
@@ -40,7 +40,7 @@ public class StartTest {
       SDKTestWorkflowRule.newBuilder().setWorkflowTypes(TestMultiArgWorkflowImpl.class).build();
 
   @Test
-  public void testStart() {
+  public void startNoArgFuncWithRejectDuplicate() {
     WorkflowOptions workflowOptions =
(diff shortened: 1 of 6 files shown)


temporal-sdk/src/test/java/io/temporal/workflow/StartTest.java
```diff
@@ -53,18 +53,26 @@ public void testStart() {
...
         "func", stubF.func()); // Check that duplicated start just returns the result.
-    WorkflowOptions options =
-        WorkflowOptions.newBuilder().setTaskQueue(testWorkflowRule.getTaskQueue()).build();
-    Test1ArgWorkflowFunc stubF1 =
-        testWorkflowRule.getWorkflowClient().newWorkflowStub(Test1ArgWorkflowFunc.class, options);
+  }
 
+  @Test
+  public void startOneArgsFuncWithDefault() {
+    // TODO why it doesn't work with external service?
```
Sushisource, line 60: Does this need to be addressed in this PR?

Sushisource (approved):

temporal-sdk/src/test/java/io/temporal/workflow/StartTest.java
Spikhalskiy: No, only this note is new, the code and exclusion of the external service were always there. Completely irrelevant to the nature of this PR. Left it for the future me mostly.

PR #1176, opened 2022-04-29
Unify the limit on gRPC incoming message across SDKs to 128Mb

Spikhalskiy:
Unify the limit on gRPC incoming message across SDKs to 128Mb.

GoSDK change: https://github.com/temporalio/sdk-go/pull/796

### DIFF
--- a/temporal-serviceclient/src/main/java/io/temporal/serviceclient/ChannelManager.java
+++ b/temporal-serviceclient/src/main/java/io/temporal/serviceclient/ChannelManager.java
@@ -39,7 +39,13 @@ import org.slf4j.LoggerFactory;
 final class ChannelManager {
   private static final Logger log = LoggerFactory.getLogger(ChannelManager.class);
 
-  private static final int MAX_INBOUND_MESSAGE_SIZE = 25_000_000;
+  /**
+   * This value sets the limit on the incoming responses from Temporal Server to Java SDK. It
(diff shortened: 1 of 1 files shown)


temporal-serviceclient/src/main/java/io/temporal/serviceclient/ChannelManager.java
```diff
@@ -39,7 +39,13 @@
...
   private static final Logger log = LoggerFactory.getLogger(ChannelManager.class);
 
-  private static final int MAX_INBOUND_MESSAGE_SIZE = 25_000_000;
+  /**
+   * This value sets the limit on the incoming responses from Temporal Server to Java SDK.
+   * It doesn't affect the limit that Temporal Server exposes on the messages coming to
+   * Temporal Frontend from Java SDK.
+   * The Server Frontend limit on incoming gRPC message value is currently 4Mb.
+   */
+  private static final int MAX_INBOUND_MESSAGE_SIZE = 128 * 1024 * 1024;
```
bergundy, line 48: What about outbound? Did we decide on max size for that?

bergundy (approved):

temporal-serviceclient/src/main/java/io/temporal/serviceclient/ChannelManager.java
Spikhalskiy: The server does it by limiting its inbound. And it's 4Mb now (gRPC default)

PR #1185, opened 2022-05-03
Fix workflow_active_thread_count, abstract WorkflowThreadExecutor instead of using a broad thread pool

### LINKED ISSUE #1179: workflow_active_thread_count metric should not be tagged with WORKFLOW_TYPE
## Expected Behavior

Every workflow per worker [emits](https://github.com/temporalio/sdk-java/blob/3db50b6d9dd60b17a3b516353cf19ddb9065abd7/temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowThreadImpl.java#L244) `workflow_active_thread_count` which is the number of active threads in the shared `ThreadPoolExecutor`.

This metric should not contain the `WORKFLOW_TYPE` tag to avoid double-counting of `workflow_active_thread_count` given that every workflow share the same `ThreadPoolExecutor`.

## Actual Behavior

This metric **contains** the `WORKFLOW_TYPE` tag which leads to double-counting, when aggregated by the host, of `workflow_active_thread_count` given that every workflow shared the same `ThreadPoolExecutor`.

  1. Run a 1 worker with 2 workflows
  1. Check `workflow_active_thread_count` metric. It will contain the `WORKFLOW_TYPE` tag

## Specifications

  - Version: v1.10.0

Spikhalskiy:
## What was changed

Abstract WorkflowThreadExecutor instead of using a broad thread pool in WorkflowThread class. WorkflowThread needs only the regular `submit` method.
Initialize WorkflowThreadExecutor in WorkerFactory to use the root worker metric scope.

## Why
WorkerThread doesn't need to log the worker_active_thread metric. This metric belongs to `WorkerFactory` level scope and logging of this metric should be performed on that level.

Closes #1179

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowThreadExecutor.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowThreadExecutor.java
@@ -1,0 +1,45 @@
+/*
+ *  Copyright (C) 2020 Temporal Technologies, Inc. All Rights Reserved.
+ *
+ *  Copyright 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ *  Modifications copyright (C) 2017 Uber Technologies, Inc.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License"). You may not
+ *  use this file except in compliance with the License. A copy of the License is
+ *  located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ *  or in the "license" file accompanying this file. This file is distributed on
+ *  an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ *  express or implied. See the License for the specific language governing
+ *  permissions and limitations under the License.
+ */
+
+package io.temporal.internal.sync;
+
+import java.util.concurrent.Future;
+import java.util.concurrent.RejectedExecutionException;
+import javax.annotation.Nonnull;
+
+/**
+ * Abstracts the {@link java.util.concurrent.ThreadPoolExecutor} that is used to submit workflow
+ * thread tasks to allow higher levels to define additional wrapping logic like worker-wide metric
+ * reporting.
+ */
+public interface WorkflowThreadExecutor {
+  /**
+   * Submits a Runnable task for execution and returns a Future representing that task. The Future's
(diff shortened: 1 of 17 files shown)


temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowThreadExecutor.java
```diff
@@ -0,0 +1,36 @@
...
+public interface WorkflowThreadExecutor {
+  /**
+   * Submits a Runnable task for execution and returns a Future representing that task. The Future's
+   * {@code get} method will return {@code null} upon <em>successful</em> completion.
+   *
+   * @param task the task to submit
+   * @return a Future representing pending completion of the task
+   * @throws RejectedExecutionException if the task cannot be scheduled for execution
+   * @throws NullPointerException if the task is null
+   */
```
Sushisource, line 34: Should the docstring change at all to reflect the more-specific nature of this executor?

temporal-sdk/src/main/java/io/temporal/worker/WorkerFactory.java
```diff
@@ -393,6 +397,24 @@ public String toString() {
         "WorkerFactory{identity=%s, uniqueId=%s}", workflowClient.getOptions().getIdentity(), id);
   }
 
+  private static WorkflowThreadExecutor newWorkflowThreadExecutor(
+      ThreadPoolExecutor workflowThreadPool, Scope metricsScope) {
```
Sushisource, line 401: Would it make sense for this to exist next to / inside the WorkflowThreadExecutor definition?

Sushisource (approved):

temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowThreadExecutor.java
Spikhalskiy: It's pretty much the same as the generic ThreadPool submit javadoc for #submit. But I will add an additional docstring to the class.

temporal-sdk/src/main/java/io/temporal/worker/WorkerFactory.java
Spikhalskiy: I don't think so. It uses the constants and metrics-related logic that is not reusable and much more appropriate to be in this class.

PR #1196, opened 2022-05-09
Local Activity uses Server time and elapsed period to update workflow task timestamp

### LINKED ISSUE #1187: LocalActivity completion advances workflow task timestamp to the current time instead of using an offset
## Expected Behavior

Finished LocalActivity execution advances workflow task timestamp by the execution time. If test framework is used and initial timestamp is overridden, it should be respected through the whole execution and not get discarded by a completion of a local activity. 

## Actual Behavior
`TestEnvironment` that has an initial time set to something other than the current time.
When LocalActivity runs, `Workflow.currentTimeMillis()` gets updated to the current DateTime that is nothing around the time set on TestEnv.

Spikhalskiy:
## What was changed
Now Local Activities when updating workflow task timestamp uses the workflow task timestamp when it was scheduled and a time elapsed in the execution instead of System.currentTimeMillis

## Why?
Local Activity updating time should respect the server time of the workflow task and don't use the local time of the worker host as a source of truth.

Closes #1187

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/statemachines/LocalActivityStateMachine.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/statemachines/LocalActivityStateMachine.java
@@ -41,6 +41,7 @@ import io.temporal.workflow.Functions;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.TimeUnit;
 
 final class LocalActivityStateMachine
     extends EntityStateMachineInitialCommand<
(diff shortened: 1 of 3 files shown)


temporal-sdk/src/main/java/io/temporal/internal/statemachines/LocalActivityStateMachine.java
```diff
@@ -239,10 +256,14 @@ private void createMarker() {
       details.put(MARKER_ACTIVITY_ID_KEY, id);
       Payloads type = dataConverter.toPayloads(activityType.getName()).get();
       details.put(MARKER_ACTIVITY_TYPE_KEY, type);
-      // TODO(maxim): Consider using elapsed since start instead of Sytem.currentTimeMillis
-      long currentTime = setCurrentTimeCallback.apply(System.currentTimeMillis());
+
+      long elapsedNanoseconds = System.nanoTime() - systemNanoTimeWhenStarted;
+      long currentTime =
+          setCurrentTimeCallback.apply(
+              workflowTimeMillisWhenStarted + TimeUnit.NANOSECONDS.toSeconds(elapsedNanoseconds));
```
Sushisource, line 263: This looks like it's adding seconds to millis

Sushisource (approved):

Issue #1203, opened 2022-05-11
Rework long poll code to use an absolute deadline instead of calculating an elapsed time


Spikhalskiy:
**Is your feature request related to a problem? Please describe.**
Currently, the long polling retrying code has to maintain a very unpleasant calculation of the elapsed and remaining time.
Instead of this, it should be reworked onto using an absolute deadline.

**Describe the solution you'd like**
`GrpcRetryer` should support an absolute deadline directly, without error-prone explicit converting of the absolute deadline to the timeout duration.
This absolute deadline should be passed into `GrpcRetryer` using some internal `RetryerOptions`.

Spikhalskiy:
Closed by #1209

PR #1209, opened 2022-05-12
Make getResult methods to throw TimeoutException instead of raw gRPC DEADLINE_EXCEEDED

### LINKED ISSUE #988: Make WorkflowServiceStubsOptions#rpcLongPollTimeout configurable

Right now WorkflowServiceStubsOptions#rpcLongPollTimeout has to be 70s or more because Server's default timeout is 60s.
It looks like now Server uses a timeout from the client that initiates the long poll to return the result (empty or non-empty) in time. And 60s is just a cap for this timeout.

> Yimin Chen 
> To be clear, if the context timeout on the connection is less than 60s, server would return empty response before timeout hit.
> there is a small gap server added to make sure the empty response is delivered before actual context timeout when it is less than 60s.
> 

Not allowing users to configure the long polls to be shorter than 70s, Temporal makes configuration tricky for some of the users who may be using middle layers with a strict configuration (like reverse proxies) that may limit the durtion of long polls.

Taking into account that the Server takes a time limit from the request timeout/deadline to return a response in an amount of time shorter than 60s, we can allow an SDK to accept rpcLongPollTimeout values less than the 70s. that and returns a response in a shorter amount of time.

GoSDK: https://github.com/temporalio/sdk-go/issues/705

### LINKED ISSUE #1177: WorkflowStub.getResult and getResultAsync do not throw TimeoutException
String result = workflow.getResult(X, TimeUnit.SECONDS, String.class);

If this call times out (workflow does not complete before X seconds):

1. if X < 10
You get GrpcSyncRetryer -> DEADLINE_EXCEEDED: deadline exceeded after Xs
and can catch WorkflowServiceException

3. if X >=10
You get two GrpcSyncRetryer logs:
a) DEADLINE_EXCEEDED: context deadline exceeded
b) Failure, out of retries
and can catch WorkflowServiceException

Same happens with getResultAsync

The only way to catch TimeoutException is if you use getResultAsync and put the timeout on Future.get, for example:

     CompletableFuture<String> future = workflow.getResultAsync(String.class);
     String result = future.get(X, TimeUnit.SECONDS);

Note that even in this case if X >=10 you do get DEADLINE_EXCEEDED in worker logs from GrpcAsyncRetryer

Spikhalskiy:
## What was changed

Make getResult methods to throw `TimeoutException` as the method contract states instead of raw gRPC DEADLINE_EXCEEDED.
getResultAsync now respects the timeout.
Unit tests are added that are covering different kinds of timeouts and retries of the long poll.

Closes #1177
Closes #988
Partially addresses #1203

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/client/WorkflowClientLongPollAsyncHelper.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/client/WorkflowClientLongPollAsyncHelper.java
@@ -23,6 +23,8 @@ import static io.temporal.internal.common.WorkflowExecutionUtils.getResultFromCl
 
 import com.google.protobuf.ByteString;
 import io.grpc.Deadline;
+import io.grpc.Status;
+import io.grpc.StatusRuntimeException;
 import io.temporal.api.common.v1.Payloads;
 import io.temporal.api.common.v1.WorkflowExecution;
 import io.temporal.api.enums.v1.EventType;
@@ -33,9 +35,9 @@ import io.temporal.api.workflowservice.v1.GetWorkflowExecutionHistoryResponse;
 import io.temporal.common.converter.DataConverter;
 import io.temporal.internal.client.external.GenericWorkflowClient;
 import io.temporal.internal.common.WorkflowExecutionUtils;
-import io.temporal.serviceclient.CheckedExceptionWrapper;
 import java.util.Optional;
 import java.util.concurrent.*;
+import javax.annotation.Nonnull;
 
 /** This class encapsulates async long poll logic of {@link RootWorkflowClientInvoker} */
 final class WorkflowClientLongPollAsyncHelper {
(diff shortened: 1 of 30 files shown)


temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowStubImpl.java
```diff
@@ -274,25 +272,30 @@ public <R> CompletableFuture<R> getResultAsync(
...
+              try {
+                return mapToWorkflowFailureException(e, resultClass);
+              } catch (TimeoutException ex) {
+                throw new CompletionException(ex);
               }
-              return r;
             });
   }
 
+  // this function never return anything, it only throws
```
Sushisource, line 285:
```suggestion
  // This function never returns anything, it only throws
```

temporal-sdk/src/main/java/io/temporal/internal/client/WorkflowClientLongPollAsyncHelper.java
```diff
@@ -48,11 +49,32 @@ static CompletableFuture<Optional<Payloads>> getWorkflowExecutionResultAsync(
...
       DataConverter converter) {
+    Deadline deadline = Deadline.after(timeout, unit);
     return getInstanceCloseEventAsync(
-            genericClient, workflowClientHelper, workflowExecution, ByteString.EMPTY, timeout, unit)
-        .thenApply(
-            (closeEvent) ->
-                getResultFromCloseEvent(workflowExecution, workflowType, closeEvent, converter));
+            genericClient, workflowClientHelper, workflowExecution, ByteString.EMPTY, deadline)
+        .handle(
+            (closeEvent, e) -> {
```
Sushisource, line 56: Maybe makes sense to break this out into a function

temporal-serviceclient/src/main/java/io/temporal/internal/retryer/GrpcRetryerUtils.java
```diff
@@ -58,14 +54,10 @@ class GrpcRetryerUtils {
...
-          return lastMeaningfulException(currentException, previousException);
-        } else {
-          // If this specific request's deadline has expired, but the higher-level deadline
-          // that was established when GrpcRetryer was initialized not been exceeded, we
-          // should keep retrying.
-          break;
-        }
+        // By default, we keep retrying with DEADLINE_EXCEEDED assuming that it's the deadline of
+        // one
+        // attempt may be expired, but not the whole sequence.
```
Sushisource, line 59:
```suggestion
        // By default, we keep retrying with DEADLINE_EXCEEDED assuming that it's the deadline of
        // one attempt which expired, but not the whole sequence.
```

temporal-serviceclient/src/test/java/io/temporal/serviceclient/functional/GetResultsAsyncOverLongPollTimeoutTest.java
```diff
@@ -0,0 +1,63 @@
...
+import io.temporal.testing.internal.SDKTestWorkflowRule;
+import io.temporal.workflow.Workflow;
+import java.time.Duration;
+import java.util.concurrent.ExecutionException;
+import org.junit.Rule;
+import org.junit.Test;
+
+/**
+ * These tests are intentionally running longer than a minute to cover a situation when getResult
+ * calls needs to survive a server timeout on a long poll of 60 seconds to get the result of
```
Sushisource, line 34: Any reason not to just crank down the poll timeout a bit?

Sushisource (approved):

temporal-serviceclient/src/test/java/io/temporal/serviceclient/functional/GetResultsAsyncOverLongPollTimeoutTest.java
Spikhalskiy: I want to test a situation when the server responds with an empty result, not when we hit a long poll timeout (we ideally should never hit it). And for that, I just have to get that one minute. Or modify the settings of our server, which I don't want to abuse.

PR #1227, opened 2022-05-18
Deescalate GrpcRetryer logging to trace level

### LINKED ISSUE #1208: Remove verbose exceptions logging from GrpcRetryer
`GrpcRetryer` shouldn't log the errors happening, even the final ones. 
It may do it on a Trace level for debugging purposes, but it shouldn't be higher than that.
Code that uses `GrpcRetryer` should take care of handling the exceptions propagated from there.
The main reason is that GrpcRetryer doesn't have a context to understand which exception is a problem and which is not, only the caller does.

Spikhalskiy:
This PR addresses the verbose logging made by GrpcRetryer. The client code of GrpcRetryer should be responsible for dealing with the exceptions, including not logging and not reacting to them at all because they may be not a problem.

Closes #1208

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
@@ -229,7 +229,16 @@ public final class WorkflowWorker
         do {
           PollWorkflowTaskQueueResponse currentTask = nextTask.get();
           WorkflowTaskHandler.Result response = handleTask(currentTask, workflowTypeMetricsScope);
-          nextTask = sendReply(service, workflowTypeMetricsScope, currentTask, response);
+          try {
+            nextTask =
+                sendReply(currentTask.getTaskToken(), service, workflowTypeMetricsScope, response);
+          } catch (Exception e) {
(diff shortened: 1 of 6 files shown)


temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
```diff
@@ -288,83 +297,85 @@ private WorkflowTaskHandler.Result handleTask(
...
+            currentTask.getStartedEventId(),
+            result,
+            e);
+      } else {
         log.warn(
-            "Workflow task failure during replying to the server. startedEventId={}, WorkflowId={}, RunId={}. If seen continuously the workflow might be stuck.",
-            task.getStartedEventId(),
-            execution.getWorkflowId(),
-            execution.getRunId(),
+            "Failure during reporting of workflow progress to the server. If seen continuously the workflow might be stuck. WorkflowId={}, RunId={}, startedEventId={}",
```
Sushisource, line 374:
```suggestion
            "Failure while reporting workflow progress to the server. If seen continuously the workflow might be stuck. WorkflowId={}, RunId={}, startedEventId={}",
```

Sushisource (approved):

PR #1240, opened 2022-06-01
Fix Version State Machine ignoring range check during replay if the history was unversioned

### LINKED ISSUE #1239: getVersion call with a range that doesn't include DEFAULT_VERSION tolerates unversioned code
## Expected Behavior

1. Originally workflow doesn't have `getVersion` call at all.
2. Workflow code is modified with a range that doesn't include `DEFAULT_VERSION`.
3. Trying to replay a history created in 1. with code from 2. should fail with UnsupportedVersion

## Actual Behavior

Code versioned with `DEFAULT_VERSION` is supported by the code from 2.

Spikhalskiy:
## What was changed

Fix Version State Machine ignoring range check during replay if the history was unversioned.

## How was this tested

Unit test verifying situation when getVersion check is added during replay and wasn't there during the original execution.

Closes #1239

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/statemachines/VersionStateMachine.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/statemachines/VersionStateMachine.java
@@ -104,17 +104,17 @@ final class VersionStateMachine {
                   State.SKIPPED,
                   CommandType.COMMAND_TYPE_RECORD_MARKER,
                   State.SKIPPED_NOTIFIED,
-                  InvocationStateMachine::cancelCommandValidateCallAndNotifyResultCachedExecuting)
+                  InvocationStateMachine::notifySkippedExecuting)
               .add(
                   State.REPLAYING,
                   ExplicitEvent.SCHEDULE,
                   new State[] {State.MARKER_COMMAND_CREATED_REPLAYING, State.SKIPPED_REPLAYING},
-                  InvocationStateMachine::createMarker_replaying)
+                  InvocationStateMachine::createMarkerReplaying)
               .add(
                   State.MARKER_COMMAND_CREATED_REPLAYING,
                   CommandType.COMMAND_TYPE_RECORD_MARKER,
                   State.RESULT_NOTIFIED_REPLAYING,
-                  InvocationStateMachine::validateCallAndNotifyPreloaded_replaying)
+                  InvocationStateMachine::notifyMarkerCreatedReplaying)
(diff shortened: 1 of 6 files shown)


temporal-sdk/src/main/java/io/temporal/internal/statemachines/VersionStateMachine.java
```diff
@@ -279,28 +284,34 @@ void flushPreloadedVersionAndUpdateFromEventReplaying() {
       flushPreloadedVersionAndUpdateFromEvent(currentEvent);
     }
 
-    void cancelCommandValidateCallAndNotifyCachedOrDefaultReplaying() {
+    void notifySkippedReplaying() {
       cancelCommand();
       if (version == null) {
+        // We are in replay and in a skipped state, which means there were no matching marker in the
+        // history
```
Sushisource, line 291:
```suggestion
        // We are in replay and in a skipped state, which means there was no matching marker in the
        // history
```

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
```diff
@@ -772,6 +772,9 @@ public void getVersion(
         maxSupported,
         (v, e) -> {
           callback.apply(v, e);
+          // without this getVersion call will trigger the end of WFT,
+          // instead we want to prepare corresponded commands and unblock the execution one more
+          // time.
```
Sushisource, line 777:
```suggestion
          // instead we want to prepare subsequent commands and unblock the execution one more
          // time.
```

temporal-sdk/src/test/java/io/temporal/internal/statemachines/VersionStateMachineTest.java
```diff
@@ -300,6 +300,56 @@ public void buildWorkflow(AsyncWorkflowBuilder<Void> builder) {
     }
   }
 
+  /**
+   * Tests when 1. an original execution is performed without any versions specified
+   * (DEFAULT_VERSION is assumed) 2. replay is performed with a code that has a version call with a
+   * range no longer including DEFAULT_VERSION
```
Sushisource, line 306:
```suggestion
   * (DEFAULT_VERSION is assumed) 2. replay is performed with code that has a version call with a
   * range no longer including DEFAULT_VERSION
```

Sushisource (approved):

PR #1247, opened 2022-06-03
Make Temporal standard non-retryable gRPC codes not configurable

Spikhalskiy:
Make Temporal standard non-retryable gRPC codes not configurable. Users can't make Temporal non-retryable code retryable after this PR, users can still add own code/failure types pairs.
Improve documentation around retrying of query rpc requests and generally around `RpcRetryOptions`.

### DIFF
--- a/temporal-serviceclient/src/main/java/io/temporal/internal/retryer/GrpcRetryerUtils.java
+++ b/temporal-serviceclient/src/main/java/io/temporal/internal/retryer/GrpcRetryerUtils.java
@@ -39,10 +39,7 @@ class GrpcRetryerUtils {
    * not.
    *
    * @param currentException exception to analyze
-   * @param previousException previous exception happened before this one, {@code null} if {@code
-   *     currentException} is the first exception in the chain
    * @param options retry options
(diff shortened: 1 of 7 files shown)


temporal-serviceclient/src/main/java/io/temporal/internal/retryer/GrpcRetryerUtils.java
```diff
@@ -56,6 +53,15 @@ class GrpcRetryerUtils {
...
         return new CancellationException("The gRPC request was cancelled");
+      case INVALID_ARGUMENT:
+      case NOT_FOUND:
+      case ALREADY_EXISTS:
+      case FAILED_PRECONDITION:
+      case PERMISSION_DENIED:
+      case UNAUTHENTICATED:
+      case UNIMPLEMENTED:
+        // never retry these codes
+        return currentException;
```
cretz, line 64: Here's what Go retries: https://github.com/temporalio/sdk-go/blob/v1.14.0/internal/common/retry/interceptor.go#L119-L124. It looks like it matches up with this.

cretz (approved):

Issue #1262, opened 2022-06-16
Workflows stuck on local activity call


Spikhalskiy:
There are user reports pointing to workflows stuck on waiting for local activity results while activity is not in progress.

Workflow method stuck with stack dump:

```
"workflow-method-571893c0-821b-30e3-a437-80ce798dff42-1e8ec9fd-6426-4fd0-bc6d-8604beb6375d" #54994 prio=5 os_prio=0 cpu=1.70ms elapsed=559.87s tid=0x00007fa6a4004450 nid=0x58ac waiting on condition  [0x00007fa6e1ddf000]
   java.lang.Thread.State: WAITING (parking)
	at jdk.internal.misc.Unsafe.park(java.base@11.0.15/Native Method)
	- parking to wait for  <0x0000000773341a78> (a java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject)
	at java.util.concurrent.locks.LockSupport.park(java.base@11.0.15/LockSupport.java:194)
	at java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.await(java.base@11.0.15/AbstractQueuedSynchronizer.java:2081)
	at io.temporal.internal.sync.WorkflowThreadContext.yield(WorkflowThreadContext.java:89)
	at io.temporal.internal.sync.WorkflowThreadImpl.yield(WorkflowThreadImpl.java:412)
	at io.temporal.internal.sync.WorkflowThread.await(WorkflowThread.java:45)
	at io.temporal.internal.sync.CompletablePromiseImpl.getImpl(CompletablePromiseImpl.java:84)
	at io.temporal.internal.sync.CompletablePromiseImpl.get(CompletablePromiseImpl.java:74)
	at io.temporal.internal.sync.ActivityStubBase.execute(ActivityStubBase.java:44)
	at io.temporal.internal.sync.LocalActivityStubImpl.execute(LocalActivityStubImpl.java:29)
	at io.temporal.internal.sync.LocalActivityInvocationHandler.lambda$getActivityFunc$0(LocalActivityInvocationHandler.java:72)
	at io.temporal.internal.sync.LocalActivityInvocationHandler$$Lambda$1955/0x00000008012df440.apply(Unknown Source)
	at io.temporal.internal.sync.ActivityInvocationHandlerBase.invoke(ActivityInvocationHandlerBase.java:70)
	at com.sun.proxy.$Proxy128.publishPayload(Unknown Source)
```

[...]

Issue #1261, opened 2022-06-16
Local Activities fail and don't retry after reaching LocalRetryThreshold


Spikhalskiy:
## Expected Behavior

Local Activity should schedule a timer on the server when it reaches LocalActivityThreshold according to its backoff strategy.

## Actual Behavior

Instead, if Local Activity fails after LocalActivityThreshold is reached, the fail propagates to the workflow

## Steps to Reproduce the Problem

Unit test https://github.com/Spikhalskiy/java-sdk/commit/ea6252a66012d0cdf3f29af1ed78ab0ae95a762d#diff-b74049b45234315b43ae6729f80ee4c54cca4cbb081f97ef9ab42cc77e9e3264R39 is expected to pass, but instead, it fails. 

## Additional context

It looks like the code that was scheduling the timer was removed by mistake in the PR from the earliest days of Temporal that was reimplementing errors handling, specifically here: https://github.com/temporalio/sdk-java/pull/101/files#diff-f6a342e5baa618f03fdb6a83c6e070e5b57b9a9d9f3e1460b928149faa12b57cR237

It wasn't caught because the unit test testing La backoff that falls back to the server timers is not tested by the existing tests.

PR #1263, opened 2022-06-16
Simplify code and error handling between ActivityWorker and LocalActivityWorker

Spikhalskiy:
Simplify code and error handling between ActivityWorker and LocalActivityWorker
Make LocalActivityWorker more robust to unexpected errors

Partially addresses #1262


temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
```diff
@@ -233,10 +288,17 @@ private ActivityTaskHandler.Result handleLocalActivity(
 
       // For small backoff we do local retry. Otherwise we will schedule timer on server side.
       // TODO(maxim): Use timer queue for retries to avoid tying up a thread.
```
Sushisource, line 290: Did you find where this implementation actually is, if anywhere?

Sushisource (approved):

cretz:
To save me a bit of time, can you clarify which parts are code moves and which parts are new code I should focus on?

Spikhalskiy:
@cretz It's all code restructuring, deduplication, and ensuring that callback is always called for LAs. There is pretty much no new code.

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
Spikhalskiy: It was actually in the code base but was removed during fat finger refactoring more than 2 years ago. https://github.com/temporalio/sdk-java/issues/1261
Will need to bring it back.

PR #1276, opened 2022-06-21
Cleanup local activities and wft state machines code

Spikhalskiy:


Quinn-With-Two-Ns:
Closing as this is a stale PR

PR #1277, opened 2022-06-22
Rework LocalActivities waiting code to use nanoseconds

Spikhalskiy:
## What was changed
Rework LocalActivities waiting code to use nanoseconds
Some other refactorings improving comprehension

## Why?
System.currentTimeMillis shouldn't be used for timeouts and it's non monotonic

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/statemachines/InternalWorkflowTaskException.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/statemachines/InternalWorkflowTaskException.java
@@ -18,9 +18,13 @@
  * limitations under the License.
  */
 
-package io.temporal.internal.replay;
+package io.temporal.internal.statemachines;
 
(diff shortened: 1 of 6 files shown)


temporal-sdk/src/main/java/io/temporal/internal/statemachines/InternalWorkflowTaskException.java
```diff
@@ -18,9 +18,13 @@
  * limitations under the License.
  */
 
-package io.temporal.internal.replay;
+package io.temporal.internal.statemachines;
 
-public final class InternalWorkflowTaskException extends RuntimeException {
+/**
+ * Originated by Temporal State Machines and happens during applying of the events inside
+ * #handleEvent call.
```
Sushisource, line 25:
```suggestion
 * Originated by Temporal State Machines and happens during application of the events inside
 * #handleEvent call.
```

PR #1278, opened 2022-06-22
Add log waring if workflow eviction / closure is taking too long time

Spikhalskiy:
## What was changed
Add log warning if workflow eviction / closure is taking too long time to close.
Add guards making sure all threads are drained.


Quinn-With-Two-Ns:
Closing as this is a stale PR

PR #1304, opened 2022-07-12
Make stacktrace query result to be serialized using standard data converter only

Spikhalskiy:
## What was changed

Currently, before the change, stacktrace query uses user-defined data converter to serialize its result.
After the change a standard default converter will always be used for serialization of stacktrace query result.

## Why?
Stacktrace query is a built-in query for UI and tctl usage, hence a default converter should be used.


Sushisource (approved):

bergundy:
I think we need to revert this. UI has remote decoder to decrypt payloads

Spikhalskiy:
@bergundy remote decoder is not our standard setup. Client encoding is. See discussion in the company’s slack. This code predates any remote encoders and is obviously wrong.

PR #1308, opened 2022-07-14
Cleanup worker creation for Spring Boot Auto Configuration and add validation of the explicit worker config

Spikhalskiy:
Cleans up and simplifies worker creation in Spring Boot Auto Configuration.
Add validation of the explicit worker config duplicate conflicts.
Avoids conflicts with auto-discovered implementations and adds detailed logging about the conflict resolution actions that were made.

Improving #8


cretz (approved):
Looks good. Confirmed that 1) configuring two workers w/ the same name is a fail, and 2) auto-implementations share workers by task queue

PR #1313, opened 2022-07-20
Implement fetching server capabilities before the first server call is made

### LINKED ISSUE #1312: Call getSystemInfo before a first client call to obtain server capabilities
Temporal Server API provides new `getSystemInfo` call that returns a set of features supported by the server. 

This call needs to be performed by the stubs or client before any other client call.

Spikhalskiy:
Adds a new interceptor that fetches systemCapabilities before the first client call can be made.

Closes #1312

### DIFF
--- a/temporal-serviceclient/src/main/java/io/temporal/serviceclient/ChannelManager.java
+++ b/temporal-serviceclient/src/main/java/io/temporal/serviceclient/ChannelManager.java
@@ -27,10 +27,12 @@ import io.grpc.health.v1.HealthCheckResponse;
 import io.grpc.health.v1.HealthGrpc;
 import io.grpc.netty.shaded.io.grpc.netty.NettyChannelBuilder;
 import io.grpc.stub.MetadataUtils;
+import io.temporal.api.workflowservice.v1.GetSystemInfoResponse;
 import io.temporal.internal.retryer.GrpcRetryer;
 import java.time.Duration;
 import java.util.Collection;
 import java.util.List;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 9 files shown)


temporal-serviceclient/src/main/java/io/temporal/serviceclient/ChannelManager.java
```diff
@@ -140,7 +145,9 @@ private Channel applyHeadStandardInterceptors(Channel channel) {
     headers.put(CLIENT_NAME_HEADER_KEY, CLIENT_NAME_HEADER_VALUE);
 
     return ClientInterceptors.intercept(
-        channel, MetadataUtils.newAttachHeadersInterceptor(headers));
+        channel,
+        MetadataUtils.newAttachHeadersInterceptor(headers),
+        new SystemInfoInterceptor(systemCapabilitiesFuture));
```
Sushisource, line 150: Just double checking that these two will go first before any user-specified interceptors

Sushisource (approved):

temporal-serviceclient/src/main/java/io/temporal/serviceclient/ChannelManager.java
Spikhalskiy: yup

temporal-serviceclient/src/main/java/io/temporal/serviceclient/SystemInfoInterceptor.java
```diff
@@ -0,0 +1,104 @@
...
+                    if (Status.UNIMPLEMENTED.getCode().equals(status.getCode())) {
+                      serverCapabilitiesFuture.complete(
+                          GetSystemInfoResponse.Capabilities.getDefaultInstance());
+                    }
+                    super.onClose(status, trailers);
+                  }
+                };
+          } else {
+            // Need to reach system capabilities, so make a getSystemInfo call in a blocking manner.
+            // We don't try to squash into one and optimize the several getSystemInfo calls that may
```
cretz, line 74: > We don't try to squash into one and optimize the several getSystemInfo calls

:+1: Same as Go. If in a racy situation the this code is reached by multiple calls and we have to make multiple getSystemInfo calls and overwrite that future one right after another, that's much more preferable than locking. https://github.com/temporalio/sdk-go/blob/942edfb5dce0dbff49e19f002470726660d54b53/internal/internal_workflow_client.go#L901-L904

cretz (approved):

temporal-serviceclient/src/main/java/io/temporal/serviceclient/SystemInfoInterceptor.java
Spikhalskiy: Yeah, in Java the first who completes wins. Great, thanks for confirming that it's aligned!

PR #1314, opened 2022-07-20
Switch from dependency version ranges to specific versions

Spikhalskiy:
By using ranges we were trying to relax requirements for users who were stuck on older lib versions by providing clear boundaries of supported versions.
Dependency version ranges create too many troubles with Maven users if some Temporal SDK dependency releases postfixed versions that are broken or incompatible and there is no existing solution to the problem. Maven supports either a range or a specific version, but not both. Which is a worse problem than the original problem we were having. It's time to move on.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -20,28 +20,23 @@ allprojects {
 }
 
 ext {
-    // Platforms.
-    // We have to use specific versions with platforms until Maven 4.0.0 is released
-    // See https://github.com/temporalio/sdk-java/issues/1033
-    //
-    // grpcVersion = '[1.34.0,)!!1.47.0'
-    // jacksonVersion = '[2.9.0,)!!2.13.1'
-    // micrometerVersion = '[1.0.0,)!!1.9.2'
-    // springBootVersion = '[2.4.0,)!!2.7.1'
-    grpcVersion = '1.47.0'
-    jacksonVersion = '2.13.1'
-    micrometerVersion = '1.9.2'
-    springBootVersion = '2.7.1'
+    // Platforms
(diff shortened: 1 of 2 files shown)


tsurdilo (approved):

cretz:
Hrmm, not sure I agree since we're a library not an executable. We are going to force a lot of people to shade or otherwise override our versions when they use this in their ecosystem. Forcing people to use specific Temporal-approved versions of third party libraries is rough.

> if some Temporal SDK dependency releases postfixed versions that are broken or incompatible and there is no existing solution to the problem

Can we tackle this specifically instead of changing them all? Just yesterday a user complained about the TS repo pinning a specific protobufjs version because now they can't upgrade easily in their own environment. Luckily in Go there is no such thing as forcing transitive dependency versions on users.

Spikhalskiy:
> TS repo pinning a specific protobufjs version because now they can't upgrade easily in their own environment. 

It's not how it works with Java deps. Gradle has two transitive modes - strict and normal (old maven way). We don't use strict. People easily can override our dep in their environment if they want it.

Spikhalskiy:
> Can we tackle this specifically instead of changing them all?

Not worth it. I already had a problem with protobuf, grpc, and guava caused by versions. It's all our major dependencies.

cretz (approved):
If it is easy/expected for users to override this version with versions they actually want, approved. I just want to make sure we're not making it difficult for users to upgrade the dependencies we share with them.

Spikhalskiy:
@cretz Oh, an upgrade is not a problem definitely. If they request a higher version number in their config - the higher version will be used. A downgrade may need an exclusion of transitive dependency, but it's totally doable.

PR #1317, opened 2022-07-21
No retrying on INTERNAL gRPC errors with new Server versions

### LINKED ISSUE #1315: Client and Workers shouldn't retry INTERNAL errors with newer versions of Server
## Expected Behavior

See https://github.com/temporalio/sdk-features/issues/6 for more details.
In newer versions of Server Internal errors are considered non-retryable. This logic has to depend on the Server version and specifically it's defined by the Server Capabilities.

Spikhalskiy:
No retrying on INTERNAL gRPC errors with the new Server versions

Closes #1315

### DIFF
--- a/temporal-serviceclient/src/main/java/io/temporal/internal/retryer/GrpcRetryerUtils.java
+++ b/temporal-serviceclient/src/main/java/io/temporal/internal/retryer/GrpcRetryerUtils.java
@@ -25,6 +25,7 @@ import static io.grpc.Status.Code.DEADLINE_EXCEEDED;
 import io.grpc.Deadline;
 import io.grpc.Status;
 import io.grpc.StatusRuntimeException;
+import io.temporal.api.workflowservice.v1.GetSystemInfoResponse;
 import io.temporal.serviceclient.RpcRetryOptions;
 import io.temporal.serviceclient.StatusUtils;
 import java.time.Duration;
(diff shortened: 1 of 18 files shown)


temporal-serviceclient/src/main/java/io/temporal/internal/retryer/GrpcRetryerUtils.java
```diff
@@ -62,6 +66,10 @@ class GrpcRetryerUtils {
       case UNIMPLEMENTED:
         // never retry these codes
         return currentException;
+      case INTERNAL:
+        if (serverCapabilities.getInternalErrorDifferentiation()) {
+          return currentException;
+        }
```
cretz, line 73: Would prefer a `break` here since fall through to deadline exceeded can be confusing to readers (ntm the deadline exceeded comment would not longer apply to just it). My normal rule of thumb for readability is to only allow case fall through if there are no statements. Your choice of course.

cretz (approved):
Left one comment about readability, but otherwise LGTM.

temporal-serviceclient/src/main/java/io/temporal/internal/retryer/GrpcRetryerUtils.java
Spikhalskiy: Yeah, it totally makes sense.

PR #1320, opened 2022-07-22
Fix Test Server erroneously propagates some timeouts from the parent  to the child workflow

### LINKED ISSUE #1319: TestServer erroneously propagates timeouts of the parent workflow to the child
## Expected Behavior

A parent workflow with a short timeout should be able to start and abandon a child workflow without explicitly specifying a timeout and such a child can take a long time to finish on its own.

## Actual Behavior

If a parent workflow doesn't specify a timeout on a child, test server propagates parent timeout to the child and times it out shortly after the parent.

Spikhalskiy:
Fix Test Server erroneously propagates some timeouts from the parent to the child workflow.

Closes #1319


temporal-test-server/src/main/java/io/temporal/internal/testservice/SelfAdvancingTimerImpl.java
```diff
@@ -139,7 +139,11 @@ private void runLocked() {
                     } catch (Throwable e) {
                       log.error("Unexpected failure in timer callback", e);
                     } finally {
-                      lockHandle.unlock();
+                      try {
+                        lockHandle.unlock();
+                      } catch (Throwable e) {
+                        log.error("Fail to unlock the timer", e);
```
Sushisource, line 145:
```suggestion
                        log.error("Failed to unlock the timer", e);
```

Sushisource (approved):

temporal-sdk/src/main/java/io/temporal/client/WorkflowOptions.java
```diff
@@ -151,9 +152,24 @@ public Builder setWorkflowIdReusePolicy(WorkflowIdReusePolicy workflowIdReusePol
...
+     * <p>If you catch yourself setting this timeout to very small values, you likely use it wrong.
+     *
+     * <p>Example: If Workflow Run Timeout is 30 seconds and network was unavailable for 1 minute,
+     * workflows that got scheduled before the network blip will never get a chance to make progress
+     * and react and will be terminated. <br>
+     * A timer that is scheduled in the workflow code using {@link Workflow#newTimer(Duration)} will
+     * handle this situation gracefully. A workflow with such a timer will start after the network
+     * blip. If it started before the network blip and the timer fires during the network blip, it
+     * will get delivered after the connectivity is restored and the workflow will be able to
+     * resume.
```
Sushisource, line 172:
```suggestion
     * The time after which a workflow run is automatically terminated by Temporal service with
     * WORKFLOW_EXECUTION_TIMED_OUT status.
     *
     * <p>When a workflow reaches Workflow Run Timeout, it can't make any progress after that. Do not
     * rely on this timeout in workflow implementation or business logic. This timeout is
     * not designed to be handled in workflow code to perform any logic in case of timeout. Consider
     * using workflow timers instead.
     *
     * <p>If you catch yourself setting this timeout to very small values, you're likely using it wrong.
     *
     * <p>Example: If Workflow Run Timeout is 30 seconds and the network was unavailable for 1 minute,
     * workflows that were scheduled before the network blip will never have a chance to make progress
     * or react, and will be terminated. <br>
     * A timer that is scheduled in the workflow code using {@link Workflow#newTimer(Duration)} will
     * handle this situation gracefully. A workflow with such a timer will start after the network
     * blip. If it started before the network blip and the timer fires during the network blip, it
     * will get delivered after connectivity is restored and the workflow will be able to
     * resume.
```

PR #1330, opened 2022-07-28
Remote Data Converter & PayloadCodec

### LINKED ISSUE #1044: Remote Data Encoder
Provide a framework for serving and consuming data encoding via HTTP across all our SDKs. 

1. Frees users to write their encryption code in whichever of our SDK-supported languages they prefer.
2. Allows users to restrict access to their encryption keys or KMS API usage to only the remote decoder process rather than developer workstations.
3. Allows `tctl` to encode payloads for `tctl workflow start`. Note: This will not work for any encoders which require workflow context. Fixing this is outside the scope of the MVP

[Remote_Data_Encoder_protocol.pdf](https://github.com/temporalio/sdk-java/files/8100346/Remote_Data_Encoder_protocol.pdf)

GoSDK reference implementation: https://github.com/temporalio/sdk-go/pull/717 https://github.com/temporalio/sdk-go/pull/733

Spikhalskiy:
Adds `PayloadCodec` interface, `CodecPayloadConverter` class and new Remote Data Encoder (RDE) module with reference implementations for RDE Codec and RDE Server/Servlet.

Closes #1044


temporal-remote-data-encoder/src/main/java/io/temporal/rde/httpserver/DataEncoderHandler.java
```diff
@@ -0,0 +1,92 @@
...
+    exchange
+        .getResponseHeaders()
+        .add(
+            HttpHeaders.CONTENT_TYPE, AbstractRemoteDataEncoderCodec.CONTENT_TYPE_APPLICATION_JSON);
+    exchange.sendResponseHeaders(HttpServletResponse.SC_OK, 0);
+    try (OutputStreamWriter out = new OutputStreamWriter(exchange.getResponseBody())) {
+      AbstractRemoteDataEncoderCodec.JSON_PRINTER.appendTo(
+          Payloads.newBuilder().addAllPayloads(outgoingPayloadsList).build(), out);
+    }
+    exchange.close();
```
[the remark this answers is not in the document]
Spikhalskiy: > just a convenience for success only?
Correct.

If body is not opened - it's not needed. If an exception is thrown during the opened body - the container will do everything during the top-level exception processing.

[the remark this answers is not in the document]
cretz: Makes sense since they are the caller.

temporal-sdk/src/main/java/io/temporal/payload/codec/PayloadCodec.java
```diff
@@ -0,0 +1,38 @@
...
+import javax.annotation.Nonnull;
+
+/**
+ * Codec that encodes or decodes the given payloads. {@link PayloadCodec} implementation may be used
+ * in {@link io.temporal.common.converter.CodecDataConverter} or in the server side of Remote Data
+ * Encoder.
+ */
+public interface PayloadCodec {
+  @Nonnull
+  List<Payload> encode(@Nonnull List<Payload> payloads);
```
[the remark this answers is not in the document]
cretz: After discussion, this is no longer needed. This was an advanced concern in Go based on expectations that the caller could reuse things after passing them here. While it's obviously best practice for people to not mess with params, no doc required.

temporal-remote-data-encoder/src/main/java/io/temporal/rde/servlet/RDEServlet4.java
```diff
@@ -0,0 +1,107 @@
...
+          HttpServletResponse.SC_NOT_FOUND, "Path should end with /encode or /decode");
+      return;
+    }
+
+    Payloads.Builder incomingPayloads = Payloads.newBuilder();
+    ServletInputStream inputStream = request.getInputStream();
+    try (InputStreamReader ioReader = new InputStreamReader(inputStream)) {
+      AbstractRemoteDataEncoderCodec.JSON_FORMAT.merge(ioReader, incomingPayloads);
+    } catch (InvalidProtocolBufferException e) {
+      response.sendError(HttpServletResponse.SC_BAD_REQUEST, e.getMessage());
```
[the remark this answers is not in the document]
Spikhalskiy: Good catch!

temporal-remote-data-encoder/README.md
```diff
@@ -0,0 +1,31 @@
+# Temporal Remote Data Encoder module
+
+## What problems does Remote Data Encoder (RDE) solves?
+
+- Allows to reuse complicated encryption logic written once between different languages
+- Allows tctl to encode payloads for tctl workflow start and Temporal WebUI to decode encrypted payloads
+- Allows creation of a service that has an access to encryption keys and performing the encryption/decryption instead of allowing developer workstations to access the keys directly.
```
Sushisource, line 7:
```suggestion
- Allows reuse of complicated encryption logic written once between different languages
- Allows `tctl` to encode payloads for `tctl workflow start` and Temporal WebUI to decode encrypted payloads
- Allows creation of a service that has an access to encryption keys for performing the encryption/decryption instead of a developer workstation or service accessing the keys directly.
```

temporal-remote-data-encoder/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+## How does RDE work?
+
+### RDE Server
+
+- It's an Http Server responding to POST requests with 'application/json' content type and sending back responses with 'application/json' content type
+- It listens on some base url which can be postfixed with /encode or /decode for encoding and decoding respectively
+- Expect and emit `io.temporal.api.common.v1.Payloads` serialized to json format using [Proto3 Json Mapping](https://developers.google.com/protocol-buffers/docs/proto3#json) and sent in Http Request / Response body
```
Sushisource, line 15:
```suggestion
- It listens on some base url which is postfixed with /encode or /decode for encoding and decoding respectively
- Expects and emits `io.temporal.api.common.v1.Payloads` serialized to json format using [Proto3 Json Mapping](https://developers.google.com/protocol-buffers/docs/proto3#json) and sent in Http Request / Response body
```

temporal-remote-data-encoder/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+Can be used as one of the codecs in `io.temporal.common.converter.CodecDataConverter` configured on `WorkflowClientOptions#dataConverter` to use Remote Data Encoder Server for payloads encoding.
+Please keep in mind that this is an optional component. If you use RDE Server to decode data for tctl or WebUI purposes, you don't have to use it for encoding/decoding of payloads on Workflow Client or Workers side. They may perform the encoding locally.
+
+This module provides a reference implementation of RDE codec that should be used by users as a base for their own RDE codecs: `io.temporal.payload.codec.AbstractRemoteDataEncoderCodec` by implementing a POST method using the HTTP client of their choice.
+This module may supply some standard implementations for popular HTTP Clients, including `io.temporal.payload.codec.OkHttpRemoteDataEncoderCodec` for [OkHttpClient](https://square.github.io/okhttp/).
+
+## Security notes
+
+Existence of RDE in a system creates an additional party that may have an access to the encryption keys and exposing APIs to encode and decode any payloads. Users should evaluate and mitigate potential security risks of this solution.
```
Sushisource, line 31: This is true sometimes, but not always. You could actually use it to _reduce_ the number of places keys exist by routing all encoding and decoding through it, but that's just a different tradeoff.

temporal-remote-data-encoder/src/main/java/io/temporal/payload/codec/AbstractRemoteDataEncoderCodec.java
```diff
@@ -0,0 +1,116 @@
...
+ *
+ * <p>
+ *
+ * <h2>Remote Data Encoder Http Server specification</h2>
+ *
+ * <p>RDE Server must:
+ *
+ * <ul>
+ *   <li>Respond on URL paths ending with "/encode" and "/decode" for encoding and decoding
+ *       functionality respectively
```
Sushisource, line 42: "ending with" implies there could be other stuff besides the base URL between it and the postfix.

Maybe instead `on the URL paths "$rdeUrl/encode" and "$rdeUrl/decode"`

temporal-remote-data-encoder/src/main/java/io/temporal/payload/codec/AbstractRemoteDataEncoderCodec.java
```diff
@@ -0,0 +1,116 @@
...
+ * </ul>
+ *
+ * <p>RDE Server may:
+ *
+ * <ul>
+ *   <li>On encoding accept <i>M</i> {@link Payload}s inside incoming {@link Payloads} and return
+ *       back {@link Payloads} with <i>N</i> {@link Payload}s and have <i>M</i> &lt;&gt; <i>N</i>.
+ *       So, encoded payloads don't have to correspond 1-1 to the incoming payloads. This enabled
+ *       better compaction if required. If this is the case, a decoding path should convert <i>N</i>
+ *       payloads back to <i>M</i>
```
Sushisource, line 57:
```suggestion
 *       So, encoded payloads don't have to correspond 1-1 to the incoming payloads. This enables
 *       better compaction if required. If this is the case, a decoding path should convert <i>N</i>
 *       payloads back to <i>M</i>
```

temporal-remote-data-encoder/src/main/java/io/temporal/rde/httpserver/RDEHttpServer.java
```diff
@@ -0,0 +1,76 @@
...
+import io.temporal.payload.codec.AbstractRemoteDataEncoderCodec;
+import io.temporal.payload.codec.PayloadCodec;
+import java.io.Closeable;
+import java.io.IOException;
+import java.net.InetSocketAddress;
+import java.util.List;
+
+/**
+ * Provides a simple standalone http server implementation of Remote Data Encoder contract. This
+ * implementation is compliant with {@link AbstractRemoteDataEncoderCodec} and it's subclasses
```
Sushisource, line 34:
```suggestion
 * implementation is compliant with {@link AbstractRemoteDataEncoderCodec} and its subclasses
```

temporal-remote-data-encoder/src/main/java/io/temporal/rde/servlet/RDEServlet4.java
```diff
@@ -0,0 +1,108 @@
...
+import javax.servlet.ServletInputStream;
+import javax.servlet.ServletOutputStream;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/**
+ * Provides a Servlet implementation of Remote Data Encoder contract that can be used in any Servlet
+ * Container supporting Servlet 3.x or 4.x specifications. This implementation is compliant with
+ * {@link AbstractRemoteDataEncoderCodec} and it's subclasses supplied by Temporal JavaSDK.
```
Sushisource, line 43:
```suggestion
 * {@link AbstractRemoteDataEncoderCodec} and its subclasses supplied by Temporal JavaSDK.
```

temporal-sdk/src/main/java/io/temporal/common/converter/CodecDataConverter.java
```diff
@@ -0,0 +1,120 @@
...
+import java.util.Optional;
+import javax.annotation.Nonnull;
+
+/**
+ * It's a delegating {@link DataConverter} implementation that wraps and chains both another {@link
+ * DataConverter} and several {@link PayloadCodec}s.
+ *
+ * <p>The underlying {@link DataConverter} is expected to be responsible for conversion between user
+ * objects and bytes represented as {@link Payloads}, while an underlying chain of codecs is
+ * responsible for a subsequent byte <-> byte manipulation such as encryption or compressions
```
Sushisource, line 41:
```suggestion
 * A delegating {@link DataConverter} implementation that wraps and chains both another {@link
 * DataConverter} and several {@link PayloadCodec}s.
 *
 * <p>The underlying {@link DataConverter} is expected to be responsible for conversion between user
 * objects and bytes represented as {@link Payloads}, while the underlying chain of codecs is
 * responsible for a subsequent byte <-> byte manipulation such as encryption or compression
```

temporal-sdk/src/main/java/io/temporal/common/converter/CodecDataConverter.java
```diff
@@ -0,0 +1,120 @@
...
+public class CodecDataConverter implements DataConverter, PayloadCodec {
+  private final DataConverter dataConverter;
+  private final PayloadCodec codec;
+
+  /**
+   * When serializing to Payloads:
+   *
+   * <ul>
+   *   <li>{@code dataConverter} is applied first, following by the chain of {@code codecs}.
+   *   <li>{@code codecs} are applied last to first meaning the earlier encoders wrap the later ones
```
Sushisource, line 52: You might consider reversing these orders.

If I supply:

`new CodecDataConverter(converter, [compress, encrypt])`

That reads naturally to me as "during serialization, serialize, then compress, then encrypt", which is definitely the order you want. But with this setup, to do that properly, you have to write the list backwards which feels a little odd to me.

temporal-sdk/src/main/java/io/temporal/payload/codec/PayloadCodec.java
```diff
@@ -0,0 +1,38 @@
...
+package io.temporal.payload.codec;
+
+import io.temporal.api.common.v1.Payload;
+import java.util.List;
+import javax.annotation.Nonnull;
+
+/**
+ * Codec that encodes or decodes the given payloads. {@link PayloadCodec} implementation may be used
+ * in {@link io.temporal.common.converter.CodecDataConverter} or in the server side of Remote Data
+ * Encoder.
```
Sushisource, line 30:
```suggestion
 * Codec that encodes or decodes the given payloads. {@link PayloadCodec} implementation may be used
 * in {@link io.temporal.common.converter.CodecDataConverter} or server side in a Remote Data
 * Encoder implementation.
```

temporal-remote-data-encoder/src/main/java/io/temporal/payload/codec/AbstractRemoteDataEncoderCodec.java
Spikhalskiy: > "ending with" implies there could be other stuff besides the base URL between it and the postfix.
Yeah, this is actually true. There may be anything in between according to our spec. "/encode" and "/decode" don't have to have the same root.

temporal-sdk/src/main/java/io/temporal/common/converter/CodecDataConverter.java
Spikhalskiy: I know what you are saying. 
But this is done this way to keep things consistent with Go and TypeScript. Also, GRPC interceptors are applied the same way, the tail is applied first.

temporal-remote-data-encoder/README.md
Spikhalskiy: Added clarifications. It reduces number of parties having access to keys, but introduces a completely new thing to protect access to (other that keys).

PR #1351, opened 2022-08-08
Replays of workflow failures caused by queries don't spam worker logs anymore

### LINKED ISSUE #1348: Query to a workflow that was failed by the workflow code produces noisy Error log with not enough context
If a workflow code throws an exception that is failing the workflow, it's getting logged by the worker.
If such a completed with a failure workflow is getting queried, the replay produces the same exact log with an ERROR level and without any additional context that would allow understanding that this exception was effectively "replayed" by the query.

An exception header:

```
2022-08-05 13:22:09.172 [ERROR] 21413 --- [af-5dce1629454d] al.internal.sync.WorkflowExecuteRunnable : Workflow execution failure WorkflowId=entityUrn=urn:indeed:icims-requisition-id:39410&&jobseekerUrn=urn:indeed:icims-candidate-id:testingWithApplicationFailureWithoutAnnotation, RunId=645d1209-6c95-4d06-aeaf-5dce1629454d, WorkflowType=HiringProfessionalWorkflow
io.temporal.failure.ApplicationFailure: message='Unsupported Chart: chartId', type='com.indeed.wayfinder.expedition.host.workflows.hiring.professional.exception.UnsupportedChartException', nonRetryable=true
```

History structure:
```
WorkflowExecutionStarted
WorkflowTaskScheduled
WorkflowTaskStarted
WorkflowTaskCompleted
MarkerRecorded (Version)
MarkerRecorded (Local Activity)
WorkflowExecutionFailed
```

The original exception should be transformed into Warning from the Error.
The exception that happens in a query-only execution should be on Debug level and contain an additional context that allows to at least track back a query that causes this execution.

Spikhalskiy:
Replays of workflow failures caused by queries don't spam worker logs anymore

Closes #1348

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/failure/TemporalException.java
+++ b/temporal-sdk/src/main/java/io/temporal/failure/TemporalException.java
@@ -26,7 +26,7 @@ package io.temporal.failure;
  * <p>Do not extend by the application code.
  */
 public class TemporalException extends RuntimeException {
-  public TemporalException(String message, Throwable cause) {
+  protected TemporalException(String message, Throwable cause) {
     super(message, cause, false, true);
   }
 }
--- a/temporal-sdk/src/main/java/io/temporal/failure/TemporalFailure.java
+++ b/temporal-sdk/src/main/java/io/temporal/failure/TemporalFailure.java
@@ -42,7 +42,7 @@ public abstract class TemporalFailure extends TemporalException {
   private Optional<Failure> failure = Optional.empty();
   private final String originalMessage;
 
-  protected TemporalFailure(String message, String originalMessage, Throwable cause) {
+  TemporalFailure(String message, String originalMessage, Throwable cause) {
     super(message, cause);
     this.originalMessage = originalMessage;
   }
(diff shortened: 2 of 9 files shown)


temporal-sdk/src/main/java/io/temporal/failure/TemporalException.java
```diff
@@ -26,7 +26,7 @@
  * <p>Do not extend by the application code.
  */
 public class TemporalException extends RuntimeException {
-  public TemporalException(String message, Throwable cause) {
+  protected TemporalException(String message, Throwable cause) {
```
Spikhalskiy, line 29: Users don't support to create or extend it directly

temporal-sdk/src/main/java/io/temporal/failure/TemporalFailure.java
```diff
@@ -42,7 +42,7 @@ public abstract class TemporalFailure extends TemporalException {
   private Optional<Failure> failure = Optional.empty();
   private final String originalMessage;
 
-  protected TemporalFailure(String message, String originalMessage, Throwable cause) {
+  TemporalFailure(String message, String originalMessage, Throwable cause) {
```
Spikhalskiy, line 45: Users don't support to create or extend it directly

Sushisource (approved):

PR #1372, opened 2022-08-17
Add an ability to register custom DataConverter in Spring Boot Autoconfiguration

### LINKED ISSUE #1370: Add an ability to register custom DataConverter in Spring Boot Autoconfiguration
Spring Boot Starter should provide the ability to register custom data converters

Spikhalskiy:
Add an ability to register custom `DataConverter` in Spring Boot Autoconfiguration
Add javadoc to the `DataConverter` interface pointing out that users shouldn't be implementing this class directly.

Closes #1370


temporal-spring-boot-autoconfigure-alpha/src/main/java/io/temporal/spring/boot/autoconfigure/RootNamespaceAutoConfiguration.java
```diff
@@ -64,11 +70,31 @@ public RootNamespaceAutoConfiguration(ConfigurableListableBeanFactory beanFactor
   public NamespaceTemplate rootNamespaceTemplate(
       TemporalProperties properties,
       WorkflowServiceStubs workflowServiceStubs,
+      @Autowired List<DataConverter> dataConverters,
```
cretz, line 73: Why does this have to be a collection?

temporal-spring-boot-autoconfigure-alpha/README.md
```diff
@@ -57,7 +57,13 @@ spring.temporal:
       # insecure-trust-manager: true # or add ca.pem to java default truststore
 ```
 
-to set the key and cert chain directly from the environment variables for example.
+## Data Converter
+
+Define a bean of type `io.temporal.common.converter.DataConverter` in Spring context to be used as a custom data converter.
+If Spring context has several beans of the `DataConverter` type, they will be ignored except the one named `mainDataConverter`.
```
cretz, line 63: Why do we allow other data converters that are just gonna be ignored? (forgive my Spring ignorance here)

temporal-spring-boot-autoconfigure-alpha/src/main/java/io/temporal/spring/boot/autoconfigure/RootNamespaceAutoConfiguration.java
Spikhalskiy: It doesn't. We allow users to have several converters. If there is only one - we just use it. If there are several - we make the user specify the "one" by giving it a specific name.

temporal-spring-boot-autoconfigure-alpha/README.md
Spikhalskiy: Because the Spring context is not just ours and not just for us. It's the context of the user app. And what people like to do (especially with XML-based configs) is to make everything beans.
Now imagine a user configures a custom data converter with codecs. This will lead a user to create a DefaultDataConverter and a CodecDataConverter, which are both DataConverter. And a very common way is to make all of these thingys beans even if they are directly referenced in only one place.

temporal-spring-boot-autoconfigure-alpha/src/main/java/io/temporal/spring/boot/autoconfigure/RootNamespaceAutoConfiguration.java
cretz: I see, it's so we can use the first one in the list if there's only 1 but fallback to main converter if there is ambiguity.


I wonder then if, instead of warn and fallback to default, we should error if there is ambiguity in custom data converters. The reason is, it may be unclear that just by adding a second data converter you are now disabling the first essentially. Someone may add a data converter in some other part of the spring app not knowing they are causing this.

Basically, we should probably require explicit data converter config/choice in the face of multiple instead of ignoring all.

Spikhalskiy: Yeah, I was debating about the same with myself and I don't have a strong preference. If you are in favor of erroring out in this situation - I will change to this approach.

cretz (approved):

PR #1375, opened 2022-08-18
Fix situation when direct query comes outdated (inconsistent) and applied to a more advanced cached workflow instance

### LINKED ISSUE #1371: Eviction of workflow due to too advanced state fails query
User [hits](https://github.com/temporalio/sdk-java/pull/1267#issuecomment-1216933564) the check located [here](https://github.com/temporalio/sdk-java/blob/da51b4852a74899a713b53d1400338c74c47b828/temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java#L182) during a direct query, which fails the query.

JavaSDK probably should handle the query with a full replay when this happens instead of failing a workflow task.

Additional context:
1. We should try to reproduce this situation when it's possible. This check was added as a guard that should never happen if Server never loses progress after responding OK to completion of a workflow task. One theory is that query may be issued and sent DURING workflow task processing by the worker, in that case, it will arrive with outdated eventIds.
2. If users actually hit this line, we should cover this behavior with unit tests.

Spikhalskiy:
Fix situation when direct query comes outdated (inconsistent) and applied to a more advanced cached workflow instance.

After this fix JavaSDK doesn't use `previousStartedWorkflowId` inside the state machines to detect a replay and uses actual history to understand that instead. This allows to accept and correctly process a query with any `previousStartedWorkflowId` value.

Closes #1371


temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowRunTaskHandler.java
```diff
@@ -132,16 +124,27 @@ class ReplayWorkflowRunTaskHandler implements WorkflowRunTaskHandler {
...
       long startTimeNanos = System.nanoTime();
-      handleWorkflowTaskImpl(workflowTask);
+
+      if (workflowTask.getPreviousStartedEventId()
+          < workflowStateMachines.getCurrentStartedEventId()) {
+        // if previousStartedEventId < currentStartedEventId - the last workflow task handled by
+        // these state machines is ahead of the last handled workflow task known by the server.
+        // Something is off, the server lost progress.
+        throw new IllegalStateException(
+            "Server history for the workflow is below the progress of the workflow on the worker, the progress needs to be discarded");
```
Sushisource, line 140: Technically, you can blow the cache and fetch history here instead. This is so rare though, it seems fine to not bother with that code.

temporal-sdk/src/main/java/io/temporal/internal/replay/ServiceWorkflowHistoryIterator.java
```diff
@@ -0,0 +1,142 @@
...
+    this.grpcRetryer = service != null ? new GrpcRetryer(service.getServerCapabilities()) : null;
+    History history = task.getHistory();
+    current = history.getEventsList().iterator();
+    nextPageToken = task.getNextPageToken();
+  }
+
+  // Returns true if more history events are available.
+  // Server can return page tokens that point to empty pages.
+  // We need to verify that page is valid before returning true.
+  // Otherwise next() method would throw NoSuchElementException after hasNext() returning true.
```
Sushisource, line 77: Worth mentioning in here this may fetch

temporal-sdk/src/main/java/io/temporal/internal/replay/ServiceWorkflowHistoryIterator.java
```diff
@@ -0,0 +1,142 @@
...
+    } catch (StatusRuntimeException ex) {
+      if (Status.DEADLINE_EXCEEDED.equals(ex.getStatus())) {
+        throw Status.DEADLINE_EXCEEDED
+            .withDescription(
+                "getWorkflowExecutionHistory pagination took longer than workflow task timeout")
+            .asRuntimeException();
+      }
+      throw ex;
+    } catch (Exception e) {
+      // TODO Error? Why?
```
Sushisource, line 138: Needs cleanup here I think

temporal-sdk/src/test/java/io/temporal/internal/replay/OutdatedDirectQueryReplayWorkflowRunTaskHandlerTest.java
```diff
@@ -0,0 +1,194 @@
...
+            createReplayWorkflow(workflowExecutionHistory),
+            wft,
+            SingleWorkerOptions.newBuilder().build(),
+            new NoopScope(),
+            mock(BiFunction.class));
+
+    stateMachines = handler.getWorkflowStateMachines();
+    QueryResult queryResult =
+        handler.handleDirectQueryWorkflowTask(
+            wft, new FullHistoryIterator(workflowExecutionHistory.getEvents()));
```
Sushisource, line 109: Some comments about what would have gone wrong were this not fixed would be helpful

Sushisource (approved):

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowRunTaskHandler.java
Spikhalskiy: I can, but it likely doesn't make much sense if it's actually an outdated WFT for some reason. I will leave a comment here that it should be done if it's ever become any common occurrence or a problem for anything.

temporal-sdk/src/main/java/io/temporal/internal/replay/ServiceWorkflowHistoryIterator.java
Spikhalskiy: This is an old code, I have a separate task to clean this up

temporal-sdk/src/main/java/io/temporal/internal/replay/ServiceWorkflowHistoryIterator.java
Spikhalskiy: This is an old code, I have a separate task to clean this up

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
```diff
@@ -267,6 +264,13 @@ private void handleSingleEvent(HistoryEvent event, boolean hasNextEvent) {
       return;
     }
 
+    if (replaying
+        && !hasNextEvent
+        && (event.getEventType() == EventType.EVENT_TYPE_WORKFLOW_TASK_STARTED
+            || WorkflowExecutionUtils.isWorkflowTaskClosedEvent(event))) {
+      replaying = false;
+    }
```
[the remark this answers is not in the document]
Spikhalskiy: Previously we were effectively switching it on WorkfowTaskScheduled, after it's processing (which was empty).
Now we switch it on WorkflowTaskStarted. It has to move up because the last WorkflowTaskStarted actually triggers an eventLoop and workflow code / state machines need to see a flipped flag.

temporal-sdk/src/test/java/io/temporal/internal/replay/OutdatedDirectQueryReplayWorkflowRunTaskHandlerTest.java
```diff
@@ -0,0 +1,194 @@
...
+        DefaultDataConverter.STANDARD_INSTANCE.fromPayloads(
+            0, queryResult.getResponsePayloads(), String.class, String.class));
+  }
+
+  private ReplayWorkflow createReplayWorkflow(WorkflowExecutionHistory workflowExecutionHistory) {
+    History history = workflowExecutionHistory.getHistory();
+    ReplayWorkflow workflowMock = mock(ReplayWorkflow.class);
+    Mockito.doAnswer(
+            invocation -> {
+              stateMachines.newTimer(
```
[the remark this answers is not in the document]
Spikhalskiy: > In core what I do is manually construct a history that that workflow would produce when operating normally
We do it here very simply, by playing a normal workflow. 

> and then chunk it up into mock WFT poll responses from the server
We do it here the same way actually, we create a fake poll

This code is needed not to generate a history or something, but to bypass the whole layer that JavaSDK has around "workflow object" and "workflow class", because it's quite scattered. Core doesn't have this problem because it has this layer in lang. Basically, I work with "activations" directly here.

temporal-sdk/src/test/java/io/temporal/internal/replay/OutdatedDirectQueryReplayWorkflowRunTaskHandlerTest.java
Spikhalskiy: The test has a reference to a github issue, the github issue has a write-up of what's happening. I stopped copying the long narratives into tests code as I hope we have github as a reliable source of truth (Somebody migrating on SVN in the future is not thanking me)

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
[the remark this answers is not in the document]
Sushisource: Got it

temporal-sdk/src/test/java/io/temporal/internal/replay/OutdatedDirectQueryReplayWorkflowRunTaskHandlerTest.java
Sushisource: Sure, that works

temporal-sdk/src/test/java/io/temporal/internal/replay/OutdatedDirectQueryReplayWorkflowRunTaskHandlerTest.java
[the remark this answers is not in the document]
Sushisource: I guess what I'm getting at is the state machine invocation itself seems a bit nitty-gritty compared to eventually seeing those commands in the WFT completion. It's not immediately clear to me why it has to operate down here instead of that level, but, if it's structurally too much stuff to make that happen I think this is still totally fine

PR #1388, opened 2022-08-27
Port config of default Jackson timestamp deserialization to kotlin

Spikhalskiy:
## What was changed

Code is reorganized in a way that the same basic Jackson config is used in java and kotlin modules. This also brings `mapper.configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false)` to default kotlin config that may help prevent non-deterministic problems in case the replaying host has a different time zone than the original executor.


Bennett-Lynch (approved):
Thanks!

bergundy (approved):

PR #1397, opened 2022-08-30
Set larger maxInboundMetadataSize to accommodate unbounded Server error messages

### LINKED ISSUE #1396: Set larger maxInboundMetadataSize to accommodate unbounded Server error messages
Server [sometimes sends unbounded error messages](https://github.com/temporalio/temporal/issues/3284) that don't fit into a standard HTTP/2 headers limit.
The solution proposed in https://github.com/temporalio/sdk-features/issues/122 is to set the `maxInboundMetadataSize` to 4Mb (from the default 8Kb) to accommodate these scenarios.

Spikhalskiy:
Sets larger `maxInboundMetadataSize = MAX_INBOUND_MESSAGE_SIZE / 8` to accommodate unbounded Server error messages

Closes #1396

### DIFF
--- a/temporal-serviceclient/src/main/java/io/temporal/serviceclient/ChannelManager.java
+++ b/temporal-serviceclient/src/main/java/io/temporal/serviceclient/ChannelManager.java
@@ -51,7 +51,16 @@ final class ChannelManager {
    * Frontend from Java SDK. The Server Frontend limit on incoming gRPC message value is currently
    * 4Mb.
    */
-  private static final int MAX_INBOUND_MESSAGE_SIZE = 128 * 1024 * 1024;
+  private static final int MAX_INBOUND_MESSAGE_SIZE = 128 * 1024 * 1024; // 128Mb
+
+  /**
+   * This value sets the limit on the metadata (HTTP/2 headers) of the incoming from Temporal Server
+   * to Java SDK. The default HTTP/2 and gRPC is 8Kb. Larger limit is set to accommodate longer
+   * Temporal Server error messages that may contain segments of unbounded variable length depending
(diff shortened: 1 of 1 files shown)


temporal-serviceclient/src/main/java/io/temporal/serviceclient/ChannelManager.java
```diff
@@ -175,7 +175,10 @@ private ManagedChannel prepareChannel() {
     NettyChannelBuilder builder =
         NettyChannelBuilder.forTarget(options.getTarget())
             .defaultLoadBalancingPolicy("round_robin")
-            .maxInboundMessageSize(MAX_INBOUND_MESSAGE_SIZE);
+            .maxInboundMessageSize(MAX_INBOUND_MESSAGE_SIZE)
+            // to accommodate unbounded Server status messages that may be over the default limit of
+            // 8Kb
+            .maxInboundMetadataSize(MAX_INBOUND_MESSAGE_SIZE / 8);
```
bergundy, line 181: Why not use a separate constant for this? Not seeing how this is tied to message size.

bergundy (approved):

temporal-serviceclient/src/main/java/io/temporal/serviceclient/ChannelManager.java
Spikhalskiy: Well, this is the problem. What should be the number? And if we don’t know, how is a separate constant really better than tangling these two values together?
The benefit of connecting them is that it’s actually easier to say how many MAX_INBOUND_MESSAGE_SIZEs do we allocate on the buffers (like 1.12x right now).
But I don’t care, we can chose some random really large size.

PR #1399, opened 2022-08-30
Align Test Server behavior with the real one if completion command is not the last

### LINKED ISSUE #1398: Align Test Server behavior with the real one if WorkflowExecutionCompleted is not the last command
## Expected Behavior

If a workflow (or an SDK) is written in the way that the workflow is completed before some other operations (like an asynchronous local activity), the Test Server should reject such a workflow task with a text message aligned with a real Temporal server. 

## Actual Behavior

TestServer actually accepts this workflow and completes the workflow.

Spikhalskiy:
Align Test Server behavior with the real one if the workflow completion command is not the last command in the commands sequence of workflow task completion request.

Closes #1398


Sushisource (commented):
This makes sense to me, but, the title/description are talking about `WorkflowExecutionFailed` and this is looking for `CompleteWorkflowExecution`.

Seems it should also apply to `FailWorkflowExecution`

Sushisource (approved):

PR #1401, opened 2022-08-31
Test Server doesn't reset sticky queue if timeout fired for an outdated workflow task

Spikhalskiy:
Test Server doesn't reset sticky queue if timeout fired for an outdated workflow task

### DIFF
--- a/temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
+++ b/temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
@@ -1060,6 +1060,7 @@ class TestWorkflowMutableStateImpl implements TestWorkflowMutableState {
 
   // TODO: insert a single  workflow task timeout into the history
   private void timeoutWorkflowTask(long scheduledEventId) {
+    StickyExecutionAttributes previousStickySettings = this.stickyExecutionAttributes;
     try {
       completeWorkflowTaskUpdate(
           ctx -> {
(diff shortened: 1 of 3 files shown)


temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -1060,13 +1060,15 @@ private void processFailWorkflowTask(
...
   private void timeoutWorkflowTask(long scheduledEventId) {
+    StickyExecutionAttributes previousStickySettings = this.stickyExecutionAttributes;
     try {
       completeWorkflowTaskUpdate(
           ctx -> {
             if (workflowTaskStateMachine == null
                 || workflowTaskStateMachine.getData().scheduledEventId != scheduledEventId
                 || workflowTaskStateMachine.getState() == State.NONE) {
               // timeout for a previous workflow task
+              stickyExecutionAttributes = previousStickySettings; // rollout sticky options
```
Sushisource, line 1071: Will this always go back to the non-sticky queue? That's what should happen in this situation, no?

Spikhalskiy: No, this branch checks if the timeout happened for a workflow task that is already done. We return sticky queue back in place.

Sushisource: Ah, got it.

Sushisource (approved):

PR #1409, opened 2022-09-07
Fix guard clause to allow passing key and cert chain as strings

yassineb:
Fixes guard clause in spring-boot-autoconfig to allow passing key and cert chain as strings.

### DIFF
--- a/temporal-spring-boot-autoconfigure-alpha/src/main/java/io/temporal/spring/boot/autoconfigure/template/ServiceStubsTemplate.java
+++ b/temporal-spring-boot-autoconfigure-alpha/src/main/java/io/temporal/spring/boot/autoconfigure/template/ServiceStubsTemplate.java
@@ -113,8 +113,10 @@ public class ServiceStubsTemplate {
       @Nullable ConnectionProperties.MTLSProperties mtlsProperties,
       WorkflowServiceStubsOptions.Builder stubsOptionsBuilder) {
     if (mtlsProperties == null
-        || (mtlsProperties.getKeyFile() == null && mtlsProperties.getCertChainFile() == null)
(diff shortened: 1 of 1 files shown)


temporal-spring-boot-autoconfigure-alpha/src/main/java/io/temporal/spring/boot/autoconfigure/template/ServiceStubsTemplate.java
```diff
@@ -113,7 +113,8 @@ private void configureMTLS(
       @Nullable ConnectionProperties.MTLSProperties mtlsProperties,
       WorkflowServiceStubsOptions.Builder stubsOptionsBuilder) {
     if (mtlsProperties == null
-        || (mtlsProperties.getKeyFile() == null && mtlsProperties.getCertChainFile() == null)) {
+        || (mtlsProperties.getKeyFile() == null && mtlsProperties.getCertChainFile() == null)
```
Spikhalskiy, line 116: I don't think this is the right condition. Think about the following: it's totally fine to have only keyFile and chainFile specified. But we will `return` here.

I think you need to && all of them. keyFile, certChainFile, key, certChain.

yassineb: you're absolutely right! fixed in https://github.com/temporalio/sdk-java/pull/1409/commits/7e4269de3834c108c535bc753d15e5f46263fbb1

Spikhalskiy (approved):

Spikhalskiy:
@yassineb Thank you for the contribution!

PR #1410, opened 2022-09-07
Deprecate addWorkflowImplementationFactory and give a register* alternative with an expected contract

### LINKED ISSUE #1408: addWorkflowImplementationFactory with no ImplementationOptions fails workflow execution on any Throwable
## Expected Behavior

The default behavior of Temporal Workers, if there is non-`TemporalFailure` exception thrown in the workflow, is to fail a workflow task.
This is configured by `WorkflowImplementationOptions#setFailWorkflowExceptionTypes`.

## Actual Behavior

If users use
```
Worker#registerWorkflowImplementationTypes(Class<?>... workflowImplementationClasses)
Worker#registerWorkflowImplementationTypes(WorkflowImplementationOptions options, Class<?>... workflowImplementationClasses)
Worker#addWorkflowImplementationFactory(WorkflowImplementationOptions options, Class<R> workflowInterface, Func<R> factory)
```
the methods behave as expected and if the user doesn't override `WorkflowImplementationOptions#setFailWorkflowExceptionTypes` non-`TemporalFailure` exceptions in Workflow code don't fail a workflow execution.

But if
```
addWorkflowImplementationFactory(Class<R> workflowInterface, Func<R> factory)
```

if used, `Throwable` is set into `WorkflowImplementationOptions#setFailWorkflowExceptionTypes` by the underlying implementation.

this difference is completely unexpected by users and undocumented and I guess it was done for a backwards compatibility in the future, but now it leads to great confusion.

Spikhalskiy:
# Why

Before this change `addWorkflowImplementationFactory(Class workflowInterface, Func<R> factory)` had a surprising behavior: it sets `WorkflowImplementationOptions.Builder#setFailWorkflowTypes` to `Throwable`. This leads to Workflow Execution failures on every exception that is thrown from the Workflow code (not just `TemporalFailures`). No other workflow implementation registration methods behave like that. Neither an overloaded `addWorkflowImplementationFactory` nor `registerWorkflowImplementationTypes`.

# What changed

`addWorkflowImplementationFactory(Class workflowInterface, Func<R> factory)` is deprecated and replaced with `registerWorkflowImplementationFactory`. New method doesn't set `WorkflowImplementationOptions.Builder#setFailWorkflowTypes` to `Throwable`.

`addWorkflowImplementationFactory(WorkflowImplementationOptions options, Class workflowInterface, Func<R> factory)` is deprecated and replaced with `registerWorkflowImplementationFactory`. 1-1 replacement.

To get an old behavior of `addWorkflowImplementationFactory(Class workflowInterface, Func<R> factory)`, users should explicitly pass `WorkflowImplementationOptions.newBuilder().setFailWorkflowExceptionTypes(Throwable.class).build()` as a third parameter of `registerWorkflowImplementationFactory(Class, Func, WorkflowImplementationOptions.newBuilder)`.

Additional clarifications were added:
1. Discouraging the usage of these methods in production code and 
2. Danger of dependency injection into workflow instances

Closes #1408

### DIFF
--- a/temporal-kotlin/src/main/kotlin/io/temporal/worker/WorkerExt.kt
+++ b/temporal-kotlin/src/main/kotlin/io/temporal/worker/WorkerExt.kt
@@ -66,7 +66,12 @@ inline fun <reified T : Any> Worker.registerWorkflowImplementationType(
  * @param factory factory that when called creates a new instance of the workflow implementation
  * object.
  * @see Worker.addWorkflowImplementationFactory
+ * @deprecated See deprecation notes on [Worker.addWorkflowImplementationFactory]
  */
+@Deprecated(
+  "Use Worker.registerWorkflowImplementationFactory instead",
+  ReplaceWith("Worker.registerWorkflowImplementationTypes(factory, options)")
+)
 inline fun <reified T : Any> Worker.addWorkflowImplementationFactory(
   options: WorkflowImplementationOptions,
   noinline factory: () -> T
@@ -75,6 +80,26 @@ inline fun <reified T : Any> Worker.addWorkflowImplementationFactory(
 }
 
 /**
+ * Configures a factory to use when an instance of a workflow implementation is created.
+ * Please read an original [Worker.registerWorkflowImplementationFactory] method doc because
+ * this method has a limited usage.
+ *
+ * @param T Workflow interface that this factory implements
+ * @param factory factory that when called creates a new instance of the workflow implementation
+ * object.
+ * @param options custom workflow implementation options for a worker
+ * @see Worker.registerWorkflowImplementationFactory
+ */
+inline fun <reified T : Any> Worker.registerWorkflowImplementationFactory(
+  noinline factory: () -> T,
+  options: WorkflowImplementationOptions
+) {
+  registerWorkflowImplementationFactory(T::class.java, factory, options)
+}
+
+/**
+ * This method may behave differently from your expectations!
+ * Read deprecation and migration notes on [Worker.addWorkflowImplementationFactory].
  * Configures a factory to use when an instance of a workflow implementation is created.
  *
  * ```kotlin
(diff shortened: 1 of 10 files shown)


temporal-kotlin/src/main/kotlin/io/temporal/worker/WorkerExt.kt
```diff
@@ -75,6 +80,26 @@ inline fun <reified T : Any> Worker.addWorkflowImplementationFactory(
...
+ * object.
+ * @param options custom workflow implementation options for a worker
+ * @see Worker.registerWorkflowImplementationFactory
+ */
+inline fun <reified T : Any> Worker.registerWorkflowImplementationFactory(
+  noinline factory: () -> T,
+  options: WorkflowImplementationOptions
+) {
+  registerWorkflowImplementationFactory(T::class.java, factory, options)
+}
```
GreyTeardrop, line 98: Is it possible to keep `factory` as the last parameter here? That would enable a nicer experience for calling this method.

**`factory` as the first parameter**
```kotlin
worker.registerWorkflowImplementationFactory({ /* code to create workflow instance */ }, options)
```

**`factory` as the last parameter**
```kotlin
worker.registerWorkflowImplementationFactory(options) { /* code to create workflow instance */ }
```

temporal-kotlin/src/main/kotlin/io/temporal/worker/WorkerExt.kt
```diff
@@ -66,7 +66,12 @@ inline fun <reified T : Any> Worker.registerWorkflowImplementationType(
  * @param factory factory that when called creates a new instance of the workflow implementation
  * object.
  * @see Worker.addWorkflowImplementationFactory
+ * @deprecated See deprecation notes on [Worker.addWorkflowImplementationFactory]
  */
+@Deprecated(
+  "Use Worker.registerWorkflowImplementationFactory instead",
+  ReplaceWith("Worker.registerWorkflowImplementationTypes(factory, options)")
+)
```
GreyTeardrop, line 74: The `ReplaceWith` annotation could probably be adjusted
* as far as I understand it should reference `registerWorkflowImplementationFactory` instead of `registerWorkflowImplementationTypes`
* the `Worker` instance acts as `this` for the extension method, so the replacement can refer to it like that. Otheriwse, IDEA would insert a literal `Worker` as a replacement
CleanShot 2022-09-08 at 14 18 18
```suggestion
@Deprecated(
  "Use Worker.registerWorkflowImplementationFactory instead",
  ReplaceWith("this.registerWorkflowImplementationFactory(factory, options)")
)
```

temporal-kotlin/src/main/kotlin/io/temporal/worker/WorkerExt.kt
```diff
@@ -90,8 +115,24 @@ inline fun <reified T : Any> Worker.addWorkflowImplementationFactory(
  * object.
  * @see Worker.addWorkflowImplementationFactory
  */
+@Deprecated("Use Worker.registerWorkflowImplementationFactory instead", ReplaceWith("Worker.registerWorkflowImplementationTypes(factory)"))
```
GreyTeardrop, line 118: And same here
```suggestion
@Deprecated("Use Worker.registerWorkflowImplementationFactory instead", ReplaceWith("this.registerWorkflowImplementationFactory(factory)"))
```

cretz (approved):

Spikhalskiy:
@GreyTeardrop thank you for keeping an eye on the kotlin module!

PR #1421, opened 2022-09-13
Rework workflow threads reporting from using blocking ThreadPoolExecutor#getActiveThreads

Spikhalskiy:
`ThreadPoolExecutor#getActiveThreads` implementation is taking a pool-wide lock and implements an actual counting inside this lock. This PR replaces it with an explicit counter to remove the points of threads serialization.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/worker/ActiveThreadReportingExecutor.java
+++ b/temporal-sdk/src/main/java/io/temporal/worker/ActiveThreadReportingExecutor.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright (C) 2022 Temporal Technologies, Inc. All Rights Reserved.
+ *
+ * Copyright (C) 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Modifications copyright (C) 2017 Uber Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this material except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


temporal-sdk/src/main/java/io/temporal/worker/ActiveThreadReportingExecutor.java
```diff
@@ -0,0 +1,60 @@
...
+  public Future<?> submit(@Nonnull Runnable task) {
+    return workflowThreadPool.submit(
+        () -> {
+          int tasksCount = tasksInFlight.incrementAndGet();
+          metricsScope.gauge(MetricsType.WORKFLOW_ACTIVE_THREAD_COUNT).update(tasksCount);
+          try {
+            task.run();
+          } finally {
+            tasksCount = tasksInFlight.decrementAndGet();
+            metricsScope.gauge(MetricsType.WORKFLOW_ACTIVE_THREAD_COUNT).update(tasksCount);
```
cretz, line 56: Technically this is racy (line 55 can run on thread A, then line 50 and 51 can run on thread B, then line 56 runs back on thread A setting to old value), but after off-PR discussion: 1) this is how the old one worked, 2) it will get right eventually, 3) a bit off is acceptable, and 4) the alternative of `synchronized` access to gauge `update` is bad.

cretz (approved):

temporal-sdk/src/main/java/io/temporal/worker/ActiveThreadReportingExecutor.java
Spikhalskiy: ACK. I consider it's fine to publish sometimes a slightly outdated value to this gauge (and override a newer value). Because alternatives either bring synchronization of all workflow threads or are blocked by uber-tally interface.

Issue #998, opened 2022-01-24
Bring Pollers & Workers on par with Core and Go implementations


Spikhalskiy:
**Is your feature request related to a problem? Please describe.**
Currently, JavaSDK creates one sticky queue per Workflow factory and one sticky poller is shared between all the workers. 
This creates several problems:
- Inability to rate limit WFT
- Inability to implement backpressure from each worker, because it will delay sticky queue for all workers
- Inability to prioritize sticky queue over the normal queue processing
- Other complexities with resources isolation

**Describe the solution you'd like**
A sticky queue and poller per each worker.
This will also open the road to implement the same backpressure semantic we have for activities when pollers don't poll until there is an available processing slot to address #760 

**Additional context**
The counter added in #997 should be reworked to a more strict reporting of a semaphore value, which will be possible after addressing this task.

PR #1423, opened 2022-09-14
Each worker now has its own sticky task queue

Spikhalskiy:
Every worker gets their own sticky task queue and their own fleet of sticky pollers.
Sticky Pollers to normal pollers ratio is fixed at 5:1 in this implementation to reduce the chance for misconfiguration. This ratio may become flexible later or completely go away if the same pollers will switch between sticky a normal queue.

Towards #998


temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
```diff
@@ -139,34 +168,40 @@ public boolean isTerminated() {
 
   @Override
   public CompletableFuture<Void> shutdown(ShutdownManager shutdownManager, boolean interruptTasks) {
-    return poller.shutdown(shutdownManager, interruptTasks);
+    return stickyPoller != null
+        ? CompletableFuture.allOf(
+            poller.shutdown(shutdownManager, interruptTasks),
+            stickyPoller.shutdown(shutdownManager, interruptTasks))
+        : poller.shutdown(shutdownManager, interruptTasks);
```
Sushisource, line 175: nit: A lot of side effects happening in one statement, which is also a return statement.

Would read easier I think with some normal conditionals and separate return.

Sushisource (approved):
Nice, net line removal :)

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
```diff
@@ -139,34 +168,40 @@ public boolean isTerminated() {
...
   }
 
   @Override
   public void awaitTermination(long timeout, TimeUnit unit) {
-    poller.awaitTermination(timeout, unit);
+    long timeoutMillis = ShutdownManager.awaitTermination(poller, unit.toMillis(timeout));
+    ShutdownManager.awaitTermination(stickyPoller, timeoutMillis);
```
cretz, line 181: In other methods you are concerned `stickyPoller` may be `null`, no similar concerns here? Is `awaitTermination` null-safe?

temporal-sdk/src/main/java/io/temporal/worker/Worker.java
```diff
@@ -493,12 +503,22 @@ private static SingleWorkerOptions toWorkflowWorkerOptions(
       Scope metricsScope) {
     Map<String, String> tags =
         new ImmutableMap.Builder<String, String>(1).put(MetricsTag.TASK_QUEUE, taskQueue).build();
+
+    Duration stickyQueueScheduleToStartTimeout = options.getStickyQueueScheduleToStartTimeout();
+    if (WorkerOptions.DEFAULT_STICKY_SCHEDULE_TO_START_TIMEOUT.equals(
+            stickyQueueScheduleToStartTimeout)
```
cretz, line 509: Didn't read the rest of the code, but assuming this is "if default fallback" code, any concerns of a user explicitly setting the same timeout as our default that we will now override?

cretz (approved):

temporal-sdk/src/main/java/io/temporal/worker/Worker.java
Spikhalskiy: Yeah. For this to "fire", user needs to set an option on the Workers into the default value AND set a DIFFERENT value on WorkerFactory too using the deprecated method. Not a likely scenario to make something special about WorkerOptions#stickyQueueScheduleToStartTimeout and incur maintenance costs in the future.

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
Spikhalskiy: It's null safe, but a good note. I will add a check just for uniformity.

PR #1431, opened 2022-09-15
Implement executor slots reservation for workflow tasks

Spikhalskiy:
Now Workflow Task pollers reserve executor slots before attempting to perform a long poll.
This behavior is now aligned with the same mechanic for activities and what happens in GoSDK & Core.

Towards #998


Sushisource (approved):
This LGTM, but I'd definitely take extra care with the semaphore - I had a few problems where it'd get out of whack that I had to solve by passing around an object which represented the workflow task and when dropped would release the permit.

You've basically done that here too, and I think it looks fine, but worth running some stress tests (which I suppose is the point of this anyway)

PR #1433, opened 2022-09-15
Optimize unneeded string concatenation

Spikhalskiy:


cretz (approved):
Definitely some micro-optimization, though I know it's on a hot path. Wonder if it'd be worth any included micro-benchmarking in the test suite.

Spikhalskiy:
@cretz It is, but it's a silly mistake to make an eager concatenation here while this string ideally should never be used at all. I think we are in pre-microbenchmarking era yet, where I'm just fixing mostly obvious silly things. Later some elements may be worth isolating and micro benchmarking. Like the cooperative multithreading runner.

PR #1434, opened 2022-09-15
Reduce eagerly executed work needed to create an ActivityStub

Spikhalskiy:
## What was changed
Reduce eagerly executed work needed to create an `ActivityStub`.
`ActivityStub`s are lighter to create. 

## Why

Currently, during the creation of an activity stub we eagerly prepare all of its methods to be called.
It leads to a huge amount of stubs preparation work every time JavaSDK creates a workflow instance.
Most of this work is lost because not all methods are called. Also if instances get evicted from the cache, it's also repetitive throwaway work. It also increases the amount of memory needed for each retained workflow instance.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/ActivityInvocationHandlerBase.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/ActivityInvocationHandlerBase.java
@@ -30,15 +30,16 @@ import io.temporal.internal.sync.AsyncInternal.AsyncMarker;
 import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.Method;
 import java.lang.reflect.Proxy;
-import java.util.HashMap;
-import java.util.Map;
 import java.util.function.Function;
 
 /** Dynamic implementation of a strongly typed activity interface. */
 @VisibleForTesting
 public abstract class ActivityInvocationHandlerBase implements InvocationHandler {
+  private final POJOActivityInterfaceMetadata activityMetadata;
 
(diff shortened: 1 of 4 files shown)


temporal-sdk/src/main/java/io/temporal/internal/sync/ActivityInvocationHandlerBase.java
```diff
@@ -50,24 +51,12 @@ public static <T> T newProxy(Class<T> activityInterface, InvocationHandler invoc
...
   @Override
   public Object invoke(Object proxy, Method method, Object[] args) {
-    Function<Object[], Object> function = methodFunctions.get(method);
-    if (function == null) {
-      throw new IllegalArgumentException("Unexpected method: " + method);
-    }
+    POJOActivityMethodMetadata methodMetadata = activityMetadata.getMethodMetadata(method);
+    MethodRetry methodRetry = methodMetadata.getMethod().getAnnotation(MethodRetry.class);
+    String activityType = methodMetadata.getActivityTypeName();
+    Function<Object[], Object> function = getActivityFunc(method, methodRetry, activityType);
```
cretz, line 59: Could memoize this is if it's frequent/expensive enough

cretz (approved):

temporal-sdk/src/main/java/io/temporal/internal/sync/ActivityInvocationHandlerBase.java
Spikhalskiy: That will be a separate PR. The main thing that we should memoize/cache is all this crazy reflection we are doing inside POJO*Metadata classes again and again for the same classes and interfaces.

PR #1439, opened 2022-09-15
Add support for Eager Activity Dispatch

### LINKED ISSUE #1339: Eager activity dispatch

mjameswh:
## What was changed
- Add support for Eager Activity Support

## Why?
- Improve performance
- Feature parity with core and go SDKs

## Checklist
1. Add metrics
2. Closes #1339
3. Note: Eager Activity Dispatch support in tes server will be pushed in a distinct PR


Spikhalskiy:
I think this PR should have a unit test where we spawn 2 Workers on the same queue, run a workflow with several sync activities, and make sure that the activities always end up scheduled and executed on the worker-originator. This may be checked from the history.

temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflowContext.java
```diff
@@ -356,6 +356,11 @@ private ExecuteActivityParameters constructExecuteActivityParameters(
                 ProtobufTimeUtils.toProtoDuration(options.getScheduleToCloseTimeout()))
             .setHeartbeatTimeout(ProtobufTimeUtils.toProtoDuration(options.getHeartbeatTimeout()));
 
+    // FIXME: Make it possible for user to opt out of eager activity execution, eg. based on worker
```
Spikhalskiy, line 359: This is too much logic for this method. This method is pretty much brainless data transferred. This decision should be a bit up in the method structure and come here as an option/flag.

temporal-sdk/src/main/java/io/temporal/internal/worker/ActivityWorker.java
```diff
@@ -363,4 +373,24 @@ private void logExceptionDuringResultReporting(
       }
     }
   }
+
+  private final class EagerActivityInjectorImpl implements EagerActivityInjector {
+    //  private final Scope metricsScope;
+
+    public boolean reserveActivitySlot(ScheduleActivityTaskCommandAttributes commandAttributes) {
```
Spikhalskiy, line 380: `tryReserve` is probably more conventional naming here

temporal-sdk/src/main/java/io/temporal/internal/worker/EagerActivitySlotsReservation.java
```diff
@@ -0,0 +1,91 @@
...
+
+  EagerActivitySlotsReservation(EagerActivityInjector eagerActivityInjector) {
+    this.eagerActivityInjector = eagerActivityInjector;
+  }
+
+  public RespondWorkflowTaskCompletedRequest applyToRequest(
+      RespondWorkflowTaskCompletedRequest request) {
+    RespondWorkflowTaskCompletedRequest.Builder mutableRequest = request.toBuilder();
+
+    for (int i = 0; i < mutableRequest.getCommandsCount(); i++) {
```
Spikhalskiy, line 42: I really don't like that we tinker with commands AFTER they got created. We don't do it with any other commands and we should avoid it here too. Why not move this into Activity StateMachine?

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
```diff
@@ -372,35 +378,44 @@ private Optional<PollWorkflowTaskQueueResponse> sendReply(
...
-            () ->
-                nextTask.set(
-                    service
-                        .blockingStub()
-                        .withOption(METRICS_TAGS_CALL_OPTIONS_KEY, workflowTypeMetricsScope)
-                        .respondWorkflowTaskCompleted(request)),
-            grpcRetryOptions);
-        if (nextTask.get().hasWorkflowTask()) {
-          return Optional.of(nextTask.get().getWorkflowTask());
+        try (EagerActivitySlotsReservation activitySlotsReservation =
```
Spikhalskiy, line 385: Not a logic for `sendReply` method.

temporal-sdk/src/main/java/io/temporal/internal/worker/EagerActivitySlotsReservation.java
```diff
@@ -0,0 +1,91 @@
...
+      Command command = mutableRequest.getCommands(i);
+      if (command.getCommandType() != CommandType.COMMAND_TYPE_SCHEDULE_ACTIVITY_TASK) continue;
+
+      ScheduleActivityTaskCommandAttributes commandAttributes =
+          command.getScheduleActivityTaskCommandAttributes();
+      if (!commandAttributes.getRequestEagerExecution()) continue;
+
+      boolean semaphoreAcquired = this.eagerActivityInjector.reserveActivitySlot(commandAttributes);
```
Spikhalskiy, line 50: What if RequestEagerExecution is already false here?

temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflowContext.java
mjameswh: Is your comment specifically about the FIXME, or is it about the equality check on task queue name?

temporal-sdk/src/main/java/io/temporal/activity/ActivityOptions.java
```diff
@@ -202,6 +204,22 @@ public Builder setCancellationType(ActivityCancellationType cancellationType) {
...
+     * eager execution may still be disabled at the worker level or eager execution may not be
+     * requested due to lack of available slots.
+     *
+     * <p>Eager activity execution means the server returns requested eager activities directly from
+     * the workflow task back to this worker which is faster than non-eager which may be dispatched
+     * to a separate worker.
+     *
+     * <p>Default is false.
+     */
+    public Builder setDisableEagerExecution(boolean disableEagerExecution) {
```
mjameswh, line 218: @bergundy In go-sdk, the activity level option for this is named `disableEagerExecution`. In TypeScript, you named it `allowEagerDispatch`, but kept the negative form later in the chain (the core-sdk use the negative form). Should I switch to the affirmative form _everywhere_? TBH that would make more sense in and would be easier to understand.

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowContextImpl.java
```diff
@@ -202,7 +206,8 @@ public Functions.Proc1<Exception> scheduleActivityTask(
       attributes.setActivityId(workflowStateMachines.randomUUID().toString());
     }
     Functions.Proc cancellationHandler =
-        workflowStateMachines.scheduleActivityTask(parameters, callback);
+        workflowStateMachines.scheduleActivityTask(
```
mjameswh, line 209: @Spikhalskiy As per our previous discussion, I passed `canRequestEagerExecution` down to the activity state machine, but I'm not that comfortable with going this deep. I think the check on `canRequestEagerExecution()` should be moved here, no deeper. What do you say?

temporal-sdk/src/test/java/io/temporal/workflow/activityTests/TestEagerActivityExecution.java
```diff
@@ -0,0 +1,253 @@
...
+import io.temporal.workflow.shared.TestActivities;
+import io.temporal.workflow.shared.TestActivities.TestActivitiesImpl;
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.Set;
+import java.util.UUID;
+import java.util.stream.Collectors;
+import org.junit.*;
+
+public class TestEagerActivityExecution {
```
mjameswh, line 47: Note: this is only a draft. Please don't review this file at this point, except for specific questions

temporal-sdk/src/main/java/io/temporal/activity/ActivityOptions.java
bergundy: In go absent struct member becomes the default value for that type which means that there’s no way to distinguish between absent and false and we want the default to be true, we don’t have that problem in TS

Spikhalskiy: It definitely should be disableEagerExecution, not enable. It's not about defaults, it's about the API naming and what is assumed to be default by the APIs. 

In java it's easy to distinguish between not set or false, but let's leave the `disableEagerExecution` naming for the mentioned reason.

temporal-sdk/src/main/java/io/temporal/internal/worker/EagerActivityInjector.java
```diff
@@ -0,0 +1,66 @@
...
+
+import io.temporal.api.command.v1.ScheduleActivityTaskCommandAttributesOrBuilder;
+import io.temporal.api.workflowservice.v1.PollActivityTaskQueueResponse;
+
+public interface EagerActivityInjector {
+  // Perform an early check at either there is any possibility that requesting eager execution on
+  // the given activity
+  // command would be accepted. This is an optimisation, to reduce late mutations of
+  // ScheduleActivityTaskCommand.
+  public boolean canRequestEagerExecution(
```
Spikhalskiy, line 31: Nit: Please cleanup `public` from interface methods. All interface methods are public.

temporal-sdk/src/main/java/io/temporal/internal/worker/EagerActivityInjector.java
```diff
@@ -0,0 +1,66 @@
...
+ */
+
+package io.temporal.internal.worker;
+
+import io.temporal.api.command.v1.ScheduleActivityTaskCommandAttributesOrBuilder;
+import io.temporal.api.workflowservice.v1.PollActivityTaskQueueResponse;
+
+public interface EagerActivityInjector {
+  // Perform an early check at either there is any possibility that requesting eager execution on
+  // the given activity
```
Spikhalskiy, line 28: line wrapping issue. With google-java-format it sometimes makes sense to write the whole comment in one line and let the formatter to the job, otherwise you end up with artifacts like this.

temporal-sdk/src/main/java/io/temporal/activity/ActivityOptions.java
bergundy: I don't see why calling the option `allowEager{Execution|Dispatch}` and saying that the default is true is bad naming.
I like using the positive form, it avoids having double negation when used in conditions later.

I even find this hard to read: `setDisableEagerExecution` and it says the default is `false`, there's some ambiguity here. Does it mean the default is default is disabled or that disable is false?

At the end of the day it's @Spikhalskiy call so do whatever feels right to you.

Spikhalskiy: Ideally, every flag should have a false default value. It makes things easy and expected from the user side. It's a flag.
A flag that is true by default is a worse problem for an API than a double negation.

temporal-sdk/src/main/java/io/temporal/internal/statemachines/ActivityStateMachine.java
```diff
@@ -243,18 +244,27 @@ public static ActivityStateMachine newInstance(
...
       Functions.Proc1<CancellableCommand> commandSink,
-      Functions.Proc1<StateMachine> stateMachineSink) {
+      Functions.Proc1<StateMachine> stateMachineSink,
+      Functions.Func1<ScheduleActivityTaskCommandAttributesOrBuilder, Boolean>
+          canRequestEagerExecution) {
     super(STATE_MACHINE_DEFINITION, commandSink, stateMachineSink);
     this.parameters = parameters;
+
     ScheduleActivityTaskCommandAttributes.Builder scheduleAttr = parameters.getAttributes();
+    if (scheduleAttr.getRequestEagerExecution() && !canRequestEagerExecution.apply(scheduleAttr))
```
Spikhalskiy, line 265: No need for the first check. Doesn't really help to understand the code either.

temporal-sdk/src/main/java/io/temporal/activity/ActivityOptions.java
mjameswh: Fair enough, I will keep it as `disableEagerExecution`.

However, I think it might be better to replace the "Default: false" by a more understandable, affirmative-form phrase... Something like "By default, eager execution is requested if possible."

Spikhalskiy: If you want, add it as a clarification, I'm all for it. But leave an explicit statement of the default value of the parameter too, please. It's the style across the whole SDK and these two styles target a bit different kinds of readers.

Spikhalskiy (approved):
LGTM!

temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflowContext.java
```diff
@@ -354,7 +354,10 @@ private ExecuteActivityParameters constructExecuteActivityParameters(
                 ProtobufTimeUtils.toProtoDuration(options.getStartToCloseTimeout()))
             .setScheduleToCloseTimeout(
                 ProtobufTimeUtils.toProtoDuration(options.getScheduleToCloseTimeout()))
-            .setHeartbeatTimeout(ProtobufTimeUtils.toProtoDuration(options.getHeartbeatTimeout()));
+            .setHeartbeatTimeout(ProtobufTimeUtils.toProtoDuration(options.getHeartbeatTimeout()))
+            .setRequestEagerExecution(
+                !options.isEagerExecutionDisabled()
+                    && Objects.equals(taskQueue, replayContext.getTaskQueue()));
```
cretz, line 360: What is the purpose of this equals check?

temporal-sdk/src/main/java/io/temporal/internal/worker/EagerActivitySlotsReservation.java
```diff
@@ -0,0 +1,88 @@
...
+    if (serverResponse.getActivityTasksCount() > this.acquiredReservationCount)
+      throw new IllegalStateException(
+          String.format(
+              "Unexpectedly received %d eager activities though we only requested %d",
+              serverResponse.getActivityTasksCount(), this.acquiredReservationCount));
+
+    if (this.acquiredReservationCount == 0) return;
+
+    for (PollActivityTaskQueueResponse act : serverResponse.getActivityTasksList())
+      this.eagerActivityDispatcher.dispatchActivity(act);
```
cretz, line 72: What if this throws after dispatching, say, 5? Instead of decrementing the outstanding count here, you're just setting to 0 at the end so when `close()` is called, it will release more than it still has. If you don't expect anything in this function to ever throw, it may be worth noting that.

It might be clearer to rename `acquiredReservationCount` to `outstandingReservedSlots` or something, so that you can decrement as you move past them and have close be exception safe. One might expect that member to _always_ have an accurate count, not just after a successful call here.

temporal-sdk/src/main/java/io/temporal/internal/worker/EagerActivitySlotsReservation.java
```diff
@@ -0,0 +1,88 @@
...
+
+    this.eagerActivityDispatcher.releaseActivitySlotReservations(
```
cretz, line 74: Probably won't affect much because the dispatch calls are fast (right?), but the logic in my head would say to move this above the dispatch step to give slots back to the activity worker as soon as possible

temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflowContext.java
mjameswh: Do not request eager activity execution if the activity's task queue is not the same as the workflow's activity queue. This check is actually redundant (it is done again in `tryReserveActivitySlot`). Doing it at this point is just an optimization, as it avoids the probability that we will need to "alter" the Schedule Activity command later on.

cretz: Gotcha, I was confused by the term "replayContext" but I see in the code above, this is what the unset task queue defaults to, :+1:

temporal-sdk/src/main/java/io/temporal/internal/worker/EagerActivitySlotsReservation.java
mjameswh: Yeah... I debated this with myself while writing this code. The idea of using a try-with-resource actually came from this concern and the difficulty of properly handling the number of outstanding slot reservations counter. (and btw, thanks for the name, its clearer)

I don't think there's actually a risk that `dispatchActivity` might ever fail in our code, but just for the sake of clarity, I'm refactoring this...

PR #1441, opened 2022-09-16
Fix MTLS configuration in spring boot autoconfiguration

yassineb:
## What was changed

* Added tests to ensure ServiceStubs is loaded properly. 
* Moved applying MTLS properties and building SSLContext to within the try-with-resource block.

## Why?

MTLS configuration fails to load certificates due to input stream already closed. The read from the stream happens within the SSLContextBuilder::build method (specifically when calling sslContextBuilder.keyManager) which is outside the try-with-resource.


temporal-spring-boot-autoconfigure-alpha/src/test/resources/pkcs8-crt-chain.pem
```diff
@@ -0,0 +1,28 @@
+-----BEGIN CERTIFICATE-----
```
yassineb, line 1: @Spikhalskiy I copied the certs from temporal-serviceclient. I noticed the certificate is expiring at the end of the year, we should probably update them to avoid unit tests failing in a few months.

Spikhalskiy:
Thank you for the contribution! This looks great.

Spikhalskiy (approved):

PR #1450, opened 2022-09-21
testing module now brings json-path as a transitive dependency

Spikhalskiy:
`json-path` is included in `temporal-sdk` module as optional with `compileOnly` scope. It's not needed for most users because it's required only for operations with Temporal History jsons.
To make things easier for users, it's helpful for the testing module to bring this dependency transitively as most users work with history jsons in tests.

### DIFF
--- a/temporal-testing/build.gradle
+++ b/temporal-testing/build.gradle
@@ -17,9 +17,14 @@ dependencies {
     api project(':temporal-sdk')
     api project(':temporal-test-server')
 
+    // This dependency is included in temporal-sdk module as optional with compileOnly scope.
+    // To make things easier for users, it's helpful for the testing module to bring this dependency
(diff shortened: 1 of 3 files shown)


temporal-testing/build.gradle
```diff
@@ -17,9 +17,14 @@ dependencies {
     api project(':temporal-sdk')
     api project(':temporal-test-server')
 
+    // This dependency is included in temporal-sdk module as optional with compileOnly scope.
+    // To make things easier for users, it's helpful for the testing module to bring this dependency
+    // transitively as most users work with history jsons in tests.
```
cretz, line 22: > most users work with history jsons in tests

_most_ is pretty strong here, but in general I support more required dependencies in testing libraries than runtime ones.

cretz (approved):

PR #1451, opened 2022-09-22
SDK doesn't allow a single WorkflowTask poller and forces the value to '2' if specified

Spikhalskiy:
After #1438 `WorkerOptions.Builder#setMaxConcurrentWorkflowTaskPollers(1)` became an illegal configuration leading a worker to listen only on a sticky task queue. In this PR worker will override 1 to 2 and log a warning about it.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/worker/Worker.java
+++ b/temporal-sdk/src/main/java/io/temporal/worker/Worker.java
@@ -50,12 +50,15 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import javax.annotation.Nonnull;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Hosts activity and workflow implementations. Uses long poll to receive activity and workflow
  * tasks and processes them in a correspondent thread pool.
  */
 public final class Worker {
+  private static final Logger log = LoggerFactory.getLogger(Worker.class);
   private final WorkerOptions options;
   private final String taskQueue;
   final SyncWorkflowWorker workflowWorker;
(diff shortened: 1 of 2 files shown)


temporal-sdk/src/main/java/io/temporal/worker/WorkerOptions.java
```diff
@@ -177,9 +177,12 @@ public Builder setMaxTaskQueueActivitiesPerSecond(double maxTaskQueueActivitiesP
...
     /**
-     * Number of simultaneous poll requests on workflow task queue. Note that the majority of the
-     * workflow tasks will be using host local task queue due to caching. So try incrementing {@link
-     * WorkerFactoryOptions.Builder#setWorkflowHostLocalPollThreadCount(int)} before this one.
+     * Sets the maximum number of simultaneous long poll requests to the Temporal Server to retrieve
+     * workflow tasks. Changing this value will affect the rate at which the worker is able to
+     * consume tasks from a task queue.
+     *
+     * <p>Due to internal logic where pollers alternate between stick and non-sticky queues, this
+     * value cannot be 1 and will be adjusted to 2 if set to that value.
```
Sushisource, line 185:
```suggestion
     * <p>Due to internal logic where pollers alternate between sticky and non-sticky queues, this
     * value cannot be 1 and will be adjusted to 2 if set to that value.
```

Sushisource (approved):
Heh, I ran into the same thing back when working on this in core

temporal-sdk/src/main/java/io/temporal/worker/Worker.java
```diff
@@ -512,11 +515,16 @@ private static SingleWorkerOptions toWorkflowWorkerOptions(
           factoryOptions.getWorkflowHostLocalTaskQueueScheduleToStartTimeout();
     }
 
+    int maxConcurrentWorkflowTaskPollers = options.getMaxConcurrentWorkflowTaskPollers();
```
mjameswh, line 518: Nit: I'd put this in `WorkerOptions::Builder::validateAndBuildWithDefaults()`...

Spikhalskiy: I thought about it and decided against it.
1. Some users may be building with just build. To catch them it will be a duplicate
2. Users may not like validateAndBuildWithDefaults CHANGING their explicitly set values. If they read it somewhere afterward. It's not what this method should do. It should error out if the value specified by the user is incorrect. And we can't error out because of backward compatibility.

After considering it I decided to carry an incorrect but tolerable value over and correct it on a later stage in the Worker.

mjameswh (approved):

cretz (approved):

PR #1452, opened 2022-09-24
Fix leaking Eager Activity Reservation

Spikhalskiy:
Recently added eager activity dispatch leads to a leaky semaphore. If an activity is actually eagerly dispatched, the semaphore gets incremented twice.

Eager activity dispatch functionality needs better tests for potential semaphore leaks.


Sushisource (approved):
I warned you! :laughing: 

Tricky to get these right.

mjameswh (approved):

PR #1454, opened 2022-09-25
Rework WorkflowExecutorCache to remove a global lock

Spikhalskiy:
# Why

WorkflowExecutorCache has a global (for all workflow task processors) lock that especially comes into play when forced evictions are needed from the cache.
The only reason for this lock is to update the `cache` and `inProcessing` collection in a thread-safe manner.
`inProcessing` in the cache is not needed, as it duplicates the functionality and the state of `WorkflowRunLockManager` that maintains all RunIds currently in-flight on the worker.

# What

This PR removes the global lock and reworks the cache to use the `WorkflowRunLockManager` as a source of truth of in-flight runIds.


Spikhalskiy:
Bypassing review requirements to put this PR to test. Code Review will be done retroactively and comments will be addressed in a separate PR before the next release of Java SDK.

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowRunLockManager.java
```diff
@@ -27,20 +27,11 @@
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.ReentrantLock;
 
-final class WorkflowRunLockManager {
+public final class WorkflowRunLockManager {
   private final Map<String, RefCountedLock> runIdLock = new ConcurrentHashMap<>();
```
cretz, line 31: I know not changed in this PR, but would change to type `ConcurrentMap` for documentation/expectation purposes. Confused me at first read.

In fact, after review, maybe the documented type for this should be `ConcurrentHashMap` or at least document the expectations around `compute`, because `Map.compute` isn't thread-safe obviously and `ConcurrentMap.compute` can run the closure multiple times (which is bad for us because we have a side-effecting closure there). But `ConcurrentHashMap.compute` seems to guarantee a single execution.

cretz (commented):
LGTM

Sushisource (commented):
Makes sense, Core has per-run locks. 

It's a much bigger change, but, the bit of code here about multiple WFTs for the same Run ID serializing application by waiting on the lock could likely be simplified by stuffing them into a per-run queue which is kinda how core operates (it's streams/channels, but same idea).

That's a bigger change obviously, food for thought.

PR #1465, opened 2022-09-29
Use longer retry interval on RESOURCE_EXHAUSTED

mjameswh:
# What was changed

Added longer retry interval on gRPC failures of type `RESOURCE_EXHAUSTED`. This interval delay is configurable by the user, but defaults to 1000 ms.

# Why

This failure code indicates the server is having a hard time. Aggressive retry will worsen the situation and could lead to more serious outage. See https://github.com/temporalio/sdk-features/issues/96

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/worker/Poller.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/worker/Poller.java
@@ -206,9 +206,11 @@ final class Poller<T> implements SuspendableWorker {
       this.task = task;
       this.pollBackoffThrottler =
           new BackoffThrottler(
-              pollerOptions.getPollBackoffInitialInterval(),
-              pollerOptions.getPollBackoffMaximumInterval(),
-              pollerOptions.getPollBackoffCoefficient());
+              pollerOptions.getBackoffInitialInterval(),
+              pollerOptions.getBackoffCongestionInitialInterval(),
+              pollerOptions.getBackoffMaximumInterval(),
+              pollerOptions.getBackoffCoefficient(),
+              pollerOptions.getBackoffMaximumJitter());
     }
 
     @Override
@@ -242,7 +244,10 @@ final class Poller<T> implements SuspendableWorker {
           Thread.currentThread().interrupt();
         } else {
           // Don't increase throttle on InterruptedException
-          pollBackoffThrottler.failure();
+          pollBackoffThrottler.failure(
+              (e instanceof StatusRuntimeException)
+                  ? ((StatusRuntimeException) e).getStatus().getCode()
+                  : Status.Code.UNKNOWN);
         }
         uncaughtExceptionHandler.uncaughtException(Thread.currentThread(), e);
       } finally {
--- a/temporal-sdk/src/main/java/io/temporal/internal/worker/Poller.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/worker/Poller.java
@@ -210,7 +210,7 @@ final class Poller<T> implements SuspendableWorker {
               pollerOptions.getBackoffCongestionInitialInterval(),
               pollerOptions.getBackoffMaximumInterval(),
               pollerOptions.getBackoffCoefficient(),
-              pollerOptions.getBackoffMaximumJitter());
+              pollerOptions.getBackoffMaximumJitterCoefficient());
     }
 
     @Override
--- a/temporal-sdk/src/main/java/io/temporal/internal/worker/PollerOptions.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/worker/PollerOptions.java
@@ -54,9 +54,11 @@ public final class PollerOptions {
 
     private int maximumPollRateIntervalMilliseconds = 1000;
     private double maximumPollRatePerSecond;
-    private double pollBackoffCoefficient = 2;
-    private Duration pollBackoffInitialInterval = Duration.ofMillis(100);
-    private Duration pollBackoffMaximumInterval = Duration.ofMinutes(1);
+    private double backoffCoefficient = 2;
+    private Duration backoffInitialInterval = Duration.ofMillis(100);
+    private Duration backoffCongestionInitialInterval = Duration.ofMillis(1000);
+    private Duration backoffMaximumInterval = Duration.ofMinutes(1);
+    private double backoffMaximumJitter = 0.1;
     private int pollThreadCount = 1;
     private String pollThreadNamePrefix;
     private Thread.UncaughtExceptionHandler uncaughtExceptionHandler;
@@ -69,9 +71,11 @@ public final class PollerOptions {
       }
       this.maximumPollRateIntervalMilliseconds = options.getMaximumPollRateIntervalMilliseconds();
       this.maximumPollRatePerSecond = options.getMaximumPollRatePerSecond();
-      this.pollBackoffCoefficient = options.getPollBackoffCoefficient();
-      this.pollBackoffInitialInterval = options.getPollBackoffInitialInterval();
-      this.pollBackoffMaximumInterval = options.getPollBackoffMaximumInterval();
+      this.backoffCoefficient = options.getBackoffCoefficient();
+      this.backoffInitialInterval = options.getBackoffInitialInterval();
+      this.backoffCongestionInitialInterval = options.getBackoffCongestionInitialInterval();
+      this.backoffMaximumInterval = options.getBackoffMaximumInterval();
+      this.backoffMaximumJitter = options.getBackoffMaximumJitter();
       this.pollThreadCount = options.getPollThreadCount();
       this.pollThreadNamePrefix = options.getPollThreadNamePrefix();
       this.uncaughtExceptionHandler = options.getUncaughtExceptionHandler();
@@ -93,23 +97,41 @@ public final class PollerOptions {
     }
 
     /** Coefficient to use when calculating exponential delay in case of failures */
-    public Builder setPollBackoffCoefficient(double pollBackoffCoefficient) {
-      this.pollBackoffCoefficient = pollBackoffCoefficient;
+    public Builder setBackoffCoefficient(double backoffCoefficient) {
+      this.backoffCoefficient = backoffCoefficient;
       return this;
     }
 
     /**
-     * Initial delay in case of failure. If backoff coefficient is 1 then it would be the constant
-     * delay between failing polls.
+     * Initial delay in case of regular failure. If backoff coefficient is 1 then it would be the
+     * constant delay between failing polls.
      */
-    public Builder setPollBackoffInitialInterval(Duration pollBackoffInitialInterval) {
-      this.pollBackoffInitialInterval = pollBackoffInitialInterval;
+    public Builder setBackoffInitialInterval(Duration backoffInitialInterval) {
+      this.backoffInitialInterval = backoffInitialInterval;
+      return this;
+    }
+
+    /**
+     * Initial delay in case of congestion-related failures (i.e. RESOURCE_EXHAUSTED errors). If
+     * backoff coefficient is 1 then it would be the constant delay between failing polls.
+     */
+    public Builder setBackoffCongestionInitialInterval(Duration backoffCongestionInitialInterval) {
+      this.backoffCongestionInitialInterval = backoffCongestionInitialInterval;
       return this;
     }
 
     /** Maximum interval between polls in case of failures. */
-    public Builder setPollBackoffMaximumInterval(Duration pollBackoffMaximumInterval) {
-      this.pollBackoffMaximumInterval = pollBackoffMaximumInterval;
+    public Builder setBackoffMaximumInterval(Duration backoffMaximumInterval) {
+      this.backoffMaximumInterval = backoffMaximumInterval;
+      return this;
+    }
+
+    /**
+     * Maximum amount of jitter to apply. 0.2 means that actual retry time can be +/- 20% of the
+     * calculated time. Set to 0 to disable jitter. Must be lower than 1. Default is 0.1.
+     */
+    public Builder setBackoffMaximumJitter(double backoffMaximumJitter) {
+      this.backoffMaximumJitter = backoffMaximumJitter;
       return this;
     }
 
@@ -151,9 +173,11 @@ public final class PollerOptions {
       return new PollerOptions(
           maximumPollRateIntervalMilliseconds,
           maximumPollRatePerSecond,
-          pollBackoffCoefficient,
-          pollBackoffInitialInterval,
-          pollBackoffMaximumInterval,
+          backoffCoefficient,
+          backoffInitialInterval,
+          backoffCongestionInitialInterval,
+          backoffMaximumInterval,
+          backoffMaximumJitter,
           pollThreadCount,
           uncaughtExceptionHandler,
           pollThreadNamePrefix);
@@ -164,9 +188,11 @@ public final class PollerOptions {
 
   private final int maximumPollRateIntervalMilliseconds;
   private final double maximumPollRatePerSecond;
-  private final double pollBackoffCoefficient;
-  private final Duration pollBackoffInitialInterval;
-  private final Duration pollBackoffMaximumInterval;
+  private final double backoffCoefficient;
+  private final double backoffMaximumJitter;
+  private final Duration backoffInitialInterval;
+  private final Duration backoffCongestionInitialInterval;
+  private final Duration backoffMaximumInterval;
   private final int pollThreadCount;
   private final Thread.UncaughtExceptionHandler uncaughtExceptionHandler;
   private final String pollThreadNamePrefix;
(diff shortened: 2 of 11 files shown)


temporal-sdk/src/main/java/io/temporal/internal/worker/PollerOptions.java
```diff
@@ -99,14 +101,24 @@ public Builder setPollBackoffCoefficient(double pollBackoffCoefficient) {
...
       this.pollBackoffInitialInterval = pollBackoffInitialInterval;
       return this;
     }
 
+    /**
+     * Initial delay in case of failures that justify longer wait period (i.e. RESOURCE_EXHAUSTED
+     * errors). If backoff coefficient is 1 then it would be the constant delay between failing
+     * polls.
+     */
+    public Builder setPollBackoffLongInitialInterval(Duration pollBackoffLongInitialInterval) {
```
mjameswh, line 117: @Spikhalskiy I'm open to name suggestions. I don't like that one, but I think it us best to avoid a name that would forever link this to "resource exhausted" failures.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/RpcRetryOptions.java
```diff
@@ -376,6 +409,9 @@ public void validate(boolean hasToBeFinite) {
     if (initialInterval == null) {
       throw new IllegalStateException("required property initialInterval not set");
     }
+    if (longInitialInterval == null) {
```
mjameswh, line 412: @Spikhalskiy Not sure what's currently the preferred way to add an option to an existing Option object.

There is probably applications out there that use `build()` instead of `validateBuildWithDefaults()`, which means that at some point, there is a risk that either the new option is null, or the option object provided by the user is used to create a new builder which is `validated()` (I have seen a few occurences of this). How do you usually deal with this?

temporal-serviceclient/src/main/java/io/temporal/internal/AbstractBackoffThrottler.java
```diff
@@ -0,0 +1,114 @@
...
+ * Used to throttle code execution in presence of failures using exponential backoff logic. The
+ * formula used to calculate the next sleep interval is:
+ *
+ * <p>
+ *
+ * <pre>
+ * min(pow(backoffCoefficient, failureCount - 1) * initialSleep, maxSleep);
+ * </pre>
+ */
+abstract class AbstractBackoffThrottler {
```
Spikhalskiy, line 39: This is a typical example where composition should be chosen over an inheritance.
You don't even need or use polymorphism here.

temporal-serviceclient/src/main/java/io/temporal/internal/AbstractBackoffThrottler.java
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  long calculateSleepTime() {
+    ThrottlerState state = lastFailure.get();
+    if (state.failureCount == 0) return 0;
+    Duration effectiveInitialSleep =
+        (state.lastFailureCode == Status.Code.RESOURCE_EXHAUSTED)
+            ? longInitialSleep
+            : regularInitialSleep;
+    double sleepMillis =
```
Spikhalskiy, line 85: Do I understand correctly, that (n+1)th sleep may be shorter than nth sleep if the gRPC code changed?
If it's the case and it's by design - a good comment explaining why it's by design is needed. Otherwise it will look like a bug for the next reader.

temporal-serviceclient/src/main/java/io/temporal/internal/AbstractBackoffThrottler.java
```diff
@@ -0,0 +1,114 @@
...
+
+  private final Duration regularInitialSleep;
+
+  private final Duration longInitialSleep;
+
+  private final Duration maxSleep;
+
+  private final double backoffCoefficient;
+
+  private final AtomicReference<ThrottlerState> lastFailure =
```
Spikhalskiy, line 49: AtomicReference? This class is not thread-safe.

temporal-serviceclient/src/main/java/io/temporal/internal/AbstractBackoffThrottler.java
```diff
@@ -0,0 +1,114 @@
...
+  public void success() {
+    lastFailure.set(INITIAL_THROTTLER_STATE);
+  }
+
+  /** Increment failure count and set last failure code. */
+  public void failure(Status.Code failureCode) {
+    lastFailure.updateAndGet((prev) -> new ThrottlerState(prev.failureCount + 1, failureCode));
+  }
+
+  static final class ThrottlerState {
```
Spikhalskiy, line 103: Why do we need a wrapper if these values can be just non-final fields?

temporal-serviceclient/src/main/java/io/temporal/internal/AbstractBackoffThrottler.java
```diff
@@ -0,0 +1,114 @@
...
+    final int failureCount;
+    final Status.Code lastFailureCode;
+
+    ThrottlerState(int failureCount, Status.Code lastFailureCode) {
+      this.failureCount = failureCount;
+      this.lastFailureCode = lastFailureCode;
+    }
+  }
+
+  static final ThrottlerState INITIAL_THROTTLER_STATE = new ThrottlerState(0, Status.Code.OK);
```
Spikhalskiy, line 113: static variables at the top of the class, should also probably be final

temporal-serviceclient/src/main/java/io/temporal/serviceclient/RpcRetryOptions.java
Spikhalskiy: We shouldn't make this property required. Too nuanced. Either attach it to a value of `initialInterval` with some reasonable multiplier or give it a default value. Users shouldn't fight their way so hard to create a functional instance.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/RpcRetryOptions.java
```diff
@@ -323,6 +348,8 @@ public RpcRetryOptions validateBuildWithDefaults() {
 
   private final Duration initialInterval;
 
+  private final Duration longInitialInterval;
```
Spikhalskiy, line 351: "Long" is terribly unhelpful in understanding the purpose. It's the quality of the value, not a description of the purpose. Let's rename to something like "congestionInitialInterval"

temporal-sdk/src/main/java/io/temporal/internal/worker/PollerOptions.java
Spikhalskiy: Let's remove "Poll" prefix from here completely. And from the neighbors.
"setCongestionInitialInterval" looks fine.
Also, this class is not a public API, we can do whatever we want with this specific class.

temporal-serviceclient/src/main/java/io/temporal/internal/AbstractBackoffThrottler.java
mjameswh: Yes, that is the expected behaviour, per consensus. I agree that it is somewhat counter intuitive and I'll add a comment making it clear that this is known and agreed. 

However, I won't try to explain why we settled on that behaviour rather than any other. TBH I'd have a hard time arguing this without falling in bike shedding, so the only rationales that stands are essentially that its what the server team asked for and that none of us as come up with a better proposition...

temporal-sdk/src/main/java/io/temporal/internal/worker/Poller.java
```diff
@@ -206,9 +206,11 @@ private class PollLoopTask implements Runnable {
       this.task = task;
       this.pollBackoffThrottler =
           new BackoffThrottler(
```
bergundy, line 208: Technically the resource exhausted retry policy could be completely different from the "normal" retry policy.
I'd instead use two different retryer instances but as long as this is all internal and we're free to customize this behavior as needed later this is fine.

bergundy (approved):
Overall LGTM, I'll wait for @Spikhalskiy's input

temporal-serviceclient/src/main/java/io/temporal/internal/BackoffThrottler.java
```diff
@@ -60,32 +71,58 @@
...
+      double backoffCoefficient,
+      double maxJitter) {
+    Objects.requireNonNull(regularInitialSleep, "regularInitialSleep");
+    Objects.requireNonNull(congestionInitialSleep, "congestionInitialSleep");
+    if (backoffCoefficient < 1.0) {
+      throw new IllegalArgumentException(
+          "backoff coefficient less than 1.0: " + backoffCoefficient);
+    }
+    if (maxJitter < 0 || maxJitter >= 1.0) {
+      throw new IllegalArgumentException("maximumJitter has to be >= 0 and < 1.0: " + maxJitter);
```
Spikhalskiy, line 110: maximumJitter -> maxJitter

temporal-serviceclient/src/main/java/io/temporal/internal/BackoffThrottler.java
```diff
@@ -60,32 +71,58 @@
...
     this.backoffCoefficient = backoffCoefficient;
+    this.maxJitter = maxJitter;
   }
 
   public long getSleepTime() {
     if (failureCount == 0) return 0;
-    double sleepMillis = Math.pow(backoffCoefficient, failureCount - 1) * initialSleep.toMillis();
+    Duration initial =
+        (lastFailureCode == Code.RESOURCE_EXHAUSTED) ? congestionInitialSleep : regularInitialSleep;
+    double jitter = Math.random() * maxJitter * 2 - maxJitter;
```
Spikhalskiy, line 123: I would add a comment here that "we generate a randomize jitter coefficient in (-1, 1) range"

Spikhalskiy: Jitter is not the most pleasant mechanic for a user to understand what's going on. I would keep it effective only for codes where jitter actually physically makes sense.
Jitter is needed for Code.RESOURCE_EXHAUSTED to avoid situations when all the parties fail and retry at the same time, getting an uneven throttle. For other codes... probably not?

temporal-serviceclient/src/main/java/io/temporal/internal/retryer/GrpcRetryerUtils.java
```diff
@@ -76,6 +76,8 @@ class GrpcRetryerUtils {
         // By default, we keep retrying with DEADLINE_EXCEEDED assuming that it's the deadline of
         // one attempt which expired, but not the whole sequence.
         break;
+      case RESOURCE_EXHAUSTED:
```
Spikhalskiy, line 79: Why do we need to add it here? Do we want to prohibit users to make RESOURCE_EXHAUSTED not retryable if they want? Why?

temporal-serviceclient/src/main/java/io/temporal/serviceclient/RpcRetryOptions.java
```diff
@@ -93,6 +95,10 @@ public static final class Builder {
 
     private Duration maximumInterval;
 
+    // 0 is a valid value for maximumJitter, and yet, it is not the default value for maximumJitter.
```
Spikhalskiy, line 98: This comment doesn't make much sense to me.
What is "Not configured jitter"? For me it means "disabled jitter". Disabled jitter I believe is exactly maximumJitter = 0. Double.NaN is a "marker for default jitter".

temporal-serviceclient/src/main/java/io/temporal/serviceclient/RpcRetryOptions.java
```diff
@@ -184,6 +206,19 @@ public Builder setMaximumInterval(Duration maximumInterval) {
       return this;
     }
 
+    /**
```
Spikhalskiy, line 209: Needs a comment about NaN that can be used to set a default.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/RpcRetryOptions.java
```diff
@@ -184,6 +206,19 @@ public Builder setMaximumInterval(Duration maximumInterval) {
...
+     * Maximum amount of jitter to apply. 0.2 means that actual retry time can be +/- 20% of the
+     * calculated time. Set to 0 to disable jitter. Must be lower than 1. Default is 0.1.
+     */
+    public Builder setMaximumJitter(double maximumJitter) {
+      if (!Double.isFinite(maximumJitter) || maximumJitter < 0 || maximumJitter >= 1.0) {
```
Spikhalskiy, line 214: It looks like we will throw an exception on NaN. We should allow/support our "magic value" that means "default".

temporal-serviceclient/src/main/java/io/temporal/serviceclient/RpcRetryOptions.java
```diff
@@ -93,6 +95,10 @@ public static final class Builder {
 
     private Duration maximumInterval;
 
+    // 0 is a valid value for maximumJitter, and yet, it is not the default value for maximumJitter.
+    // Use NaN instead as a marker for not configured jitter
+    private double maximumJitter = Double.NaN;
```
Spikhalskiy, line 100: I'm not a big fan of using NaN as a magic value for default. The main reason is that it may be tricky to pass / specify NaN through an external configuration system. 
I would advice using `-1` as a magic value for default here.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/RpcRetryOptions.java
mjameswh: I'm not convinced about this one.

This is similar to `backoffCoefficient`. Marker value for "not explicitly configured" is 0, which will be substituted by a default value of 2.0 in `validateBuildWithDefaults()`, but `setBackoffCoefficient()` wouldn't accept that marker 0 as a valid input. Same with `setInitialInterval()` and `setCongestionInitialInterval()`.

Now, I don't mind fixing all of these, but my feeling is that one is not expected to call `setX` methods uselessly on these builders. If you want to use default value for some property, then just don't call the setter. As far as I can see, the only reason to actually have those "not explicitly configured" markers is to properly handle merging in `setRetryOptions()`, which makes these an internal implementation detail, not something that users should be aware of.

temporal-serviceclient/src/main/java/io/temporal/internal/BackoffThrottler.java
mjameswh: >Jitter is not the most pleasant mechanic for a user to understand what's going on.

Can you elaborate on what is your concern here, from user's POV? Exact timing in unit tests?

In most cases, for non-congestion errors, I'd expect that the numbers we are talking here should be less that some other external factors that might add delay (DNS lookup, network level congestion, garbage collection...). Assuming defaults values, jitter would add at most +/- 20 ms on the third attempt (i.e. effective sleep time in the range [180, 220] ms). Does that sounds bad?

temporal-serviceclient/src/main/java/io/temporal/internal/retryer/GrpcRetryerUtils.java
mjameswh: Don't remember why and can't reproduce now. Removing.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/RpcRetryOptions.java
Spikhalskiy: >but setBackoffCoefficient() wouldn't accept that marker 0 as a valid input.

This should be fixed.

> Now, I don't mind fixing all of these, but my feeling is that one is not expected to call setX methods uselessly on these builders.

With remote configurations, it's fairly standard and needed. Typically, people take remote values and call setters. So, there should be a value that means "default".

temporal-serviceclient/src/main/java/io/temporal/internal/BackoffThrottler.java
Spikhalskiy: > Can you elaborate on what is your concern here, from user's POV? Exact timing in unit tests?

Users may be surprised by unexpected unstable retry periods and may percieve it as a problem/bug. Stable retry periods are easier to grasp and more standard.

> Does that sounds bad?
It's not bad, it's just more complicated than it has to be.

How is it done in other SDKs? Do we apply jitter for all failures?

mjameswh: TypeScript applies a jitter of 10% by default (user configurable) on _every_ grpc failures.

Go applies a jitter of 20% (non configurable) on _every_ grpc failures. The formula is little bit different though (actual sleep time is random in the range `[ sleep before jitter * 0.8, sleep before jitter ]`)

Spikhalskiy (approved):
LGTM! Thanks

PR #1466, opened 2022-09-30
Fix incorrect handling of -RC suffix in version

mjameswh:
## What was changed
Rewrite the way the gradle build script determine the version number.

## Why?
Previous logic would produce incorrect version number if git tag was of the form `1.17.0-RC1-2`, which would cause further build to fail.
See: https://buildkite.com/temporal/java-sdk-public/builds/1877#01838a2b-de0c-42e7-b44a-fe541f1e4570/791-853

### DIFF
--- a/gradle/versioning.gradle
+++ b/gradle/versioning.gradle
@@ -8,11 +8,14 @@ ext.getTag = { ->
     return stdout.toString().trim().substring(1)
 }
 
-// 0.20.1-g000a42a -> 0.21.0-SNAPSHOT
-// 0.20.0-RC1-1-g000a42a -> 0.20.0-RC2-1-SNAPSHOT
-// 0.20.1-some-postfix-2-g000a42a -> 0.21.0-some-postfix-SNAPSHOT
-// 0.20.0 -> 0.20.0
-// Used to name jar files
+// Determine a version name based on git tag. It is used to name JAR files
+// The following rules applies:
+// 0.20.2 -> 0.20.2
+// 0.20.2-RC1 -> 0.20.2-RC1
+// 0.20.2-g000a42a -> 0.21.0-SNAPSHOT
+// 0.20.2-somepostfix-g000a42a -> 0.21.0-SNAPSHOT
+// 0.20.2-RC1-g000a42a -> 0.20.2-SNAPSHOT
+// 0.20.2-RC1-somepostfix-g000a42a -> 0.20.2-SNAPSHOT
 ext.getVersionName = { ->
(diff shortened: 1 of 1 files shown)


Spikhalskiy:
> 0.20.0-RC1-1-g000a42a -> 0.20.0-RC2-1-SNAPSHOT

will it be hard to change the behavior to
0.20.0-RC1-1-g000a42a -> 0.21.0-SNAPSHOT
instead?

mjameswh:
> > 0.20.0-RC1-1-g000a42a -> 0.20.0-RC2-1-SNAPSHOT
> 
> will it be hard to change the behavior to 0.20.0-RC1-1-g000a42a -> 0.21.0-SNAPSHOT instead?

Very easy, but I think it is wrong. Conceptually, 0.20.0-RC1 specifically means that the previous version was 0.19.0 and that you are now in the process of moving to 0.20.0, not 0.21.0...

But I won't fight on this. Your call.

Spikhalskiy:
@mjameswh you are right, corrected the proposal.

gradle/versioning.gradle
```diff
@@ -8,42 +8,28 @@ ext.getTag = { ->
...
-                result += split[i]
-            }
-            return result + '-SNAPSHOT'
+    // The last element of describe should start with g according to git describe format
+    // if we are not exactly on a tag. If it doesn't - we are on a tag
+    def (_, major, minor, patch, rc, extra, git) = (tag =~ /^(\d+)[.](\d+)[.](\d+)(?:-RC(\d+))?(-.*?)?(-g[0-9a-f]+)?$/)[0];
 
+    if (git) {
+        if (rc) {
+            return "${major}.${minor}.${patch}-RC${(rc as int) + 1}-SNAPSHOT";
```
Spikhalskiy, line 28: Should it be “ ${major}.${minor}.0-SNAPSHOT”

mjameswh: I'd be very surprised to ever see a RC on something other than a `x.y.0`, but supposing we do... lets say `1.18.3-RC1-g01234`... then I'd think that `1.18.3-SNAPSHOT` is the proper name. `1.18.0-SNAPSHOT` would break version ordering.

Spikhalskiy: My comment was mostly about current presence of RC postfix in the result of this transformation.

mjameswh: Oh, right! Fixed.

Spikhalskiy (approved):

PR #1467, opened 2022-09-30
Don't use BackoffThrottler in multithread contexts

mjameswh:
## What was changed
- Avoid using BackoffThrottler instances in multi-thread contexts (Poller)
- Refactor `GrpcAsyncRetryer` to carry its state in fields instead of passing them as arguments
- Pass the appropriate ScheduledExecutor to `GrpcAsyncRetryer`, instead of having a global one
- Modify `Grpc*RetryerTest` to more strongly validate either the test is expected to retry, and how long these retries should take (checking for either max or min possible time)

## Why?
- `BackoffThrottler` was not designed to provide an external thread safe semantic; consequently, a race condition between several threads could result in skipping throttling or in throttling more than expected.
- These changes will make it much easier to implement [longer backoff time on RESOURCE_EXHAUSTION failures](https://github.com/temporalio/sdk-java/pull/1465).


temporal-serviceclient/src/main/java/io/temporal/internal/BackoffThrottler.java
```diff
@@ -54,17 +55,38 @@
...
+ *         // some code that can fail and should be throttled
+ *         ...
+ *         throttler.success();
+ *     }
+ *     catch (Exception e) {
+ *         throttler.failure();
+ *     }
+ * }
+ * </pre>
+ *
```
Spikhalskiy, line 75: Please add @NotThreadSafe doc annotation on this class

Spikhalskiy (approved):

Spikhalskiy (approved):

Issue #1494, opened 2022-10-24
Multi-history replay


Sushisource:
Multi-history replay support a-la what's already been done in TS/Python

PR #1497, opened 2022-10-25
Mass replay API

Sushisource:
## What was changed
Added API to workflow replayer accepting multiple histories at once

## Why?
Parity with changes to other SDKs

## Checklist

1. Closes https://github.com/temporalio/sdk-java/issues/1494
2. How was this tested:
Added unit tests

3. Any docs updates needed?

### DIFF
--- a/temporal-testing/src/main/java/io/temporal/testing/WorkflowReplayer.java
+++ b/temporal-testing/src/main/java/io/temporal/testing/WorkflowReplayer.java
@@ -213,7 +213,7 @@ public final class WorkflowReplayer {
         replayWorkflowExecution(history, worker);
       } catch (Exception e) {
         if (!failFast) {
-          results.addError(history.getWorkflowId(), history.getRunId(), e);
+          results.addError(history.getWorkflowExecution().getWorkflowId(), e);
         } else {
           throw e;
         }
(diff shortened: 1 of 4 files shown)


temporal-testing/src/main/java/io/temporal/testing/WorkflowReplayer.java
```diff
@@ -172,6 +172,56 @@ public static void replayWorkflowExecution(WorkflowExecutionHistory history, Wor
...
+   *
+   * @param histories The histories to be replayed
+   * @param failFast If true, throws upon the first error encountered (if any) during replay. If
+   *     false, all histories will be replayed and the returned object contains information about
+   *     any failures.
+   * @return If `failFast` is false, contains any replay failures encountered.
+   * @throws Exception If replay failed and `failFast` is true.
+   */
+  public static ReplayResults replayWorkflowExecutions(
+      Iterable<WorkflowExecutionHistory> histories, boolean failFast, Class<?>... workflowClasses)
```
Sushisource, line 186: Not sure if we want the fail fast flag in here, or more overloads, that way there can be ones that don't throw checked exceptions.

Spikhalskiy (approved):

PR #1507, opened 2022-11-08
Rework Local Activity scheduling

### LINKED ISSUE #1004: Local Activity Worker ignores startToClose and scheduleToClose timeouts
## Expected Behavior

Local Activity Worker should respect `startToClose` timeout and initiate a new activity execution if the activity call doesn't finish during a user-defined timeout. 
A timed-out execution should be abandoned.

## Actual Behavior

`startToClose` is ignored and not enforced anyhow by JavaSDK. The local activity can run up to scheduleToClose timeout is reached even if much shorter `startToClose` is provided.

Spikhalskiy:
## What was changed

This PR implements a timer thread that independently of the local activity executors performs lifecycle management of local activities and allows to enforce timeouts.
The interface between LA workers and workflow code / state machines is reworked to allow LA workers to communicate different failure reasons / retry back to the workflow.
`startToClose` timeout and `scheduleToClose` timeout supports are implemented.
The ground is prepared for adding scheduleToStart timeout for local activities #1512
The workflow/LA interface is improved to prepare for #1261

## Why?

Previously Local Activities scheduling and management code were trivial.
One thread executes the activities, waits with sleep until the next attempt, and executes the next attempt.
While being simple, there is a bunch of Cons to this approach:
1. Executor thread is retained for the whole time of LA execution from the start until the finish.
2. This model allows us to enforce neither `startToClose` nor `scheduleToClose` timeouts because the only thread that manages an activity invocation is tied to the execution of the activity's code.
3. The is no way to start a new attempt until the end of the previous attempt.

Closes #1004

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/common/RetryOptionsUtils.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/common/RetryOptionsUtils.java
@@ -1,0 +1,91 @@
+/*
+ * Copyright (C) 2022 Temporal Technologies, Inc. All Rights Reserved.
+ *
+ * Copyright (C) 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Modifications copyright (C) 2017 Uber Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this material except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.temporal.internal.common;
+
+import io.grpc.Deadline;
+import io.temporal.api.common.v1.RetryPolicy;
+import io.temporal.common.RetryOptions;
+import io.temporal.failure.ActivityFailure;
+import io.temporal.failure.ApplicationFailure;
+import io.temporal.failure.ChildWorkflowFailure;
+import java.time.Duration;
+import java.util.concurrent.TimeUnit;
+import javax.annotation.Nullable;
+
+public class RetryOptionsUtils {
+  public static boolean isNotRetryable(RetryOptions o, @Nullable Throwable e) {
+    if (e == null) {
+      return false;
+    }
+    if (e instanceof ActivityFailure || e instanceof ChildWorkflowFailure) {
+      e = e.getCause();
+    }
+    String type =
+        e instanceof ApplicationFailure
+            ? ((ApplicationFailure) e).getType()
+            : e.getClass().getName();
+    if (o.getDoNotRetry() != null) {
+      for (String doNotRetry : o.getDoNotRetry()) {
+        if (doNotRetry.equals(type)) {
+          return true;
+        }
+      }
+    }
+    return false;
+  }
+
+  public static boolean areAttemptsReached(RetryOptions o, long attempt) {
+    return (o.getMaximumAttempts() != 0 && attempt >= o.getMaximumAttempts());
+  }
+
+  public static boolean isDeadlineReached(@Nullable Deadline deadline, long sleepTimeMs) {
+    return deadline != null && deadline.timeRemaining(TimeUnit.MILLISECONDS) < sleepTimeMs;
+  }
+
+  public static RetryOptions toRetryOptions(RetryPolicy retryPolicy) {
+    RetryOptions.Builder roBuilder = RetryOptions.newBuilder();
+
+    Duration maximumInterval = ProtobufTimeUtils.toJavaDuration(retryPolicy.getMaximumInterval());
+    if (!maximumInterval.isZero()) {
+      roBuilder.setMaximumInterval(maximumInterval);
+    }
+
+    Duration initialInterval = ProtobufTimeUtils.toJavaDuration(retryPolicy.getInitialInterval());
+    if (!initialInterval.isZero()) {
+      roBuilder.setInitialInterval(initialInterval);
+    }
+
+    if (retryPolicy.getBackoffCoefficient() >= 1) {
+      roBuilder.setBackoffCoefficient(retryPolicy.getBackoffCoefficient());
+    }
+
+    if (retryPolicy.getMaximumAttempts() > 0) {
+      roBuilder.setMaximumAttempts(retryPolicy.getMaximumAttempts());
+    }
+
+    roBuilder.setDoNotRetry(
+        retryPolicy
+            .getNonRetryableErrorTypesList()
+            .toArray(new String[retryPolicy.getNonRetryableErrorTypesCount()]));
+
+    return roBuilder.validateBuildWithDefaults();
+  }
+}
--- a/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowRunTaskHandler.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowRunTaskHandler.java
@@ -57,7 +57,6 @@ import java.util.concurrent.LinkedBlockingDeque;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
-import java.util.function.BiFunction;
 
 /**
  * Implements workflow executor that relies on replay of a workflow code. An instance of this class
@@ -70,12 +69,12 @@ class ReplayWorkflowRunTaskHandler implements WorkflowRunTaskHandler {
 
   private final Lock lock = new ReentrantLock();
 
-  private final Functions.Proc1<ActivityTaskHandler.Result> localActivityCompletionSink;
+  private final Functions.Proc1<LocalActivityResult> localActivityCompletionSink;
 
-  private final BlockingQueue<ActivityTaskHandler.Result> localActivityCompletionQueue =
+  private final BlockingQueue<LocalActivityResult> localActivityCompletionQueue =
       new LinkedBlockingDeque<>();
 
-  private final BiFunction<LocalActivityTask, Duration, Boolean> localActivityTaskPoller;
+  private final LocalActivityDispatcher localActivityDispatcher;
 
   private final ReplayWorkflow workflow;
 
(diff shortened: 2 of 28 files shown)


temporal-sdk/src/main/java/io/temporal/internal/common/RetryOptionsUtils.java
```diff
@@ -0,0 +1,91 @@
...
+    if (e == null) {
+      return false;
+    }
+    if (e instanceof ActivityFailure || e instanceof ChildWorkflowFailure) {
+      e = e.getCause();
+    }
+    String type =
+        e instanceof ApplicationFailure
+            ? ((ApplicationFailure) e).getType()
+            : e.getClass().getName();
```
Sushisource, line 44: Is it possible to set an explicit do not retry field in Java or is it only the string matching?

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowRunTaskHandler.java
```diff
@@ -286,10 +285,10 @@ private void processLocalActivityRequests(long startTimeNs) throws InterruptedEx
         // much sense. I believe we should add ScheduleToStart timeout for the local activities
         // as well.
```
Sushisource, line 286: Todo now addressed?

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowRunTaskHandler.java
```diff
@@ -286,10 +285,10 @@ private void processLocalActivityRequests(long startTimeNs) throws InterruptedEx
...
         long maxWaitTimeNs = Math.max(nextWFTHeartbeatTimeNs - System.nanoTime(), 0);
-        boolean accepted =
-            localActivityTaskPoller.apply(
-                new LocalActivityTask(laRequest, localActivityCompletionSink),
-                Duration.ofNanos(maxWaitTimeNs));
+        laRequest.setScheduleToStartTimeout(Duration.ofNanos(maxWaitTimeNs));
```
Sushisource, line 288: Can user not set sched-to-start timeout? (Honestly, makes sense)

temporal-sdk/src/main/java/io/temporal/internal/statemachines/ExecuteLocalActivityParameters.java
```diff
@@ -20,42 +20,113 @@
...
 import java.time.Duration;
+import java.util.Objects;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 
 public class ExecuteLocalActivityParameters {
-  private final PollActivityTaskQueueResponse.Builder activityTask;
+  public static final long NOT_SCHEDULED = -1;
+
+  // It doesn't have all the fields published yet.
```
Sushisource, line 35: What doesn't? This builder? Change to "This builder won't..."

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
```diff
@@ -20,55 +20,87 @@
...
   private static final Logger log = LoggerFactory.getLogger(LocalActivityWorker.class);
 
-  @Nonnull private SuspendableWorker poller = new NoopSuspendableWorker();
+  // RETRY_STATE_IN_PROGRESS shows that it's not the end
+  // for this local activity execution from the workflow point of view.
+  // It's also not conflicting with any other situations
+  // and uniquely identifies the reach of the local retries
+  // and a need to schedule a timer.
+  private static final RetryState LOCAL_RETRY_LIMIT_RETRY_STATE =
+      RetryState.RETRY_STATE_IN_PROGRESS;
```
Sushisource, line 65: Why re-bind this here?

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
```diff
@@ -100,67 +132,96 @@ public void start() {
...
+      LocalActivityExecutionContext executionContext =
+          new LocalActivityExecutionContext(
+              params, resultCallback, localRetryDeadline, scheduleToCloseDeadline);
+
+      LocalActivityAttemptTask task =
+          new LocalActivityAttemptTask(executionContext, params.getInitialActivityTask());
+      boolean accepted = submitAnAttempt(task);
+
+      if (accepted) {
+        if (scheduleToCloseDeadline != null) {
```
Sushisource, line 210: Just `&&` ?

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
```diff
@@ -180,149 +243,375 @@ public void handle(LocalActivityTask task) throws Exception {
...
+              handler.handle(new ActivityTask(activityTask, () -> {}), metricsScope, true);
+        } finally {
+          sw.stop();
+        }
+
+        // Making sure that the result handling code following this statement is mutual exclusive
+        // with the start to close timeout handler.
+        boolean startToCloseTimeoutFired =
+            startToCloseTimeoutFuture != null && !startToCloseTimeoutFuture.cancel(false);
+
```
Sushisource, line 281: Since this is coming after `handle`, does this mean the start-to-close timeout cannot "interrupt" the running activity if it fires - but does it tell the workflow about the timeout right away?

So here we're just saying "OK we (elsewhere) told the WF the LA timed out, and now we're just discarding the result when the LA eventually finishes"?

Do we try to notify the LA of cancel / interrupt it at all? I'm guessing not since there's not a good mechanism for it

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
```diff
@@ -180,149 +243,375 @@ public void handle(LocalActivityTask task) throws Exception {
...
+                  activityId,
+                  RetryState.RETRY_STATE_NON_RETRYABLE_FAILURE,
+                  executionFailure,
+                  currentAttempt,
+                  null));
+          return;
+        }
+
+        if (executionThrowable instanceof Error) {
+          // TODO Error inside Local Activity shouldn't be failing the local activity call.
```
Sushisource, line 322: Meant to be done in this PR?

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
```diff
@@ -180,149 +243,375 @@ public void handle(LocalActivityTask task) throws Exception {
...
+        if (RetryOptionsUtils.areAttemptsReached(retryOptions, currentAttempt)) {
+          executionContext.callback(
+              failed(
+                  activityId,
+                  RetryState.RETRY_STATE_MAXIMUM_ATTEMPTS_REACHED,
+                  executionFailure,
+                  currentAttempt,
+                  null));
+          return;
+        }
```
Sushisource, line 363: Could combine these all by just setting a retry state local var and then passing that into the failed call -- looks like the next few too

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
```diff
@@ -180,149 +243,375 @@ public void handle(LocalActivityTask task) throws Exception {
...
+  private class StartToCloseTimeoutHandler implements Runnable {
+    private final LocalActivityAttemptTask attemptTask;
+
+    private StartToCloseTimeoutHandler(LocalActivityAttemptTask attemptTask) {
+      this.attemptTask = attemptTask;
+    }
+
+    @Override
+    public void run() {
+      LocalActivityExecutionContext executionContext = attemptTask.getExecutionContext();
```
Sushisource, line 461: FYI: I realized yesterday Core isn't retrying start-to-close timeouts right now, we opened a bug to do it

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
```diff
@@ -180,149 +243,375 @@ public void handle(LocalActivityTask task) throws Exception {
...
-        return handleLocalActivity(
-            activityTask, activityStartTimeMs, localRetryThresholdMs, metricsScope);
-      } else {
-        return result;
+      if (executionContext.getLocalRetryDeadline().timeRemaining(TimeUnit.MILLISECONDS)
+          <= sleepMillis) {
+        executionContext.callback(
+            failed(
+                activityId,
+                LOCAL_RETRY_LIMIT_RETRY_STATE,
```
Sushisource, line 513: Is this the state where we schedule a timer?

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowRunTaskHandler.java
Spikhalskiy: Nope! https://github.com/temporalio/sdk-java/issues/1512 Will be in a separate PR.

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
Spikhalskiy: It used in couple of places and I wanted to write a long comment on this value choice. And I don't want the comment to be duplicated.

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
Spikhalskiy: I expect there may be more logic later here and it makes sense to make just an "accepted" code structure, it will be less error-prone later.

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
Spikhalskiy: It's old existing behavior. It's incorrect, but it's not trivial to fix correctly. And directly relevant to the goal of this PR. I will open a separate task for it.

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
Spikhalskiy: Good eye, will refactor it.

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
Spikhalskiy: Where we WILL schedule a timer in one of the upcoming PRs, yeah.

Sushisource (approved):

PR #1514, opened 2022-11-15
Revisit Activity Timeouts

### LINKED ISSUE #1427: Test server heartbeat timeout issue
## Expected Behavior

An activity with a heartbeat timeout of 1s that does a heartbeat then sleeps for 3s should give the workflow a heartbeat timeout

## Actual Behavior

Does not heartbeat timeout

Spikhalskiy:
## What was changed

Fix discrepancies in how Test Server creates activity timeouts with the real Temporal Server.
Fix heartbeat timeouts scheduling if `scheduleToClose` timeout is not present: #1427
Main tests related to Activity Timeouts are unified under `ActivityTimeoutTest`.
Added more tests and checks into this test set.
This test suite now runs with both Test Server and Docker Server.

Bonus: Refactored configuration of the external docker service in unit tests to simplify writing tests running on both the Test Server and Dockerized Server in CI. Reduce code duplication, and allow more conventional ways to pass a boolean flag in env vars.

Closes #1427


temporal-sdk/src/main/java/io/temporal/internal/common/env/EnvironmentVariableUtils.java
```diff
@@ -0,0 +1,43 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.temporal.internal.common.env;
+
+import javax.annotation.Nullable;
+
+public class EnvironmentVariableUtils {
```
cretz, line 25: I usually toss private constructors on these things (not required of course)

cretz (approved):

PR #1542, opened 2022-11-27
Implement retries of local activities that break local retry threshold

### LINKED ISSUE #1261: Local Activities fail and don't retry after reaching LocalRetryThreshold

Spikhalskiy:
Implemented retries of local activities that break the local retry threshold.
Local activities now persist additional information into local activities marker (original scheduled time, last attempt, backoff duration).
If activity execution kept failing, and has more retries to do according to LocalActivityOptions, but is over the Local Retry Threshold, now workflow code will schedule a timer and continues retrying the local activities after the timer is fired. This will be done transparently for users. The timer will be present in the workflow history but is not observable from the user workflow code.

TODO: backward compatibility tests with old histories that don't have new marker fields.

Closes #1261

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/activity/LocalActivityOptions.java
+++ b/temporal-sdk/src/main/java/io/temporal/activity/LocalActivityOptions.java
@@ -81,8 +81,9 @@ public final class LocalActivityOptions {
     }
 
     /**
-     * Maximum time to retry locally, while keeping the Workflow Task open via a Heartbeat. Default
-     * value is Workflow Task timeout multiplied by 6.
+     * Maximum time to wait between retries locally, while keeping the Workflow Task open via a
+     * Heartbeat. If the delay between the attempts becomes larger that this threshold, a Workflow
+     * Timer will be scheduled. Default value is Workflow Task Timeout multiplied by 3.
      */
     public Builder setLocalRetryThreshold(Duration localRetryThreshold) {
       if (localRetryThreshold.isZero() || localRetryThreshold.isNegative()) {
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflowContext.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflowContext.java
@@ -40,6 +40,7 @@ import io.temporal.api.common.v1.SearchAttributes;
 import io.temporal.api.common.v1.WorkflowExecution;
 import io.temporal.api.common.v1.WorkflowType;
 import io.temporal.api.enums.v1.ParentClosePolicy;
+import io.temporal.api.enums.v1.RetryState;
 import io.temporal.api.failure.v1.Failure;
 import io.temporal.api.history.v1.HistoryEvent;
 import io.temporal.api.taskqueue.v1.TaskQueue;
@@ -51,10 +52,7 @@ import io.temporal.common.converter.DataConverter;
 import io.temporal.common.interceptors.Header;
 import io.temporal.common.interceptors.WorkflowInboundCallsInterceptor;
 import io.temporal.common.interceptors.WorkflowOutboundCallsInterceptor;
-import io.temporal.failure.CanceledFailure;
-import io.temporal.failure.ChildWorkflowFailure;
-import io.temporal.failure.FailureConverter;
-import io.temporal.failure.TemporalFailure;
+import io.temporal.failure.*;
 import io.temporal.internal.common.ActivityOptionUtils;
 import io.temporal.internal.common.OptionsUtils;
 import io.temporal.internal.common.ProtobufTimeUtils;
@@ -62,10 +60,7 @@ import io.temporal.internal.common.SearchAttributesUtil;
 import io.temporal.internal.replay.ChildWorkflowTaskFailedException;
 import io.temporal.internal.replay.ReplayWorkflowContext;
 import io.temporal.internal.replay.WorkflowContext;
-import io.temporal.internal.statemachines.ExecuteActivityParameters;
-import io.temporal.internal.statemachines.ExecuteLocalActivityParameters;
-import io.temporal.internal.statemachines.StartChildWorkflowExecutionParameters;
-import io.temporal.internal.statemachines.UnsupportedVersion;
+import io.temporal.internal.statemachines.*;
 import io.temporal.worker.WorkflowImplementationOptions;
 import io.temporal.workflow.CancellationScope;
 import io.temporal.workflow.ChildWorkflowOptions;
@@ -77,6 +72,7 @@ import io.temporal.workflow.Promise;
 import io.temporal.workflow.Workflow;
 import java.lang.reflect.Type;
 import java.time.Duration;
+import java.time.Instant;
 import java.util.*;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
@@ -84,6 +80,8 @@ import java.util.function.BiPredicate;
 import java.util.function.Supplier;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 // TODO separate WorkflowOutboundCallsInterceptor functionality from this class into
 // RootWorkflowOutboundInterceptor
@@ -94,6 +92,8 @@ import javax.annotation.Nullable;
  * WorkflowTask
  */
 final class SyncWorkflowContext implements WorkflowContext, WorkflowOutboundCallsInterceptor {
+  private static final Logger log = LoggerFactory.getLogger(SyncWorkflowContext.class);
+
   private final WorkflowImplementationOptions workflowImplementationOptions;
   private final DataConverter dataConverter;
   private final List<ContextPropagator> contextPropagators;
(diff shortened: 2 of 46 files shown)


temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflowContext.java
```diff
@@ -297,45 +296,118 @@ public void invoke(Optional<Payloads> output, Failure failure) {
...
+                RuntimeException temporalFailure =
+                    FailureConverter.failureToException(
+                        laException.getFailure(), getDataConverter());
+                result.completeExceptionally(temporalFailure);
+              }
+            } else {
+              log.warn(
+                  "[BUG] executeLocalActivity() returned an exception other than LocalActivityFailedException",
+                  e);
+              result.completeExceptionally(
```
Sushisource, line 363: Will this fail the workflow task?

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
```diff
@@ -158,16 +158,17 @@ private boolean submitAnAttempt(LocalActivityAttemptTask task) {
   /**
    * @param executionContext execution context of the activity
    * @param activityTask activity task
-   * @param executionThrowable exception happened during the activity execution. Can be null (for
-   *     startToClose timeout)
+   * @param executionThrowable exception happened during the activity execution. Can be null.
+   * @param
```
Sushisource, line 162: Empty param

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
```diff
@@ -193,6 +194,13 @@ private RetryDecision shouldRetry(
       return new RetryDecision(RetryState.RETRY_STATE_NON_RETRYABLE_FAILURE, null);
     }
 
+    if (executionFailure != null
+        && executionFailure.hasApplicationFailureInfo()
+        && RetryOptionsUtils.isNotRetryable(
+            retryOptions, executionFailure.getApplicationFailureInfo().getType())) {
+      return new RetryDecision(RetryState.RETRY_STATE_NON_RETRYABLE_FAILURE, null);
+    }
```
Sushisource, line 202: Consider assigning this and the expressions above which all result in the same `RETRY_STATE_NON_RETRYABLE_FAILURE` decision  to bools and just or-ing them all together so there's only one return path per unique decision.

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
```diff
@@ -260,8 +262,56 @@ public boolean dispatch(
           new LocalActivityExecutionContext(
               params, resultCallback, localRetryDeadline, scheduleToCloseDeadline);
 
-      LocalActivityAttemptTask task =
-          new LocalActivityAttemptTask(executionContext, params.getInitialActivityTask());
+      PollActivityTaskQueueResponse activityTask = params.getInitialActivityTask();
+      LocalActivityAttemptTask task = new LocalActivityAttemptTask(executionContext, activityTask);
+
+      final Failure previousExecutionFailure = executionContext.getPreviousExecutionFailure();
+      if (previousExecutionFailure != null) {
```
Sushisource, line 269: Maybe consider breaking the inner part of the if block out into a function

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
```diff
@@ -260,8 +262,56 @@ public boolean dispatch(
...
+              // encountering
+              // RetryState#TIMEOUT during calculation of the next attempt (which is effectively a
+              // scheduleToClose
+              // timeout).
+              // See how StartToCloseTimeoutHandler or
+              // io.temporal.internal.testservice.StateMachines#timeoutActivityTask
+              // discard startToClose in this case and replaces it with scheduleToClose
+              failure =
+                  newTimeoutFailure(
+                      TimeoutType.TIMEOUT_TYPE_SCHEDULE_TO_CLOSE,
```
Sushisource, line 293: Weird. I wouldn't expect this to suddenly become `SCHEDULE_TO_CLOSE`

I would expect the failure to be whatever the reason would be because of the current retry policy. EX: if they changed the max attempts to be fewer than were already done, it should be max attempts reached. If they shortened some timeout, it should be timeout failure of that type, etc.

temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflowContext.java
Spikhalskiy: It will fail the workflow execution for now. I mean, this code path should never happen.

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
Spikhalskiy: Not possible, this specific order is important to make sense of what's going on. The RETRY_STATE_RETRY_POLICY_NOT_SET outcome has to divide two code paths leading to RETRY_STATE_NON_RETRYABLE_FAILURE for different reasons.

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
Spikhalskiy: Yeah, I will work on readability & doc side of this PR more.

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
Spikhalskiy: Weird. But this is how Temporal Server works right now.
See this compliance test:
https://github.com/temporalio/sdk-java/blob/c5f0ebbe8a5cd52ee018809ecdd466379b5a7339/temporal-sdk/src/test/java/io/temporal/workflow/activityTests/ActivityTimeoutTest.java#L325
And this piece in Test Server that I wrote to comply with the real server behavior:
https://github.com/temporalio/sdk-java/blob/7b526740eeff0d0e3568716281d0aaf86a344bc8/temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java#L1614

temporal-sdk/src/test/java/io/temporal/workflow/activityTests/LocalActivityRetryOverLocalThresholdTest.java
```diff
@@ -0,0 +1,168 @@
...
+              .setScheduleToCloseTimeout(Duration.ofSeconds(100))
+              .setLocalRetryThreshold(Duration.ofSeconds(5))
+              .setRetryOptions(
+                  RetryOptions.newBuilder()
+                      .setInitialInterval(Duration.ofSeconds(1))
+                      .setBackoffCoefficient(1)
+                      .setMaximumAttempts(12)
+                      .setDoNotRetry(AssertionError.class.getName())
+                      .build())
+              .build();
```
Sushisource, line 135: We discussed on the call, but, for PR reference:

In core this wouldn't ever cause timer backoff to happen. That only happens if the amount we would back off exceeds the local retry threshold.

We agreed to adjust the meaning of local retry threshold to fit that concept. We'll also make it not user-adjustable, since it should really just be a function of workflow task timeout length.

temporal-sdk/src/main/java/io/temporal/activity/LocalActivityOptions.java
```diff
@@ -81,8 +81,9 @@ public Builder setScheduleToCloseTimeout(Duration timeout) {
     }
 
     /**
-     * Maximum time to retry locally, while keeping the Workflow Task open via a Heartbeat. Default
-     * value is Workflow Task timeout multiplied by 6.
+     * Maximum time to wait between retries locally, while keeping the Workflow Task open via a
+     * Heartbeat. If the delay between the attempts becomes larger that this threshold, a Workflow
+     * Timer will be scheduled. Default value is Workflow Task Timeout multiplied by 3.
      */
     public Builder setLocalRetryThreshold(Duration localRetryThreshold) {
```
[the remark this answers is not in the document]
Spikhalskiy: I'm on the fence and not sure it's a good idea to do in this release or potentially at all. I think about a scenario when users may really not like the timer mechanic and want to keep WFT heartbeating for a really long time even if it "makes sense" to schedule timers. For example, to keep the history short, predictable, and easy to read.
Let's catch up on it separately. It's not the last PR in the LA epic anyway.

temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflowContext.java
```diff
@@ -416,10 +495,16 @@ private ExecuteLocalActivityParameters constructExecuteLocalActivityParameters(
         toRetryPolicy(RetryOptions.newBuilder(retryOptions).validateBuildWithDefaults()));
     Duration localRetryThreshold = options.getLocalRetryThreshold();
     if (localRetryThreshold == null) {
-      localRetryThreshold = replayContext.getWorkflowTaskTimeout().multipliedBy(6);
+      localRetryThreshold = replayContext.getWorkflowTaskTimeout().multipliedBy(3);
```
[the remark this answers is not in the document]
Spikhalskiy: No. An arbitrary number is less than 6 because the meaning of this parameter changed from the duration of retries to a duration of a single timer. At the same time, I don't want to keep it close to 1 because of the potential accuracy issues that we discussed.

Issue #1510, opened 2022-11-09
Investigate and fix graceful shutdown problems 


Spikhalskiy:
## Actual Behavior

From the investigation of a user problem:
- [x] `worker.suspendPolling` stops local activity worker from polling. This leads to workflow tasks with local activities getting stuck in a heartbeat loop without making progress. Local activity poller should never stop polling. Local activity poller is an implementation detail and the only source it polls from is an internal worker queue.
- [ ] User reported fact: During worker shutdown some activities are getting distributed to the worker (server registers the dispatch and an attempt), but never start execution on this worker that is shutting down. Theory: Worker long-poll thread can be interrupted when the server has already registered an attempt and already sent the task to the worker, but the worker didn't receive the task yet. Potential mitigation to think through: Server in this scenario should've to get CANCELLED (https://gitter.im/grpc/grpc?at=5afd98b79ad18967b9f2c464) and may register an activity failure.

PR #1560, opened 2022-12-06
Add scheduleToStart timeout to Local Activities

### LINKED ISSUE #1512: Add scheduleToStart to LocalActivityOptions
Local Activities Workers can be saturated in the same way as a regular Activity Workers. 
Temporal SDK should provide a way to set and limit `scheduleToStart` for Local Activities.

Spikhalskiy:
LocalActivityWorker is reworked to remove the unneeded poller and two queues between the workflow code and the executor have been reduced to one.
Backpressure for new executions is implemented with retries always having a green light.
Additional guards were added that prevent retries when activity execution is already completed by a timeout.

Partially addresses #1510
Closes #1512


temporal-sdk/src/main/java/io/temporal/activity/LocalActivityOptions.java
```diff
@@ -81,18 +89,27 @@ public Builder setScheduleToCloseTimeout(Duration timeout) {
     }
 
     /**
-     * Maximum time to wait between retries locally, while keeping the Workflow Task open via a
-     * Heartbeat. If the delay between the attempts becomes larger that this threshold, a Workflow
-     * Timer will be scheduled. Default value is Workflow Task Timeout multiplied by 3.
+     * Time that the Activity Task can stay in the Task Queue before it is picked up by a Worker.
```
Sushisource, line 92:
```suggestion
     * Time that the Activity Task can stay in the worker's internal queue of Local Activities.
```

I think this could be misread to imply there's interaction with a real Task Queue here.

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowRunTaskHandler.java
```diff
@@ -267,28 +272,21 @@ public void close() {
...
-        long maxWaitTimeNs = Math.max(nextWFTHeartbeatTimeNs - System.nanoTime(), 0);
-        laRequest.setScheduleToStartTimeout(Duration.ofNanos(maxWaitTimeNs));
-        boolean accepted = localActivityDispatcher.dispatch(laRequest, localActivityCompletionSink);
-        // TODO this needs to be reworked when we implement a proper scheduleToStart to report a
-        //  Failure in the callback. A proper test is also needed for this scenario.
+        boolean accepted =
+            localActivityDispatcher.dispatch(
+                laRequest, localActivityCompletionSink, wftHeartbeatDeadline);
+        // TODO do we have to fail? if we didn't fit in a potentially tight timeout left until
+        // wftHeartbeatDeadline,
```
Sushisource, line 288: I don't think you do. In core there is no bound on the local activity queue, just the number of slots.

Sushisource (approved):
Looks good to me. This definitely improved the readability in the worker a bit

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowRunTaskHandler.java
Spikhalskiy: Yeah, it's not really a question, it's just a representation of my thinking. I think we should limit it by one WFT timeout though and fail WFT after to get backpressure all the way up. May be especially important if one worker experiences issues that cause long LA execution and exhausting of LA executor threads.
I will think more about how to improve it. This preserves the existing behavior for now in this regard.

Issue #1552, opened 2022-12-01
High level API to list workflows


bergundy:
For the SDK side safe rollout feature we need a user-friendly API for listing workflows.

We've recently added this to TS and Python, Go already exposed a somewhat friendly API.
Java needs this too.

This API belongs on `WorkflowClient` and should return some form of iterator that can be mapped into histories for the mass replayer. The history mapper ideally would support a `concurrency` option to speed up history fetching.

cretz:
My implementation suggestion:

* Add `io.temporal.client.WorkflowExecution` class that is a high-level wrapper over `temporal.api.workflow.v1.WorkflowExecutionInfo`
  * Make sure it has raw proto access
* Add `io.temporal.client.WorkflowExecutionDescription` class that extends `WorkflowExecution` and is a high-level wrapper over `temporal.api.workflowservice.v1.DescribeWorkflowExecutionResponse` 
  * Make sure it has raw proto access (it has a singular workflow execution info inside it which is why it can extend the other one)
* Add `io.temporal.common.WorkflowExecutionHistory` that combines a raw `io.temporal.api.history.v1.History` and a `workflowId`
  * There is already `io.temporal.internal.common.WorkflowExecutionHistory` that is used by the replayer, I suggest this replace that
  * I think it'd make sense to put this in `io.temporal.client` too, but I could understand
  * Add a static `fromJson(String workflowId, String jsonString)` that builds this from our UI/tctl JSON format (there's already a helper)
  * Ideally we could have a `toJson` that doesn't store the workflow ID but does a normal proto-JSON on the history, but we either have to have that use our messed-up Go form of enum storage or have `fromJson` _also_ support the non-messed-up proto form which @dmitry has resisted
* Add `WorkflowClient.describe(String workflowId)` (and overload `describe(String workflowId, Optional<String> runId)`) that returns `WorkflowExecutionDescription`
* Add `WorkflowClient.list(String query)` that returns `Stream<WorkflowExecution>`. I don't expect people to care about options to this function at this time (can add overloads later). I do not expect users need to care about pagination, but if they did, we could easily add `WorkflowClient.listPaginated` that returned `Stream<WorkflowExecutionPage>` or something.
* Add `WorkflowClient.fetchHistoryEvents(String workflowId)`, `WorkflowClient.fetchHistoryEvents(String workflowId, Optional<String> runId)`, and `WorkflowClient.fetchHistoryEvents(WorkflowExecution execution)`
[...]

PR #1566, opened 2022-12-12
Open WorkflowExecutionHistory as a public class

Spikhalskiy:
## What was changed
New `io.temporal.common.WorkflowExecutionHistory` was created that extends a deprecated `io.temporal.internal.common.WorkflowExecutionHistory` as a step to open this class as a part of the official API.

## Why?
Originally WorkflowExecutionHistory wasn't intended as an API class and was created for internal usage only, but after that, it leaked as part of a user-facing test API. To make things right we should move it into an open package.

This change is a prerequisite for #1552


temporal-sdk/src/main/java/io/temporal/internal/common/WorkflowExecutionHistory.java
```diff
@@ -57,7 +62,7 @@ public WorkflowExecutionHistory(History history, String workflowId) {
     this.workflowId = workflowId;
   }
 
-  public static WorkflowExecutionHistory fromJson(String serialized) {
+  public static io.temporal.common.WorkflowExecutionHistory fromJson(String serialized) {
```
cretz, line 65: Might as well also expose this on the `io.temporal.common.WorkflowExecutionHistory` class unless we are trying to hide it intentionally in internal

temporal-sdk/src/main/java/io/temporal/internal/common/WorkflowExecutionHistory.java
```diff
@@ -33,11 +33,16 @@
...
+// To fix the mistake, it was republished in a public package as
+// io.temporal.common.WorkflowExecutionHistory class
+// that extends this class.
 /**
- * Provides a wrapper with convenience methods over raw protobuf {@link History} object representing
- * workflow history
+ * @deprecated use {@link io.temporal.common.WorkflowExecutionHistory} instead.
  */
-public final class WorkflowExecutionHistory {
+@Deprecated
```
cretz, line 44: Is there any value in keeping this class around at all? It's just internal so we can delete at will and just move everything over right?

Spikhalskiy: See my note right above the line you commented on

Spikhalskiy: See my note right above the line you commented on.
Users are already using and depending on `io.temporal.internal.common.WorkflowExecutionHistory`.
Just deleting it without deprecating will bring them to a compile time error that doesn't have a clear remedy (as users will not know where to find the migration information)

temporal-sdk/src/main/java/io/temporal/internal/common/WorkflowExecutionHistory.java
Spikhalskiy: `static` methods are inherited, so it's already done.

temporal-sdk/src/main/java/io/temporal/internal/common/WorkflowExecutionHistory.java
cretz: > Users are already using and depending on [...] internal

:-( But maybe we can make _this_ internal one extend the now-public one then? And remove everything from this internal one onto the public one.

Spikhalskiy: It's possible, but this is worse and moving users in a right direction.
To don't break compile time for users in the hierarchy you propos, I will need our util methods to return INTERNAL class . This way:
1. APIs will be returning deprecated classes.
2. Users using standard IDE autocompletion will continue to get vars defined with a deprecated class.

temporal-sdk/src/main/java/io/temporal/internal/common/WorkflowExecutionHistory.java
Spikhalskiy: Yeah, I can explicitly add a delegate for the doc reason.

temporal-sdk/src/main/java/io/temporal/internal/common/WorkflowExecutionHistory.java
Spikhalskiy: I don't see a reason to support two completely different classes that are essentially copy-pastes of each other.

> Unfortunately it means that you haven't really made a public type with intention to get rid of non-public type. You've made a half-public half-internal type.

I disagree with that and I don't see it like that. I don't see any problem for a user with the current structure. If we want to carry the methods explicitly into the new public class and make them delegating to the super - it can be done and it's not a big issue.

temporal-sdk/src/main/java/io/temporal/internal/common/WorkflowExecutionHistory.java
cretz: :1: Maybe for `toJson` too

Spikhalskiy: Yeah, I will carry all the public methods as delegates for doc purposes and will write javadocs on them in the public class only.

PR #1573, opened 2022-12-16
Remove usage of Workflow time from tracing spans

Spikhalskiy:
## Why?
There is no big reason to use workflow time in spans. It actually may be confusing even during replay to get spans "started in the past".
Usage of Workflow Time creates problems for users that attempt to use span structures and data for profiling, especially on the milliseconds level.


cretz:
Can we confirm people are not using time to deduplicate replayed spans?

cretz (approved):

PR #1581, opened 2023-01-01
Set heartbeat details for TestActivityEnvironment

### LINKED ISSUE #1310: Provide a way to set Heartbeat details in TestActivityEnvironment
It doesn't seem easy (or possible) to test Activity methods which make use of HeartbeatDetails for resuming an aborted process, etc. The `TestActivityEnvironment` does not allow for Activity retries (all exceptions are set to non-retryable), and there is no method for "faking" a previous heartbeat call before calling the Activity method under test.

I'd like to be able to call a method to simulate a previous heartbeat. Whatever mechanism this is would have to be accessible from the `executeActivity` method [here](https://github.com/temporalio/sdk-java/blob/9d843cd927cc0039eb77d30182080f40926eac62/temporal-testing/src/main/java/io/temporal/testing/TestActivityEnvironmentInternal.java#L270). Since the only real externally-facing part of the test environment is the `TestActivityEnvironment` object itself, that would probably be the place to hook into. We'd have to add something like a `setHeartbeatDetails` with parameters indicating the Activity name to apply this heartbeat to as well as the heartbeat payload. Then, inside `executeActivity` it would call a `getHeartbeatDetails` for the given Activity name which would return the heartbeat payload and clear the buffer. (Definitely not married to any of the specifics here, just my initial thought)

Allowing the `TestActivityEnvironment` to retry Activity executions could also work, as long as heartbeats are persisted and available to the retry as normal.
The final option would be to work around this entirely and use `TestWorkflowEnvironment` together with an ad-hoc Workflow that only executes the Activity being tested.

tylercunnion:
## What was changed
Adds a `setHeartbeatDetails` method to `TestActivityEnvironment`. This will make the provided details object available with the next activity execution's call to `ActivityExecutionContext#getHeartbeatDetails`. The details object is cleared when it is applied to an activity execution.

As written this only applies to Activity Executions, not Local Activity Executions. My understanding is that Local Activities do not support heartbeating; if I'm wrong here it is trivial to extend support.

## Why?
Activities can send a "details" object when heartbeating, and that details object is made available to the next run of that activity, should it need to be retried. One of the obvious applications of this as a progress marker; e.g., an activity which is performing a number of tasks can potentially use the details marker to resume partway after being interrupted and retried.

However, there is currently no way to test this functionality within the `TestActivityEnvironment`, since the execution contexts it creates never include heartbeat details. This PR addresses this gap and makes resumption logic and other applications of heartbeat details testable. 

## Checklist

1. Closes #1310 

2. How was this tested:
New set of unit tests

3. Any docs updates needed?
Javadoc is included. I don't believe `TestActivityEnvironment` is documented outside of the Javadoc

### DIFF
--- a/temporal-testing/src/main/java/io/temporal/testing/TestActivityEnvironmentInternal.java
+++ b/temporal-testing/src/main/java/io/temporal/testing/TestActivityEnvironmentInternal.java
@@ -95,6 +95,7 @@ public final class TestActivityEnvironmentInternal implements TestActivityEnviro
   private final TestEnvironmentOptions testEnvironmentOptions;
   private final WorkflowServiceStubs workflowServiceStubs;
   private ClassConsumerPair<Object> activityHeartbeatListener;
+  private Object heartbeatDetails;
 
   public TestActivityEnvironmentInternal(@Nullable TestEnvironmentOptions options) {
     // Initialize an in-memory mock service.
@@ -246,6 +247,11 @@ public final class TestActivityEnvironmentInternal implements TestActivityEnviro
   }
 
   @Override
+  public <T> void setHeartbeatDetails(T details) {
+    heartbeatDetails = details;
+  }
+
+  @Override
   public void close() {
     heartbeatExecutor.shutdownNow();
     activityWorkerExecutor.shutdownNow();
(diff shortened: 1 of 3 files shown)


temporal-testing/src/main/java/io/temporal/testing/TestActivityEnvironmentInternal.java
```diff
@@ -95,6 +95,7 @@ public final class TestActivityEnvironmentInternal implements TestActivityEnviro
   private final TestEnvironmentOptions testEnvironmentOptions;
   private final WorkflowServiceStubs workflowServiceStubs;
   private ClassConsumerPair<Object> activityHeartbeatListener;
+  private Object heartbeatDetails;
```
Spikhalskiy, line 98: Could you replace it with AtomicReference and read-following-with-write in executeActivity with `getAndSet`?
It's mostly to eliminate a possibility of a write happened in `setHeartbeatDetails` being overridden with null without ever being read if there is a race.

Spikhalskiy (commented):
Great contribution, thank you!
There is just one little comment about thread-safety, but it looks great otherwise

temporal-testing/src/main/java/io/temporal/testing/TestActivityEnvironmentInternal.java
tylercunnion: Makes sense, thanks! New commit should address

Spikhalskiy (approved):

PR #1583, opened 2023-01-02
List workflow API

### LINKED ISSUE #1552: High level API to list workflows

Spikhalskiy:
## What was changed

`WorkflowClient#listWorkflowExecutions` was added to provide an easy-to-use wrapper over the underlying `{WorkflowServiceStub#listWorkflowExecutions(ListWorkflowExecutionsRequest)}`

Closes #1552

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/client/ListWorkflowExecutionIterator.java
+++ b/temporal-sdk/src/main/java/io/temporal/client/ListWorkflowExecutionIterator.java
@@ -1,0 +1,133 @@
+/*
+ * Copyright (C) 2022 Temporal Technologies, Inc. All Rights Reserved.
+ *
+ * Copyright (C) 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Modifications copyright (C) 2017 Uber Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this material except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.temporal.client;
+
+import com.google.protobuf.ByteString;
+import io.temporal.api.workflow.v1.WorkflowExecutionInfo;
+import io.temporal.api.workflowservice.v1.ListWorkflowExecutionsRequest;
+import io.temporal.api.workflowservice.v1.ListWorkflowExecutionsResponse;
+import io.temporal.internal.client.external.GenericWorkflowClient;
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+import java.util.Objects;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import javax.annotation.Nonnull;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+class ListWorkflowExecutionIterator implements Iterator<WorkflowExecutionInfo> {
+  private static final Logger log = LoggerFactory.getLogger(ListWorkflowExecutionIterator.class);
+
+  private final @Nonnull String query;
+  private final @Nonnull String namespace;
+  private final int pageSize;
+  private final @Nonnull GenericWorkflowClient genericClient;
+  private ListWorkflowExecutionsResponse activeResponse;
+  private int nextActiveResponseIndex;
+  private CompletableFuture<ListWorkflowExecutionsResponse> nextResponse;
+
+  ListWorkflowExecutionIterator(
+      @Nonnull String query,
+      @Nonnull String namespace,
+      int pageSize,
+      @Nonnull GenericWorkflowClient genericClient) {
+    this.query = Objects.requireNonNull(query, "query");
+    this.namespace = Objects.requireNonNull(namespace, "namespace");
+    this.pageSize = pageSize;
+    this.genericClient = Objects.requireNonNull(genericClient, "genericClient");
+  }
+
+  @Override
+  public boolean hasNext() {
+    if (nextActiveResponseIndex < activeResponse.getExecutionsCount()) {
+      return true;
+    }
+    fetch();
+    return nextActiveResponseIndex < activeResponse.getExecutionsCount();
+  }
+
+  @Override
+  public WorkflowExecutionInfo next() {
+    if (hasNext()) {
+      return activeResponse.getExecutions(nextActiveResponseIndex++);
+    } else {
+      throw new NoSuchElementException();
+    }
+  }
+
+  void fetch() {
+    if (nextResponse == null) {
+      // if nextResponse is null, it's the end of the iteration through the pages
+      return;
+    }
+
+    ListWorkflowExecutionsResponse response;
+    try {
+      response = this.nextResponse.get();
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+      throw new RuntimeException(e);
+    } catch (ExecutionException e) {
+      throw new RuntimeException(e.getCause());
+    }
+
+    ByteString nextPageToken = response.getNextPageToken();
+    if (!nextPageToken.isEmpty()) {
+      ListWorkflowExecutionsRequest request =
+          ListWorkflowExecutionsRequest.newBuilder()
+              .setNamespace(namespace)
+              .setQuery(query)
+              .setPageSize(pageSize)
+              .setNextPageToken(nextPageToken)
+              .build();
+      this.nextResponse = genericClient.listWorkflowExecutionsAsync(request);
+    } else {
+      this.nextResponse = null;
+    }
+
+    if (response.getExecutionsCount() == 0 && nextResponse != null) {
+      log.warn(
+          "[BUG] listWorkflowExecutions received an empty collection with a non-empty nextPageToken");
+      // shouldn't be happening, but we want to tolerate it if it does, so we just effectively
+      // skip the empty response and wait for the next one in a blocking manner.
+      // If this actually ever happens as a normal scenario, this skipping should be reworked to
+      // be done asynchronously on a completion of nextResponse future.
+      fetch();
+      return;
+    }
+
+    activeResponse = response;
+    nextActiveResponseIndex = 0;
+  }
+
+  public void init() {
+    ListWorkflowExecutionsRequest request =
+        ListWorkflowExecutionsRequest.newBuilder()
+            .setNamespace(namespace)
+            .setQuery(query)
+            .setPageSize(pageSize)
+            .build();
+    nextResponse = new CompletableFuture<>();
+    nextResponse.complete(genericClient.listWorkflowExecutions(request));
+    fetch();
+  }
+}
--- a/temporal-sdk/src/main/java/io/temporal/client/WorkflowClient.java
+++ b/temporal-sdk/src/main/java/io/temporal/client/WorkflowClient.java
@@ -23,7 +23,6 @@ package io.temporal.client;
 import io.temporal.activity.Activity;
 import io.temporal.activity.ActivityExecutionContext;
 import io.temporal.api.common.v1.WorkflowExecution;
-import io.temporal.internal.sync.WorkflowClientInternal;
 import io.temporal.serviceclient.WorkflowServiceStubs;
 import io.temporal.workflow.Functions;
 import io.temporal.workflow.Functions.Func;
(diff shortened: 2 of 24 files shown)


temporal-sdk/src/main/java/io/temporal/client/ListWorkflowExecutionIterator.java
```diff
@@ -0,0 +1,133 @@
...
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+import java.util.Objects;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import javax.annotation.Nonnull;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ListWorkflowExecutionIterator implements Iterator<WorkflowExecutionInfo> {
```
cretz, line 37: Do we have an every-public-class-needs-javadoc rule?

temporal-sdk/src/main/java/io/temporal/client/ListWorkflowExecutionIterator.java
```diff
@@ -0,0 +1,133 @@
...
+  private final @Nonnull GenericWorkflowClient genericClient;
+  private ListWorkflowExecutionsResponse activeResponse;
+  private int nextActiveResponseIndex;
+  private CompletableFuture<ListWorkflowExecutionsResponse> nextResponse;
+
+  public ListWorkflowExecutionIterator(
+      @Nonnull String query,
+      @Nonnull String namespace,
+      int pageSize,
+      @Nonnull GenericWorkflowClient genericClient) {
```
cretz, line 52: Since this takes an internal type, could justify `ListWorkflowExecutionIterator` only being an interface and moving this impl back to `io.temporal.internal.client` or `io.temporal.internal.client.external` somewhere. But this works for me as is.

temporal-sdk/src/main/java/io/temporal/client/ListWorkflowExecutionIterator.java
```diff
@@ -0,0 +1,133 @@
...
+          "[BUG] listWorkflowExecutions received an empty collection with a non-empty nextPageToken");
+      // shouldn't be happening, but we want to tolerate it if it does, so we just effectively
+      // skip the empty response and wait for the next one in a blocking manner.
+      // If this actually ever happens as a normal scenario, this skipping should be reworked to
+      // be done asynchronously on a completion of nextResponse future.
+      fetch();
+      return;
+    }
+
+    activeResponse = response;
```
cretz, line 118: Hrmm, could probably only have `CompletableFuture<ListWorkflowExecutionsResponse> activeResponse` and not have both "next" and "active", but I think if I'm following this code correctly `activeResponse` should never be null, so LGTM.

temporal-sdk/src/main/java/io/temporal/client/ListWorkflowExecutionIterator.java
```diff
@@ -0,0 +1,133 @@
...
+    nextActiveResponseIndex = 0;
+  }
+
+  public void init() {
```
cretz, line 122: As a public iterator, it can be hard to know that his must be called by a user. Wonder if we want to make the constructor and this package private. Or maybe just the whole class package private.

temporal-sdk/src/main/java/io/temporal/client/WorkflowClient.java
```diff
@@ -237,7 +239,29 @@ WorkflowStub newUntypedWorkflowStub(
...
+
+  /**
+   * A wrapper around {WorkflowServiceStub#listWorkflowExecutions(ListWorkflowExecutionsRequest)}
+   *
+   * @param query Temporal Visibility Query, for syntax see <a
+   *     href="https://docs.temporal.io/visibility#list-filter">Visibility docs</>
+   * @param pageSize a page size to use for the underlying remote pagination
+   * @return iterable that performs remote pagination under the hood
+   */
+  Iterable<WorkflowExecutionMetadata> listWorkflowExecutions(String query, int pageSize);
```
cretz, line 264: This seems right to me, but I just want to confirm it's still normal in modern Java to return iterables/iterators and have the user convert to streams if they want as opposed to returning a stream for them?

temporal-sdk/src/main/java/io/temporal/client/WorkflowClient.java
```diff
@@ -237,7 +239,29 @@ WorkflowStub newUntypedWorkflowStub(
...
+   * A wrapper around {WorkflowServiceStub#listWorkflowExecutions(ListWorkflowExecutionsRequest)}
+   *
+   * <p>Note: This method uses 1000 as the underlying page size. To customize the page size, use
+   * {@link #listWorkflowExecutions(String, int)}
+   *
+   * @param query Temporal Visibility Query, for syntax see <a
+   *     href="https://docs.temporal.io/visibility#list-filter">Visibility docs</>
+   * @return iterable that performs remote pagination under the hood
+   */
+  default Iterable<WorkflowExecutionMetadata> listWorkflowExecutions(String query) {
```
cretz, line 252: Wonder if we should call this list `list` or `listExecutions`. We are already in a `WorkflowClient` and other service calls don't seem to include the extra `Workflow` term.

cretz (commented):
Some things from my comment on #1552 I want to confirm we intentionally don't want before we close that issue:

* `WorkflowStub.describe()` or client-level equivalent to describe a workflow getting `WorkflowExecutionInfo` introduced here (but probably extended to have potentially more details) instead of the raw proto from `WorkflowStub.getExecution()`.
* `WorkflowStub.getHistory()` or client-level equivalent for getting the history + workflow ID for use in a replayer (sorry if I missed it)

Specifically, in all SDKs we are trying to make it easy to "run histories through replayer based on list query" (so technically first bullet point isn't _required_ but might as well). What might user code to do that today look like? Can open another issue for the history side of this if needed, just don't want to lose that part of the reqs.

temporal-sdk/src/main/java/io/temporal/client/ListWorkflowExecutionIterator.java
Spikhalskiy: Doesn't need to be public 👍

temporal-sdk/src/main/java/io/temporal/client/ListWorkflowExecutionIterator.java
Spikhalskiy: I start the `next` request asynchronously at the same time when we only START iterating through `active` to reduce a delay caused by a synchronous remote request in a trivial implementation.

temporal-sdk/src/main/java/io/temporal/client/ListWorkflowExecutionIterator.java
Spikhalskiy: The user shouldn't be creating this class, made package-private

temporal-sdk/src/main/java/io/temporal/client/WorkflowClient.java
Spikhalskiy: I'm a little reserved about using Stream in public APIs of libraries if stream manipulation is not the sole purpose of these APIs. If the producer and consumer of the Stream are in the same codebase - sure, why not, it's handy.

Otherwise, Stream instance carries a bunch of configurations with it that the producer did. Specifically, Spliterator characteristics and if the Stream is parallel or not. These configurations is often an implementation detail of a CONSUMER and important for a CONSUMER, not a producer. Exposing Streams, we make consumers to
1. guess and expect that parameters are correct
2. assert that the returned Stream has expected characteristics
3. or recreate the Stream with the correct characteristics.


I think it's a better and unambiguous experience to return an Iterable and let the consumer create a Stream with the parameters that the consumer expects / relies on.

But I may see how users may no want to do that and expect a simpler higher-level API and if you have a strong opinion - I can rework on returning a Stream.

temporal-sdk/src/main/java/io/temporal/client/WorkflowClient.java
Spikhalskiy: I'm a little reserved about using Stream in public APIs of libraries if stream manipulation is not the sole purpose of these APIs. If the producer and consumer of the Stream are in the same codebase - sure, why not, it's handy.

Otherwise, Stream instance carries a bunch of configurations with it that the producer would define if Stream is in the method definition. Specifically, Spliterator characteristics and if the Stream is parallel or not. These configurations are often an implementation detail of a CONSUMER and important for a CONSUMER, not a producer. Exposing Stream, we make consumers who care about these parameters of the Stream to

guess and expect that parameters are correct
assert that the returned Stream has expected characteristics
or recreate the Stream with the correct characteristics.
I think it's a less error-prone experience to return an unambiguous Iterable and let the consumer create a Stream with the parameters that the consumer expects / relies on.

But I may see how users may no want to do that and expect a simpler higher-level API and if you have a strong opinion - I can rework on returning a Stream.

temporal-sdk/src/main/java/io/temporal/client/WorkflowClient.java
Spikhalskiy: I really prefer it to don't be just `list`. Just because there may be so many other stuff we may want to `list` that may be handy on WorkflowClient.

Belonging to `Workflow`Client is a bit of a weak argument, because one day we may want to list some other executions, that are a part of a workflow, but not a workflow itself, like an Activity.

But I'm fine with listExecutions if you prefer it.

Spikhalskiy:
WorkflowStub.describe() and WorkflowStub.fetchHistory() are not subjects for this PR because they are separate APIs that are not coupled with this one and don't rely on this one, so they will be added in their separate PRs to keep PRs cohesive.

cretz (approved):
> so they will be added in their separate PRs to keep PRs cohesive.

Works for me, just wanted to make sure the need wasn't lost since the original issue will be closed with this merge.

temporal-sdk/src/main/java/io/temporal/client/WorkflowClient.java
cretz: I don't have a strong opinion here. The only thing I might suggest is returning your own `WorkflowExecutionMetadataIterable` here with a `asStream()` on it that knows we don't support parallel. But I think we can add that later and still be considered compatible (I figure we're allowed to raise to higher interface in covariant returns even though it technically breaks explicit implementers of the interface). The real benefit for me for streams was the ability to map and collect and such. From that issue, I was hoping something nice like:

```java
public void runBeforeDeployment(WorkflowClient client, Worker worker, String workflowType) {
  client
    .list("WorkflowType = '" + workflowType + "'")
    .map(client::fetchHistory)
    .parallel()
    .map(worker::replayWorkflowExecution);
}
```

Could exist. Here it adds an extra step, but not that big of a deal.

temporal-sdk/src/main/java/io/temporal/client/WorkflowClient.java
cretz: It just seems off to add `Workflow` in the high-level service wrapper when we don't for start/execute, signal with start, signal, query, etc. In Python we add it because we have a generic client that has workflow and async activity completion and other things in it, but here you're already scoped to a workflow. I can't personally imagine how any other use of `list` on a "workflow client" would make more sense than listing workflows, but I think `listExecutions` at least is in spirit with the other method names.

temporal-sdk/src/main/java/io/temporal/client/WorkflowClient.java
Spikhalskiy: > asStream() on it that knows we don't support parallel

That's exactly my point for not doing this. Users should define if they want a parallel or not parallel stream. There is nothing wrong to create a parallel stream from this iterator.

cretz: That works for me

temporal-sdk/src/main/java/io/temporal/internal/client/external/GenericWorkflowClientImpl.java
```diff
@@ -40,7 +40,8 @@
 
 public final class GenericWorkflowClientImpl implements GenericWorkflowClient {
 
-  private static final ScheduledExecutorService executor =
+  // TODO we need to shutdown this executor
```
Sushisource, line 43: This or next PR?

Sushisource (approved):
:+1:

temporal-sdk/src/main/java/io/temporal/internal/client/external/GenericWorkflowClientImpl.java
Spikhalskiy: another PR, at it's irrelevent to this PR. Just left a note that it should be done.

temporal-sdk/src/main/java/io/temporal/client/WorkflowClient.java
Spikhalskiy: Reconsidered it and will end up exposing the Stream here. The main reason is that correct support of Iterable would mean that iterator() should create a new iterator if called several times and it becomes a bit too much for this method.

PR #1589, opened 2023-01-07
Exposing an ability to fail WorkflowTask for any callback in executor code

### LINKED ISSUE #1533: Error in a local activity should fail the WorkflowTask instead of failing an activity execution
# Current behavior

Currently, Local Activities lack graceful support for Errors that may happen in the Activity thread. If this happens we fail the local activity execution with `RetryState.RETRY_STATE_INTERNAL_SERVER_ERROR`. 

# Expected behavior

If Local Activity throws an `Error`, it should immediately fail Workflow Task. Java Error signals a problem with the Worker and shouldn't lead to a failure of a Local Activity execution or a Workflow.

Spikhalskiy:
## What was changed
Exposing an ability to fail WorkflowTask for any callback in executor code
Refactored a clean separate WorkflowMutableState.
After this change, `ReplayWorkflowContext` has a clean logic over two complementary mutable entities - `WorkflowStateMachines` and `WorkflowMutableState`. Workflow mutable state fields are not scattered over the workflow control (executor) codebase anymore.

This also creates a solid ground to migrate `WorkflowMutableState` into states of state machines in `WorkflowStateMachines` (defining WorkflowStateMachine)

### DIFF
--- a/gradle/licensing.gradle
+++ b/gradle/licensing.gradle
@@ -2,7 +2,8 @@ subprojects {
     apply plugin: 'org.cadixdev.licenser'
     license {
         header rootProject.file('LICENSE.header')
-        exclude '**/*.puml'
+        exclude '**/*.puml', 'io/temporal/api', 'gogoproto/Gogo.java'
+
     }
     tasks.check.dependsOn('checkLicenseMain')
 }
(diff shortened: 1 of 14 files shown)


Sushisource (approved):
This makes sense to me, it's a bit more similar to how core works.

I find `ReplayWorkflowContext` inheriting from `WorkflowMutableState` a bit odd, since this feels more like a has-a than an is-a thing (and I'm used to living without any inheritance at all), but I understand how it's useful here.

Spikhalskiy:
@Sushisource Yeah, I know what you are saying. It does still feel a bit odd to me too. The fact that methods of WorkflowMutableState are EXACTLY the same as ReplayWorkflowContext is more of a coincidence and may not hold in the future.
An alternative is pretty much a copy-paste. I will think about it more. Maybe a copy-paste is actually the right solution here.

PR #1591, opened 2023-01-08
Fail WFT if Local Activity execution experienced an Error

### LINKED ISSUE #1533: Error in a local activity should fail the WorkflowTask instead of failing an activity execution

Spikhalskiy:
## What was changed
Now if a Local Activity throws an `Error`, it will lead to WFT failure, not a failure of activity execution.

Closes #1533

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/replay/WorkflowRunTaskHandler.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/replay/WorkflowRunTaskHandler.java
@@ -21,6 +21,7 @@
 package io.temporal.internal.replay;
 
 import io.temporal.api.workflowservice.v1.PollWorkflowTaskQueueResponseOrBuilder;
+import io.temporal.worker.NonDeterministicException;
 
 /**
  * Task handler that encapsulates a cached workflow and can handle multiple calls to
@@ -35,10 +36,12 @@ public interface WorkflowRunTaskHandler {
    *
    * @param workflowTask task to handle
    * @return an object that can be used to build workflow task completion or failure response
+   * @throws Throwable if processing experienced issues that are considered unrecoverable inside the
+   *     current workflow task. {@link NonDeterministicException} or {@link Error} are such cases.
    */
   WorkflowTaskResult handleWorkflowTask(
       PollWorkflowTaskQueueResponseOrBuilder workflowTask, WorkflowHistoryIterator historyIterator)
-      throws InterruptedException;
+      throws Throwable;
 
   /**
    * Handles a Direct Query (or Legacy Query) scenario. In this case, it's not a real workflow task
--- a/build.gradle
+++ b/build.gradle
@@ -10,7 +10,7 @@ plugins {
     id 'org.cadixdev.licenser' version '0.6.1'
     id 'com.palantir.git-version' version "${palantirGitVersionVersion}" apply false
     id 'io.github.gradle-nexus.publish-plugin' version '1.1.0'
-    id 'com.diffplug.spotless' version '6.12.0' apply false
+    id 'com.diffplug.spotless' version '6.12.1' apply false
     id 'com.github.nbaztec.coveralls-jacoco' version "1.2.15" apply false
 
     //    id 'org.jetbrains.kotlin.jvm' version '1.4.32'
@@ -39,7 +39,7 @@ ext {
     guavaVersion = '31.1-jre' // [10.0,)
     tallyVersion = '0.11.1' // [0.4.0,)
 
-    gsonVersion = '2.10' // [2.0,)
+    gsonVersion = '2.10.1' // [2.0,)
 
     jsonPathVersion = '2.7.0' // compileOnly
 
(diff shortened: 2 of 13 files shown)


temporal-sdk/src/main/java/io/temporal/internal/replay/WorkflowRunTaskHandler.java
```diff
@@ -35,10 +35,11 @@ public interface WorkflowRunTaskHandler {
    *
    * @param workflowTask task to handle
    * @return an object that can be used to build workflow task completion or failure response
+   * @throws Throwable if processing experienced unrecoverable unexpected issues
```
Sushisource, line 38: Should this say "unrecoverable"? Since the idea is to let it try again with WFT failure? Unrecoverable for some smaller scope, but recoverable in the bigger sense.

Sushisource (approved):
Makes sense to me -- this behavior should probably be added to the docstring for `newLocalActivityStub` / wherever else you can start LAs from `Workflow`

temporal-sdk/src/main/java/io/temporal/internal/replay/WorkflowRunTaskHandler.java
Spikhalskiy: Yeah, it may be unclear with this wording. I actually mean "Unrecoverable on this worker in the context of this Workflow Task". Because you are totally correct, it doesn't mean that it's unrecoverable for the workflow, hence WFT failure. I will add a clarification.

Spikhalskiy:
@Sushisource Java devs never throw Errors, so I feel it's overcommunication and just an implied correct behavior. But it may be worth mentioning in LA section of our Developer Guide for Java.

bergundy:
I think I'm missing context here, why is this the intended behavior?

Spikhalskiy:
@bergundy it's explained in the original issue and in the test comment:

```
 * If Local Activity throws an {@link Error}, it should immediately fail Workflow Task. Java Error
 * signals a problem with the Worker and shouldn't lead to a failure of a Local Activity execution
 * or a Workflow.
```

Users wouldn't expect a failure of a Workflow just because their worker is OutOfMemory. Now it leads to a final Local Activity failure.

Spikhalskiy:
@bergundy 
There is no IOError, there is IOException.
There are no exceptions here pretty much, any Error happening in Local Activity is an unrecoverable failure of the JVM by definition of Error.

bergundy:
Looks like there is an `IOError` class that is thrown in extreme cases https://docs.oracle.com/javase/7/docs/api/java/io/IOError.html

According to [Stackoverflow](https://stackoverflow.com/questions/29202584/when-might-an-ioerror-be-thrown), it usually would be thrown from `Console` methods and it's much more typical to get an `IOException`.

Is `AssertionError` supposed to fail the workflow task? I'm not certain.

Anyways, I'm not sure why we need this special behavior for `Error`s but it seems like it should be relatively harmless to merge.

Spikhalskiy:
@bergundy Yeah, I forgot about this guy. IOError is thrown if the file system is broken or some interface is completely inaccessible. For example, your app is reading from console, but the app has no more STDIN (you will get it with Ctrl + D I believe). Anyway, an Error means that the app can't do its job anymore. If a Worker can't do its job, it should never be able to fail an Execution.

bergundy:
Might as well throw a critical failure from the worker in this case instead of failing a WFT.
This behavior might affect low latency applications negatively preventing workflows from completing in a timely fashion.

Spikhalskiy:
@bergundy Nah, there may be some problems with Control code logic that is specific for this specific workflow. We don't want such a workflow to kill the whole workers fleet. WFT failure is the right reaction on such a "Bug" or an "Unexpected state".

Spikhalskiy:
@bergundy Yes, this may be added, while it's tricky. Most java frameworks don't do it. The problem is that there is no reliable way to do anything after an Error, including shutting down a worker. There is also no easy standard way to notify user's code about it. 
But we still should at least try to explicitly fail WFT, and actually specifically for low-latency applications usecase. Otherwise, the Server will have to wait until WFT timeout.

bergundy:
I agree not waiting for WFT timeout is better here but the server also throttles WFT retries with backoff so failing local activity and reporting to the workflow ASAP is the better outcome in (what I believe is) most latency sensitive cases.

Spikhalskiy:
@bergundy 
1.I don't think that even for most even latency sensitive usecases a failure is better than a delay. In some - sure, but not for the majority. 
2. After an Error, reliable reporting of Workflow failure is not possible by Error spec in Java. By this PR I made sure Error will always lead to either a timeout or an explicit WFT failure.
3. To correctly address your situation a low latency app that is absolutely not ready to wait should set scheduleToClose on the workflow. If case of Error, activity didn't really fail. It is the worker who failed.

Unrelated, In addition, for low latency mode of operation we will probably have to reconsider WFT backoff. WFT failure in low latency mode should cause immediate scheduling.

bergundy:
Let's discuss off PR.

PR #1608, opened 2023-01-22
Added ActivityOutput.activityId

mfateev:
## What was changed
Added ActivityOutput.activityId. 

## Why?
The ActivityOutput is used by interceptors, and knowing `activityId` is important for some use cases.
The change is not backward compatible if an interceptor uses ActivityOutput constructor. I believe that there are very few (if any) interceptors that create ActivityOutput explicitly.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkflowOutboundCallsInterceptor.java
+++ b/temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkflowOutboundCallsInterceptor.java
@@ -108,10 +108,16 @@ public interface WorkflowOutboundCallsInterceptor {
   }
 
   final class ActivityOutput<R> {
+    private final String activityId;
     private final Promise<R> result;
 
-    public ActivityOutput(Promise<R> result) {
+    public ActivityOutput(String activityId, Promise<R> result) {
+      this.activityId = activityId;
       this.result = result;
+    }
+
+    public String getActivityId() {
+      return activityId;
     }
 
     public Promise<R> getResult() {
--- a/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowContext.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowContext.java
@@ -51,6 +51,25 @@ import javax.annotation.Nullable;
  */
 public interface ReplayWorkflowContext extends ReplayAware {
 
+  class ScheduleActivityTaskOutput {
+    private final String activityId;
+    private final Functions.Proc1<Exception> cancellationHandle;
+
+    public ScheduleActivityTaskOutput(
+        String activityId, Functions.Proc1<Exception> cancllationHandle) {
+      this.activityId = activityId;
+      this.cancellationHandle = cancllationHandle;
+    }
+
+    public String getActivityId() {
+      return activityId;
+    }
+
+    public Functions.Proc1<Exception> getCancellationHandle() {
+      return cancellationHandle;
+    }
+  }
+
   WorkflowExecution getWorkflowExecution();
 
   WorkflowExecution getParentWorkflowExecution();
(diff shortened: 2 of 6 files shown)


Spikhalskiy (approved):

bergundy:
This is inconsistent with the rest of our SDKs where the activityID is available in the input (or the context in the case of Go).

Spikhalskiy:
Allowing setting ActivityId on the input was deemed a bad practice by us on numerous discussions and we agreed that exposing it was a mistake. 
JavaSDK never allowed setting ActivityId and there are no plans for it. ActivityId is also used as a unique identifier in JavaSDK. 
The only plan is to make the workflow code aware of the activity IDs of activities it started.

bergundy:
What if I want to create a logger that tracks all activity invocations?
I agree that we typically don't want users messing around with the activity IDs, but since we're talking about in interceptor-land, I'd say we should empower users and give them as much information and control as possible.
I don't think that we made a mistake in all of our other SDKs.

Spikhalskiy:
@bergundy they can easily do it with this functionality Max just added in this PR. I suspect this was the reason for this PR.
scheduleActivityTask returns immediately with the metadata of the invocation, so the users can take the activityId and log it.

> I'd say we should empower users and give them as much information and control as possible.
We should give them the information, but I don't think we should give them control over a pretty internal ID field that is easily can be "cooked wrong".

> I don't think that we made a mistake in all of our other SDKs.
Just to reiterate, we had this discussion a couple of times and every time we had a common team-wide understanding that exposing an explicit way to set an activityId was a wrong design choice. If you are not on the same page with the team, we should raise the discussion again.

bergundy:
> It can easily be done with this functionality that Max just added in this PR. I suspect this was actually the reason for this PR.
> `scheduleActivityTask` returns immediately with the metadata of the invocation, so the users can take the `activityId` and log it.

The question is what happens if scheduling fails and the user wants to log the intent, they won't have the `actvityId` available.

Preventing users from accessing the `activityId` in the interceptor is nerfing the capabilities of interceptors which IMHO should have as much freedom as possible and should not be subject to the same safety constraints we put for workflows.

> We should give them (and workflow code) the information, but I don't think we should give them control over an internal ID field that easily can be "cooked wrong".
> 

Whether or not they should be able to mutate that field is irrelevant IMHO.
This is interceptors, anything is possible.

> I have to reiterate, we had this discussion a couple of times. And every time I believe we had a common team-wide understanding that exposing an explicit way to set an `activityId` was a wrong error-prone design choice. There is not much gain for users from it, but quite a bit that can be done wrong. If we are not on the same page, it may be worth raising the discussion again and documenting the outcome this time, so it can be referenced.

I don't think there's agreement here.
There's value in exposing this for manual activity completion.
The issues I see with exposing the ID are:
1. How it's surfaced to users in the Java SDK where activity options are often reused due to relying on the proxy pattern. In Python this is less of a concern.
2. In Go we made the mistake of relying on a user supplied ID for correlating the started and completion events of the activity (we did the same for child workflows there).

If it weren't for the manual activity completion API, I'd agree with you.

Spikhalskiy:
> The question is what happens if scheduling fails 
It's not possible. If it does - it's an SDK bug.

> Preventing users from accessing the activityId in the interceptor is nerfing the capabilities of interceptors which IMHO should have as much freedom as possible and should not be subject to the same safety constraints we put for workflows.

This PR provides an access to it pretty much immediately the activityId is generated. Hard to be faster.

> Whether or not they should be able to mutate that field is irrelevant IMHO.
This is interceptors, anything is possible.

Anything that we will allow. We shouldn't allow unsafe operations compromising state machines functionality.

> I don't think there's agreement here.

We just had a conversation both in Slack and on standup and there is pretty much an agreement here, that the main activityId should not be user-settable.
We can discuss other problems and methods of solving them.

This discussion overgrown the PR. PR delivers the desired functionality in the correct way in accordance how activity machinery is designed in Java.
It looks like majority is in agreement that Java does it right and we should bring other SDKs in the same page, but it's not the point of this PR.

bergundy:
Sure, let’s continue off PR

PR #1612, opened 2023-01-23
Now query on workflow stub that was was used for start follows runIds 

### LINKED ISSUE #1601: Query on a stub used to start a workflow unexpectedly targets only the first run
## Expected Behavior

The query for a WorkflowStub that wasn't created for a specific runId should target the latest run of the workflow and follow ContinueAsNew chain.

## Actual Behavior

The expected behavior holds for a WorkflowStub explicitly created without a runId. But if this stub is used to start a workflow, after this start call the query targets the specific run that was started by this stub.

Spikhalskiy:
## What was changed

Previously if workflow stub was created and used to start a workflow, query on this stub was targeting the specific runId that was created by this stub and didn't follow the chain of runs.
Now the workflow stub follows the chain of runIds until it is explicitly created for a specific runId.

This PR also cleans up the messy internals that provided potentially outdated runIds in logs of operations that were not targeting a specific runId.

Closes #1601

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -43,11 +43,11 @@ ext {
 
     jsonPathVersion = '2.7.0' // compileOnly
 
-    springBootVersion = '2.7.6'// [2.4.0,)
+    springBootVersion = '2.7.8'// [2.4.0,)
 
     // test scoped
     logbackVersion = '1.2.11' // we don't upgrade to 1.3 and 1.4 because they require slf4j 2.x
-    mockitoVersion = '4.11.0'
+    mockitoVersion = '5.0.0'
     junitVersion = '4.13.2'
 }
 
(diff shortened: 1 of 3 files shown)


cretz:
> query on this stub was targeting the specific runId that was created by this stub and didn't follow the chain of runs.

In (some) other SDKs, by default the result from start will not set run ID on query so it'll auto-choose the latest for the workflow ID. But if they did create a handle/stub with an explicit run ID, then we use that. I assume this works similarly? Does it actually "follow runs" instead of just leaving run ID off?

Spikhalskiy:
@cretz 
>  I assume this works similarly?

Exactly. After the change it will still focus on a specific runId if it was created with specific runId. But it will follow runIds if it wasn't.
Before if the stub is created without a runId and after that was used for a start, query suddenly targets only that first runId, which is highly counter-intuitive.

cretz (approved):

PR #1614, opened 2023-01-24
Fix event type that's being used to start Spring Boot workers

### LINKED ISSUE #1579: [Feature Request] Improving replay issue for Timers
Currently timers create a different execution history if timer / sleep duration is <=0 or > 0. A number of users a running into problems on internal workflow replay for this. It would be nice to add some checks to reduce the number of non-deterministic errors timer duration can lead to.

Spikhalskiy:
# Why

Worker's should start on context initialization and should do it on a context start instead

Closes #1579


temporal-spring-boot-autoconfigure-alpha/src/main/java/io/temporal/spring/boot/autoconfigure/RootNamespaceAutoConfiguration.java
```diff
@@ -149,15 +150,23 @@ public WorkerFactoryStarter workerFactoryStarter(WorkerFactory workerFactory) {
     return new WorkerFactoryStarter(workerFactory);
   }
 
-  public static class WorkerFactoryStarter implements ApplicationListener<ContextRefreshedEvent> {
+  // It needs to listed on ContextStartedEvent, not ContextRefreshedEvent.
+  // Using ContextRefreshedEvent will cause start of workers early, during the context
+  // initialization
+  // and potentially incorrect order of bean initialization.
+  // Refresh event can also be fired multiple times during the context initialization.
+  // For this listener to ever work, Soring context needs to be actually started.
```
cretz, line 158:
```suggestion
  // For this listener to ever work, Spring context needs to be actually started.
```

cretz (approved):

temporal-spring-boot-autoconfigure-alpha/src/main/java/io/temporal/spring/boot/autoconfigure/RootNamespaceAutoConfiguration.java
```diff
@@ -149,15 +150,23 @@ public WorkerFactoryStarter workerFactoryStarter(WorkerFactory workerFactory) {
...
+  // Note that a lot of online samples for Spring don't start the context at all:
+  // https://stackoverflow.com/questions/48099355/contextstartedevent-not-firing-in-custom-listener
```
cretz, line 160: I wonder if it's worth having a test that would break if we hadn't changed this code (i.e. a test that would fail if refreshed during initialization or context not started or whatever).

Spikhalskiy: Added a separate test verifying the influence of initialization/refresh and start on WorkerFactory.

PR #1620, opened 2023-01-25
Spring Boot 3 support

### LINKED ISSUE #1619: Support for Spring Boot 3.0 in integration

Temporal Spring Boot Integration is not compatible with Spring Boot 3.0. One specific issue is that `spring.factories` is no longer supported for enabling autoconfiguration, it was deprecated in 2.7 and removed in 3.0 - see https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.0-Migration-Guide#auto-configuration-files. However there may be other issues.

Adding support, making changes where necessary (backwards compatibility may be possible).

Spikhalskiy:
Fix incompatibilities with Sprint Boot 3.
Add `edgeDepsTest` profile that is going to use the latest dependencies versions for compatibility testing.

Closes #1619

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -3,7 +3,7 @@ steps:
     agents:
       queue: "default"
       docker: "*"
-    command: "./gradlew --no-daemon test -x checkLicenseMain -x checkLicenses -x spotlessCheck -x spotlessApply -x spotlessJava"
+    command: "./gradlew --no-daemon test -x checkLicenseMain -x checkLicenses -x spotlessCheck -x spotlessApply -x spotlessJava -P edgeDepsTest"
     timeout_in_minutes: 15
     plugins:
       - docker-compose#v3.8.0:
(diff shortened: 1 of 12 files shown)


.buildkite/pipeline.yml
```diff
@@ -3,7 +3,7 @@ steps:
     agents:
       queue: "default"
       docker: "*"
-    command: "./gradlew --no-daemon test -x checkLicenseMain -x checkLicenses -x spotlessCheck -x spotlessApply -x spotlessJava"
+    command: "./gradlew --no-daemon test -x checkLicenseMain -x checkLicenses -x spotlessCheck -x spotlessApply -x spotlessJava -P edgeDepsTest"
```
cretz, line 6: By only adding `-P edgeDepsTest` here and not the other test below, does this also kinda serve as a way to test with and without these deps?

When we move to GH actions, I think a matrix of Java versions (probably just 8 and latest which the GH action probably supports...it's 19 atm I think) and test servers would make sense. That's what we do in Python (multi version, multi platform, multi server type). There may not be much value in testing different platforms for Java though.

cretz (approved):

.buildkite/pipeline.yml
Spikhalskiy: Yeah, I kinda use "test environment" based config as an "edge" already by putting it on java 18. Now it also uses the new dependencies. "Test server" is a "legacy" environment on java8 and uses the deps that we pack in the release.

Honestly, I don't think a lot of reasons to do a fair matrix here yet.
SDK either works correctly on old/new Java or not. Also, if something breaks on java 11, it will be broken on java 17. So covering the edges is typically enough.
"Edge" dependencies added here only work on new JDKs (Spring Boot 3 requires Java 17 for example).

But we can add [java 8,11,17 : test,real server] matrix to master merges to be sure in (java17;real server) coverage for example.

PR #1623, opened 2023-01-26
Add auto discovery binding strategy by worker name to Spring Boot

Spikhalskiy:
## What?

Added a concept of Worker Name. Worker Name defaults to Worker Task Queue, but can be explicitly set. Worker Name can be used by auto-discovery annotations instead of Task Queues. 

## Why?

Application that incorporates [Task Queue based Versioning strategy](https://community.temporal.io/t/workflow-versioning-strategies/6911) may choose to use explicit Worker names to reference because it adds a level of indirection. 
This way Task Queue name is specified only once in the config and can be easily changed when needed, while all the auto-discovery annotations reference the Worker by its static name.

### DIFF
--- a/temporal-spring-boot-autoconfigure-alpha/src/main/java/io/temporal/spring/boot/autoconfigure/template/WorkersTemplate.java
+++ b/temporal-spring-boot-autoconfigure-alpha/src/main/java/io/temporal/spring/boot/autoconfigure/template/WorkersTemplate.java
@@ -36,7 +36,6 @@ import io.temporal.worker.Worker;
 import io.temporal.worker.WorkerFactory;
 import io.temporal.worker.WorkerFactoryOptions;
 import java.util.*;
-import java.util.stream.Collectors;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import org.slf4j.Logger;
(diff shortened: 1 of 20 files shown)


temporal-spring-boot-autoconfigure-alpha/src/main/java/io/temporal/spring/boot/autoconfigure/template/WorkersTemplate.java
```diff
@@ -115,90 +114,126 @@ WorkerFactory createWorkerFactory(WorkflowClient workflowClient) {
...
+
+  private void configureWorkflowImplementationsByWorkerName(
+      Workers workers, Collection<Class<?>> autoDiscoveredWorkflowImplementationClasses) {
+    for (Class<?> clazz : autoDiscoveredWorkflowImplementationClasses) {
+      WorkflowImpl annotation = clazz.getAnnotation(WorkflowImpl.class);
+
+      for (String workerName : annotation.workers()) {
+        Worker worker = workers.getByName(workerName);
+        if (worker == null) {
+          throw new BeanDefinitionValidationException(
```
cretz, line 203: I almost think we should encourage this indirection just because I find the "worker with name and task queue required" exception to be clearer than creating a worker with default settings when you don't find one for a task queue (you won't know a mistyped task queue name is in your annotation until too late).

temporal-spring-boot-autoconfigure-alpha/src/main/java/io/temporal/spring/boot/autoconfigure/template/WorkersTemplate.java
```diff
@@ -287,4 +372,72 @@ public void setBeanFactory(@Nonnull BeanFactory beanFactory) throws BeansExcepti
...
+  private static class Workers {
+    private final Map<String, Worker> workersByName = new HashMap<>();
+    private final Map<String, Worker> workersByTaskQueue = new HashMap<>();
+    private final List<Worker> workers = new ArrayList<>();
+
+    public void addWorker(WorkerDescriptor workerDescriptor) {
+      Worker newWorker = workerDescriptor.getWorker();
+      Worker existingWorker = workersByName.get(workerDescriptor.getName());
+      if (existingWorker != null
+          && !existingWorker.getTaskQueue().equals(newWorker.getTaskQueue())) {
```
cretz, line 385: Shouldn't name be globally unique regardless of task queue?

cretz (commented):
This change makes sense.

When worker versioning comes around, they may want to run the same worker for different versions. I think users then will definitely want to put their workflows/activities on differently-named workers with the same task queue but maybe different versions.

temporal-spring-boot-autoconfigure-alpha/src/main/java/io/temporal/spring/boot/autoconfigure/template/WorkersTemplate.java
Spikhalskiy: I think that default workers are mostly "quick project to try things" type of deal. Production users will have a config section. But I think it's fine to make workers auto-discoverable for people who want things up and running as easy as possible.

temporal-spring-boot-autoconfigure-alpha/src/main/java/io/temporal/spring/boot/autoconfigure/template/WorkersTemplate.java
Spikhalskiy: It's to catch a situation when someone gives a worker an explicit name of ANOTHER worker's task queue.

cretz (approved):

PR #1628, opened 2023-01-29
Unregistered Local Activity now fails Workflow Task

### LINKED ISSUE #1394: [Spring Boot] Support user defined *Customizers for all *Options classes
Spring boot supports externalized configuration, but for full flexibility, JavaSDK Spring Boot should allow users to intercept the creation of all *Options instances to support modification

Expose all properties through Spring externalized config. This will work for most cases, but it's not trivial when property values are complicated objects.

### LINKED ISSUE #1561: Unregistered Local Activity should fail Workflow Task
We should fail workflow task if workflow code requests the execution of a local activity that is not registered with the worker.
Such a scenario shouldn't fail activity execution.

Spikhalskiy:
# What

Unregistered Local Activity now fails Workflow Task

# Why

Closes #1561


cretz:
Still not personally the biggest fan of deviating from normal activity behavior here from the workflow caller POV. Granted as a dev I almost never want to fail the workflow task, and especially in a DSL case where workflow input has a typo and that input is used to derive the activity name.

Sushisource (approved):

PR #1645, opened 2023-02-10
Implement FailureConverter API

### LINKED ISSUE #1553: FailureConverter API
Add public `FailureConverter` API to allow customizing error <-> failure conversion and `DefaultFailureConverter` implementation that supports encrypting common failure attributes (`message`, `stack_trace`).

This API was added to the other 3 major SDKs already.

mjameswh:
## What was changed

- Exception/Failure conversions now goes through the `DataConverter`;
- The Exception/Failure conversion process can now be customized by passing a custom implementation of the `FailureConverter` interface to DefaultDataConverter;
- The `PayloadDataConverter` now supports encoding of Failure's `message` and `stackTrace` attributes. This is an opt-in feature, that can be enabled by setting `encodeFailureAttributes = true` on constructor.

## Why?

- This API was added to the other 3 major SDKs already.
- Fix #1553


temporal-sdk/src/main/java/io/temporal/common/converter/CodecDataConverter.java
```diff
@@ -100,21 +106,162 @@ public <T> T fromPayloads(
...
+    if (decodedFailure.hasEncodedAttributes()) {
+      encodedFailure.setEncodedAttributes(encode(decodedFailure.getEncodedAttributes()));
+    }
+    switch (decodedFailure.getFailureInfoCase()) {
+      case APPLICATION_FAILURE_INFO:
+        {
+          ApplicationFailureInfo decodedInfo = decodedFailure.getApplicationFailureInfo();
+          ApplicationFailureInfo.Builder encodedInfo = decodedInfo.toBuilder();
+          if (decodedInfo.hasDetails()) {
+            encodedInfo.setDetails(encode(decodedInfo.getDetails()));
```
mjameswh, line 133: @bergundy Can you please confirm that `Failure` details also have to be codec encoded, and that it is _non-optional_ (provided that codecs have indeed been registered)? That means that updated clients that are configured with codecs may suddenly be unable to decode failure details from history generated by previous workers.

Spikhalskiy: > That means that updated clients that are configured with codecs may suddenly be unable to decode failure details from history generated by previous workers.

That's not how Codecs should be implemented though. If implemented correctly. Codecs should check the `encoding` field and don't try to decode that they didn't encode in the first place.

temporal-sdk/src/main/java/io/temporal/failure/DefaultFailureConverter.java
```diff
@@ -0,0 +1,383 @@
...
+      ((TemporalFailure) result).setFailure(failure);
+    }
+    if (failure.getSource().equals(JAVA_SDK) && !failure.getStackTrace().isEmpty()) {
+      StackTraceElement[] stackTrace = parseStackTrace(failure.getStackTrace());
+      result.setStackTrace(stackTrace);
+    }
+    return result;
+  }
+
+  private RuntimeException failureToExceptionImpl(Failure failure, DataConverter dataConverter) {
```
mjameswh, line 120: I'm hesitant here on either this method (and the equivalent one below) should be `protected`, so that users may more easily redefine only the "decision chain", while reusing the `encoded_attributes` and stack trace bits below.

On the opposite side, we could make this class `final` and encourage customization by decorator paradigm instead.

I personally have a slight preference for the second approach, for it promote a safer customization mechanism for the vast majority of users (ie. users are more likely to convert their custom exceptions to Temporal exceptions). I however expect diverging opinions on that question.

temporal-sdk/src/main/java/io/temporal/common/converter/CodecDataConverter.java
mjameswh: Very good point. Dismiss my argument.

temporal-sdk/src/main/java/io/temporal/failure/DefaultFailureConverter.java
```diff
@@ -0,0 +1,383 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DefaultFailureConverter implements FailureConverter {
+
+  private static final Logger log = LoggerFactory.getLogger(DefaultFailureConverter.class);
+
+  public static final String JAVA_SDK = "JavaSDK";
```
cretz, line 59: IMO no need to make this const public. I can understand why people need it, but in the world of optional string interning, people can just use a string themselves. If you do want to leave this as public API, please rename to `JAVA_SDK_FAILURE_SOURCE`.

temporal-sdk/src/main/java/io/temporal/failure/DefaultFailureConverter.java
```diff
@@ -0,0 +1,383 @@
...
+   * compact as it omits most of framework level code.
+   */
+  private static final ImmutableSet<String> CUTOFF_METHOD_NAMES =
+      ImmutableSet.of(
+          "io.temporal.internal.worker.POJOActivityImplementationFactory$POJOActivityImplementation.execute",
+          "io.temporal.internal.sync.POJOWorkflowTaskHandler$POJOWorkflowImplementation.execute");
+
+  /** Used to parse a stack trace line. */
+  private static final String TRACE_ELEMENT_REGEXP =
+      "((?<className>.*)\\.(?<methodName>.*))\\(((?<fileName>.*?)(:(?<lineNumber>\\d+))?)\\)";
```
cretz, line 79: IMO you don't need all these single-reference static fields, but especially this one. Just inline it in the field below.

temporal-sdk/src/main/java/io/temporal/failure/DefaultFailureConverter.java
```diff
@@ -0,0 +1,383 @@
...
+import java.lang.reflect.Type;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DefaultFailureConverter implements FailureConverter {
```
cretz, line 55: Add javadoc on all public types

temporal-sdk/src/main/java/io/temporal/common/converter/DataConverter.java
```diff
@@ -108,6 +109,24 @@ <T> T fromPayloads(
...
       throws DataConverterException;
 
+  /**
+   * Instantiate an appropriate Java Exception from a serialized Failure object. The default
+   * implementation delegates the conversion process to an instance of {@link
+   * io.temporal.failure.FailureConverter}, using this data converter for payload decoding.
+   *
+   * @param failure Failure protobuf object to deserialize into an exception
+   */
+  RuntimeException failureToException(Failure failure);
```
cretz, line 119: I saw a null check later. Do you need to add nullability annotations?

temporal-sdk/src/main/java/io/temporal/failure/DefaultFailureConverter.java
```diff
@@ -0,0 +1,383 @@
...
+
+  @Override
+  public RuntimeException failureToException(Failure failure, DataConverter dataConverter) {
+    if (failure == null) {
+      return null;
+    }
+    if (failure.hasEncodedAttributes()) {
+      Payload encodedAttributesPayload = failure.getEncodedAttributes();
+      Map<String, String> encodedAttributes =
+          DefaultDataConverter.STANDARD_INSTANCE.fromPayload(
```
cretz, line 101: I think in our other SDKs this is expected to use the data/payload converter from the parameter

temporal-sdk/src/main/java/io/temporal/failure/DefaultFailureConverter.java
cretz: > so that users may more easily redefine only the "decision chain", while reusing the encoded_attributes and stack trace bits below.

Nah, IMO these failure converters need to be wrapped not inherited IMO. I'd consider making this whole class "final" but that is a bit much :-). IMO users need to delegate to other failure converters and their entire behavior or not.

temporal-sdk/src/main/java/io/temporal/failure/DefaultFailureConverter.java
```diff
@@ -0,0 +1,383 @@
...
+              encodedAttributesPayload, HashMap.class, HASH_MAP_STRING_STRING_TYPE);
```
cretz, line 102: I think you should use a static inner pojo that represents this attribute as a Jackson type for clearer conversion.

temporal-sdk/src/main/java/io/temporal/common/converter/CodecDataConverter.java
```diff
@@ -100,21 +106,162 @@ public <T> T fromPayloads(
...
-      List<Payload> decodedPayloads = codec.decode(content.get().getPayloadsList());
-      content = Optional.of(Payloads.newBuilder().addAllPayloads(decodedPayloads).build());
+      content = Optional.of(decode(content.get()));
     }
     return dataConverter.fromPayloads(index, content, valueType, valueGenericType);
   }
 
+  @Override
+  public Failure exceptionToFailure(Throwable e) {
+    return this.encodeFailure(dataConverter.exceptionToFailure(e));
```
Spikhalskiy, line 116: This should be configurable. I totally see users who may want to encode normal payloads and leave Failures unencoded. Pretty much have what they have currently.

temporal-sdk/src/main/java/io/temporal/common/converter/CodecDataConverter.java
```diff
@@ -100,21 +106,162 @@ public <T> T fromPayloads(
...
+    return decodedFailure.build();
+  }
+
   @Nonnull
   @Override
   public List<Payload> encode(@Nonnull List<Payload> payloads) {
     return codec.encode(payloads);
   }
 
+  private Payload encode(Payload payload) {
```
Spikhalskiy, line 244: Could you please give these private encode/decode methods more specific names? The amount of overloads drives me a bit crazy on this piece and I don't think it's really needed here, because they are private. I think it will be more readable if it's `encodePayload` and `encodePayloads`

temporal-sdk/src/main/java/io/temporal/common/converter/DataConverter.java
Spikhalskiy: I would suggest marking this method non-null, throw on failure=null and add a javadoc about it. 

I even may see some value for exceptionToFailure to support null. While it shouldn't, but we have situations where we explicitly send encoded "null" Payload, for example as a workflow or activity result.

But there is no reason for a decoding method to support null and I think it will make the contract of this method unclear. Callers of this method should be dealing with null and avoid calling it.

temporal-sdk/src/main/java/io/temporal/common/converter/DefaultDataConverter.java
```diff
@@ -82,6 +88,10 @@ public static void setDefaultDataConverter(DataConverter converter) {
   /**
    * Creates a new instance of {@code DefaultDataConverter} populated with the default list of
    * payload converters.
+   *
+   * <p>The newly created instance will use the default failure converter, configured to not encode
```
Spikhalskiy, line 92: > The newly created instance will use the default failure converter

Nitpicking, but this method returns `DefaultDataConverter`, so this reads strange and feels like a duplication. I think what you try to say is said on the line above: "populated with the default list of payload converters."

Spikhalskiy, line 92:
```suggestion
   * <p>The instance is configured to not encode
```

Nitpicking, but this method returns DefaultDataConverter, so this reads strange and feels like a duplication. I think what you try to say is said on the line above: "populated with the default list of payload converters."

temporal-sdk/src/main/java/io/temporal/common/converter/DefaultDataConverter.java
```diff
@@ -91,9 +101,14 @@ public static DefaultDataConverter newDefaultInstance() {
...
    * array of converters is iterated from the beginning until one of the converters successfully
    * converts the value.
+   *
+   * <p>The newly created instance will use the default failure converter, configured to not encode
+   * message and stack trace. Call {@code withFailureConverterOverride} on the instance to set a
+   * different failure converter.
    */
   public DefaultDataConverter(PayloadConverter... converters) {
     Collections.addAll(this.converters, converters);
+    this.failureConverter = new DefaultFailureConverter();
```
Spikhalskiy, line 111: nit: leave line 110 and 112 next to each other, they work together :)

temporal-sdk/src/main/java/io/temporal/common/converter/DefaultDataConverter.java
```diff
@@ -91,9 +101,14 @@ public static DefaultDataConverter newDefaultInstance() {
    * Creates instance from ordered array of converters. When converting an object to payload the
    * array of converters is iterated from the beginning until one of the converters successfully
    * converts the value.
+   *
+   * <p>The newly created instance will use the default failure converter, configured to not encode
+   * message and stack trace. Call {@code withFailureConverterOverride} on the instance to set a
```
Spikhalskiy, line 106: You may prefer `@link` over `@code` especially when you are linking inside the same class and don't try express "calling a method on a VARIABLE". This will make it much nicer for users to travel around the javadocs, as it will be clickable.

temporal-sdk/src/main/java/io/temporal/common/converter/DefaultDataConverter.java
```diff
@@ -119,6 +134,14 @@ public DefaultDataConverter withPayloadConverterOverrides(
     return this;
   }
 
+  /** Modifies this {@code DefaultDataConverter} by overriding its {@link FailureConverter}. */
+  public DefaultDataConverter withFailureConverterOverride(
```
Spikhalskiy, line 138: Let's make it just `withFailureConverter`. In PayloadConverter that `Override` word was added to show that the passed converters with not replace the full collection, but only the relevant "overrode"e elements. I don't see any reason to add it here, it will only make the comprehension of this method harder.

temporal-sdk/src/main/java/io/temporal/common/converter/CodecDataConverter.java
Quinn-With-Two-Ns: +1 this is how go failure converters work

temporal-sdk/src/main/java/io/temporal/failure/DefaultFailureConverter.java
Spikhalskiy: I think giving cryptic strings a name and a separate place that makes it easier to write a javadoc next to it is a good practice for documentation purposes. I would leave it like this.

temporal-sdk/src/main/java/io/temporal/failure/DefaultFailureConverter.java
Spikhalskiy: concur, I would expect a custom converter here.

temporal-sdk/src/main/java/io/temporal/common/converter/CodecDataConverter.java
Quinn-With-Two-Ns: Oh I see it is configurable it is just in the `encodeFailure` function

temporal-sdk/src/main/java/io/temporal/failure/DefaultFailureConverter.java
Spikhalskiy: Agree with Chad. It's definitely not a piece a user should ever customize. If users really want to implement their CORE logic of FailureConvert - they should be forced to implement their own from scratch and take full responsibility for it. Let's not provide an easy way to hijack the core-core logic as it's intended to be overridable. It really is not.

temporal-sdk/src/main/java/io/temporal/common/converter/DefaultDataConverter.java
```diff
@@ -82,6 +88,10 @@ public static void setDefaultDataConverter(DataConverter converter) {
   /**
    * Creates a new instance of {@code DefaultDataConverter} populated with the default list of
    * payload converters.
+   *
+   * <p>The newly created instance will use the default failure converter, configured to not encode
+   * message and stack trace. Call {@code withFailureConverterOverride} on the instance to set a
```
Spikhalskiy, line 93: This sends an incorrect message to users. I think we should actually remove this doc completely to don't give wrong ideas.
Instead, on CodecConverter we should provide information that it takes care of encoding failures if configured to do so with the right constructor flag.

temporal-sdk/src/main/java/io/temporal/common/converter/DefaultDataConverter.java
Spikhalskiy: > I think we should also add a VERY load and clear doc that most users don't need this method. And something like "If you think about encoding the Failures, see CodecPayloadConverter". Users SHOULDN'T typically use this method for failure encoding.

cretz: Agree. I can't think of a good reason anyone needs a failure converter and since its inception a few months ago I haven't heard of one use of it in the wild (which is why we've been fine for so long not exposing failure conversion semantics to users until now).

temporal-sdk/src/main/java/io/temporal/failure/DefaultFailureConverter.java
mjameswh: The [spec says](https://github.com/temporalio/api/blob/65015c07229047e82645923cc16285548e642794/temporal/api/failure/v1/message.proto#L97) it should be using the default data converter.

temporal-sdk/src/main/java/io/temporal/failure/DefaultFailureConverter.java
mjameswh: Great! Then I'm actually making this class final. There is just no reason to extend this class rather than implements the interface.

cretz: I think that's a bit much unless we change `DefaultDataConverter` to be final too. I think we can discourage inheritance without preventing it. While I generally can understand the final-by-default stance, that is almost a project-wide design decision and I think we generally aren't adding final to all of our classes (@Spikhalskiy correct me if I'm wrong)

temporal-sdk/src/main/java/io/temporal/failure/DefaultFailureConverter.java
cretz: I see in Go and Python that we use the converter given to the default failure converter to convert this two-field JSON object. Can you check TypeScript? If there is a difference we can discuss how to reconcile. For the most part it makes little sense to use a non-default converter I know, but we already have it given to us so why not.

temporal-sdk/src/main/java/io/temporal/common/converter/DataConverter.java
mjameswh: After verification, `exceptionToFailure` is never called with a null exception. I added `@Nonnull` on that one too.

mjameswh:
I just found out that state machines create a lot of `Failure` directly, without going through `FailureConverter`. I will probably have to fix those too. Converting back to draft in the mean time.

mjameswh:
> I just found out that state machines create a lot of `Failure` directly, without going through `FailureConverter`. I will probably have to fix those too. Converting back to draft in the mean time.

After discussion with @Spikhalskiy on various scenarios, we agree that it is preferable _not_ to modify the way state machines make up `Failure` themselves.

That means that those Failures will not be encrypted, but that seems acceptable since those Failures themselves do not contain any sensible information by themselves (no stack trace, messages are constant from the SDK). In case where the failure's cause might contain sensible data, that cause will come directly from the corresponding history event, and thus, will still be encoded anyway.

temporal-sdk/src/main/java/io/temporal/failure/DefaultFailureConverter.java
mjameswh: TypeScript also use the provided payload converter. Modified the Java code to match.

temporal-sdk/src/main/java/io/temporal/common/converter/CodecDataConverter.java
```diff
@@ -100,21 +155,160 @@ public <T> T fromPayloads(
...
+    return dataConverter.failureToException(this.decodeFailure(failure.toBuilder()).build());
+  }
+
+  private Failure.Builder decodeFailure(Failure.Builder failure) {
+    if (failure.hasCause()) {
+      failure.setCause(decodeFailure(failure.getCause().toBuilder()));
+    }
+    if (failure.hasEncodedAttributes()) {
+      Map<String, String> encodedAttributes =
+          fromPayload(failure.getEncodedAttributes(), HashMap.class, HASH_MAP_STRING_STRING_TYPE);
```
cretz, line 242: I think you should use a static inner pojo that represents this attribute as a Jackson type for clearer conversion and field naming.

(restating previous comment since it was outdated)

temporal-sdk/src/main/java/io/temporal/failure/DefaultFailureConverter.java
cretz: This is not that case of cryptic string used elsewhere. This is just a string that's a single parameter to a regex creation below. If Java had auto-regex strings this wouldn't be a separate field. Not sure it makes sense to have every static final regex pattern be two fields.

```java
  /** Used to parse a stack trace line. */
  private static final String TRACE_ELEMENT_REGEXP =
      "((?<className>.*)\\.(?<methodName>.*))\\(((?<fileName>.*?)(:(?<lineNumber>\\d+))?)\\)";

  private static final Pattern TRACE_ELEMENT_PATTERN = Pattern.compile(TRACE_ELEMENT_REGEXP);
```

Provides 0 value over:

```java
  /** Used to parse a stack trace line. */
  private static final Pattern TRACE_ELEMENT_PATTERN = Pattern.compile(
      "((?<className>.*)\\.(?<methodName>.*))\\(((?<fileName>.*?)(:(?<lineNumber>\\d+))?)\\)");
```

cretz (commented):
2 minor comments

Spikhalskiy:
> > I just found out that state machines create a lot of `Failure` directly, without going through `FailureConverter`. I will probably have to fix those too. Converting back to draft in the mean time.
> 
> After discussion with @Spikhalskiy on various scenarios, we agree that it is preferable _not_ to modify the way state machines make up `Failure` themselves.
> 
> That means that those Failures will not be encrypted, but that seems acceptable since those Failures themselves do not contain any sensible information by themselves (no stack trace, messages are constant from the SDK). In case where the failure's cause might contain sensible data, that cause will come directly from the corresponding history event, and thus, will still be encoded anyway.

Yeah. Also to add to this decision and reasoning, Failures created by State Machines can be approached like History Events. 
For example, a failure created by an Activity State machine from `ActivityTaskCancelled` is unencrypted, but the correspondent Event is also unencrypted already. And the only information in the failure comes from this History Event.

The only exception is the Local Activity State Machine which creates its Failure "out of thin air" and publishes it into a marker. But the type and extent of the data published into this Failure is exactly the same as present in Regular Activity's lifecycle history event and mimicks it. So it's only reasonable to don't encode this Failure, as raw history event fields are unencrypted. The underlying actual cause will be encrypted though.

temporal-sdk/src/main/java/io/temporal/common/converter/CodecDataConverter.java
```diff
@@ -60,14 +76,53 @@ public class CodecDataConverter implements DataConverter, PayloadCodec {
...
+   * Setting {@code encodeDefaultAttributes} to true enables codec encoding of Failure's default
+   * attributes. This can be used in conjunction with an encrypting codec to enable encryption of
+   * failures message and stack traces. Note that failure's details are always codec-encoded,
+   * without regard to {@code encodeDefaultAttributes}.
+   *
+   * @param dataConverter to delegate data conversion to
+   * @param codecs to delegate bytes encoding/decoding to. When encoding, the codecs are applied
+   *     last to first meaning the earlier encoders wrap the later ones. When decoding, the decoders
+   *     are applied first to last to reverse the effect
+   * @param encodeDefaultAttributes enable encoding of Failure's default attributes (message and
```
Spikhalskiy, line 116: Let's rename to `encodeFailures`. `default attributes` is a waaaay to internal-api level of naming to be clear for users.

temporal-sdk/src/main/java/io/temporal/common/converter/DataConverter.java
```diff
@@ -108,6 +110,28 @@ <T> T fromPayloads(
...
+   */
+  @Nonnull
+  RuntimeException failureToException(@Nonnull Failure failure);
+
+  /**
+   * Serialize an existing Java Exception into a Failure object. The default implementation
+   * delegates the conversion process to an instance of {@link
+   * io.temporal.failure.FailureConverter}, using this data converter for payload encoding.
+   *
+   * @param e A Java Exception object to serialize into a Failure protobuf object
```
Spikhalskiy, line 129: I believe we don't capitalize the first letters of param sentences. Also, it's any throwable, not just an exception.

temporal-sdk/src/main/java/io/temporal/common/converter/DataConverter.java
```diff
@@ -108,6 +110,28 @@ <T> T fromPayloads(
...
+  /**
+   * Instantiate an appropriate Java Exception from a serialized Failure object. The default
+   * implementation delegates the conversion process to an instance of {@link
+   * io.temporal.failure.FailureConverter}, using this data converter for payload decoding.
+   *
+   * @param failure Failure protobuf object to deserialize into an exception
+   * @throws NullPointerException if failure is null
+   */
+  @Nonnull
+  RuntimeException failureToException(@Nonnull Failure failure);
```
Spikhalskiy, line 122: The return value can and should be changed to a more specific TemporalFailure, will give much greater peace of mind for the calling code.

temporal-sdk/src/main/java/io/temporal/failure/DefaultFailureConverter.java
```diff
@@ -0,0 +1,350 @@
...
+
+  /** Used to parse a stack trace line. */
+  private static final String TRACE_ELEMENT_REGEXP =
+      "((?<className>.*)\\.(?<methodName>.*))\\(((?<fileName>.*?)(:(?<lineNumber>\\d+))?)\\)";
+
+  private static final Pattern TRACE_ELEMENT_PATTERN = Pattern.compile(TRACE_ELEMENT_REGEXP);
+
+  @Override
+  @Nonnull
+  public RuntimeException failureToException(
```
Spikhalskiy, line 78: RuntimeFailure -> TemporalFailure

temporal-sdk/src/main/java/io/temporal/failure/DefaultFailureConverter.java
```diff
@@ -0,0 +1,350 @@
...
+      ((TemporalFailure) result).setFailure(failure);
+    }
+    if (failure.getSource().equals(JAVA_SDK) && !failure.getStackTrace().isEmpty()) {
+      StackTraceElement[] stackTrace = parseStackTrace(failure.getStackTrace());
+      result.setStackTrace(stackTrace);
+    }
+    return result;
+  }
+
+  private RuntimeException failureToExceptionImpl(Failure failure, DataConverter dataConverter) {
```
Spikhalskiy, line 93: RuntimeFailure -> TemporalFailure

temporal-sdk/src/main/java/io/temporal/failure/DefaultFailureConverter.java
```diff
@@ -0,0 +1,350 @@
...
+  private static final Pattern TRACE_ELEMENT_PATTERN = Pattern.compile(TRACE_ELEMENT_REGEXP);
+
+  @Override
+  @Nonnull
+  public RuntimeException failureToException(
+      @Nonnull Failure failure, @Nonnull DataConverter dataConverter) {
+    Preconditions.checkNotNull(failure, "failure");
+    Preconditions.checkNotNull(dataConverter, "dataConverter");
+    RuntimeException result = failureToExceptionImpl(failure, dataConverter);
+    if (result instanceof TemporalFailure) {
```
Spikhalskiy, line 83: This block will become unconditional once you make `RuntimeFailure -> TemporalFailure` change.

temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflowContext.java
```diff
@@ -844,7 +839,7 @@ public SignalExternalOutput signalExternalWorkflow(SignalExternalInput input) {
                     "child workflow failure callback",
                     () ->
                         result.completeExceptionally(
-                            FailureConverter.failureToException(failure, getDataConverter())));
+                            getDataConverter().failureToException(failure)));
```
Spikhalskiy, line 842: nit: I would just use field directly, it belongs to this object :)

temporal-sdk/src/test/java/io/temporal/common/converter/FailureConverterTest.java
```diff
@@ -0,0 +1,323 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+import javax.annotation.Nonnull;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.Timeout;
+
+public class FailureConverterTest {
```
Spikhalskiy, line 65: This test belongs under io.temporal.workflow
It's an integration test, all integration tests are there.
We put tests into the package of the tested class if it's a unit test.
Also, a short description will be very appreciated by the future editor. What situations this test covers and fixates?

temporal-sdk/src/main/java/io/temporal/common/converter/DataConverter.java
cretz: I think _technically_ we're allowing users to convert these to/from any exception they want

Spikhalskiy: It doesn't make much sense. Any exception other than TemporalFailure in workflow code will cause a workflow task failure, while this thing is persisted in the history already. I don't see any legitimate usecase for non-TemporalFailure to be here.
A user that will try to deserialize a Failure into non-TemporalFailure will face issues down the line.

mjameswh: I agree. Other exceptions should be wrapped into some `TemporalFailure`, so that makes sense as the return value from both `DataConverter.failureToException` and `FailureConverter.failureToException`.

cretz (approved):
LGTM, but please wait for @Spikhalskiy approval before merging.

Spikhalskiy (approved):

Spikhalskiy:
Great job, that came together rather neatly.

Issue #1646, opened 2023-02-10
Eager Workflow Task Dispatch


Spikhalskiy:
**Is your feature request related to a problem? Please describe.**

WorkflowClient should request eager workflow task dispatch on #start and #signalWithStart calls from the Server if it's aware of the local worker listening on the correct task queue.

cretz:
May need a features-level issue that this issue is referenced from to for tracking across SDKs

Issue #1230, opened 2022-05-20
Allow dynamic creation and shutdown of workers


Spikhalskiy:
**Is your feature request related to a problem? Please describe.**
In some situations, users may have dynamic task queues. 

**Describe the solution you'd like**
To support this use case without a huge overhead we should allow starting and stopping of individual workers without the need to recreate the whole WorkersFactory

**Describe alternatives you've considered**

**Additional context**
A related Slack discussion: https://temporalio.slack.com/archives/CTRCR8RBP/p1652713381216549
A related task that should be approached together with this one: https://github.com/temporalio/sdk-java/issues/966

Spikhalskiy:
Subtask of #966

Quinn-With-Two-Ns:
Closing as this is currently not supported in any other SDK and not currently planned

PR #1654, opened 2023-02-14
Fix Eager Activity Dispatch Worker state check, rework states of internal Workers

Spikhalskiy:
## What was changed
`isStarted` method was removed from internal Workers because of the potential duality of its meaning.
Internal SDK workers got explicit states in addition to `isX` methods.
Eager Activity Dispatch was fixed to be requested if Activity Worker is in an ACTIVE state.

## Why
Needed to correctly implement #1646 and #1230 
The current `isStarted` method already caused at least one potentially incorrect usage that may lead an eager dispatch to be requested for a suspended worker.


cretz (approved):
Looks all internal, so didn't concern myself with details

PR #1659, opened 2023-02-16
Implement nonfirst attempt LA metering changes

Sushisource:
## What was changed
Attach nonfirst LA attempt count to WFT completes

## Why?
In support of metering changes

## Checklist

1. Closes 

2. How was this tested:
Added integration test

3. Any docs updates needed?
Nope

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowRunTaskHandler.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowRunTaskHandler.java
@@ -284,9 +284,6 @@ class ReplayWorkflowRunTaskHandler implements WorkflowRunTaskHandler {
     }
   }
 
-  /**
-   * @Override public void markWFTClosed() { }
-   */
   private void processLocalActivityRequests(Deadline wftHeartbeatDeadline)
       throws InterruptedException, Throwable {
 
--- a/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowRunTaskHandler.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowRunTaskHandler.java
@@ -48,13 +48,11 @@ import io.temporal.worker.MetricsType;
 import io.temporal.worker.WorkflowImplementationOptions;
 import io.temporal.workflow.Functions;
 import java.time.Duration;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
+import java.util.*;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.LinkedBlockingDeque;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 
@@ -75,6 +73,8 @@ class ReplayWorkflowRunTaskHandler implements WorkflowRunTaskHandler {
       new LinkedBlockingDeque<>();
 
   private final LocalActivityDispatcher localActivityDispatcher;
+
+  private final LocalActivityMeteringHelper localActivityMeteringHelper;
 
   private final ReplayWorkflow workflow;
 
@@ -122,6 +122,7 @@ class ReplayWorkflowRunTaskHandler implements WorkflowRunTaskHandler {
     this.replayWorkflowExecutor =
         new ReplayWorkflowExecutor(workflow, workflowStateMachines, context);
     this.localActivityCompletionSink = localActivityCompletionQueue::add;
+    this.localActivityMeteringHelper = new LocalActivityMeteringHelper();
   }
 
   @Override
@@ -130,6 +131,8 @@ class ReplayWorkflowRunTaskHandler implements WorkflowRunTaskHandler {
       throws Throwable {
     lock.lock();
     try {
+      localActivityMeteringHelper.newWFTStarting();
+
       Deadline wftHearbeatDeadline =
           Deadline.after(
               (long)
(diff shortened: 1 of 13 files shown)


Sushisource:
Draft until proto PR can be merged to master

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowRunTaskHandler.java
```diff
@@ -363,4 +373,36 @@ public void cancel(HistoryEvent cancelEvent) {
       replayWorkflowExecutor.handleWorkflowExecutionCancelRequested(cancelEvent);
     }
   }
+
+  private static class LocalActivityMeteringHelper {
+    private final Map<String, AtomicInteger> firstWftActivities = new HashMap<>();
+    private final Map<String, AtomicInteger> nonFirstWftActivities = new HashMap<>();
+    private final Set<String> completed = new HashSet<>();
+
+    private void newWFTStarting() {
```
Spikhalskiy, line 382: getNonfirstAttempts and newWFTStarting are not atomic, I think it's better to make them thread-safe, because it's fairly simple here.

To do that, getNonfirstAttempts should decrease the nonFirstWftActivities counters by the counts it read from them.
newWFTStarting should set 0 only in counters that are copied from firstWftActivities.

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
```diff
@@ -129,6 +130,7 @@ private Result handleWorkflowTaskWithQuery(
         WorkflowTaskResult wftResult =
             workflowRunTaskHandler.handleWorkflowTask(workflowTask, historyIterator);
         finalCommand = wftResult.isFinalCommand();
+        log.warn("WFT completing, {}", wftResult.getNonfirstLocalActivityAttempts());
```
Spikhalskiy, line 133: I guess it's your debug print, shouldn't be in the final version.

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityExecutionContext.java
```diff
@@ -39,6 +39,7 @@
 class LocalActivityExecutionContext {
   private final @Nonnull ExecuteLocalActivityParameters executionParams;
   private final @Nonnull AtomicInteger currentAttempt;
+  private final @Nonnull AtomicInteger attemptsDuringThisWFT;
```
Spikhalskiy, line 42: why do we need a field? just call executionParams.getAttemptsDuringThisWFT() in the getAttemptsDuringThisWFT

temporal-sdk/src/main/java/io/temporal/internal/statemachines/ExecuteLocalActivityParameters.java
```diff
@@ -49,20 +50,23 @@ public class ExecuteLocalActivityParameters {
   private final @Nonnull Duration localRetryThreshold;
   private final boolean doNotIncludeArgumentsIntoMarker;
   private final @Nullable Duration scheduleToStartTimeout;
+  private final @Nonnull AtomicInteger attemptsDuringThisWFT;
```
Spikhalskiy, line 53: Looking at this pr and at the namings, I think it will be much cleaner to replace it with just an "attemptCompletion" callback. And implement callback in the WorkflowExecutor code.

Sushisource: Yep this looks a bit nicer

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowRunTaskHandler.java
Sushisource: Done :+1:

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
```diff
@@ -106,6 +106,7 @@ private void submitAttempt(
               TimeUnit.MILLISECONDS);
     }
 
+    executionContext.newAttempt();
```
Spikhalskiy, line 109: It should be inside AttemptTaskHandlerImpl#AttemptTaskHandlerImpl#handle right after `attemptTask.markAsTakenFromQueue();`

Spikhalskiy (approved):
Two minor comments, looks very neat otherwise!

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
Sushisource: Done

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowRunTaskHandler.java
```diff
@@ -363,4 +370,41 @@ public void cancel(HistoryEvent cancelEvent) {
...
+      for (String activityId : firstWftActivities.keySet()) {
+        AtomicInteger removed = firstWftActivities.remove(activityId);
+        removed.set(0);
+        nonFirstWftActivities.put(activityId, removed);
+      }
+    }
+
+    private void addNewLocalActivity(ExecuteLocalActivityParameters params) {
+      AtomicInteger attemptsDuringWFTCounter = new AtomicInteger(0);
+      params.setOnNewAttemptCallback(attemptsDuringWFTCounter::incrementAndGet);
```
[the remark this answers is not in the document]
Sushisource: Eh, I think since you end up having to pass both the id and the atomic int itself, it ends up being kinda messy - if the class weren't already nested I think I'd agree but the two are closely associated in a way I think makes sense.

PR #1674, opened 2023-02-24
Eager Workflow Task Dispatch

### LINKED ISSUE #1646: Eager Workflow Task Dispatch

Spikhalskiy:
# What

This PR implements Eager Workflow Task Dispatch support in both JavaSDK and Test Server. 

# Why

Closes #1646

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/client/WorkflowClient.java
+++ b/temporal-sdk/src/main/java/io/temporal/client/WorkflowClient.java
@@ -121,7 +121,8 @@ public interface WorkflowClient {
    * @param service client to the Temporal Service endpoint.
    */
   static WorkflowClient newInstance(WorkflowServiceStubs service) {
-    return WorkflowClientInternal.newInstance(service, WorkflowClientOptions.getDefaultInstance());
+    return WorkflowClientInternalImpl.newInstance(
+        service, WorkflowClientOptions.getDefaultInstance());
   }
 
   /**
@@ -132,7 +133,7 @@ public interface WorkflowClient {
    *     configuring client.
    */
   static WorkflowClient newInstance(WorkflowServiceStubs service, WorkflowClientOptions options) {
-    return WorkflowClientInternal.newInstance(service, options);
+    return WorkflowClientInternalImpl.newInstance(service, options);
   }
 
   WorkflowClientOptions getOptions();
@@ -308,7 +309,7 @@ public interface WorkflowClient {
    * @return WorkflowExecution that contains WorkflowId and RunId of the started workflow.
    */
   static WorkflowExecution start(Functions.Proc workflow) {
-    return WorkflowClientInternal.start(workflow);
+    return WorkflowClientInternalImpl.start(workflow);
   }
 
   /**
@@ -320,7 +321,7 @@ public interface WorkflowClient {
    * @return WorkflowExecution that contains WorkflowId and RunId of the started workflow.
    */
   static <A1> WorkflowExecution start(Functions.Proc1<A1> workflow, A1 arg1) {
-    return WorkflowClientInternal.start(workflow, arg1);
+    return WorkflowClientInternalImpl.start(workflow, arg1);
   }
 
   /**
@@ -333,7 +334,7 @@ public interface WorkflowClient {
    * @return WorkflowExecution that contains WorkflowId and RunId of the started workflow.
    */
   static <A1, A2> WorkflowExecution start(Functions.Proc2<A1, A2> workflow, A1 arg1, A2 arg2) {
-    return WorkflowClientInternal.start(workflow, arg1, arg2);
+    return WorkflowClientInternalImpl.start(workflow, arg1, arg2);
   }
 
   /**
@@ -348,7 +349,7 @@ public interface WorkflowClient {
    */
   static <A1, A2, A3> WorkflowExecution start(
       Functions.Proc3<A1, A2, A3> workflow, A1 arg1, A2 arg2, A3 arg3) {
-    return WorkflowClientInternal.start(workflow, arg1, arg2, arg3);
+    return WorkflowClientInternalImpl.start(workflow, arg1, arg2, arg3);
   }
 
   /**
@@ -364,7 +365,7 @@ public interface WorkflowClient {
    */
   static <A1, A2, A3, A4> WorkflowExecution start(
       Functions.Proc4<A1, A2, A3, A4> workflow, A1 arg1, A2 arg2, A3 arg3, A4 arg4) {
-    return WorkflowClientInternal.start(workflow, arg1, arg2, arg3, arg4);
+    return WorkflowClientInternalImpl.start(workflow, arg1, arg2, arg3, arg4);
   }
 
   /**
@@ -381,7 +382,7 @@ public interface WorkflowClient {
    */
   static <A1, A2, A3, A4, A5> WorkflowExecution start(
       Functions.Proc5<A1, A2, A3, A4, A5> workflow, A1 arg1, A2 arg2, A3 arg3, A4 arg4, A5 arg5) {
-    return WorkflowClientInternal.start(workflow, arg1, arg2, arg3, arg4, arg5);
+    return WorkflowClientInternalImpl.start(workflow, arg1, arg2, arg3, arg4, arg5);
   }
 
   /**
@@ -405,7 +406,7 @@ public interface WorkflowClient {
       A4 arg4,
       A5 arg5,
       A6 arg6) {
-    return WorkflowClientInternal.start(workflow, arg1, arg2, arg3, arg4, arg5, arg6);
+    return WorkflowClientInternalImpl.start(workflow, arg1, arg2, arg3, arg4, arg5, arg6);
   }
 
   /**
@@ -416,7 +417,7 @@ public interface WorkflowClient {
    * @return WorkflowExecution that contains WorkflowId and RunId of the started workflow.
    */
   static <R> WorkflowExecution start(Functions.Func<R> workflow) {
-    return WorkflowClientInternal.start(workflow);
+    return WorkflowClientInternalImpl.start(workflow);
   }
 
   /**
@@ -428,7 +429,7 @@ public interface WorkflowClient {
    * @return WorkflowExecution that contains WorkflowId and RunId of the started workflow.
    */
   static <A1, R> WorkflowExecution start(Functions.Func1<A1, R> workflow, A1 arg1) {
-    return WorkflowClientInternal.start(workflow, arg1);
+    return WorkflowClientInternalImpl.start(workflow, arg1);
   }
 
   /**
@@ -442,7 +443,7 @@ public interface WorkflowClient {
    */
   static <A1, A2, R> WorkflowExecution start(
       Functions.Func2<A1, A2, R> workflow, A1 arg1, A2 arg2) {
-    return WorkflowClientInternal.start(workflow, arg1, arg2);
+    return WorkflowClientInternalImpl.start(workflow, arg1, arg2);
   }
 
   /**
@@ -457,7 +458,7 @@ public interface WorkflowClient {
    */
   static <A1, A2, A3, R> WorkflowExecution start(
       Functions.Func3<A1, A2, A3, R> workflow, A1 arg1, A2 arg2, A3 arg3) {
-    return WorkflowClientInternal.start(workflow, arg1, arg2, arg3);
+    return WorkflowClientInternalImpl.start(workflow, arg1, arg2, arg3);
   }
 
   /**
@@ -473,7 +474,7 @@ public interface WorkflowClient {
    */
   static <A1, A2, A3, A4, R> WorkflowExecution start(
       Functions.Func4<A1, A2, A3, A4, R> workflow, A1 arg1, A2 arg2, A3 arg3, A4 arg4) {
-    return WorkflowClientInternal.start(workflow, arg1, arg2, arg3, arg4);
+    return WorkflowClientInternalImpl.start(workflow, arg1, arg2, arg3, arg4);
   }
 
   /**
@@ -495,7 +496,7 @@ public interface WorkflowClient {
       A3 arg3,
       A4 arg4,
       A5 arg5) {
-    return WorkflowClientInternal.start(workflow, arg1, arg2, arg3, arg4, arg5);
+    return WorkflowClientInternalImpl.start(workflow, arg1, arg2, arg3, arg4, arg5);
   }
 
   /**
@@ -519,7 +520,7 @@ public interface WorkflowClient {
       A4 arg4,
       A5 arg5,
       A6 arg6) {
-    return WorkflowClientInternal.start(workflow, arg1, arg2, arg3, arg4, arg5, arg6);
+    return WorkflowClientInternalImpl.start(workflow, arg1, arg2, arg3, arg4, arg5, arg6);
   }
 
   /**
@@ -530,7 +531,7 @@ public interface WorkflowClient {
    * @return future becomes ready upon workflow completion with null value or failure
    */
   static CompletableFuture<Void> execute(Proc workflow) {
-    return WorkflowClientInternal.execute(workflow);
+    return WorkflowClientInternalImpl.execute(workflow);
   }
 
   /**
@@ -542,7 +543,7 @@ public interface WorkflowClient {
    * @return future becomes ready upon workflow completion with null value or failure
    */
   static <A1> CompletableFuture<Void> execute(Proc1<A1> workflow, A1 arg1) {
-    return WorkflowClientInternal.execute(workflow, arg1);
+    return WorkflowClientInternalImpl.execute(workflow, arg1);
   }
 
   /**
@@ -555,7 +556,7 @@ public interface WorkflowClient {
    * @return future becomes ready upon workflow completion with null value or failure
    */
   static <A1, A2> CompletableFuture<Void> execute(Proc2<A1, A2> workflow, A1 arg1, A2 arg2) {
-    return WorkflowClientInternal.execute(workflow, arg1, arg2);
+    return WorkflowClientInternalImpl.execute(workflow, arg1, arg2);
   }
 
   /**
@@ -570,7 +571,7 @@ public interface WorkflowClient {
    */
   static <A1, A2, A3> CompletableFuture<Void> execute(
       Proc3<A1, A2, A3> workflow, A1 arg1, A2 arg2, A3 arg3) {
-    return WorkflowClientInternal.execute(workflow, arg1, arg2, arg3);
+    return WorkflowClientInternalImpl.execute(workflow, arg1, arg2, arg3);
   }
 
   /**
@@ -586,7 +587,7 @@ public interface WorkflowClient {
    */
   static <A1, A2, A3, A4> CompletableFuture<Void> execute(
       Proc4<A1, A2, A3, A4> workflow, A1 arg1, A2 arg2, A3 arg3, A4 arg4) {
-    return WorkflowClientInternal.execute(workflow, arg1, arg2, arg3, arg4);
+    return WorkflowClientInternalImpl.execute(workflow, arg1, arg2, arg3, arg4);
   }
 
   /**
@@ -603,7 +604,7 @@ public interface WorkflowClient {
    */
   static <A1, A2, A3, A4, A5> CompletableFuture<Void> execute(
       Proc5<A1, A2, A3, A4, A5> workflow, A1 arg1, A2 arg2, A3 arg3, A4 arg4, A5 arg5) {
-    return WorkflowClientInternal.execute(workflow, arg1, arg2, arg3, arg4, arg5);
+    return WorkflowClientInternalImpl.execute(workflow, arg1, arg2, arg3, arg4, arg5);
   }
 
   /**
@@ -627,7 +628,7 @@ public interface WorkflowClient {
       A4 arg4,
       A5 arg5,
       A6 arg6) {
-    return WorkflowClientInternal.execute(workflow, arg1, arg2, arg3, arg4, arg5, arg6);
+    return WorkflowClientInternalImpl.execute(workflow, arg1, arg2, arg3, arg4, arg5, arg6);
   }
 
   /**
@@ -638,7 +639,7 @@ public interface WorkflowClient {
    * @return future that contains workflow result or failure
    */
   static <R> CompletableFuture<R> execute(Func<R> workflow) {
-    return WorkflowClientInternal.execute(workflow);
+    return WorkflowClientInternalImpl.execute(workflow);
   }
 
   /**
@@ -650,7 +651,7 @@ public interface WorkflowClient {
    * @return future that contains workflow result or failure
    */
   static <A1, R> CompletableFuture<R> execute(Func1<A1, R> workflow, A1 arg1) {
-    return WorkflowClientInternal.execute(workflow, arg1);
+    return WorkflowClientInternalImpl.execute(workflow, arg1);
   }
 
   /**
@@ -664,7 +665,7 @@ public interface WorkflowClient {
    */
   static <A1, A2, R> CompletableFuture<R> execute(
       Functions.Func2<A1, A2, R> workflow, A1 arg1, A2 arg2) {
-    return WorkflowClientInternal.execute(workflow, arg1, arg2);
+    return WorkflowClientInternalImpl.execute(workflow, arg1, arg2);
   }
 
   /**
@@ -679,7 +680,7 @@ public interface WorkflowClient {
    */
   static <A1, A2, A3, R> CompletableFuture<R> execute(
       Functions.Func3<A1, A2, A3, R> workflow, A1 arg1, A2 arg2, A3 arg3) {
-    return WorkflowClientInternal.execute(workflow, arg1, arg2, arg3);
+    return WorkflowClientInternalImpl.execute(workflow, arg1, arg2, arg3);
   }
 
   /**
@@ -695,7 +696,7 @@ public interface WorkflowClient {
    */
   static <A1, A2, A3, A4, R> CompletableFuture<R> execute(
       Functions.Func4<A1, A2, A3, A4, R> workflow, A1 arg1, A2 arg2, A3 arg3, A4 arg4) {
-    return WorkflowClientInternal.execute(workflow, arg1, arg2, arg3, arg4);
+    return WorkflowClientInternalImpl.execute(workflow, arg1, arg2, arg3, arg4);
   }
 
   /**
@@ -717,7 +718,7 @@ public interface WorkflowClient {
       A3 arg3,
       A4 arg4,
       A5 arg5) {
-    return WorkflowClientInternal.execute(workflow, arg1, arg2, arg3, arg4, arg5);
+    return WorkflowClientInternalImpl.execute(workflow, arg1, arg2, arg3, arg4, arg5);
   }
 
   /**
@@ -741,6 +742,13 @@ public interface WorkflowClient {
       A4 arg4,
       A5 arg5,
       A6 arg6) {
-    return WorkflowClientInternal.execute(workflow, arg1, arg2, arg3, arg4, arg5, arg6);
+    return WorkflowClientInternalImpl.execute(workflow, arg1, arg2, arg3, arg4, arg5, arg6);
   }
+
+  /**
+   * For SDK Internal usage only. This method should <b>not</b> be used by users. If implementing a
+   * proxy or an adapter over a {@link WorkflowClient} provided by the SDK, users should pass an
+   * object returned by this method as-is.
+   */
+  Object getInternal();
 }
--- a/temporal-sdk/src/main/java/io/temporal/client/WorkflowClientInternalImpl.java
+++ b/temporal-sdk/src/main/java/io/temporal/client/WorkflowClientInternalImpl.java
@@ -36,12 +36,15 @@ import io.temporal.common.interceptors.WorkflowClientCallsInterceptor;
 import io.temporal.common.interceptors.WorkflowClientInterceptor;
 import io.temporal.internal.WorkflowThreadMarker;
 import io.temporal.internal.client.RootWorkflowClientInvoker;
+import io.temporal.internal.client.WorkerFactoryRegistry;
+import io.temporal.internal.client.WorkflowClientInternal;
 import io.temporal.internal.client.external.GenericWorkflowClient;
 import io.temporal.internal.client.external.GenericWorkflowClientImpl;
 import io.temporal.internal.client.external.ManualActivityCompletionClientFactory;
 import io.temporal.internal.sync.StubMarker;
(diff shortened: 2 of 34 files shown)


temporal-sdk/src/main/java/io/temporal/worker/WorkflowTaskDispatchHandle.java
```diff
@@ -0,0 +1,57 @@
...
+
+import io.temporal.api.workflowservice.v1.PollWorkflowTaskQueueResponse;
+import io.temporal.internal.worker.WorkflowTask;
+import java.io.Closeable;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Semaphore;
+import java.util.function.Consumer;
+
+public class WorkflowTaskDispatchHandle implements Closeable {
+  private final CompletableFuture<Void> completed = new CompletableFuture<>();
```
cretz, line 31: Took me a sec to understand what was happening here. I think an `AtomicBoolean` would be clearer

temporal-sdk/src/main/java/io/temporal/worker/WorkflowTaskDispatchHandle.java
```diff
@@ -0,0 +1,57 @@
...
+      Consumer<WorkflowTask> dispatchCallback, Semaphore executorSlotsSemaphore) {
+    this.dispatchCallback = dispatchCallback;
+    this.executorSlotsSemaphore = executorSlotsSemaphore;
+  }
+
+  /**
+   * @param workflowTask to be fed directly into the workflow worker
+   * @throws IllegalArgumentException if the workflow task doesn't belong to this worker
+   */
+  public void dispatch(PollWorkflowTaskQueueResponse workflowTask) {
```
cretz, line 45: Would be curious to know what situations dispatch is not called. Arguably (and maybe this is in favor of the no-interceptor argument), the building of this handle and the invocation of dispatch are so close together that every handle is expected to call dispatch and this doesn't have to be closeable.

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
```diff
@@ -104,8 +109,6 @@ public boolean start() {
               options.getTaskExecutorThreadPoolSize(),
               workerMetricsScope,
               true);
-      Semaphore workflowTaskExecutorSemaphore =
-          new Semaphore(options.getTaskExecutorThreadPoolSize());
```
cretz, line 108: Can you confirm that `start` is never called more than once on this class? (i.e. no restarting that may cause issues with reusing the semaphore)

temporal-sdk/src/main/java/io/temporal/internal/client/WorkflowClientInternal.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * From OOP point of view, there is no reason for this interface to don't extend {@link
+ * WorkflowClient}. It's not extending only to make sure that the calls that can come through a
+ * user-supplied WorkflowClient are made through that entity whenever possible instead of the entity
+ * obtained by {@link WorkflowClient#getInternal()}. This will make sure DI/AOP frameworks or user's
+ * wrapping code implemented in WorkflowClient proxy or adapter gets executed when possible and
+ * {@link WorkflowClient#getInternal()} is used only for internal functionality.
+ */
+public interface WorkflowClientInternal {
```
cretz, line 34: If you want, I'd support just removing this interface and casting directly to the internal class impl. But I understand if you feel better about the abstraction.

temporal-sdk/src/main/java/io/temporal/client/WorkflowOptions.java
```diff
@@ -281,6 +285,23 @@ public Builder setContextPropagators(@Nullable List<ContextPropagator> contextPr
...
+     *   <li>started
+     *   <li>has a non paused worker on the right task queue
+     * </ul>
+     *
+     * TODO
+     *
+     * @param disableEagerExecution if true, an eager local execution of the workflow task will
+     *     never be requested even if it is possible.
+     */
+    public void setDisableEagerExecution(boolean disableEagerExecution) {
```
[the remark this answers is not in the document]
Spikhalskiy: We already have `disableEagerExecution` there. And yeah, with this name... So I think I will just respect this flag for Workflow Tasks too. It looks quite reasonable.
The only problem is that it's a bit backward from implementation perspective, because this flag is on Worker, but used by a Client. But nevertheless I think WorkerOptions is the right place for it from a user intuition perspective.

temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkflowClientCallsInterceptor.java
```diff
@@ -84,18 +86,31 @@ final class WorkflowStartInput {
...
-        String workflowType,
-        Header header,
-        Object[] arguments,
-        WorkflowOptions options) {
+        @Nonnull String workflowId,
+        @Nonnull String workflowType,
+        @Nonnull Header header,
+        @Nonnull Object[] arguments,
+        @Nonnull WorkflowOptions options,
+        @Nullable Boolean requestEagerExecution) {
```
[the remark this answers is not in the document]
Spikhalskiy: I was debating between two approaches. The approach that you are talking about is actually much EASIER to implement. But after a bit of thinking, I decided on the current approach.
Several main motives for it:
1. User interceptors should KNOW about the SDK decision if this workflow task is requested eagerly. I think it's important for logging, metrics, whatever users may want to implement in their interceptors. That's why I put it as a first interceptor in the chain.
2. I think users should be able to override the decision into "false" depending on the custom logic that they may want to have.

The current approach provides both of ^ without significant shortcomings.

temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkflowClientCallsInterceptor.java
```diff
@@ -117,18 +132,33 @@ public Object[] getArguments() {
...
 
   final class WorkflowStartOutput {
-    private final WorkflowExecution workflowExecution;
+    private final @Nonnull WorkflowExecution workflowExecution;
+    private final @Nullable PollWorkflowTaskQueueResponse eagerlyDispatchedWorkflowTask;
 
-    public WorkflowStartOutput(WorkflowExecution workflowExecution) {
+    public WorkflowStartOutput(
+        @Nonnull WorkflowExecution workflowExecution,
+        @Nullable PollWorkflowTaskQueueResponse eagerlyDispatchedWorkflowTask) {
```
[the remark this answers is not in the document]
Spikhalskiy: It's done this way not to make it "mutable", but to "publish" the result to the interceptors and pass it back to the control code.
I also think it should be, totally see users that may want to get a custom metric of something like "eagerly dispatched / eagerly requested" ratio or use it for logging or reporting.

temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkflowClientCallsInterceptor.java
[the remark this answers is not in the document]
cretz: > User interceptors should KNOW about the SDK decision if this workflow task is requested eagerly.

I'm not convinced this is true. And if it is true, can we provide it on some task-level info available to workflow interceptor instead?

> I think users should be able to override the decision into "false" depending on the custom logic that they may want to have.

I think they should be able to do that on the way down the chain, but not after already accepted. And they already can by mutating the workflow options and setting disable eager as true.

temporal-sdk/src/main/java/io/temporal/client/WorkflowClientInternalImpl.java
```diff
@@ -102,6 +109,21 @@ public static WorkflowClient newInstance(
...
 
+  private List<WorkflowClientInterceptor> initializeInterceptors() {
+    List<WorkflowClientInterceptor> interceptors =
+        new ArrayList<>(options.getInterceptors().length + 1);
+    interceptors.add(
+        new WorkflowClientInterceptorBase() {
+          @Override
+          public WorkflowClientCallsInterceptor workflowClientCallsInterceptor(
+              WorkflowClientCallsInterceptor next) {
+            return new LocalDispatchWorkflowClientInterceptor(next, workerFactories);
```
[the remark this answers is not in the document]
Spikhalskiy: Can be and it's much easier to do. Originally I did it this way, but after that I realized it should be better as an interceptor:
1. It allows user to implement their custom disabling logic
2. It provides an access for any kind of reporting interceptor user may want to implement in the interceptors.
3. can be an interceptor which helps code modularity. This keeps implementation highly cohesive and better structured.
4. Should be an interceptor, because there is no significant shortfalls while we are getting ^

temporal-sdk/src/main/java/io/temporal/client/WorkflowClientInternalImpl.java
```diff
@@ -505,4 +527,19 @@ Stream<HistoryEvent> streamHistory(WorkflowExecution execution) {
...
+  }
+
+  @Override
+  public void registerWorkerFactory(WorkerFactory workerFactory) {
+    workerFactories.add(workerFactory);
+  }
+
+  @Override
+  public void deregisterWorkerFactory(WorkerFactory workerFactory) {
+    workerFactories.remove(workerFactory);
```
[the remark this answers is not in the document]
Spikhalskiy: An order doesn't really matter. For most users, this collection should have 1 element most of the time. There shouldn't be more that one active WorkferFactory for a single namespace. The only exception is when they "restart" a WorkerFactory. 

> when looking up by task queue
Lookup by task queue is INSIDE the WorkerFactory and it's a map there.

> confirming list has value over set
There should be no difference here really. I expect in normal scenarios for this collection to have only one element always. And if not, the access pattern is iteration. Set may be preferred over list to optimize insertion/deletion, but it shouldn't matter, as these operations SHOULD be very rare.

temporal-sdk/src/main/java/io/temporal/worker/WorkflowTaskDispatchHandle.java
Spikhalskiy: Sure, I don't see a big difference with a uncompleted/completed CompletableFuture or an AtomicBoolean for this usage.

temporal-sdk/src/main/java/io/temporal/worker/WorkflowTaskDispatchHandle.java
Spikhalskiy: It's a normal scenario where it's not called:
1. We obtain the handler
2. Server doesn't make an eager dispatch
3. We call `close` and don't call `dispatch`.

temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkflowClientCallsInterceptor.java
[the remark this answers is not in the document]
cretz: Still unsure that knowing about eager dispatch occurring in whatever form should allow them to intercept the task. I also think that this large gap in code path from when the slot is reserved to when it is freed is a bit dangerous.

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
Spikhalskiy: Yeah, it's protected by an AtomicBoolean in Worker. Start on the internal workers is always called once.

temporal-sdk/src/main/java/io/temporal/internal/client/WorkflowClientInternal.java
Spikhalskiy: I envision that some very advanced and invasive user may want to intercept getInternal to create a wrapper for logging/tracing purposes. It's not something we may ever recommend, but some users may want it anyway. It will be cleaner to have an interface for them.
But I don't have a strong opinion on this one and can replace it with a class usage, I see a value in actually PREVENTING the above scenario in doing this too.

temporal-sdk/src/main/java/io/temporal/worker/WorkflowTaskDispatchHandle.java
cretz: Ah, duh, ok. I wonder if this is clearer:

```java
public void resolve(@Nullable PollWorkflowTaskQueueResponse workflowTask) {
```

Just a bit concerned that there isn't a single clear call to this unfortunately-user-facing class. But this does work as is today.

temporal-sdk/src/main/java/io/temporal/client/WorkflowClient.java
```diff
@@ -741,6 +742,13 @@ static <A1, A2, A3, A4, A5, A6, R> CompletableFuture<R> execute(
       A4 arg4,
       A5 arg5,
       A6 arg6) {
-    return WorkflowClientInternal.execute(workflow, arg1, arg2, arg3, arg4, arg5, arg6);
+    return WorkflowClientInternalImpl.execute(workflow, arg1, arg2, arg3, arg4, arg5, arg6);
   }
+
+  /**
+   * For SDK Internal usage only. This method should be used by users. If implementing a proxy or an
```
Sushisource, line 749:
```suggestion
   * For SDK Internal usage only. This method should not be used by users. If implementing a proxy or an
```

temporal-sdk/src/main/java/io/temporal/client/WorkflowOptions.java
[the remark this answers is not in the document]
Sushisource: I'm not sure I get the scenario where a client would want to set this flag?

temporal-sdk/src/main/java/io/temporal/internal/client/WorkflowClientInternal.java
```diff
@@ -0,0 +1,38 @@
...
+ * limitations under the License.
+ */
+
+package io.temporal.internal.client;
+
+import io.temporal.client.WorkflowClient;
+import io.temporal.worker.WorkerFactory;
+
+/**
+ * From OOP point of view, there is no reason for this interface to don't extend {@link
```
Sushisource, line 27:
```suggestion
 * From OOP point of view, there is no reason for this interface not to extend {@link
```

temporal-sdk/src/main/java/io/temporal/internal/client/WorkflowClientInternal.java
Sushisource: I think I'd just prevent that for now and if anyone ever makes a fuss consider it then. Seems like a really edge situation to me.

temporal-sdk/src/main/java/io/temporal/worker/WorkflowTaskDispatchHandle.java
```diff
@@ -0,0 +1,57 @@
...
+
+  public WorkflowTaskDispatchHandle(
+      Consumer<WorkflowTask> dispatchCallback, Semaphore executorSlotsSemaphore) {
+    this.dispatchCallback = dispatchCallback;
+    this.executorSlotsSemaphore = executorSlotsSemaphore;
+  }
+
+  /**
+   * @param workflowTask to be fed directly into the workflow worker
+   * @throws IllegalArgumentException if the workflow task doesn't belong to this worker
```
Sushisource, line 43: It'd be worth clarifying what "doesn't belong" means in this case

Sushisource (approved):
This looks pretty clean to me, just some minor stuff.

I don't really have an opinion on the interceptor or not part. I can see both sides, but I do think the interceptor makes for a nice clean implementation.

temporal-sdk/src/main/java/io/temporal/client/WorkflowClient.java
bergundy: I'd even suggest putting the `not` in bold.

temporal-sdk/src/main/java/io/temporal/worker/WorkflowTaskDispatchHandle.java
Spikhalskiy: I can move it into an internal package to show that users typically shouldn't be working with it.

> Just a bit concerned that there isn't a single clear call to this unfortunately-user-facing class. 
I'm not sure. I like the current structure. We call dispatch only if there is something to dispatch. We call close always. It makes the class safe to use and the eager path is distinct, which is good for readability.

When Local Completion is implemented and this method will start to return a LocalCompleteion instance, it will become even more clear that it's good to have a separate eager-path.

temporal-sdk/src/main/java/io/temporal/client/WorkflowOptions.java
```diff
@@ -281,6 +285,23 @@ public Builder setContextPropagators(@Nullable List<ContextPropagator> contextPr
...
 
+    /**
+     * If {@link WorkflowClient} is used to create a {@link WorkerFactory} that is
+     *
+     * <ul>
+     *   <li>started
+     *   <li>has a non paused worker on the right task queue
+     * </ul>
+     *
+     * TODO
```
bergundy, line 296: Is the docstring TODO here?

temporal-sdk/src/main/java/io/temporal/client/WorkflowOptions.java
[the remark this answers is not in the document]
cretz: I think it's the general safety of being able to turn off this new unbalancing feature. Granted I think it may need to be disable-able at a higher level, but I could _technically_ see doing it for cert workflows.

One use case I can see is that we've been telling people they can use poll success/timeout metrics to determine they have too many workers and can use that to scale down. Well, that's gonna completely change now for people with colocated clients and workers. Especially in cases where the client calls are not evenly distributed, workers without those client calls will all the sudden start hitting poll timeouts more because work distribution isn't as uniform as it once was (not that it was that uniform/fair in the first place though).

temporal-sdk/src/main/java/io/temporal/client/WorkflowOptions.java
```diff
@@ -411,6 +438,10 @@ public Map<String, Object> getMemo() {
     return contextPropagators;
   }
 
+  public boolean isDisableEagerExecution() {
```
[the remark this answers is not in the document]
bergundy: +1, this reads better to me.

temporal-sdk/src/main/java/io/temporal/client/WorkflowOptions.java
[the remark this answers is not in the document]
bergundy: I can see a case where a worker can process latency sensitive and insensitive workflows and you'd want to disable eager workflows for the latency insensitive ones.

temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkflowClientCallsInterceptor.java
```diff
@@ -84,18 +86,31 @@ final class WorkflowStartInput {
     private final Header header;
     private final Object[] arguments;
     private final WorkflowOptions options;
+    private final @Nullable Boolean requestEagerExecution;
 
+    /**
+     * @param workflowId workflowId of the workflow to be started
```
bergundy, line 92:
```suggestion
     * @param workflowId ID of the workflow to be started
```

temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkflowClientCallsInterceptor.java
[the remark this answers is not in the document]
bergundy: Seems like this will change after our team standup discussion, waiting to see the modifications.

temporal-sdk/src/main/java/io/temporal/client/WorkflowOptions.java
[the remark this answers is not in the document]
Spikhalskiy: It matches the ActivityOptions, yeah. Users never will use getter too, so shouldn't be a big deal

temporal-sdk/src/main/java/io/temporal/internal/client/WorkflowClientInternal.java
Spikhalskiy: Nah, after second consideration, this solution would require opening WorkflowClientInternal implementation to `public`. I think it's cleaner to open only an internal interface than the whole implementation class.

temporal-sdk/src/main/java/io/temporal/internal/client/IteratorWithOffset.java
```diff
@@ -0,0 +1,76 @@
...
+    // startingIndex theoretically can be out of bound at the moment we got here already, so
+    // we need to write implementation that tolerates that
+    try {
+      headIterator = list.subList(0, startingIndex).iterator();
+    } catch (IndexOutOfBoundsException e) {
+      headIterator = list.iterator();
+    }
+    // Initializations of headIterator and iterator are not atomic, safe.
+    // It can lead to an item to be skipped or subbed in some rare edge case
+    // if the collection is modified in just the right time. Considering it not a big deal.
```
cretz, line 42: Agreed not a big deal, but if you can trust the list has random access (which it seems like you can), just store the list and the index as fields and increment/wrap manually instead of two iterators. Can throw no such element when the index reaches start a second time (keep bool if needed to track that). I think it's easier to understand code.

Granted, I'd be tempted at the known cost of performance for every start workflow, to just copy to array list and shuffle so you can have real random. With this implementation, if two worker factories that can accept the same workflow are next to each other in a list of 20, you'll have to land right on the starting index of the second one or the first one will be prioritized 19 times out of 20 (assuming uniform random).

temporal-sdk/src/main/java/io/temporal/internal/client/EagerWorkflowTaskDispatcher.java
```diff
@@ -0,0 +1,50 @@
...
+
+  public EagerWorkflowTaskDispatcher(WorkerFactoryRegistry workerFactories) {
+    this.workerFactories = workerFactories;
+  }
+
+  @Nullable
+  public WorkflowTaskDispatchHandle tryGetLocalDispatchHandler(
+      WorkflowClientCallsInterceptor.WorkflowStartInput workflowStartInput) {
+    for (WorkerFactory workerFactory : workerFactories.workerFactoriesWithRandomOffset()) {
+      Worker worker = workerFactory.tryGetWorker(workflowStartInput.getOptions().getTaskQueue());
```
cretz, line 40: Nit: Consider moving this to a `tryGetWorker` on the registry itself or maybe even `reserveWorkflowExecutor(String taskQueue)`. No real reason except that you can maybe get rid of this class and encapsulate worker selection where worker factories are. Feel free to ignore.

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
```diff
@@ -27,28 +27,52 @@
...
+    this.eagerWorkflowTaskDispatcher = new EagerWorkflowTaskDispatcher(workerFactoryRegistry);
     this.requestsHelper = new WorkflowClientRequestFactory(clientOptions);
   }
 
   @Override
   public WorkflowStartOutput start(WorkflowStartInput input) {
-    StartWorkflowExecutionRequest request =
-        requestsHelper.newStartWorkflowExecutionRequest(input).build();
-    return new WorkflowStartOutput(genericClient.start(request));
+    try (WorkflowTaskDispatchHandle eagerDispatchHandle = obtainDispatchHandle(input)) {
```
cretz, line 55:
```suggestion
    // Handle is null if user disabled via options or no worker slot available
    try (WorkflowTaskDispatchHandle eagerDispatchHandle = obtainDispatchHandle(input)) {
```

Was a bit unclear to me before I started digging into other code more

temporal-sdk/src/main/java/io/temporal/worker/WorkflowTaskDispatchHandle.java
```diff
@@ -0,0 +1,72 @@
...
+  /**
+   * @param workflowTask to be fed directly into the workflow worker
+   * @return true is the workflow task was successfully dispatched
+   * @throws IllegalArgumentException if the workflow task doesn't belong to the task queue of the
+   *     worker provided this {@link WorkflowTaskDispatchHandle}
+   */
+  public boolean dispatch(@Nonnull PollWorkflowTaskQueueResponse workflowTask) {
+    Preconditions.checkNotNull(workflowTask, "workflowTask");
+    if (completed.compareAndSet(false, true)) {
+      return dispatchCallback.apply(
```
cretz, line 59: How much is done in this blocking call? While I know there is future talk of returning single-task results from start, I just want to check that a workflow task completed response isn't required to be sent to the server before a user is told a workflow is started. (or is that what we want? even with local activities of WFT failure?)

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
```diff
@@ -27,28 +27,52 @@
...
+              .setRunId(response.getRunId())
+              .setWorkflowId(request.getWorkflowId())
+              .build();
+      @Nullable
+      PollWorkflowTaskQueueResponse eagerWorkflowTask =
+          requestEagerExecution && response.hasEagerWorkflowTask()
+              ? response.getEagerWorkflowTask()
+              : null;
+      if (eagerWorkflowTask != null) {
+        eagerDispatchHandle.dispatch(eagerWorkflowTask);
```
cretz, line 72: Is there a reason to notify the user if this returns false?

cretz (commented):
LGTM. Added couple small things, but nothing major. But will wait for responses before marking approved.

temporal-sdk/src/main/java/io/temporal/worker/WorkflowTaskDispatchHandle.java
Spikhalskiy: Almost none, it puts task in the executor that is guaranteed to have a slot because we reserved it. But I will add a javadoc for future reader that it may be important to keep this call non-blocking.

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
Spikhalskiy: No. I made the flag mostly as a "placeholder" for future LocalCompletionHandle / LocalCompletionFuture

temporal-sdk/src/main/java/io/temporal/internal/client/EagerWorkflowTaskDispatcher.java
Spikhalskiy: I don't want WorkerFactoryRegistry to know about WorkflowTaskDispatchHandle, I prefer to keep two cleaner classes with separate concerns.

temporal-sdk/src/main/java/io/temporal/internal/client/IteratorWithOffset.java
Spikhalskiy: > Agreed not a big deal, but if you can trust the list has random access (which it seems like you can), just store the list and the index as fields and increment/wrap manually instead of two iterators.

Not much simpler, because we will have to handle all the out-of-bound exceptions, handle overflow and implement hasNext ourselves. I don't think it actually will give a simpler implementation.

> With this implementation, if two worker factories that can accept the same workflow are next to each other in a list of 20, you'll have to land right on the starting index of the second one or the first one will be prioritized 19 times out of 20 (assuming uniform random).

Honestly, in vast majority of cases, the first selected factory will be the last. Because we deregister factories when we shut them down. So pretty much the only way the described scenario is possible is when there are many WorkerFactories and most of them are paused or out of slots. And the "good" factories need to be next to each other. Very-very edgy.

But maybe I will just implement it with a array-copy and Collections.shuffle honestly for simplicity of reading and maintaining, because rare users should have more than one WorkflowFactory on one client anyway.

cretz (approved):

temporal-sdk/src/main/java/io/temporal/internal/client/IteratorWithOffset.java
Spikhalskiy: We do need to randomize. Otherwise imagine a user that currently stats two WorkerFactories on the same client. For some strange reason, but we totally allow it. only one of these workers will get tasks from the client without randomization. So I think the fact of randomization is important, but for this situation only pretty much.

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
```diff
@@ -139,11 +147,15 @@ public boolean start() {
...
+    // TODO wait for returning all slots to semaphore to make sure all eager dispatch handlers are
+    // respected
   }
 
   @Override
   public void awaitTermination(long timeout, TimeUnit unit) {
     ShutdownManager.awaitTermination(poller, unit.toMillis(timeout));
+    // TODO wait for returning all slots to semaphore to make sure all eager dispatch handlers are
+    // respected
   }
```
bergundy, line 159: Are these going to be addressed in the next PR or are you going to open issues?

Also, pardon my ignorance, but why do we need to wait both in awaitTermination and shutdown?

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
```diff
@@ -182,12 +194,42 @@ private PollerOptions getPollerOptions(SingleWorkerOptions options) {
...
+    return executor != null && !isSuspended() && wftExecutorSlotsSemaphore.tryAcquire()
+        ? new WorkflowTaskDispatchHandle(
+            workflowTask -> {
+              String queueName =
+                  workflowTask.getResponse().getWorkflowExecutionTaskQueue().getName();
+              TaskQueueKind queueKind =
+                  workflowTask.getResponse().getWorkflowExecutionTaskQueue().getKind();
+              Preconditions.checkArgument(
+                  this.taskQueueName.equals(queueName)
+                      || TaskQueueKind.TASK_QUEUE_KIND_STICKY.equals(queueKind)
```
bergundy, line 216: Right now there's no way you'll get back an eager task on a sticky queue but I agree it doesn't hurt to accept it.

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
Spikhalskiy: > Are these going to be addressed in the next PR or are you going to open issues?

I will do it in a separate PR. It's a larger scope than it seems because it will need to change/fix the shutdown implementation of a Poller. 

> Also, pardon my ignorance, but why do we need to wait both in awaitTermination and shutdown?

`shutdown` will wait for semaphore to perform the last stage of shutdown after the mentioned PR and change.
`awaitTermination` will wait for that last state to be executed.

PR #1691, opened 2023-03-07
Fix incorrect triggering of eventLoop in case of activity cancelation before all events from WFT are applied

### LINKED ISSUE #1558: Issue with ReplayAwareLogger and CanceledFailure
When catching ActivityFailure with cause io.temporal.failure.CanceledFailure it seems ReplayAwareLogger thinks workflow is in replay mode, looks like issue with setting WorkflowUnsafe.isReplaying() in this case. 

Full repro (could run in within java samples) [here](https://gist.github.com/tsurdilo/4eb48c5f87fe8797b4a7a4399375ee85). 
Relevant lines where issue happens (only system outs get printed) [here](https://gist.github.com/tsurdilo/4eb48c5f87fe8797b4a7a4399375ee85#file-gistfile1-txt-L77-L82).

Spikhalskiy:
# What
Fix incorrect triggering of eventLoop in case of activity cancelation before all events from WFT are applied

# Why
May close #1558

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


dano:
I can confirm that these changes fix the issue I was seeing. FWIW, here is the test I've been using. It fails with 1.18.2, but passes with these changes.

``` java 
  @Test
  public void doIt() throws InterruptedException {
    WorkflowServiceStubs service = WorkflowServiceStubs.newServiceStubs(
                    .validateAndBuildWithDefaults()
    );
    WorkflowClient client = WorkflowClient.newInstance(service);
    new Thread(() -> {
      var factory = WorkerFactory.newInstance(client, WorkerFactoryOptions.newBuilder()
              .validateAndBuildWithDefaults()
      );
      var worker = factory.newWorker(TASK_QUEUE);
      worker.registerWorkflowImplementationTypes(CancelWfImpl.class);
      worker.registerActivitiesImplementations(new MyActivityImpl());
      factory.start();
    }).start();
    sleep(1000L);
    var wf = client.newWorkflowStub(CancelWf.class, WorkflowOptions.newBuilder()
            .setTaskQueue(TASK_QUEUE)
            .build());
    var stub = WorkflowStub.fromTyped(wf);
    stub.start();
    MyActivityImpl.l.await();
    stub.cancel();
    try {
      stub.getResult(Void.class);
    } catch (WorkflowFailedException e) {
      assertFalse(CancelWfImpl.isReplaying);
    }
  }

  @WorkflowInterface
  public interface CancelWf {
    @WorkflowMethod
    void execute();
  }

  public static class CancelWfImpl implements CancelWf {
    public static Boolean isReplaying = null;

    @Override
    public void execute() {
      var act = Workflow.newActivityStub(MyActivity.class, ActivityOptions.newBuilder()
          .setCancellationType(ActivityCancellationType.WAIT_CANCELLATION_COMPLETED)
          .setHeartbeatTimeout(Duration.ofSeconds(2))
          .setStartToCloseTimeout(Duration.ofSeconds(10))
          .validateAndBuildWithDefaults());
      try {
        act.execute();
      } catch (Exception e) {
        isReplaying = WorkflowUnsafe.isReplaying();
      }
    }
  }

  @ActivityInterface
  public interface MyActivity {
    @ActivityMethod
    void execute();
  }

  public static class MyActivityImpl implements MyActivity {
    public static CountDownLatch l = new CountDownLatch(1);

    public void execute() {
      l.countDown();
      var ctx = Activity.getExecutionContext();
      while (true) {
        ctx.heartbeat("");
        try {
        Thread.sleep(200L);
          throw new RuntimeException(e);
        }
      }
    }
  }
```

temporal-sdk/src/main/java/io/temporal/internal/statemachines/ActivityStateMachine.java
```diff
@@ -288,20 +289,57 @@ private void setStartedCommandEventId() {
...
+    }
+  }
+
+  /**
+   * Workflow code doesn't need to wait for the cancellation event if {@link
+   * ActivityCancellationType#TRY_CANCEL}, immediately notify the workflow code.
+   */
+  private void notifyCanceledIfTryCancelImmediately() {
+    if (cancellationType == ActivityCancellationType.TRY_CANCEL) {
+      notifyCanceled(false);
```
Sushisource, line 312: The abandon handling looks a little odd to me here and elsewhere.

Abandoned activities should still have the workflow code get notified that they were cancelled - or is the idea that this has already happened at a higher level?

If so, probably makes more sense to just deal with not calling the callback inside `notifyCanceled` itself

Sushisource (approved):
This makes sense to me. Just the one thing.

temporal-sdk/src/main/java/io/temporal/internal/statemachines/ActivityStateMachine.java
Spikhalskiy: Yes, Abandon is handled even earlier, on the entry point of requesting cancelation: https://github.com/temporalio/sdk-java/blob/7f2679c4ac53b19f6dfe54913cc13ad0dec84179/temporal-sdk/src/main/java/io/temporal/internal/statemachines/ActivityStateMachine.java#L280

Spikhalskiy: But it's actually an argument for simplifying the code and not even checking for `ABANDON` here.

Spikhalskiy: I add this TODO here

```
  private void cancelCommandNotifyCanceledImmediately() {
    cancelCommand();
    // TODO With {@link ActivityCancellationType#ABANDON} we shouldn't even get here as it gets
    //  handled in #cancel.
    //  It's a code path for TRY_CANCEL and WAIT_CANCELLATION_COMPLETED only.
    //  Was the original design to cancel a not-yet-sent
    //  {@link CommandType#COMMAND_TYPE_SCHEDULE_ACTIVITY_TASK} in case of ABANDON too?
    if (cancellationType != ActivityCancellationType.ABANDON) {
      notifyCanceled(false);
    }
  }
```

and not gonna touch it in this PR. It looks like this if for ABANDON is not needed, we are not getting here with ABANDON. But I feel like by cleaning it out I may lose some original context here.

PR #1695, opened 2023-03-09
Provide SerializationContext for PayloadConverter and PayloadCodec

### LINKED ISSUE #1694: Make Serialization Context available for PayloadCodecs and PayloadConverters
Some users need to have an access to some basic info like workflow id and activity id and task queue names in their `PayloadConverter` or `PayloadCodec`.
One of the typical scenarios is implementing a custom data conversion for a specific task queue or incorporating a `workflowId` into an encrypting codec as salt to sign payloads to be used in this workflow execution only.

Spikhalskiy:
## What was changed

Provides `SerializationContext` for `PayloadConverter` and `PayloadCodec` and `ActivityCompletionClient`.
This allows users to use workflowId as a parameter and an input to their serialization and encoding. Which enables the usage of workflowId for signing payloads to avoid replay attacks.

## Why?

Closes #1694


temporal-sdk/src/main/java/io/temporal/payload/context/SerializationContext.java
```diff
@@ -0,0 +1,63 @@
...
+
+import io.temporal.api.common.v1.Payload;
+import io.temporal.client.WorkflowClient;
+import io.temporal.common.Experimental;
+import io.temporal.common.converter.DataConverter;
+import io.temporal.common.converter.PayloadConverter;
+import io.temporal.payload.codec.PayloadCodec;
+
+/**
+ * Temporal SDK may provide one of subclasses of {@link SerializationContext} during data
```
Sushisource, line 31:
```suggestion
 * Temporal SDK may provide a subclass of {@link SerializationContext} during data
```

temporal-sdk/src/main/java/io/temporal/payload/context/SerializationContext.java
```diff
@@ -0,0 +1,63 @@
...
+ * serialization to identify the context of this serialization. This context is composed in a way to
+ * guarantee that a user object, and it's corresponded Payload will always be handled by the
```
Sushisource, line 33:
```suggestion
 * guarantee that a user object, and its corresponding Payload will always be handled by the
```

temporal-sdk/src/main/java/io/temporal/payload/context/WorkflowSerializationContext.java
```diff
@@ -0,0 +1,50 @@
...
+
+@Experimental
+public class WorkflowSerializationContext implements SerializationContext {
+  private final @Nonnull String namespace;
+  private final @Nonnull String workflowId;
+  // We can't currently reliably and consistency provide workflowType to the DataConverter.
+  // 1. Signals and queries don't know workflowType when they are sent.
+  // 2. WorkflowStub#getResult call is not aware of the workflowType, workflowType is an optional
+  // parameter for a workflow stub that is not used to start a workflow.
+  //    private final @Nullable String workflowTypeName;
```
Sushisource, line 34: Can get rid of this commented out line?

temporal-sdk/src/test/java/io/temporal/functional/serialization/WorkflowIdSignedPayloadsTest.java
```diff
@@ -0,0 +1,250 @@
...
+import java.util.stream.Collectors;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import org.junit.Rule;
+import org.junit.Test;
+
+/**
+ * This test emulates a scenario when users may be using WorkflowId in their encoding to sign every
+ * payload to be used for a specific workflow only to prevent a "payload replay" (had nothing to do
+ * with Temporal Replay) attack. For this scenario it's important that every entity during it's
```
Sushisource, line 56:
```suggestion
 * with Temporal Replay) attack. For this scenario it's important that every entity during its
```

temporal-sdk/src/main/java/io/temporal/payload/context/WorkflowSerializationContext.java
```diff
@@ -0,0 +1,50 @@
...
+    this.workflowId = workflowId;
+  }
+
+  @Nonnull
+  public String getNamespace() {
+    return namespace;
+  }
+
+  @Nonnull
+  public String getWorkflowId() {
```
Sushisource, line 47: I think this needs a docstring explaining that it means the child workflow ID when being used on either end of a child WF call.

temporal-sdk/src/main/java/io/temporal/payload/context/SerializationContext.java
```diff
@@ -0,0 +1,63 @@
...
+ * PayloadCodec#withContext(SerializationContext)}, and {@link
+ * PayloadConverter#withContext(SerializationContext)} and using the modified instance when
+ * applicable.
+ *
+ * <p>Note: Serialization Context is experimental feature, the class and field structure of {@link
+ * SerializationContext} objects may change in the future. There may be also situation where the
+ * context is expected, but is not currently provided.
+ */
+@Experimental
+public interface SerializationContext {}
```
Sushisource, line 63: We could potentially lift workflow id and namespace up into this interface. It's hard to imagine a situation where they ever wouldn't be available, but at the same time I can get wanting to keep this empty to avoid a problem with that later on.

temporal-sdk/src/test/java/io/temporal/functional/serialization/WorkflowIdSignedPayloadsTest.java
```diff
@@ -0,0 +1,250 @@
...
+
+    private Payload encodePayload(final Payload originalPayload) {
+      String workflowId = null;
+      assertNotNull(
+          "everything in this test should go through contextualized codecs", serializationContext);
+      if (serializationContext != null) {
+        if (serializationContext instanceof WorkflowSerializationContext) {
+          workflowId = ((WorkflowSerializationContext) serializationContext).getWorkflowId();
+        } else if (serializationContext instanceof ActivitySerializationContext) {
+          workflowId = ((ActivitySerializationContext) serializationContext).getWorkflowId();
```
Sushisource, line 189: Probably worth doing something activity specific in here too just to make sure that's getting piped into the context properly.

Sushisource (commented):
The general approach makes perfect sense to me. It's a bit ugly how much some of the constructor arities exploded. I'm thinking it might be reasonable to introduce some way to fetch an implicit context and use something like `SomeClass.getWorkflowDataConverter()` and `SomeClass.getActivityDataConverter` where we used to use `clientOptions.getDataConverter()`.

But, there's obviously downsides to that too.

temporal-sdk/src/main/java/io/temporal/payload/context/WorkflowSerializationContext.java
Spikhalskiy: I actually put and left it intentionally with this comment above. Having workflowType here is very natural and I'm sure it will be the next ask in line. And the developer who will consider or pick it up will be able to see why it wasn't added in the first place and what are the challenges here.

Spikhalskiy:
@Sushisource So, doing this

> I'm thinking it might be reasonable to introduce some way to fetch an implicit context and use something like SomeClass.getWorkflowDataConverter() and SomeClass.getActivityDataConverter where we used to use clientOptions.getDataConverter().

was my original thinking and design. Half way implementing it I realized that it has a VERY significant downside. Users will start to write data converters that RELY on the presence of the context and it will fail them in situations when they may be calling the converter themselves without any context at all for example.
That's why I implemented this `withContext` way and added this specific doc-line:
```
   * A correct implementation of this interface should have a fully functional "contextless"
   * implementation. Temporal SDK will call this method when a knowledge of the context exists, but
   * {@link DataConverter} can be used directly by user code and sometimes SDK itself without any
   * context.
```

If we expose the context as static methods, we pretty much set up users to use them as they are always given, while they are not.

temporal-sdk/src/main/java/io/temporal/internal/client/WorkflowClientRequestFactory.java
```diff
@@ -55,26 +53,30 @@ public WorkflowClientRequestFactory(WorkflowClientOptions clientOptions) {
...
   StartWorkflowExecutionRequest.Builder newStartWorkflowExecutionRequest(
-      WorkflowClientCallsInterceptor.WorkflowStartInput input) {
-    WorkflowOptions options = input.getOptions();
-
+      @Nonnull String workflowId,
+      @Nonnull String workflowTypeName,
+      @Nonnull io.temporal.common.interceptors.Header header,
+      @Nonnull WorkflowOptions options,
+      @Nullable Payloads inputArgs,
+      @Nullable Memo memo) {
```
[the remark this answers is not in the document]
Spikhalskiy: Nah. Across this whole PR I was doing work to pull usages of data converters together. "On entry points from the user code" in the call trees. Instead of spreading it all over the code. This method is cleaner now because it accepts already serialized entities and doesn't have to do it inside at all. And data conversion is done in RootWorkflowClientInvoker which is already doing data conversions anyway.

temporal-sdk/src/main/java/io/temporal/payload/context/WorkflowSerializationContext.java
Spikhalskiy: I think I explain this idea in https://github.com/temporalio/sdk-java/blob/ac95cc93ad4e114417cdda4d764fc344ca40718d/temporal-sdk/src/main/java/io/temporal/payload/context/SerializationContext.java#L44

But I hear what you are saying, I will try to also spread this idea among the field's javadocs.

temporal-sdk/src/main/java/io/temporal/payload/context/SerializationContext.java
Spikhalskiy: I'm not sure if I like that `getWorkflowId()` is one situation will actually mean an ID of the target entity (WorkflowSerializationContext) and just an attribute in another case (ActivitySerializationContext).
Let's maybe discuss on the call, there are different + for different approaches here. And there are >2 approaches.

Sushisource (approved):
Pre-approved per our on-call discussion :)

PR #1723, opened 2023-03-29
Allow Data Converter code to escape deadlock detection

### LINKED ISSUE #1301: Implement an ability to disable deadlock detector for data converters performing IO
Some users have data converters that perform an IO and they sometimes trigger deadline detector. 

JavaSDK should have a way to disable deadline detector for a piece of code (under `WorkflowUnsafe`) and for a specific data converter.
See implementation in GoSDK as a reference: https://github.com/temporalio/sdk-go/pull/772

Spikhalskiy:
## What was changed
Add `WorkflowUnsafe#deadlockDetectorOff` that allows running pieces of code with disabled deadlock detection

## Why?
Closes #1301

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowInternal.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowInternal.java
@@ -38,7 +38,9 @@ import io.temporal.common.interceptors.WorkflowOutboundCallsInterceptor;
 import io.temporal.common.metadata.POJOWorkflowImplMetadata;
 import io.temporal.common.metadata.POJOWorkflowInterfaceMetadata;
 import io.temporal.common.metadata.POJOWorkflowMethodMetadata;
+import io.temporal.internal.WorkflowThreadMarker;
 import io.temporal.internal.common.ActivityOptionUtils;
+import io.temporal.internal.common.NonIdempotentHandle;
 import io.temporal.internal.common.SearchAttributesUtil;
 import io.temporal.internal.logging.ReplayAwareLogger;
 import io.temporal.serviceclient.CheckedExceptionWrapper;
@@ -395,14 +397,6 @@ public final class WorkflowInternal {
     return result.get();
   }
 
-  private static WorkflowOutboundCallsInterceptor getWorkflowOutboundInterceptor() {
-    return getRootWorkflowContext().getWorkflowOutboundInterceptor();
-  }
-
-  static SyncWorkflowContext getRootWorkflowContext() {
-    return DeterministicRunnerImpl.currentThreadInternal().getWorkflowContext();
-  }
-
   public static void await(String reason, Supplier<Boolean> unblockCondition)
       throws DestroyWorkflowThreadError {
     getWorkflowOutboundInterceptor().await(reason, unblockCondition);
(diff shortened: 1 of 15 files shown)


temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowInternal.java
```diff
@@ -520,6 +506,27 @@ public static void sleep(Duration duration) {
     getWorkflowOutboundInterceptor().sleep(duration);
   }
 
+  public static boolean isWorkflowThread() {
+    return WorkflowThreadMarker.isWorkflowThread();
+  }
+
+  public static <T> T deadlockDetectorOff(Functions.Func<T> func) {
+    if (isWorkflowThread()) {
+      Runnable deadlockDetectorUnlockHandle = getWorkflowThread().lockDeadlockDetector();
```
cretz, line 515: Arguably something that returns `Closeable` here so you could use try-with-resources would be clearer

temporal-sdk/src/main/java/io/temporal/workflow/unsafe/WorkflowUnsafe.java
```diff
@@ -66,6 +66,34 @@ public static boolean isReplaying() {
...
   }
 
+  /**
+   * Runs the supplied procedure in the calling thread with disabled deadlock detection if called
+   * from the workflow thread. Does nothing except the procedure execution if called from a
+   * non-workflow thread.
+   *
+   * @param proc to run with disabled deadlock detection
+   */
+  public static void deadlockDetectorOff(Functions.Proc proc) {
```
cretz, line 76: We chose not to expose these in Go, just left them in internal code read to expose if/when we wanted. Instead we offered the more specific way to create a data converter with deadlock disabled. I wonder if it'd be worth adding a `default DataConverter withWorkflowDeadlockDetectorDisabled()` to the `DataConverter` interface (though I can understand if you don't want that class referencing the workflow package).

But I have no strong opinion.

cretz (approved):

temporal-sdk/src/main/java/io/temporal/workflow/unsafe/WorkflowUnsafe.java
Spikhalskiy: I think users should strive to isolate the "deadlock disabled" segment of code as much as only possible. So, pieces of codec or payload converter that are not performing IO shouldn't go under the disabled deadlock.
I think we shouldn't encourage to just disabling of deadlock detector for the whole DataConverter assembly.
We can raise it on a standup maybe.

temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowInternal.java
Spikhalskiy: I can do AutoCloseable. I don't want to make this thingy idempotent, because it will increase synchronization costs without real benefit.

temporal-sdk/src/main/java/io/temporal/workflow/unsafe/WorkflowUnsafe.java
cretz: > I think we shouldn't encourage to just disabling of deadlock detector for the whole DataConverter assembly.

It's a tough tradeoff, because arguably this encourages disabling deadlock detector everywhere.

> Let's sync up on standup.

:+1:

Spikhalskiy: > It's a tough tradeoff, because arguably this encourages disabling deadlock detector everywhere.

I'm not exactly worried about it putting it into WorkflowUnsafe. All tools in that class are expected to be used with huge caution and not treated lightly. But good note, I will actually add an explicit doc that these new methods shouldn't be used in actual workflow code implementations.

PR #1724, opened 2023-03-29
DRAFT: Add update annotation

Quinn-With-Two-Ns:
Opening this draft PR to review with @Spikhalskiy some details on the `Validator` implementation for Synchronous Update

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowInternal.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowInternal.java
@@ -50,11 +50,15 @@ import io.temporal.workflow.CompletablePromise;
 import io.temporal.workflow.ContinueAsNewOptions;
 import io.temporal.workflow.DynamicQueryHandler;
 import io.temporal.workflow.DynamicSignalHandler;
+import io.temporal.workflow.DynamicUpdateHandler;
 import io.temporal.workflow.ExternalWorkflowStub;
 import io.temporal.workflow.Functions;
 import io.temporal.workflow.Functions.Func;
+import io.temporal.workflow.Functions.Proc1;
 import io.temporal.workflow.Promise;
 import io.temporal.workflow.QueryMethod;
+import io.temporal.workflow.UpdateMethod;
+import io.temporal.workflow.Validator;
 import io.temporal.workflow.Workflow;
 import io.temporal.workflow.WorkflowInfo;
 import io.temporal.workflow.WorkflowMethod;
@@ -71,6 +75,8 @@ import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import java.lang.reflect.ParameterizedType; 
+
 
 /**
  * Never reference directly. It is public only because Java doesn't have internal package support.
@@ -141,7 +147,7 @@ public final class WorkflowInternal {
   /**
    * Register query or queries implementation object. There is no need to register top level
    * workflow implementation object as it is done implicitly. Only methods annotated with @{@link
-   * QueryMethod} are registered.
+   * QueryMethod} are registered. TODO(quinn) LIES!
    */
   public static void registerListener(Object implementation) {
     if (implementation instanceof DynamicSignalHandler) {
@@ -156,6 +162,13 @@ public final class WorkflowInternal {
           .registerDynamicQueryHandler(
               new WorkflowOutboundCallsInterceptor.RegisterDynamicQueryHandlerInput(
                   (DynamicQueryHandler) implementation));
+      return;
+    }
+    if (implementation instanceof DynamicUpdateHandler) {
+      getWorkflowOutboundInterceptor()
+          .registerDynamicUpdateHandler(
+              new WorkflowOutboundCallsInterceptor.RegisterDynamicUpdateHandlerInput(
+                  (DynamicUpdateHandler) implementation));
       return;
     }
     Class<?> cls = implementation.getClass();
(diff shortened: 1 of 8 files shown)


temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowInternal.java
```diff
@@ -197,6 +210,41 @@ public static void registerListener(Object implementation) {
...
+
+      updateRequests.add(
+          new WorkflowOutboundCallsInterceptor.UpdateRegistrationRequest(
+              methodMetadata.getName(),
+              method.getParameterTypes(),
+              method.getGenericParameterTypes(),
+              (args) -> {
+                try {
+                  Validator<?> validator = updateMethod.validator().getDeclaredConstructor().newInstance();
+                  // TODO(quinn) cast implementation to <?> don't think this is possible
```
Quinn-With-Two-Ns, line 226: @Spikhalskiy This is the part I wasn't sure was possible if the `Validator` had  a generic argument

temporal-sdk/src/main/java/io/temporal/workflow/UpdateMethod.java
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Indicates that the method is a update handler method. A update method gets executed when a
+ * workflow receives a update after the validator is called.
+ *
+ * <p>This annotation applies only to workflow interface methods.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.METHOD)
+public @interface UpdateMethod {
```
cretz, line 32: I wonder if it's clearer to just have an update validator annotation than a validator class. This annotation can _require_ a name, and we can validate at registration time (along w/ our other annotation validations) that the name references an update and that the arguments are exactly the same and return type is void. This is better than `Object[]` arguments.

I think this:

```java
public class MyWorkflow {
  private String thing = "<unset>;

  // ...

  @UpdateMethod
  public String doUpdate(int someArg) {
    // ...
  }

  @UpdateValidationMethod("doUpdate")
  public void validateUpdate(int someArg) {
    Preconditions.checkArgument(someArg >= 0, "Argument can't be negative");
    Preconditions.checkState(!"<unset>".equals(thing), "Thing not ready");
  }
}
```

Looks better than:

```java
public class MyWorkflow {
  private String thing = "<unset>;

  // ...

  @UpdateMethod(validator=DoUpdateValidator.class)
  public String doUpdate(int someArg) {
    // ...
  }

  public static class DoUpdateValidator implements Validator<MyWorkflow> {
    public void apply(MyWorkflow workflowObject, Object[] args) {
      Preconditions.checkArgument((int)args[0] >= 0, "Argument can't be negative");
      Preconditions.checkState(!"<unset>".equals(MyWorkflow.thing), "Thing not ready");
    }
  }
}
```

temporal-sdk/src/main/java/io/temporal/workflow/DynamicUpdateHandler.java
```diff
@@ -0,0 +1,38 @@
...
+ * <p>When registered any signals which don't have a specific handler will be delivered to it.
+ *
+ * @see DynamicQueryHandler
+ * @see DynamicSignalHandler
+ * @see DynamicWorkflow
+ */
+public interface DynamicUpdateHandler {
+  void validate(String updateName, EncodedValues args);
+
+  EncodedValues execute(String updateName, EncodedValues args);
```
cretz, line 37: This should be called `handle` to be consistent with other "handlers"

Really though I think `DynamicSignalHandler` and `DynamicQueryHandler` got it a bit wrong. They should have qualified their method names to `handleDynamicSignal` and `handleDynamicQuery` respectively. Then this should have methods named `validateDynamicUpdate` and `handleDynamicUpdate`. So that way I could implement all three interfaces in the same class without ambiguity.

temporal-sdk/src/main/java/io/temporal/workflow/DynamicUpdateHandler.java
```diff
@@ -0,0 +1,38 @@
...
+ * DynamicUpdateListener. Only one such listener can be registered per workflow execution.
+ *
+ * <p>When registered any signals which don't have a specific handler will be delivered to it.
+ *
+ * @see DynamicQueryHandler
+ * @see DynamicSignalHandler
+ * @see DynamicWorkflow
+ */
+public interface DynamicUpdateHandler {
+  void validate(String updateName, EncodedValues args);
```
cretz, line 35: This should have a default no-op implementation.

temporal-sdk/src/main/java/io/temporal/workflow/Validator.java
```diff
@@ -0,0 +1,31 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Validator are used during update to decided weather to accept or reject an update workflow
+ * execution.
+ */
+@FunctionalInterface
+public interface Validator<W> {
```
cretz, line 22: This should be called `UpdateValidator`. This is too generic of a term in a generic package to assume it's for updates.

temporal-sdk/src/main/java/io/temporal/workflow/UpdateMethod.java
Quinn-With-Two-Ns: That was my initial idea as well. @Spikhalskiy argued the validator approach was the more common Java pattern and we should support it, but we can add other in the future.

cretz: I think we have to weigh common Java pattern with losing argument types and general user annoyance. I would rather not have to create a separate class and cast all my arguments manually even if that's the "common Java pattern".

temporal-sdk/src/main/java/io/temporal/workflow/DynamicUpdateHandler.java
cretz: > You still can through overloading no?

I don't believe so. For example, given the `DynamicQueryHandler.handle` and `DynamicSignalHandler.handle` methods define the same name and same arg types but different return types, I cannot think of a way for a single Java class to implement both (my Java may be rusty though).

PR #1732, opened 2023-04-06
typo and conditional argument check fix

ManuverGujjar:
## What was changed
Changed argument checking with Preconditions. and fixed a typo

## Why?
Reading through code, those changes were needed for consitency across codebase

## Checklist
NA

2. How was this tested:
NA

3. Any docs updates needed?
NA


Spikhalskiy:
This change doesn't have any substance and is just a stylistic refactoring. We typically perform such refactorings when we perform another meaningful work on the affected piece of code. One of the reasons is to preserve a somewhat-accurate "blame".

PR #1735, opened 2023-04-11
remove the explicit dependency upon ServiceStubsAutoConfiguration

osi:
## What was changed
Modify the Spring AutoConfiguration to be conditional on what is actually required, the service stubs

## Why?
This allows users to potentially disable the `ServiceStubsAutoConfiguration` and build a `WorkflowServiceStubs` themselves while still using the rest of the library. This may be needed if, for instance, further customization of the gRPC connection is required.

## Checklist
This was tested by running the unit tests. They all pass.


Spikhalskiy (approved):

Spikhalskiy:
It would be great if you can add an integration test that covers and solidifies your scenario by creating an applicable profile under resources/application.yml and a test with the right beans structure under `io.temporal.spring.boot.autoconfigure.*` like it's done in `CustomDataConverterTest` class. But feel free to ignore.

Issue #1745, opened 2023-04-18
Support sync update request


Quinn-With-Two-Ns:
**Is your feature request related to a problem? Please describe.**
Add support for sending synchronous update request to external workflows from the java client

PR #1749, opened 2023-04-21
Add support for sync update

Quinn-With-Two-Ns:
## What was changed
Add support for registering update handles to workflows and triggering updates from the client and stub

## Checklist

1. Closes (https://github.com/temporalio/sdk-java/issues/1745)

2. How was this tested:
Added some integration and unit test to verify the changes, more will be added in the feature test to verify consistency with Go

All functionality related to update should be tagged with `@Experimental`


temporal-sdk/src/main/java/io/temporal/client/UpdateReference.java
```diff
@@ -0,0 +1,54 @@
...
+import io.temporal.api.common.v1.WorkflowExecution;
+import io.temporal.common.Experimental;
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * UpdateReference is a reference to an update workflow execution request that can be used to get
+ * the status of that update request.
+ */
+@Experimental
+public interface UpdateReference<T> {
```
Quinn-With-Two-Ns, line 32: I called this a Reference instead of a handle because the Java SDK already refers to the annotated methods for processing signals, queries and updates as handles.

temporal-sdk/src/main/java/io/temporal/common/metadata/POJOWorkflowMethod.java
```diff
@@ -65,6 +66,20 @@ final class POJOWorkflowMethod {
...
+    if (updateMethod != null) {
+      type = WorkflowMethodType.UPDATE;
+      count++;
+      name = updateMethod.name();
+    }
+    if (updateValidatorMethod != null) {
+      type = WorkflowMethodType.UPDATE_VALIDATOR;
+      if (method.getReturnType() != Void.TYPE) {
+        throw new IllegalArgumentException(
+            "Method annotated with @UpdateValidateMethod must have a void return type: " + method);
```
Quinn-With-Two-Ns, line 78: `UpdateValidateMethod ` -> `UpdateValidatorMethod `

temporal-sdk/src/main/java/io/temporal/internal/client/external/GenericWorkflowClient.java
```diff
@@ -38,6 +38,8 @@ SignalWithStartWorkflowExecutionResponse signalWithStart(
 
   QueryWorkflowResponse query(QueryWorkflowRequest queryParameters);
 
+  UpdateWorkflowExecutionResponse update(UpdateWorkflowExecutionRequest updateParameters);
```
Quinn-With-Two-Ns, line 41: `@Experimental`

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
```diff
@@ -228,16 +247,60 @@ private void handleEventsBatch(List<HistoryEvent> events, boolean hasNextEvent)
...
+        }
+      }
+    }
+
     for (Iterator<HistoryEvent> iterator = events.iterator(); iterator.hasNext(); ) {
       HistoryEvent event = iterator.next();
+
+      // On replay the messages are available after the workflow task schedule event, so we
+      // need to handle them before workflow task started event to maintain a consistent order.
+      for (Message msg : this.takeLTE(event.getEventId() - 1)) {
```
Quinn-With-Two-Ns, line 274: This is not very elegant, but It tried a few ways and this is the best I could come up with that doesn't require a lot of refactoring. Keep in mine once we have proper commands for messages this code will need to change so I didn't think it was worth the effort at this moment.

temporal-sdk/src/main/java/io/temporal/workflow/UpdateValidateMethod.java
```diff
@@ -0,0 +1,50 @@
...
+public @interface UpdateValidateMethod {
+
+  /**
+   * Name of the update handler the validator should be used for. Must not be null.
+   *
+   * <p>Be careful about names that contain special characters. These names can be used as metric
+   * tags. And systems like prometheus ignore metrics which have tags with unsupported characters.
+   */
+  @Nonnull
+  String updateName();
```
Quinn-With-Two-Ns, line 49: The other option I thought of was to have the update declare the validator it wanted, more similar to Go

temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflow.java
```diff
@@ -139,6 +144,35 @@ public void handleSignal(String signalName, Optional<Payloads> input, long event
...
+              callbacks.reject(this.dataConverter.exceptionToFailure(e));
+              return;
+            }
+          }
+          callbacks.accept();
+          try {
+            Optional<Payloads> result =
+                workflowProc.handleExecuteUpdate(updateName, input, eventId);
+            callbacks.complete(result, null);
+          } catch (Exception e) {
```
cretz, line 170: Hrmm, it is a curious question whether an update exception deserves to fail the update or fail the task. In my head I would think, like workflow functions and signal handlers IIUC, any non-temporal-failure exception is a task failure not an update failure. But for queries it would make sense that any exception is a query failure.

Sure some of us are not fans of the task-failure concept when applied to generic user exceptions, but I think we need to be consistent. So I think any validator exception is a rejection, but exceptions in the update handler itself are treated like exceptions in workflow or signal handlers.

temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowInternal.java
```diff
@@ -199,6 +188,73 @@ public static void registerListener(Object implementation) {
           .registerSignalHandlers(
               new WorkflowOutboundCallsInterceptor.RegisterSignalHandlersInput(requests));
     }
+
+    // Get all validators and lazily assign them to update handlers as we see them.
+    Map<String, POJOWorkflowMethodMetadata> validators = new HashMap<>();
```
cretz, line 193:
```suggestion
    Map<String, POJOWorkflowMethodMetadata> validators = new HashMap<>(workflowMetadata.getUpdateValidatorMethods().size());
```

Usually best to give a capacity to these collections where you know what they'll be. Especially short-lived ones like this. There are other collections in this review this same logic should apply to.

temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowInternal.java
```diff
@@ -199,6 +188,73 @@ public static void registerListener(Object implementation) {
...
+        new ArrayList<>();
+    for (POJOWorkflowMethodMetadata methodMetadata : workflowMetadata.getUpdateMethods()) {
+      Method method = methodMetadata.getWorkflowMethod();
+      UpdateMethod updateMethod = method.getAnnotation(UpdateMethod.class);
+      String updateMethodName = updateMethod.name();
+      if (updateMethodName.isEmpty()) {
+        updateMethodName = method.getName();
+      }
+      // Check if any validators claim they are the validator for this update
+      POJOWorkflowMethodMetadata validatorMethodMetadata = validators.get(updateMethodName);
```
cretz, line 214:
```suggestion
      POJOWorkflowMethodMetadata validatorMethodMetadata = validators.remove(updateMethodName);
```

No need for a separate remove call below

temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowInternal.java
```diff
@@ -199,6 +188,73 @@ public static void registerListener(Object implementation) {
...
+        }
+      } else {
+        validatorMethod = null;
+      }
+      updateRequests.add(
+          new WorkflowOutboundCallsInterceptor.UpdateRegistrationRequest(
+              methodMetadata.getName(),
+              method.getParameterTypes(),
+              method.getGenericParameterTypes(),
+              (args) -> {
```
cretz, line 233: This kind of stuff makes stack traces ugly. I wonder if we should just declare a checked `Exception` on the callbacks (or use one of those Lombok style cheats). Meh, this is probably fine, just a bit ugly.

temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowInternal.java
```diff
@@ -199,6 +188,73 @@ public static void registerListener(Object implementation) {
...
+                }
+              }));
+    }
+    if (!updateRequests.isEmpty()) {
+      getWorkflowOutboundInterceptor()
+          .registerUpdateHandlers(
+              new WorkflowOutboundCallsInterceptor.RegisterUpdateHandlersInput(updateRequests));
+    }
+    if (!validators.isEmpty()) {
+      throw new IllegalArgumentException("Unused validators declared: " + validators);
```
cretz, line 256:
```suggestion
      throw new IllegalArgumentException("Missing update methods for update validator(s): " + Joiner.on(", ").join(validators.keySet()));
```

temporal-sdk/src/main/java/io/temporal/workflow/UpdateValidateMethod.java
cretz: This is better IMO since it's often the case an update may not have a validator, it's best to have the validator attach itself vs the update attach itself to the validator.

cretz (commented):
Looks good! I admit I didn't dig too deeply into the state machine stuff due to lack of familiarity.

temporal-sdk/src/main/java/io/temporal/client/WorkflowUpdateExecutionFailedException.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+@Experimental
+public final class WorkflowUpdateExecutionFailedException extends WorkflowException {
+
+  private final String updateId;
+  private final String updateName;
+
+  public WorkflowUpdateExecutionFailedException(
+      WorkflowExecution execution,
+      String workflowType,
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: `workflowType`? it is used by the base calasss constructor?

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
mmcshane: This hack was necessary because of a quirk in how Go handles events, namely that WFTScheduled events don't even go through the normal event handler with the big event type switch. If Java doesn't have that same quirk then you don't need to replicate that weirdness.

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
```diff
@@ -416,6 +513,22 @@ public List<Command> takeCommands() {
...
   }
 
+  public void sendMessage(Message message) {
+    checkEventLoopExecuting();
+    if (!isReplaying()) {
+      messageOutbox.add(message);
+    }
+  }
+
+  public List<Message> takeMessages() {
```
mmcshane, line 523: Minor: Probably cheaper to grab a temp ref to messageOutbox, reassign messageOutbox = new ArrayList<>() and then return the temp ref here.

temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflow.java
mmcshane: Not exactly clear if you're asking for a change here, @cretz but the current implementation to treat exceptions from update execution as failures _of the update_ seems correct to me.

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowTaskHandler.java
```diff
@@ -76,6 +80,13 @@ public boolean isCompletionCommand() {
       return completionCommand;
     }
 
+    public Functions.Proc1<Long> getEventIdSetHandle() {
```
mmcshane, line 83: I have a slight preference for requiring this field to be non-null and passing a do-nothing value in places where we don't want to do the reset. Minor though.

temporal-sdk/src/main/java/io/temporal/workflow/UpdateValidateMethod.java
mmcshane: Only thing is that I don't think this could support reusing validators  for multiple update functions, could it?

temporal-sdk/src/test/java/io/temporal/internal/statemachines/TestHistoryBuilder.java
```diff
@@ -30,34 +30,7 @@
...
-import io.temporal.api.history.v1.TimerStartedEventAttributes;
-import io.temporal.api.history.v1.WorkflowExecutionCompletedEventAttributes;
-import io.temporal.api.history.v1.WorkflowExecutionSignaledEventAttributes;
-import io.temporal.api.history.v1.WorkflowExecutionStartedEventAttributes;
-import io.temporal.api.history.v1.WorkflowTaskCompletedEventAttributes;
-import io.temporal.api.history.v1.WorkflowTaskFailedEventAttributes;
-import io.temporal.api.history.v1.WorkflowTaskScheduledEventAttributes;
-import io.temporal.api.history.v1.WorkflowTaskStartedEventAttributes;
-import io.temporal.api.history.v1.WorkflowTaskTimedOutEventAttributes;
+import io.temporal.api.history.v1.*;
```
mmcshane, line 33: I suspect it's IntelliJ do this - suggest you commit the project config files in .idea (not for this PR, ob)

temporal-sdk/src/test/java/io/temporal/workflow/updateTest/UpdateTest.java
```diff
@@ -0,0 +1,218 @@
...
+          .setWorkerOptions(WorkerOptions.newBuilder().build())
+          .setWorkflowTypes(TestUpdateWorkflowImpl.class)
+          .setActivityImplementations(new ActivityImpl())
+          .setTestTimeoutSeconds(60)
+          .build();
+
+  @Test
+  public void testUpdate() {
+    Assume.assumeTrue(
+        "skipping because test server does not support update",
```
mmcshane, line 64: You've run these though, right?

temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflow.java
```diff
@@ -139,6 +144,35 @@ public void handleSignal(String signalName, Optional<Payloads> input, long event
...
   }
 
+  @Override
+  public void handleUpdate(
+      String updateName, Optional<Payloads> input, long eventId, UpdateProtocolCallback callbacks) {
+    runner.executeInWorkflowThread(
+        "update " + updateName,
+        () -> {
+          // Skip validator on replay
+          if (!callbacks.isReplaying()) {
```
mmcshane, line 154: Is isReplaying not available anywhere else in this scope? It would be nice not to have that on the interface.

mmcshane (approved):
Overall lgtm. Comments are mostly nits or improvements that are not required.

temporal-sdk/src/main/java/io/temporal/client/WorkflowUpdateExecutionFailedException.java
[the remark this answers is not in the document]
cretz: Just set it as null when calling the base class constructor

temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflow.java
cretz: It does not seem correct to me and I think @Quinn-With-Two-Ns and I agreed to change this.

In non-Go SDKs, only Temporal exceptions fail the thing (workflow, signal handler, etc) and non-Temporal exceptions fail the task. Task failure vs workflow/update failure is an annoying thing to deal with as a user, but it is consistent with the choices we've made elsewhere.

temporal-sdk/src/main/java/io/temporal/workflow/UpdateValidateMethod.java
Quinn-With-Two-Ns: correct that is a drawback @mmcshane

cretz: I think that's an acceptable drawback, same as not reusing signal handlers or update handlers for multiple purposes.

temporal-sdk/src/main/java/io/temporal/client/WorkflowStub.java
```diff
@@ -73,6 +74,92 @@ static <T> WorkflowStub fromTyped(T typed) {
...
+   * io.temporal.workflow.UpdateMethod}.
+   *
+   * @param updateName name of the update handler. Usually it is a method name.
+   * @param resultClass class of the update return value
+   * @param <R> type of the update return value
+   * @param args update method arguments
+   * @return update reference that can be used to get the result of the update.
+   */
+  @Experimental
+  <R> UpdateReference<R> updateAsync(String updateName, Class<R> resultClass, Object... args);
```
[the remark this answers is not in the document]
cretz: > Or maybe the utility is so dumb and easy that it's not necessary

That's true in Go, but Java with it's overloads and default interface impls it may be worth it. In other languages we often have "start" + "handle get", and then that combination we make a shortcut for, but not Go.

temporal-sdk/src/main/java/io/temporal/client/UpdateReference.java
```diff
@@ -0,0 +1,54 @@
...
+   */
+  String getId();
+
+  /**
+   * Returns a {@link CompletableFuture} with the update workflow execution request result
+   * potentially waiting for the update to complete.
+   *
+   * @return future completed with the result of the update or an exception
+   */
+  <R> CompletableFuture<R> getResultAsync();
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: Consistent with other usage in the SDK and a richer API then future.

temporal-sdk/src/main/java/io/temporal/client/UpdateReference.java
Quinn-With-Two-Ns: user visibly

temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflow.java
Quinn-With-Two-Ns: yeah I am not a huge fan o favas decision to make exceptions fail the WFT, but that is the decision so I agree we should be consistent.

temporal-sdk/src/test/java/io/temporal/workflow/updateTest/UpdateTest.java
Quinn-With-Two-Ns: yes they are only skipped if ` testWorkflowRule.isUseExternalService()` so CI will run them when testing against docker

temporal-sdk/src/main/java/io/temporal/workflow/UpdateValidateMethod.java
mmcshane: That being the case, I'd throw my vote with specifying the validator method from the update annotation with a default validator of "__unused" or something like that. It's a mild preference and I would defer to the sdk owner.

temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflow.java
mmcshane: I misunderstood the original statement but I agree with the the above two comments :+1:

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
```diff
@@ -228,16 +247,60 @@ private void handleEventsBatch(List<HistoryEvent> events, boolean hasNextEvent)
       }
     }
 
+    // Look ahead to infer protocol messages
+    if (EventType.EVENT_TYPE_WORKFLOW_TASK_STARTED.equals(events.get(0).getEventType())) {
+      if (this.isReplaying()) {
+        for (HistoryEvent event : events) {
+          if (event.hasWorkflowExecutionUpdateAcceptedEventAttributes()) {
```
Sushisource, line 254: This loop could be merged with the above one which is already looking ahead for version markers

temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflow.java
Quinn-With-Two-Ns: it is not

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
```diff
@@ -258,6 +321,32 @@ private RuntimeException createEventProcessingException(RuntimeException e, Hist
     }
   }
 
+  private void handleSingleMessage(Message message) {
+    // Get or create protocol state machine based on Instance ID and protocolName
+    EntityStateMachine stateMachine =
+        protocolStateMachines.computeIfAbsent(
+            message.getProtocolInstanceId(),
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: The state machines are per protocol instance ID so that is what the map is, maybe I don't understand your question

temporal-sdk/src/main/java/io/temporal/internal/client/external/GenericWorkflowClientImpl.java
```diff
@@ -248,4 +248,21 @@ private static <T> CompletableFuture<T> toCompletableFuture(
...
+            .build();
+    Scope scope = metricsScope.tagged(tags);
+
+    return grpcRetryer.retryWithResult(
+        () ->
+            service
+                .blockingStub()
+                .withOption(METRICS_TAGS_CALL_OPTIONS_KEY, scope)
+                .updateWorkflowExecution(updateParameters),
+        grpcRetryerOptions);
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: It will timeout based on the `grpcRetryerOptions`

temporal-sdk/src/main/java/io/temporal/internal/common/ProtocolUtils.java
```diff
@@ -0,0 +1,40 @@
...
+package io.temporal.internal.common;
+
+import io.temporal.api.protocol.v1.Message;
+
+public class ProtocolUtils {
+  public static String getProtocol(Message message) {
+    String typeUrl = message.getBody().getTypeUrl();
+    int slashIndex = typeUrl.lastIndexOf("/");
+    if (slashIndex == -1) {
+      throw new IllegalArgumentException("Message body type URL invalid");
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: No and it is not recommend to use `Preconditions` in this way

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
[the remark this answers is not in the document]
Sushisource: What I mean is (and maybe `protocol_instance_id` isn't what I think it is) the normal state machine map maps event IDs to state machines. If `protocol_instance_id` is something like `UPDATE_PROTOCOL` or whatever, how can we have two updates running at the same time? Seemingly they ought to be differentiated by their `id` rather than `protocol_instance_id`.

I'm sure I'm just missing something here, would be good to just clear it up, maybe that means changing the doc comment here https://github.com/temporalio/api/blob/e40424b5e4dd109fb3278e76820dc7990d009486/temporal/api/protocol/v1/message.proto#L44 since it sounds like that field isn't what I thought it was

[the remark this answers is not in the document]
Quinn-With-Two-Ns: `protocol_instance_id ` is not `UPDATE_PROTOCOL ` . From the docs it says `Identifies the specific instance of a protocol` it says it identifies an instance of a protocol, not the protocol type itself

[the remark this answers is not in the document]
mmcshane: Elsewhere in this PR you can see ProtocolUtils and ProtocolType which reflect the concepts you're thinking of, Spencer. Specifically, the ProtocolType is "temporal.api.update.v1" for this version of the protocol. Then we have _instances_ of the protocol. In practice these use the initially requested UpdateID to identify the collection of messages and state for a single update.  If you're running 10 updates then each will be identifiable by a protocolInstanceID

temporal-sdk/src/main/java/io/temporal/internal/statemachines/UpdateProtocolStateMachine.java
```diff
@@ -0,0 +1,196 @@
...
+          StateMachineDefinition.<State, ExplicitEvent, UpdateProtocolStateMachine>newInstance(
+                  "Update", State.NEW, State.COMPLETED)
+              .add(
+                  State.NEW,
+                  ProtocolType.UPDATE_V1,
+                  State.REQUEST_INITIATED,
+                  UpdateProtocolStateMachine::triggerUpdate)
+              .add(State.REQUEST_INITIATED, ExplicitEvent.ACCEPT, State.ACCEPTED)
+              .add(State.REQUEST_INITIATED, ExplicitEvent.REJECT, State.COMPLETED)
+              .add(State.ACCEPTED, ExplicitEvent.COMPLETE, State.COMPLETED);
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: For various reasons adding a bunch of transitions from states to themselves is not really feasible, I'll add protection before I actually send the message to the state machine

temporal-sdk/src/main/java/io/temporal/client/WorkflowUpdateExecutionFailedException.java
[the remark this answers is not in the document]
Quinn-With-Two-Ns: oh yeah I see your point

temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflow.java
```diff
@@ -139,6 +145,35 @@ public void handleSignal(String signalName, Optional<Payloads> input, long event
...
+              callbacks.reject(this.dataConverter.exceptionToFailure(e));
+              return;
+            }
+          }
+          callbacks.accept();
+          try {
+            Optional<Payloads> result =
+                workflowProc.handleExecuteUpdate(updateName, input, eventId);
+            callbacks.complete(result, null);
+          } catch (WorkflowExecutionException e) {
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: No, at this level `TemporalFailure` is converted to a failure held by `WorkflowExecutionException`

temporal-sdk/src/main/java/io/temporal/client/WorkflowStubImpl.java
```diff
@@ -292,6 +293,81 @@ public <R> R query(String queryType, Class<R> resultClass, Type resultType, Obje
     return result.getResult();
   }
 
+  @Override
+  public <R> R update(String updateName, Class<R> resultClass, Object... args) {
+    return update(updateName, "", "", resultClass, resultClass, args);
+  }
+
+  @Override
+  public <R> R update(
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: yes, I would prefer to take up a refactor like that when I implement the async path

cretz (approved):

PR #1762, opened 2023-05-10
Add support for update to the Java test server

Quinn-With-Two-Ns:
Add support for sync update to the test server.

### DIFF
--- a/temporal-sdk/src/test/java/io/temporal/client/functional/UpdateTest.java
+++ b/temporal-sdk/src/test/java/io/temporal/client/functional/UpdateTest.java
@@ -20,16 +20,18 @@
 
 package io.temporal.client.functional;
 
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertThrows;
 
 import io.temporal.api.common.v1.WorkflowExecution;
 import io.temporal.client.*;
 import io.temporal.testing.internal.SDKTestOptions;
 import io.temporal.testing.internal.SDKTestWorkflowRule;
+import io.temporal.workflow.CompletablePromise;
+import io.temporal.workflow.Workflow;
 import io.temporal.workflow.shared.TestWorkflows;
 import java.util.concurrent.ExecutionException;
 import org.junit.Assume;
-import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 
@@ -38,15 +40,10 @@ public class UpdateTest {
   @Rule
   public SDKTestWorkflowRule testWorkflowRule =
       SDKTestWorkflowRule.newBuilder()
-          .setWorkflowTypes(UpdateTest.QuickWorkflowWithUpdateImpl.class)
+          .setWorkflowTypes(
+              UpdateTest.QuickWorkflowWithUpdateImpl.class,
+              UpdateTest.LongWorkflowWithUpdateImpl.class)
           .build();
-
-  @Before
-  public void checkExternalService() {
-    Assume.assumeTrue(
-        "skipping because test server does not support update",
-        testWorkflowRule.isUseExternalService());
-  }
 
   @Test
   public void updateNonExistentWorkflow() {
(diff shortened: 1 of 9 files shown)


temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
```diff
@@ -461,6 +425,27 @@ public String toString() {
     }
   }
 
+  /** Represents an accepted update workflow execution request */
+  static final class UpdateWorkflowExecutionData {
+    final String Id;
```
cretz, line 430: Don't capitalize fields and variables in Java

temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
```diff
@@ -1222,6 +1221,16 @@ private static void bufferQuery(
     data.queryBuffer.put(query.getKey(), query);
   }
 
+  private static void bufferUpdate(
+      RequestContext ctx, WorkflowTaskData data, UpdateWorkflowExecution update, long notUsed) {
+    data.updateRequestBuffer.put(update.getId(), update);
```
cretz, line 1226: Need to validate one not already present for this ID here or on "add"?

temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
```diff
@@ -1729,6 +1748,92 @@ private static void heartbeatActivityTask(
...
+
+      HistoryEvent event =
+          HistoryEvent.newBuilder()
+              .setEventType(EventType.EVENT_TYPE_WORKFLOW_EXECUTION_UPDATE_ACCEPTED)
+              .setWorkflowExecutionUpdateAcceptedEventAttributes(acceptedAttribute)
+              .build();
+      // If the workflow is finished we can't write more events
+      // to history so if the message was processed after the workflow
+      // was closed there is nothing we can do.
+      // The real server also has this same problem
```
cretz, line 1775: What does the client-side caller of the update see in this situation?

Would love a to see tests for workflow-complete-before-update-accept (probably have to terminate while worker not running) and workflow-complete-before-update-complete (can just hang the update handler and finish the workflow).

temporal-sdk/src/test/java/io/temporal/client/functional/UpdateTest.java
```diff
@@ -102,6 +99,44 @@ public void updateCompletedWorkflowUntyped() {
     assertThrows(ExecutionException.class, () -> updateRef.getResultAsync().get());
   }
 
+  @Test
+  public void updateWorkflowDuplicateId() {
+    Assume.assumeTrue(
+        "skipping for real server because the real server does not handle duplicate update ID correctly",
```
cretz, line 105: Can we get a `TODO` + a link to a GH issue where they will solve it?

cretz (commented):
Added a couple of comments. Not reviewing _too_ closely because 1) the tests run against real server and this server so as we find things they can just be made tests, and 2) there's gonna be a lot more coming on workflow update in the near future I'm sure

temporal-sdk/src/test/java/io/temporal/client/functional/UpdateTest.java
Quinn-With-Two-Ns: I would like to, but the issue is tracked in Jira. I can link to the PR itself if you want https://github.com/temporalio/temporal/pull/4297?

cretz: Ah, so long as tracked, no problem.

temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
Quinn-With-Two-Ns: >What does the client-side caller of the update see in this situation?

This does not impact what the caller side sees. I'll note this shouldn't be an issue once we have message commands.

Quinn-With-Two-Ns: >Would love a to see tests for workflow-complete-before-update-accept (probably have to terminate while worker not running)

I agree but we need the server to support `UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_ADMITTED` to write this test.

> workflow-complete-before-update-complete (can just hang the update handler and finish the workflow).

I also agree, I was planning on covering this in the test server and the real server once they get support for `UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_ACCEPTED`

cretz: I wonder if we should just error in workflow-completed situations then instead of silently succeeding (unless I'm misreading)

temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
cretz: If you check there, good enough. I admit not following the flow all the way through.

temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
Quinn-With-Two-Ns: I don't understand why we would error. I am emulating the real server and the real server does not error. Erroring would also not help us?

temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
Quinn-With-Two-Ns: I added a check anyway, can't hurt

temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
cretz: I thought the real server just improperly handled this. Surely an update will error when workflow closed while processing.  I was just saying error here as not-yet-properly implemented on real server so intentionally isn't here either until we see the shape that error takes.

cretz (approved):

PR #1766, opened 2023-05-17
Add support for async update

Quinn-With-Two-Ns:
Add support for async update to test server and client.

Includes some breaking changes to interceptors and where exceptions are returned. This should be fine as all the APIs are marked experimental and never released.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/client/UpdateHandle.java
+++ b/temporal-sdk/src/main/java/io/temporal/client/UpdateHandle.java
@@ -23,6 +23,7 @@ package io.temporal.client;
 import io.temporal.api.common.v1.WorkflowExecution;
 import io.temporal.common.Experimental;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
 
 /**
  * UpdateHandle is a handle to an update workflow execution request that can be used to get the
@@ -51,4 +52,14 @@ public interface UpdateHandle<T> {
    * @return future completed with the result of the update or an exception
    */
   CompletableFuture<T> getResultAsync();
+
+  /**
+   * Returns a {@link CompletableFuture} with the update workflow execution request result
+   * potentially waiting for the update to complete.
+   *
+   * @param timeout maximum time to wait and perform the background long polling
+   * @param unit unit of timeout
+   * @return future completed with the result of the update or an exception
+   */
+  CompletableFuture<T> getResultAsync(long timeout, TimeUnit unit);
 }
--- a/temporal-sdk/src/main/java/io/temporal/client/WorkflowStub.java
+++ b/temporal-sdk/src/main/java/io/temporal/client/WorkflowStub.java
@@ -22,6 +22,7 @@ package io.temporal.client;
 
 import io.temporal.api.common.v1.WorkflowExecution;
 import io.temporal.api.enums.v1.QueryRejectCondition;
+import io.temporal.api.enums.v1.UpdateWorkflowExecutionLifecycleStage;
 import io.temporal.common.Experimental;
 import io.temporal.failure.CanceledFailure;
 import io.temporal.failure.TerminatedFailure;
@@ -103,8 +104,7 @@ public interface WorkflowStub {
    *     the update request with an error.
    * @param resultClass class of the update return value.
    * @param <R> type of the update return value.
-   * @param resultType type of the workflow return value. Differs from resultClass for generic
-   *     types.
+   * @param resultType type of the update return value. Differs from resultClass for generic types.
    * @param args update method arguments
    * @return update result
    * @throws WorkflowNotFoundException if the workflow execution doesn't exist or completed and
@@ -130,7 +130,7 @@ public interface WorkflowStub {
    * @param resultClass class of the update return value
    * @param <R> type of the update return value
    * @param args update method arguments
-   * @return update reference that can be used to get the result of the update.
+   * @return update handle that can be used to get the result of the update.
    */
   @Experimental
   <R> UpdateHandle<R> startUpdate(String updateName, Class<R> resultClass, Object... args);
@@ -145,21 +145,49 @@ public interface WorkflowStub {
    * @param firstExecutionRunId specifies the RunID expected to identify the first run in the
    *     workflow execution chain. If this expectation does not match then the server will reject
    *     the update request with an error.
+   * @param waitPolicy specifies at what point in the update request life cycles this request should
+   *     return.
    * @param resultClass class of the update return value.
    * @param <R> type of the update return value.
-   * @param resultType type of the workflow return value. Differs from resultClass for generic
-   *     types.
+   * @param resultType type of the update return value. Differs from resultClass for generic types.
    * @param args update method arguments
-   * @return update reference that can be used to get the result of the update.
+   * @return update handle that can be used to get the result of the update.
    */
   @Experimental
   <R> UpdateHandle<R> startUpdate(
       String updateName,
       String updateId,
       String firstExecutionRunId,
+      UpdateWorkflowExecutionLifecycleStage waitPolicy,
       Class<R> resultClass,
       Type resultType,
       Object... args);
+
+  /**
+   * Get an update handle to an update request previously started. Getting an update handle does not
+   * guarantee the update ID exists.
+   *
+   * @param updateId the identifier for the requested update.
+   * @param resultClass class of the update return value.
+   * @param <R> type of the update return value.
+   * @return update handle that can be used to get the result of the update.
+   */
+  @Experimental
+  <R> UpdateHandle<R> getUpdateHandle(String updateId, Class<R> resultClass);
+
+  /**
+   * Get an update handle to an update request previously started.Getting an update handle does not
+   * guarantee the update ID exists.
+   *
+   * @param updateId is an application-layer identifier for the requested update. It must be unique
+   *     within the scope of a workflow execution.
+   * @param resultClass class of the update return value.
+   * @param <R> type of the update return value.
+   * @param resultType type of the update return value. Differs from resultClass for generic types.
+   * @return update handle that can be used to get the result of the update.
+   */
+  @Experimental
+  <R> UpdateHandle<R> getUpdateHandle(String updateId, Class<R> resultClass, Type resultType);
 
   WorkflowExecution start(Object... args);
 
--- a/temporal-sdk/src/main/java/io/temporal/client/WorkflowStubImpl.java
+++ b/temporal-sdk/src/main/java/io/temporal/client/WorkflowStubImpl.java
@@ -24,22 +24,20 @@ import com.google.common.base.Strings;
 import io.grpc.Status;
 import io.grpc.StatusRuntimeException;
 import io.temporal.api.common.v1.WorkflowExecution;
+import io.temporal.api.enums.v1.UpdateWorkflowExecutionLifecycleStage;
 import io.temporal.api.errordetails.v1.QueryFailedFailure;
 import io.temporal.api.errordetails.v1.WorkflowExecutionAlreadyStartedFailure;
 import io.temporal.api.errordetails.v1.WorkflowNotReadyFailure;
+import io.temporal.api.update.v1.WaitPolicy;
 import io.temporal.common.interceptors.Header;
 import io.temporal.common.interceptors.WorkflowClientCallsInterceptor;
-import io.temporal.common.interceptors.WorkflowClientCallsInterceptor.UpdateOutput;
 import io.temporal.failure.CanceledFailure;
 import io.temporal.serviceclient.CheckedExceptionWrapper;
 import io.temporal.serviceclient.StatusUtils;
 import java.lang.reflect.Type;
 import java.util.Optional;
 import java.util.UUID;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.CompletionException;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
+import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicReference;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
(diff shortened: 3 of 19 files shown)


Quinn-With-Two-Ns:
Test are failing because polling is not enabled on the namespace for the server being tested hmm

temporal-sdk/src/main/java/io/temporal/client/UpdateHandle.java
```diff
@@ -51,4 +52,14 @@
...
+
+  /**
+   * Returns a {@link CompletableFuture} with the update workflow execution request result
+   * potentially waiting for the update to complete.
+   *
+   * @param timeout maximum time to wait and perform the background long polling
+   * @param unit unit of timeout
+   * @return future completed with the result of the update or an exception
+   */
+  CompletableFuture<T> getResultAsync(long timeout, TimeUnit unit);
```
cretz, line 64: Re https://github.com/temporalio/sdk-go/pull/1088#discussion_r1166787604, we also need this in Java at some point. We need a way to wait on a workflow handle to reach a stage even after submitted to server. But that doesn't have to be this PR.

temporal-sdk/src/main/java/io/temporal/client/WorkflowStub.java
```diff
@@ -145,22 +145,50 @@ <R> R update(
...
    * @param args update method arguments
-   * @return update reference that can be used to get the result of the update.
+   * @return update handle that can be used to get the result of the update.
    */
   @Experimental
   <R> UpdateHandle<R> startUpdate(
       String updateName,
       String updateId,
       String firstExecutionRunId,
+      UpdateWorkflowExecutionLifecycleStage waitPolicy,
```
cretz, line 161: I'm a bit torn on whether to wrap this in a friendlier enum. I know Java doesn't in some other places, but this name is nothing like the parameter name. I guess it can stay as is, it's just very ugly.

temporal-sdk/src/main/java/io/temporal/client/WorkflowStub.java
```diff
@@ -145,22 +145,50 @@ <R> R update(
...
    * @param <R> type of the update return value.
-   * @param resultType type of the workflow return value. Differs from resultClass for generic
-   *     types.
+   * @param resultType type of the update return value. Differs from resultClass for generic types.
    * @param args update method arguments
-   * @return update reference that can be used to get the result of the update.
+   * @return update handle that can be used to get the result of the update.
    */
   @Experimental
   <R> UpdateHandle<R> startUpdate(
```
cretz, line 157: I think the parameter count on this overload is large enough that you need `WorkflowUpdateOptions` + builder (like `WorkflowOptions`, `ChildWorkflowOptions`, etc). This will allow us to add options as needed too.

temporal-sdk/src/main/java/io/temporal/client/WorkflowStub.java
```diff
@@ -145,22 +145,50 @@ <R> R update(
...
+   * Get an update handle to an update request previously started. Getting an update handle does not
+   * guarantee the update ID exists.
+   *
+   * @param updateId the identifier for the requested update.
+   * @param resultClass class of the update return value.
+   * @param <R> type of the update return value.
+   * @return update handle that can be used to get the result of the update.
+   */
+  @Experimental
+  <R> UpdateHandle<R> getUpdateHandle(String updateId, Class<R> resultClass);
```
cretz, line 176: I wonder if this would make more sense as `newUpdateStub` and move start under there (but can still leave shortcuts out here). This matches other Java approaches, though would make it unlike other SDKs. Hrmm.

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
```diff
@@ -356,20 +349,111 @@ public <R> UpdateOutput<R> update(UpdateInput<R> input) {
...
+                        "Received unexpected outcome from poll update request: "
+                            + result.getOutcome().getValueCase());
+                }
+              }
+              //
+              throw new RuntimeException("Received no outcome from server");
+            }));
+  }
+
+  private void pollWorkflowUpdateHelper(
```
cretz, line 423: Just want to note I reviewed this carefully and while it doesn't exactly match what waiting for history close event is doing (which is synchronous, passes timeout/unit to timeout exception, etc), I think what is here is good.

cretz (commented):
Looks good, only minor things

temporal-sdk/src/main/java/io/temporal/client/UpdateHandle.java
Quinn-With-Two-Ns: Yes when the server supports that we can have another method that waits on accepted.

temporal-sdk/src/main/java/io/temporal/client/UpdateHandle.java
```diff
@@ -51,4 +52,14 @@
    * @return future completed with the result of the update or an exception
    */
   CompletableFuture<T> getResultAsync();
+
+  /**
+   * Returns a {@link CompletableFuture} with the update workflow execution request result
+   * potentially waiting for the update to complete.
```
Sushisource, line 58:
```suggestion
   * Returns a {@link CompletableFuture} with the update workflow execution request result,
   * potentially waiting for the update to complete.
```

temporal-sdk/src/main/java/io/temporal/client/WorkflowStub.java
Sushisource: I +1 that. This name is a mouthful

temporal-sdk/src/main/java/io/temporal/client/WorkflowStub.java
```diff
@@ -145,22 +145,50 @@ <R> R update(
...
       String updateId,
       String firstExecutionRunId,
+      UpdateWorkflowExecutionLifecycleStage waitPolicy,
       Class<R> resultClass,
       Type resultType,
       Object... args);
 
+  /**
+   * Get an update handle to an update request previously started. Getting an update handle does not
+   * guarantee the update ID exists.
```
Sushisource, line 168:
```suggestion
   * Get an update handle to a previously started update request. Getting an update handle does not
   * guarantee the update ID exists.
```

temporal-sdk/src/main/java/io/temporal/client/WorkflowStub.java
```diff
@@ -145,22 +145,50 @@ <R> R update(
...
+   * @param resultClass class of the update return value.
+   * @param <R> type of the update return value.
+   * @return update handle that can be used to get the result of the update.
+   */
+  @Experimental
+  <R> UpdateHandle<R> getUpdateHandle(String updateId, Class<R> resultClass);
+
+  /**
+   * Get an update handle to an update request previously started.Getting an update handle does not
+   * guarantee the update ID exists.
```
Sushisource, line 180:
```suggestion
   * Get an update handle to a previously started update request. Getting an update handle does not
   * guarantee the update ID exists.
```

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
```diff
@@ -356,20 +349,111 @@ public <R> UpdateOutput<R> update(UpdateInput<R> input) {
...
+                        input.getUpdateName(),
+                        dataConverterWithWorkflowContext.failureToException(
+                            result.getOutcome().getFailure()));
+                  default:
+                    throw new RuntimeException(
+                        "Received unexpected outcome from poll update request: "
+                            + result.getOutcome().getValueCase());
+                }
+              }
+              //
```
Sushisource, line 418:
```suggestion
```

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
Quinn-With-Two-Ns: yeah I spent most of the time of this PR on this method and the retry logic. I can include timeout/unit to the timeout exception if the team think it is useful to have.

cretz: Nah, no need. I am good with this. And I checked deadline vs user given timeout, how it's implemented on Go, etc and all looks good.

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -2155,6 +2136,102 @@ public UpdateWorkflowExecutionResponse updateWorkflowExecution(
...
+            inflightUpdate.get().getAcceptance(),
+            inflightUpdate.get().getCompletion());
+      }
+      StateMachine<UpdateWorkflowExecutionData> acceptedUpdate = updates.get(updateId);
+      if (acceptedUpdate != null) {
+        return new UpdateHandle(
+            acceptedUpdate.getData().id,
+            acceptedUpdate.getData().acceptance,
+            acceptedUpdate.getData().complete);
+      }
```
Sushisource, line 2225: nit: Looks like this could be de-duped with the above fairly reasonably

Sushisource (approved):
Sweet

Nothing important, a few docstring improvements

temporal-sdk/src/main/java/io/temporal/client/WorkflowStub.java
Quinn-With-Two-Ns: I think the only other java approach this matches is what we do for workflow with the untyped stub? I strongly prefer the API of handle over stub personally so I would keep it as is.

temporal-sdk/src/main/java/io/temporal/client/WorkflowStubImpl.java
```diff
@@ -307,65 +305,108 @@ public <R> R update(
...
       String firstExecutionRunId,
+      UpdateWorkflowExecutionLifecycleStage waitPolicy,
       Class<R> resultClass,
       Type resultType,
       Object... args) {
     checkStarted();
     WorkflowExecution targetExecution = execution.get();
+    try {
+      if (updateId == null || updateId.isEmpty()) {
+        updateId = UUID.randomUUID().toString();
```
Quinn-With-Two-Ns, line 352: @mmcshane We should probably do this in Go as well, instead of letting the server generate an ID.

cretz: Hrmm, this is worth a discussion. What is the benefit of generating client side if the server side must return a result anyways before you can return a handle?

Quinn-With-Two-Ns: Generating server side can be dangerous if the client retries

cretz: Ok, I completely agree. I know other APIs have `request_id`, but IIRC we are intentionally choosing to use update ID as that ID here. Yes, server side we should probably require this and client side we should default it.

temporal-sdk/src/main/java/io/temporal/client/WorkflowStub.java
cretz: I agree, I think I like the handle approach (though probably less strongly than you). Just wanted to note that the other things that have start-then-complete semantics do use stubs (with shortcuts/helpers as needed).

temporal-sdk/src/main/java/io/temporal/client/WorkflowStubImpl.java
```diff
@@ -307,65 +304,101 @@ public <R> R update(
...
+            .setWaitPolicy(UpdateWaitPolicy.ACCEPTED)
+            .setResultClass(resultClass)
+            .setResultType(resultClass)
+            .build();
+
+    return startUpdate(options, args);
+  }
+
+  @Override
+  public <R> UpdateHandle<R> startUpdate(StartUpdateOptions<R> options, Object... args) {
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: I don't want to enumerate every possible type of override for update. `update` should not take `update` because it defeats the purpose of the API

temporal-sdk/src/main/java/io/temporal/client/WorkflowStubImpl.java
```diff
@@ -307,65 +304,101 @@ public <R> R update(
...
+      return startUpdate(options, args).getResultAsync().get();
+    } catch (InterruptedException e) {
+      throw new RuntimeException(e);
+    } catch (ExecutionException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  @Override
+  public <R> UpdateHandle<R> startUpdate(String updateName, Class<R> resultClass, Object... args) {
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: Yes, this is already implemented

temporal-sdk/src/main/java/io/temporal/client/WorkflowStubImpl.java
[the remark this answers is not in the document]
cretz: Two things:

First, we have `update(String updateName, String updateId, String firstExecutionRunId, Class<R> resultClass, Type resultType, Object... args)`. So can we at least make update and start update overloads consistent?

Second, do we want to encourage the use of update ID? If not, no problem, I agree the update-id overload can be removed. So you have two overloads for each.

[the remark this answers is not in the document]
Quinn-With-Two-Ns: >First, we have update(String updateName, String updateId, String firstExecutionRunId, Class<R> resultClass, Type resultType, Object... args). So can we at least make update and start update overloads consistent?

At this point in the update feature I don't want `update` to take `UpdateOptions` because `UpdateOptions` only are meant to apply for `StartUpdate`. I would encourage the use of `startUpdate` vs `update`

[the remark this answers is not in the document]
Quinn-With-Two-Ns: IN the future if we receive user request we can add more overloads for usability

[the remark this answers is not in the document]
cretz: Why are update options only meant for start update? Is it just because of wait policy? We can enforce that be unset for the execute one at runtime. We accept workflow options for start or execute workflow, and activity options for start and execute activity. Granted those are called `execute` and `executeAsync`. Maybe we want to change the naming here too to `update` and `updateAsync`?

We should not wait for the future for consistency I don't think, and we shouldn't let wait policy affect the consistency that much. We should simply let you execute or start update with options or with simpler shortcuts, nothing more complicated hopefully.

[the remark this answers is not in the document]
cretz: After off-PR discussion here, we can just remove the many-param overload for `update` and make that one sugar `update` be the only sync way and anyone that wants to do anything more advanced needs to use `startUpdate`.

cretz (approved):

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
```diff
@@ -356,20 +349,110 @@ public <R> UpdateOutput<R> update(UpdateInput<R> input) {
...
+        .whenComplete(
+            (r, e) -> {
+              if ((e instanceof StatusRuntimeException
+                      && ((StatusRuntimeException) e).getStatus().getCode()
+                          == Status.Code.DEADLINE_EXCEEDED)
+                  || pollTimeoutDeadline.isExpired()
+                  || (e == null && !r.hasOutcome())) {
+                // if the request has timed out, stop retrying
+                if (!deadline.isExpired()) {
+                  pollWorkflowUpdateHelper(resultCF, request, deadline);
```
mmcshane, line 439: Is `deadline` clamped somewhere? Obviously there's a stack depth concern here but let's say someone wanted to wait for 24h, with 1m polls we're only looking at something on the order of 1440 frames so ... maybe ok? I guess we're prevented from using an iterative approach because of whenComplete?

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
```diff
@@ -356,20 +349,110 @@ public <R> UpdateOutput<R> update(UpdateInput<R> input) {
...
+                        input.getUpdateName(),
+                        dataConverterWithWorkflowContext.failureToException(
+                            result.getOutcome().getFailure()));
+                  default:
+                    throw new RuntimeException(
+                        "Received unexpected outcome from poll update request: "
+                            + result.getOutcome().getValueCase());
+                }
+              }
+              throw new RuntimeException("Received no outcome from server");
```
mmcshane, line 418: If I'm understanding correctly, this RuntimeException would be if the future _didn't_ complete exceptionally but also _did_ return a null value for Outcome. So basically if the server broke contract, is that right?

mmcshane (approved):
Two notes but they're more questions for my own understanding than changes that are needed.

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
Quinn-With-Two-Ns: This is actually not recursive. `pollWorkflowUpdateHelper`  will return immediately , it just launches another poll request that will write to the result future.

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
Quinn-With-Two-Ns: Yes exactly

Issue #1005, opened 2022-01-25
Provide ActivityExecutionContext#isCancelled


Spikhalskiy:
**Is your feature request related to a problem? Please describe.**
Currently, the only way for an activity to know that it's canceled is heartbeating and getting a CancellationException from the heartbeat.
1. Not all activities need to heartbeat and making them do it just to check for cancelation is ugly.
2. Local Activities don't heartbeat, so now there is no way to notify them that they should cancel (for example, because there is a timeout). GoSDK has go context for it, Java has nothing. Using Interruption here is not right, because we need to distinguish it from a thread interruption during the shutdown to keep things clean. 

**Describe the solution you'd like**
ActivityExecutionContext#isCancelled method that activity can periodically check as an alternative to listening cancelations through heart beating. This solution will be close to the native Interruption design in Java: We will have an exception if Activity calls a heartbeat and a flag otherwise.

**Additional context**
This is related to #1004. There is no way to notify Local Activity about the cancelation.

Quinn-With-Two-Ns:
We should also add a more detailed method (name TBD) that returns an object that contains the reason/reasons for the activity being cancelled.

Quinn-With-Two-Ns:
Resolved by the addition on `CancellationToken`

PR #1771, opened 2023-05-21
Throw on heartbeat when activity task times out

### LINKED ISSUE #1770: Timeouts should be detected locally during activity execution

Activity timeouts are only detectable by activity code when a heartbeat attempt reaches the Temporal server. Heartbeat throttling can significantly delay timeout detection because it prevents or delays some heartbeats from reaching the server. This in turn causes activity attempts to continue running even when any result will be rejected.

While this exact model isn't directly applicable in Java, the Go SDK sets a context deadline when executing activities in `internal.activity.WithActivityTask` ([source](https://github.com/temporalio/sdk-go/blob/master/internal/activity.go#L251)) that allows for local detection of start-to-close and schedule-to-close timeouts being reached.

Instead, I propose having the `heartbeat` method on the activity context throw a subclass of `ActivityCompletionException` when these timeouts are violated. All the information for calculating this is already present on the activity task (start-to-close timeout, schedule-to-close timeout, task attempt start time, task first scheduled time). This approach is consistent with other exceptions already thrown by this method for worker shutdown and activity cancellation.

*Interrupt the activity thread.* While it might be more thorough, this would be a much more invasive change to the existing activity programming model. Developers are already encouraged to catch `ActivityCompletionException` upon heartbeating to detect a need to exit activity execution early.

*Add methods to the activity context for timeout detection.* This could be done in addition to or instead of the above. An exception does not have to be thrown, but is consistent with existing behavior of `heartbeat` around other activity completion actions.

**Additional context**

When I chatted with Maxim a while back about this on the Temporal Community Slack, he was supportive. Unfortunately, the thread appears to have been retentioned away. I don't believe an issue was ever opened for this.

jeffschoner:
## What was changed

- Adds a new `ActivityTaskTimedOutException` subclassed from `ActivityCompletionException`
- This exception is now thrown on heartbeating when the activity task has exceeded the start-to-close or schedule-to-close timeout
- Adds a `getStartedTimestamp`  method to `ActivityInfo` for the `started_time` field from `PollActivityTaskQueueResponse`. This is needed to calculate if the start-to-close timeout has been exceeded.
- Added two new test cases in `ActivityTimeoutTest` and updated an existing test case in `WorkflowClosedRunningActivityTest`

## Why?

Because of heartbeat throttling, an activity task can keep running even though it has exceeded its timeouts and is heartbeating. This change moves these failures up in time, for more efficient activity processing and to reduce cases where multiple attempts of the same activity are executing simultaneously.

## Checklist

1. Closes #1770 

2. How was this tested:

Ran above unit tests using gradle:
```
./gradlew :temporal-sdk:test --tests "io.temporal.workflow.activityTests.ActivityTimeoutTest"
./gradlew :temporal-sdk:test --tests "io.temporal.workflow.activityTests.cancellation.WorkflowClosedRunningActivityTest.activitySeesActivityNotExistException[*]"

```

3. Any docs updates needed?

https://docs.temporal.io/dev-guide/java/features#activity-heartbeats

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/activity/ActivityInfo.java
+++ b/temporal-sdk/src/main/java/io/temporal/activity/ActivityInfo.java
@@ -67,6 +67,13 @@ public interface ActivityInfo {
   long getScheduledTimestamp();
 
   /**
+   * Time when the activity task (this attempt) was started
+   *
+   * @return Timestamp in milliseconds (UNIX Epoch time)
+   */
+  long getStartedTimestamp();
+
+  /**
    * Time when the Activity Task (current attempt) was scheduled by the Temporal Server.
    *
    * @return Timestamp in milliseconds (UNIX Epoch time)
--- a/temporal-sdk/src/main/java/io/temporal/client/ActivityTaskTimedOutException.java
+++ b/temporal-sdk/src/main/java/io/temporal/client/ActivityTaskTimedOutException.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright (C) 2022 Temporal Technologies, Inc. All Rights Reserved.
+ *
+ * Copyright (C) 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Modifications copyright (C) 2017 Uber Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this material except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.temporal.client;
+
+import io.temporal.activity.ActivityInfo;
+import io.temporal.api.enums.v1.TimeoutType;
+
+/**
+ * Indicates that the activity task has timed out while executing.
+ *
+ * <p>Catching this exception directly is discouraged and catching the parent class {@link
+ * ActivityCompletionException} is recommended instead.<br>
+ */
+public final class ActivityTaskTimedOutException extends ActivityCompletionException {
+  private final TimeoutType timeoutType;
+  private final long timeoutDeadline;
+
+  public ActivityTaskTimedOutException(
+      ActivityInfo info, TimeoutType timeoutType, long timeoutDeadline) {
+    super(info);
+    this.timeoutType = timeoutType;
+    this.timeoutDeadline = timeoutDeadline;
+  }
+
+  /**
+   * Gets the type of timeout that was exceeded during activity execution
+   *
+   * @return only TIMEOUT_TYPE_START_TO_CLOSE or TIMEOUT_TYPE_SCHEDULE_TO_CLOSE
+   */
+  public TimeoutType getTimeoutType() {
+    return timeoutType;
+  }
+
+  /**
+   * Gets the timeout deadline that was exceeded during activity execution
+   *
+   * @return Timestamp in milliseconds (UNIX Epoch time)
+   */
+  public long getTimeoutDeadline() {
+    return timeoutDeadline;
+  }
+}
--- a/temporal-sdk/src/main/java/io/temporal/internal/activity/ActivityInfoImpl.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/activity/ActivityInfoImpl.java
@@ -88,6 +88,11 @@ final class ActivityInfoImpl implements ActivityInfoInternal {
   }
 
   @Override
+  public long getStartedTimestamp() {
+    return Timestamps.toMillis(response.getStartedTime());
+  }
+
+  @Override
   public Duration getScheduleToCloseTimeout() {
     return ProtobufTimeUtils.toJavaDuration(response.getScheduleToCloseTimeout());
   }
--- a/temporal-sdk/src/main/java/io/temporal/internal/activity/HeartbeatContextImpl.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/activity/HeartbeatContextImpl.java
@@ -26,6 +26,7 @@ import io.grpc.StatusRuntimeException;
 import io.temporal.activity.ActivityExecutionContext;
 import io.temporal.activity.ActivityInfo;
 import io.temporal.api.common.v1.Payloads;
+import io.temporal.api.enums.v1.TimeoutType;
 import io.temporal.api.workflowservice.v1.RecordActivityTaskHeartbeatResponse;
 import io.temporal.client.*;
 import io.temporal.common.converter.DataConverter;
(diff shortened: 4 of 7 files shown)


cretz:
Hrmm. I don't think this is a custom exception type, I think this is just a cancellation. Today timeout results in cancellation to the activity, correct? We can't break people's code that have that existing expectation, e.g. `catch (CanceledFailure`.

Also, @Quinn-With-Two-Ns, I am curious if Go and Core do client-side preempted timeouts. Does that apply for heartbeat timeout too? I think they _may_. But my initial thought is that there is existing behavior that, besides worker shut down, a Java activity can't receive a cancel outside of the heartbeat throttle window and I am unsure it's worth changing that for everyone.

If you rely on more precise activity timeouts, I think you should maintain that timer in the activity yourself.

jeffschoner:
> Hrmm. I don't think this is a custom exception type, I think this is just a cancellation. Today timeout results in cancellation to the activity, correct? We can't break people's code that have that existing expectation, e.g. `catch (CanceledFailure`.

As far as I can tell today, this becomes `ActivityNotExistsException` when a `NOT_FOUND` status is returned from Temporal server, indicating that the activity task's token cannot be found since it's no longer valid. A "not found" message here is confusing because it does not clearly communicate the underlying cause of the failure. It makes it look like the activity got lost somehow by Temporal server. Moreover, this exception is only thrown when the heartbeat reached the server, which due to heartbeat throttling can be unpredictable.

I haven't been able to reproduce this at an SDK level, but perhaps cancellation occurs in certain other timeout situations? In that case `ActivityCanceledException` would be thrown when the heartbeat is successful and the `cancel_requested` bit is set on the GRPC response. Source is all [here](https://github.com/temporalio/sdk-java/blob/1be0cee5837037610179675c028145691c5df456/temporal-sdk/src/main/java/io/temporal/internal/activity/HeartbeatContextImpl.java#L209).

Either way, these are all subclasses of `ActivityCompletionException`, and do not have `CanceledFailure` in their hierarchy. Existing guidance on `.heartbeat()` suggests catching the superclass anyways. This new exception type also extends that same superclass. I suppose this could be a subclass of `ActivityNotExistsException` to preserve existing behavior. That seems odd, but maybe is worth the tradeoff for consistency?

As for handling this in activity code (or an interceptor), I could certainly do that, but the activity task coming from Temporal server has all the information to calculate and enforce these deadlines accurately. This change will make timeout failures more consistent. Due to heartbeat throttling, you get into the situation where activity code continues to run even when it's already 100% known that any heartbeat or success/failure will be rejected by the server.

General guidance seems to be to heartbeat regularly in order to detect when your activity needs to exit early. From a user perspective, it is clear when your code could be pre-empted and if you don't try/catch anything, you're going to get the right before (in this case, the activity will retry on start-to-close and simply exit on schedule-to-close). This change works well within that existing behavior. I do agree that starting to raise `InterruptedException` is a much more severe and risky change because it would change behavior substantially and because it's easy to mishandle `InterruptedException` in code. But this change doesn't go down that path.

cretz:
> I could certainly do that, but the activity task coming from Temporal server has all the information to calculate and enforce these deadlines accurately.

This timeout information is made available on `ActivityInfo`, and you can probably write an interceptor that does this for you.

Besides the obvious fear of changing everyone's behavior to client-side timeouts by default, there are two other concerns: 1) client side cannot report timeout to server side so the workflow-side error may change if client side happens first (or client side has to just not send anything after timeout), and 2) there has been talk of allowing activity option updates server side and/or after the activity starts, which can include timeouts.

cretz:
I think if we make start and schedule time visible, then this is just a simple matter of subtraction from timeouts. I hesitate to make the helper visible because I am not sure we want to encourage users writing their own timeout checks that race with the server (it can be racy whether someone failing based on this timeout responds to the server in time with their failure before the server hits the same timeout).

Quinn-With-Two-Ns:
hm so in `Go` [we do cancel the context based on the activity deadline ](https://github.com/temporalio/sdk-go/blob/55ef398fe18e09855212cbcd3fbb76b3ddf1ed98/internal/internal_task_handlers.go#L1943) that we do derive from the [timeout](https://github.com/temporalio/sdk-go/blob/55ef398fe18e09855212cbcd3fbb76b3ddf1ed98/internal/activity.go#L251) in `Go` we effectively do have a `hasTaskTimedOut` by checking if the deadline is expired. If the context is deadline is expired in `Go` I believe the heartbeat will fail.


I am not sure we should be returning a different exception on a "local" timeout. We document that `heartbeat` returns `ActivityCompletionException` on any reason to cancel the activity. Not sure what the "racey" concerns are @cretz ? Doesn't seem any worse then the current race with `heartbeat` throttling and timeouts?

Quinn-With-Two-Ns:
I'll note we also have [this](https://github.com/temporalio/sdk-java/issues/1005) issue open

cretz:
> Not sure what the "racey" concerns are @cretz ? Doesn't seem any worse then the current race with heartbeat throttling and timeouts?

What are you gonna return to the server on activity timeout from the client side? If nothing, no problem. I can get behind adding this if it's the case in Go. Would like to know what Core does too.

jeffschoner:
> What are you gonna return to the server on activity timeout from the client side? If nothing, no problem.

Let me look into this some more too. In a similar change in temporal-ruby, I just don't send an activity task response back because I know it's already timed out (or about to) on the server anyways and is likely to be rejected. If there's a small amount of clock skew, this should get both the worker and server into the same state.

Even without that, if it wins the race with the server, the attempt will be failed, and the server will retry the activity according to the retry policy. As far as I can tell, it's the same outcome as the activity task timing out on the server, with some differences possibly in what gets logged and emitted to metrics. Having a message about a timeout being logged is better than not found (the current experience).

Quinn-With-Two-Ns:
In Go we should return [the cancelled error ](https://github.com/temporalio/sdk-go/blob/55ef398fe18e09855212cbcd3fbb76b3ddf1ed98/internal/internal_task_handlers.go#L1951) I believe

cretz:
> Even without that, if it wins the race with the server, the attempt will be failed, and the server will retry the activity according to the retry policy.

@jeffschoner - This is not necessarily true. For instance a schedule to close timeout is not retryable, but if client side wins then it may retry only to fail immediately but potentially be another (cloud-chargeable) activity attempt. Also, different error types of a retry policy can be marked non-retryable, so if it's racy on what that error type is, then different retry policy may apply. And what gets emitted to metrics does matter. And in general what the activity failure event looks like in history matters too (it says what timeout type it hit, etc).

If we aren't ok with allowing the server to decide timeouts (which is the safest, even if Go didn't, need to check core), then we should probably just consider dropping the task. Otherwise the user can fail themselves if they want to do client-side failure for any reason, including timeout.

We have future plans of delivering cancellation unrelated to heartbeating which makes this unnecessary (there are talks of bidirectional gRPC streams here).

jeffschoner:
> We have future plans of delivering cancellation unrelated to heartbeating which makes this unnecessary (there are talks of bidirectional gRPC streams here).

I was talking with @antstorm elsewhere about the new approach to cancellation APIs in the Rust core-based SDKs too. It sounds like there are some shorter and longer term plans to take some better approaches overall.

I'll send out a new PR to add start time to `ActivityInfo`. The scheduled time is already present. With that, I can implement this as an interceptor included in all workers to get where I need to go in the shorter term. I'm willing to take on more consistent timeout failures over subtle retry behavior changes, but I understand that may not be more broadly acceptable.

jeffschoner:
Looks like @Quinn-With-Two-Ns got this in in #1798. Thanks!

PR #1777, opened 2023-06-03
Correctly resolve null workflow locals to Optionals

### LINKED ISSUE #1761: Calling WorkflowLocal.set(null) results in the next call to WorkflowLocal.get() to throw a NullPointerException
## Expected Behavior
I expect this:
``` java
public static WorkflowLocal<String> wl = WorkflowLocal.withInitial(() -> "value");
...
wl.get();
wl.set(null);
String s = wl.get();
```
To result in the String `s` being initialized to `null`.

## Actual Behavior
The call to `wl.get()` throws an NullPointerException:
```
Caused by: java.lang.NullPointerException
	at java.util.Objects.requireNonNull(Objects.java:221) ~[?:?]
	at java.util.Optional.<init>(Optional.java:107) ~[?:?]
	at java.util.Optional.of(Optional.java:120) ~[?:?]
	at io.temporal.internal.sync.DeterministicRunnerImpl.getRunnerLocal(DeterministicRunnerImpl.java:597) ~[temporal-sdk-1.19.1.jar:?]
	at io.temporal.internal.sync.RunnerLocalInternal.get(RunnerLocalInternal.java:30) ~[temporal-sdk-1.19.1.jar:?]
	at io.temporal.workflow.WorkflowLocal.get(WorkflowLocal.java:68) ~[temporal-sdk-1.19.1.jar:?]
        at <line that called wl.get()>
```

Note that this works fine:
``` java
public static WorkflowLocal<String> wl = WorkflowLocal.withInitial(() -> null);
...
wl.get();
```
The problem only occurs once you've called `set(null)`.

See the code snippet above. The root cause of the bug is that `DeterministicRunnerImpl.getRunnerLocal.getRunnerLocal` returns an `Optional<T>`, which will return `Optional.empty()` if `WorkflowLocal.set` has not been called yet, or use `Optional.of` to wrap whatever was passed to `WorkflowLocal.set` in an `Optional`. When `set(null)` is used, the next `get()` results in a call to `Optional.of(null)`, which throws the NPE. I think that the `Optional<T>` return type just doesn't work here, since this method needs to differentiate between two states: `set` has not been called, or `set` has been called, inclusive of `set` being called with `null`, which `Optional` can't handle.

I also noticed that when `get()` is called for the first time, without `set` being called first, the value returned by the `withInitial` callback is not actually stored anywhere. So repeated calls to `get()` result in repeated calls to the `withInitial` callback. This is surprising behavior to me - not sure if it's intended or not.

## Specifications

  - Version: Java SDK 1.19.1.
  - Platform: Linux

bbqbaron:
Java optional can't itself contain null, so use `ofNullable` to flatten Optionality when retrieving null values from the map of runner locals. 

Closes #1761

## What was changed
Use `ofNullable` to construct Optionals from values that may be null.

## Why?
There are three cases when getting values from a map where nullity matters:
* it's absent
* it's null
* it's non-null

The code handled the first with `empty` and tried to combine the second and third with `of`, (probably) not realizing `of` would not handle the second case.

### But I Don't Want to Call the Supplier

As #1761 points out, calling `get` over and over with no `set` results in repeated supplier calls. That means the supplier must be cheap and ideally pure, or at least idempotent. To me, this means it's fine to call the supplier whenever the locals have a null value. We just can't crash. If the user wants to differentiate between "affirmative absence" and nullity, they can use whatever type they want.

If we want the supplier to be called at most once regardless of what it returns, we're making what could be seen as an API change, albeit a small one. This is just fixing an immediate bug. Does that make sense?

If we want to go that way, I have a [more ambitious](https://github.com/bbqbaron/sdk-java/compare/master...bbqbaron:sdk-java:issue-1761-supplier-reuse?expand=1) branch that tries to disambiguate the cases and invoke the supplier only when needed.

## Checklist

1. Closes #1761

2. How was this tested:

Added a unit test. This code seemed to be untested, so I wasn't sure if there was a preferred/better idiom, but it was incredibly cheap to write, so I did it.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/DeterministicRunnerImpl.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/DeterministicRunnerImpl.java
@@ -594,7 +594,7 @@ class DeterministicRunnerImpl implements DeterministicRunner {
     if (!runnerLocalMap.containsKey(key)) {
       return Optional.empty();
     }
-    return Optional.of((T) runnerLocalMap.get(key));
+    return Optional.ofNullable((T) runnerLocalMap.get(key));
   }
 
   <T> void setRunnerLocal(RunnerLocalInternal<T> key, T value) {
--- a/temporal-sdk/src/test/java/io/temporal/internal/sync/DeterministicRunnerTest.java
+++ b/temporal-sdk/src/test/java/io/temporal/internal/sync/DeterministicRunnerTest.java
@@ -890,4 +890,14 @@ public class DeterministicRunnerTest {
     d.close();
     assertTrue("Close should return only after the thread finished", threadFinished.get());
   }
+
+  @Test
+  public void testGetRunnerLocalOfNull() {
+    DeterministicRunnerImpl d =
+        new DeterministicRunnerImpl(
+            threadPool::submit, DummySyncWorkflowContext.newDummySyncWorkflowContext(), () -> {});
+    RunnerLocalInternal<String> runnerLocalInternal = new RunnerLocalInternal<>();
+    d.setRunnerLocal(runnerLocalInternal, null);
+    d.getRunnerLocal(runnerLocalInternal);
+  }
 }


Quinn-With-Two-Ns:
Using `ofNullable` does not really fix the issue. If you wrote tests for `WorkflowLocal` with a supplier you'd see that it will end up calling the callback passed to `withInitial` again if you return an empty Optional. The problem is `Optional` only can represent two states, not set or has a (not null) value. As you pointed out need to represent 3 states not set, null or a (not null) value

bbqbaron:
> Using `ofNullable` does not really fix the issue. If you wrote tests for `WorkflowLocal` with a supplier you'd see that it will end up calling the callback passed to `withInitial` again if you return an empty Optional. The problem is `Optional` only can represent two states, not set or has a (not null) value. As you pointed out need to represent 3 states not set, null or a (not null) value

Right, that's what my other more invasive branch does. It sounds like you don't just want to fix the NPE, but regard the re-calling of the initial supplier as a bug. I'll open a second PR from my other branch shortly.

bbqbaron:
> Using `ofNullable` does not really fix the issue. If you wrote tests for `WorkflowLocal` with a supplier you'd see that it will end up calling the callback passed to `withInitial` again if you return an empty Optional. The problem is `Optional` only can represent two states, not set or has a (not null) value. As you pointed out need to represent 3 states not set, null or a (not null) value

https://github.com/temporalio/sdk-java/pull/1779

bbqbaron:
Superseded by #1779

PR #1779, opened 2023-06-05
Issue 1761 fix supplier reuse

### LINKED ISSUE #1761: Calling WorkflowLocal.set(null) results in the next call to WorkflowLocal.get() to throw a NullPointerException

bbqbaron:
See https://github.com/temporalio/sdk-java/pull/1777 for background/most thoughts. This is the more ambitious fix that tries to resolve the "not set"/"null" ambiguity.

Closes #1761 

Nested `Optional` didn't feel great, but I preferred it over the other solutions like "an `Optional` that is also nullable".

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowThreadImpl.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowThreadImpl.java
@@ -405,7 +405,7 @@ class WorkflowThreadImpl implements WorkflowThread {
   }
 
   /**
-   * Retrieve data from runner locals. Returns 1. not found (an empty Optional) 2. found but null
+   * Retrieve data from thread locals. Returns 1. not found (an empty Optional) 2. found but null
    * (an Optional of an empty Optional) 3. found and non-null (an Optional of an Optional of a
    * value) The type nesting is because Java Optionals cannot understand "Some null" vs "None",
    * which is exactly what we need here.
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowThreadImpl.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowThreadImpl.java
@@ -407,7 +407,7 @@ class WorkflowThreadImpl implements WorkflowThread {
   /**
    * Retrieve data from thread locals. Returns 1. not found (an empty Optional) 2. found but null
    * (an Optional of an empty Optional) 3. found and non-null (an Optional of an Optional of a
-   * value) The type nesting is because Java Optionals cannot understand "Some null" vs "None",
+   * value). The type nesting is because Java Optionals cannot understand "Some null" vs "None",
    * which is exactly what we need here.
    *
    * @param key
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowThreadImpl.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowThreadImpl.java
@@ -404,13 +404,22 @@ class WorkflowThreadImpl implements WorkflowThread {
     threadLocalMap.put(key, value);
   }
 
+  /**
+   * Retrieve data from runner locals. Returns 1. not found (an empty Optional) 2. found but null
+   * (an Optional of an empty Optional) 3. found and non-null (an Optional of an Optional of a
+   * value) The type nesting is because Java Optionals cannot understand "Some null" vs "None",
+   * which is exactly what we need here.
+   *
+   * @param key
+   * @return one of three cases
+   * @param <T>
+   */
   @SuppressWarnings("unchecked")
-  @Override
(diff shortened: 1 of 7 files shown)


Quinn-With-Two-Ns:
Can we add some tests for `WorkflowLocal`? Just to verify the original issue is resolved and we don't break any existing behaviour?

bbqbaron:
Added a test that I think shows the various required behaviors:

* null initial values are fine
* the supplier is called once
* `set` is reflected in a subsequent `get`

bbqbaron:
Learning about the SDK as I go, it looks like maybe `WorkflowThreadLocal` could have the same problem! Want me to throw that into this PR? Not sure what your PR-scoping preferences are.

Quinn-With-Two-Ns:
Yeah I think it will as well, If you don't mind feel free to include a fix in this PR.

temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowThreadImpl.java
```diff
@@ -404,13 +404,22 @@ public <T> void setThreadLocal(WorkflowThreadLocalInternal<T> key, T value) {
     threadLocalMap.put(key, value);
   }
 
+  /**
+   * Retrieve data from runner locals. Returns 1. not found (an empty Optional) 2. found but null
```
bbqbaron, line 408:
```suggestion
   * Retrieve data from thread locals. Returns 1. not found (an empty Optional) 2. found but null
```

bbqbaron:
Added. It ends up being basically a copypaste, but only two instances of the same behavior aren't an abstraction etc etc.

Quinn-With-Two-Ns:
I believe CI is failing here because your branch does not have tags. I believe you can run `git fetch --tags --all`

bbqbaron:
> I believe CI is failing here because your branch does not have tags. I believe you can run `git fetch --tags --all`

Sorry @Quinn-With-Two-Ns, despite having used git for a while, I'm not sure what to do here. I can fetch tags and push them to my fork, but I'm not sure how that is affecting tests, and I'm not sure how to affect a change on this PR. Does my branch need to contain some new tag or something?

Quinn-With-Two-Ns:
Sorry it is not your fault, most repositories testing don't require git tags to work, but this one does for now. The reason your branch needs to contain tags is the `gradle` files use git to lookup the tags to figure out what current version is. Just pushing tags to your branch should fix the issue

https://github.com/temporalio/sdk-java/blob/18120b164ac5fdbaac894b84353552671b9a0ec9/gradle/versioning.gradle#L5

bbqbaron:
Ah, there we go; tag-pushing doesn't trigger a CI build. Had to push an inconsequential change to source.

Quinn-With-Two-Ns (approved):

PR #1780, opened 2023-06-06
Add update protocol commands

Quinn-With-Two-Ns:
Add support for `ProtocolMessageCommands` to the Java SDK. `UpdateProtocolStateMachine` will now generate commands and event on update accepted and completion to verify determinism of update requests.


temporal-sdk/src/main/java/io/temporal/internal/statemachines/UpdateProtocolStateMachine.java
```diff
@@ -53,7 +57,14 @@ enum State {
     NEW,
     REQUEST_INITIATED,
     ACCEPTED,
+    ACCEPTED_COMMAND_CREATED,
```
Quinn-With-Two-Ns, line 60: I'll note I didn't remove the look ahead for protocol messages. It **may** be possible to with message commands, but it would cause the state machine to have two totally different paths depending on if it is replaying or not and it would need different special logic inside the `WorkflowStateMachine` to handle determinism and the event loop.

temporal-sdk/src/main/java/io/temporal/internal/statemachines/UpdateProtocolStateMachine.java
```diff
@@ -165,7 +240,8 @@ public void complete(Optional<Payloads> payload, Failure failure) {
 
     Response outcomeResponse =
         Response.newBuilder().setOutcome(outcome).setMeta(initialRequest.getMeta()).build();
-    // TODO send ProtocolMessage command when server supports
+
+    messageId = requestMsgId + "/complete";
     sendHandle.apply(
         Message.newBuilder()
             .setId(requestMsgId + "/complete")
```
Quinn-With-Two-Ns, line 247: typo, I pushed a fix but it isn't showing up here for some reason

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
```diff
@@ -952,6 +939,42 @@ public Functions.Proc scheduleLocalActivityTask(
...
+      ProtocolMessageCommandAttributes commandAttributes =
+          command.getProtocolMessageCommandAttributes();
+      switch (event.getEventType()) {
+        case EVENT_TYPE_WORKFLOW_EXECUTION_UPDATE_ACCEPTED:
+          assertMatch(
+              command,
+              event,
+              "messageType",
+              true,
+              commandAttributes.getMessageId().endsWith("accept"));
```
mmcshane, line 955: I don't love it but I get it. Is it possible to add the message set to the args to validateCommand? Not blocking but it would be better to test the actual message type.

mmcshane (approved):
Just the one possible improvement. Which might not be possible.

temporal-sdk/src/main/java/io/temporal/internal/statemachines/UpdateProtocolStateMachine.java
```diff
@@ -67,19 +78,74 @@ enum State {
...
+              // Handle the validation failure case
+              .add(State.REQUEST_INITIATED, ExplicitEvent.REJECT, State.COMPLETED_COMMAND_RECORDED)
+              // Handle an edge case when the update handle completes immediately. The state machine
+              // should then expect
+              // to see two protocol command messages back to back then two update events.
+              .add(
+                  State.ACCEPTED,
+                  ExplicitEvent.COMPLETE,
+                  State.COMPLETED_IMMEDIATELY,
+                  UpdateProtocolStateMachine::sendCommandMessage)
```
Sushisource, line 127: IMO these ones that are doing send command message would be more clear as having the method that does the explicit event do only that, and then having different implementations of `sendCommandMessage` for the different outcomes.

This way, you don't set the `messageId` class field and then rely on it at a distance a moment later, it can avoid being a class field entirely.

Not blocking by any means and I get the Java state machines do this pattern not infrequently but this approach might be a little more clear / safer.

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
Sushisource: +1 to that

Sushisource (approved):

temporal-sdk/src/main/java/io/temporal/internal/statemachines/UpdateProtocolStateMachine.java
Quinn-With-Two-Ns: yeah totally agree it isn't unusual, I just don't think it actually is better since patch handing SM keeps track of replay in the state machine and still needs look ahead.

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
Quinn-With-Two-Ns: I'll create a follow up issue to improve

PR #1782, opened 2023-06-07
Typed search attributes

cretz:
## What was changed

Implementation of [typed search attributes](https://github.com/temporalio/proposals/blob/master/sdk-typed-search-attributes.md). Specifically:

* Marked all existing user-facing search attribute usage as deprected
* Added `SearchAttributes` (with builder) and `SearchAttributeKey` and `SearchAttributeUpdate` to `io.temporal.common`
* Added getters, setters, and upserters for typed search attributes

The tests were just written to have the same coverage as existing tests, but are admittedly probably not enough and more could be written for edge cases (e.g. using old API to set multiple int attributes and trying to use them with new API which only allows one).


temporal-sdk/src/main/java/io/temporal/internal/statemachines/ActivityStateMachine.puml
cretz, line 1: This was a mistake (auto done on build, but meant to not include it or any other puml file), will commit fix early tomorrow to put it back the way it was

temporal-sdk/src/main/java/io/temporal/common/SearchAttributes.java
```diff
@@ -0,0 +1,129 @@
...
+    private Builder(SearchAttributes copyFrom) {
+      untypedValues = new TreeMap<>(copyFrom.untypedValues);
+    }
+
+    /** Set a search attribute key and typed value. */
+    public <T> Builder set(SearchAttributeKey<T> key, @Nonnull T value) {
+      Objects.requireNonNull(untypedValues, "Collection already built");
+      Objects.requireNonNull(value, "Value cannot be null");
+      // TODO(cretz): Prevent duplicate key name?
+      untypedValues.put(key, value);
```
Sushisource, line 63: IMO override and mention it in the docstring. Mildly annoying to have to strip all-but-last if you want to map a series of changes to the final upsert, which is probably a common case.

temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflowContext.java
```diff
@@ -723,7 +725,14 @@ private StartChildWorkflowExecutionParameters createChildWorkflowParameters(
 
     Map<String, Object> searchAttributes = options.getSearchAttributes();
     if (searchAttributes != null && !searchAttributes.isEmpty()) {
+      if (options.getTypedSearchAttributes() != null) {
+        throw new IllegalArgumentException(
+            "Cannot have typed search attributes and search attributes");
```
Sushisource, line 730: Just a touch clearer -- applies in a few places
```suggestion
            "Cannot have both typed search attributes and search attributes");
```

Sushisource (approved):
This looks really good to me, big improvement.

temporal-sdk/src/main/java/io/temporal/internal/common/SearchAttributePayloadConverter.java
```diff
@@ -46,6 +47,84 @@ final class SearchAttributePayloadConverter {
...
+        break;
+      default:
+        log.warn(
+            "[BUG] Unrecognized indexed value type {} on search attribute key {}", indexType, name);
+        return;
+    }
+
+    // Attempt conversion to key type or if not keyword list, single-value list and extract out
+    try {
+      Object value =
```
bergundy, line 95: What happens if there are multiple values in the payload or the payload is an array with a single item and the typed attribute is not a list?
Seems like we should log a warning and (possibly) put the first value from the array in the attribute, at least for single item arrays.
In any case, just verifying this won't fail WFTs.
Also acceptable if we warn and users have to use untyped search attributes to access the values.

temporal-sdk/src/main/java/io/temporal/common/SearchAttributes.java
cretz: I concur, this jives with `put` (and probably don't need to even mention it since "set" implies it as opposed to "add")

temporal-sdk/src/main/java/io/temporal/internal/common/SearchAttributePayloadConverter.java
bergundy: > It will, by intention. If you have chosen to move to typed search attributes, you have chosen the strict guarantees that come with it (and that come with our new SA validations we do now). It is only possible to violate these if you used non-typed search attribute API (and advanced self-hosted validation IIUC) which I don't believe needs to lead to silent truncate behavior.

What if a client from another language or SDK version starts a workflow with extra values in a keyword array?
The server accepts this but workflows will get stuck.

temporal-sdk/src/main/java/io/temporal/internal/common/SearchAttributePayloadConverter.java
```diff
@@ -46,6 +47,84 @@ final class SearchAttributePayloadConverter {
...
+      case INDEXED_VALUE_TYPE_DOUBLE:
+        key = SearchAttributeKey.forDouble(name);
+        break;
+      case INDEXED_VALUE_TYPE_BOOL:
+        key = SearchAttributeKey.forBoolean(name);
+        break;
+      case INDEXED_VALUE_TYPE_DATETIME:
+        key = SearchAttributeKey.forOffsetDateTime(name);
+        break;
+      case INDEXED_VALUE_TYPE_KEYWORD_LIST:
```
Quinn-With-Two-Ns, line 84: Before this PR `INDEXED_VALUE_TYPE_KEYWORD_LIST` didn't work in Java. I am not totally confident this fixes it. Can we add tests for the list cases? Maybe you did add them and I just missed them

temporal-sdk/src/main/java/io/temporal/internal/common/SearchAttributePayloadConverter.java
cretz: > What if a client from another language or SDK version starts a workflow with extra values in a keyword array?

They chose to mix non-typed-search-attribute API usage with strict typed-search-attribute usage. Yes I can understand the concern about half of a user's code using non-strict and the other half using strict, but we want typed search attributes to be strict here. Untyped is not "all value types", untyped is "all values".

temporal-sdk/src/main/java/io/temporal/internal/common/SearchAttributePayloadConverter.java
cretz: Yes, I need to add those for sure

temporal-sdk/src/main/java/io/temporal/internal/common/SearchAttributePayloadConverter.java
bergundy: Okay, I buy that, we'll just have to explain this limitation to users and how to fix get their workflows "unstuck".

temporal-sdk/src/main/java/io/temporal/internal/statemachines/ActivityStateMachine.puml
cretz: Fixed

temporal-sdk/src/main/java/io/temporal/common/SearchAttributes.java
cretz: Removed TODO update doc

temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflowContext.java
cretz: Updated

temporal-sdk/src/main/java/io/temporal/internal/common/SearchAttributePayloadConverter.java
cretz: :+1: Limitation accepted for using deprecated API

temporal-sdk/src/main/java/io/temporal/internal/common/SearchAttributePayloadConverter.java
cretz: Added tests here. You were correct, there was a problem decoding side on Java test server which was fixed.

Quinn-With-Two-Ns (approved):

PR #1784, opened 2023-06-12
Treat UpdateWorkflowExecution as a long poll

Quinn-With-Two-Ns:
Treat `UpdateWorkflowExecution` as a long poll.

See also https://github.com/temporalio/sdk-go/pull/1123

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
@@ -49,6 +49,7 @@ import org.slf4j.LoggerFactory;
 
 public class RootWorkflowClientInvoker implements WorkflowClientCallsInterceptor {
   private static final Logger log = LoggerFactory.getLogger(RootWorkflowClientInvoker.class);
+  private static final long POLL_UPDATE_TIMEOUT_S = 60L;
 
   private final GenericWorkflowClient genericClient;
   private final WorkflowClientOptions clientOptions;
@@ -324,7 +325,9 @@ public class RootWorkflowClientInvoker implements WorkflowClientCallsInterceptor
             .setFirstExecutionRunId(input.getFirstExecutionRunId())
             .setRequest(request)
             .build();
-    UpdateWorkflowExecutionResponse result = genericClient.update(updateRequest);
+    Deadline pollTimeoutDeadline = Deadline.after(POLL_UPDATE_TIMEOUT_S, TimeUnit.SECONDS);
+    UpdateWorkflowExecutionResponse result =
+        genericClient.update(updateRequest, pollTimeoutDeadline);
 
     if (result.hasOutcome()) {
       switch (result.getOutcome().getValueCase()) {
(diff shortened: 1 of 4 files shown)


temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
```diff
@@ -49,6 +49,7 @@
 
 public class RootWorkflowClientInvoker implements WorkflowClientCallsInterceptor {
   private static final Logger log = LoggerFactory.getLogger(RootWorkflowClientInvoker.class);
+  private static final long POLL_UPDATE_TIMEOUT_S = 60l;
```
Sushisource, line 52: Why 601?

Quinn-With-Two-Ns: that is an `l`, let me change it to caps.

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
```diff
@@ -424,7 +427,8 @@ private void pollWorkflowUpdateHelper(
       PollWorkflowExecutionUpdateRequest request,
       Deadline deadline) {
 
-    Deadline pollTimeoutDeadline = Deadline.after(60L, TimeUnit.SECONDS).minimum(deadline);
+    Deadline pollTimeoutDeadline =
+        Deadline.after(POLL_UPDATE_TIMEOUT_S, TimeUnit.SECONDS).minimum(deadline);
```
cretz, line 431: How long until the server times this out? I just saw Go SDK does this too. I really think we need to be letting server time these out with blank successes like they do our worker polling and history fetching.

Quinn-With-Two-Ns: @mmcshane

cretz: Doesn't affect this PR, but does effect behavior. To ask another way, what happens when my update takes 90 seconds? Do we error or do we turn to async result fetching?

Quinn-With-Two-Ns: I'll let Matt comment on changing the server behavior. This PR is just to get Go and Java aligned.

>what happens when my update takes 90 seconds? Do we error or do we turn to async result fetching?

I believe we would time out, there is no logic to switch to "async result fetching". Depending on what exactly you mean  by "async result fetching" that might not be possible.

mmcshane (approved):

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
mmcshane: That's an application semantic. The underlying transport may or may not support that desire. Best you can do is try to do what is requested and react.

PR #1786, opened 2023-06-14
Build id versioning

Sushisource:
Add build id versioning client methods and worker/command options in support of the new feature


temporal-sdk/src/main/java/io/temporal/client/BuildIDOperation.java
```diff
@@ -0,0 +1,137 @@
...
+  abstract void augmentBuilder(UpdateWorkerBuildIdCompatibilityRequest.Builder builder);
+
+  /**
+   * This operation adds a new Build ID into a new set, which will be used as the default set for
+   * the queue. This means all new workflows will start on this Build ID.
+   *
+   * @param buildID The Build ID to add as the new overall default.
+   */
+  public static BuildIDOperation newIDInNewDefaultSet(@Nonnull String buildID) {
+    return new BuildIDOperation() {
```
cretz, line 44: For legibility in stack traces and elsewhere, static inner classes are often clearer than anonymous classes (they don't add many more lines, don't need to be exposed (unless you prefer that over methods), etc)

temporal-sdk/src/main/java/io/temporal/worker/Worker.java
```diff
@@ -561,15 +561,23 @@ private static SingleWorkerOptions toLocalActivityOptions(
...
       WorkerFactoryOptions factoryOptions,
       WorkerOptions options,
       WorkflowClientOptions clientOptions,
       List<ContextPropagator> contextPropagators) {
+    String buildID = null;
+    if (options.getBuildID() != null) {
+      buildID = options.getBuildID();
+    } else if (clientOptions.getBinaryChecksum() != null) {
+      buildID = clientOptions.getBinaryChecksum();
+    }
```
cretz, line 575:
```suggestion
    String buildID = clientOptions.getBinaryChecksum();
    if (options.getBuildID() != null) {
      buildID = options.getBuildID();
    }
```

(style choice, doesn't really matter)

temporal-sdk/src/main/java/io/temporal/client/BuildIDOperation.java
```diff
@@ -0,0 +1,137 @@
...
+import io.temporal.api.workflowservice.v1.UpdateWorkerBuildIdCompatibilityRequest;
+import javax.annotation.Nonnull;
+
+/**
+ * The implementations of this class can be passed as parameters to {@link
+ * WorkflowClient#updateWorkerBuildIDCompatability(String, BuildIDOperation)}
+ *
+ * <p>See each public static method to learn about and construct the available operations.
+ */
+public abstract class BuildIDOperation {
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: Yeah I dislike this about the Java SDK, but it is pretty consistent  with not capitalizing both letter so we should  be too.

temporal-sdk/src/main/java/io/temporal/worker/WorkerOptions.java
```diff
@@ -319,6 +323,33 @@ public Builder setDisableEagerExecution(boolean disableEagerExecution) {
       return this;
     }
 
+    /**
+     * Opts the worker in to the Build-ID-based versioning feature. This ensures that the worker
+     * will only recieve tasks which it is compatible with. For more information see: TODO: Doc link
+     *
+     * <p>Defaults to false
+     *
+     * <p>This is an EXPERIMENTAL API.
```
Quinn-With-Two-Ns, line 332: Can we use the annotation for this `@Experimental`

temporal-sdk/src/main/java/io/temporal/workflow/ChildWorkflowOptions.java
```diff
@@ -315,6 +300,15 @@ public Builder setCronSchedule(CronSchedule c) {
       return this;
     }
 
+    /**
+     * Specifies whether this child workflow should run on a worker with a compatible build ID or
+     * not. See the variants of {@link VersioningIntent}.
+     */
+    public Builder setVersioningIntent(VersioningIntent versioningIntent) {
```
Quinn-With-Two-Ns, line 307: Is there a reason normal workflow options do not need versioning intent? Maybe I am misunderstanding something.

temporal-sdk/src/main/java/io/temporal/worker/WorkerOptions.java
Sushisource: I'd need to upgrade the annotation, as it wasn't working in some spots for whatever reason, but seems like I can make it work there without much effort.

temporal-sdk/src/main/java/io/temporal/workflow/ChildWorkflowOptions.java
Sushisource: Because new workflows always start on the default. "Compatible" doesn't really make any sense for a new workflow, at least not without specifying a version, but we don't want to do that.

temporal-sdk/src/main/java/io/temporal/client/BuildIDOperation.java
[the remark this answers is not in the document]
Sushisource: Guess I'll rename it all. I actually prefer `Id` to `ID`. `ID` is not an acronym, the `D` doesn't stand for anything by itself. Not sure on what axis `ID` makes more sense, but I guess there are some strong feelings there lol. I was just sticking with what I'd done in Go, but should've just looked around.

temporal-sdk/src/main/java/io/temporal/worker/WorkerOptions.java
Sushisource: Huh, I guess that works now. Weird.

temporal-test-server/src/test/java/io/temporal/testserver/functional/WorkflowCachingTest.java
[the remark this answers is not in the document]
Sushisource: https://github.com/temporalio/sdk-java/issues/1788

temporal-sdk/src/main/java/io/temporal/client/BuildIDOperation.java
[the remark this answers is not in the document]
cretz: I personally like this about the Java SDK and in general. Otherwise you can't tell boundaries of consecutive acronyms/abbreviations like `XMLRPC` or `GRPC`. In .NET it's "all upper if only two letters, only first if more" which is a little weird too.

cretz (approved):

Quinn-With-Two-Ns (approved):

PR #1795, opened 2023-06-20
Avoid NPE on malformed/truncated failure stack traces

mjameswh:
## What changed

- While rehydrating Protobuf Failures to Java Exception objects, don't NPE on malformed/truncated stack traces

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/failure/DefaultFailureConverter.java
+++ b/temporal-sdk/src/main/java/io/temporal/failure/DefaultFailureConverter.java
@@ -44,6 +44,7 @@ import io.temporal.internal.sync.POJOWorkflowImplementationFactory;
 import io.temporal.serviceclient.CheckedExceptionWrapper;
 import java.io.PrintWriter;
 import java.io.StringWriter;
+import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


temporal-sdk/src/main/java/io/temporal/failure/DefaultFailureConverter.java
```diff
@@ -298,11 +299,14 @@ private StackTraceElement[] parseStackTrace(String stackTrace) {
     try {
       @SuppressWarnings("StringSplitter")
       String[] lines = stackTrace.split("\r\n|\n");
-      StackTraceElement[] result = new StackTraceElement[lines.length];
+      ArrayList<StackTraceElement> result = new ArrayList<>(lines.length);
       for (int i = 0; i < lines.length; i++) {
-        result[i] = parseStackTraceElement(lines[i]);
+        StackTraceElement elem = parseStackTraceElement(lines[i]);
+        if (elem != null) {
+          result.add();
```
Quinn-With-Two-Ns, line 306: should be `result.add(elem);`?

Quinn-With-Two-Ns (approved):

PR #1796, opened 2023-06-26
Generate bom using java-platform plugin

sghill:
## What was changed

Added a `temporal-bom` subproject that publishes a bom using Gradle's [java-platform plugin](https://docs.gradle.org/current/userguide/java_platform_plugin.html).

Most of the script plugins (`dependencyManagement.gradle`, `java.gradle`, etc) are built around the assumption each module is a java module. I excluded this project by name for this change.

## Why?
BOMs are useful for keeping versions aligned, especially when they are brought in only as transitive dependencies.

Example without a bom:
```
runtimeClasspath
+--- org.example:some-library:1.0.0
|    +--- io.temporal:temporal-serviceclient:1.15 (misaligned)
\--- io.temporal:temporal-sdk:1.20
```

Other examples of java libraries producing similar boms: [junit-bom](https://repo1.maven.org/maven2/org/junit/junit-bom/5.9.3/junit-bom-5.9.3.pom), [okhttp-bom](https://repo1.maven.org/maven2/com/squareup/okhttp3/okhttp-bom/4.11.0/okhttp-bom-4.11.0.pom)

## Checklist

How was this tested:

I ran `./gradlew :temporal-bom:generatePomFileForMavenJavaPublication` and viewed the generated pom was expected.

Any docs updates needed?

Possibly an example of usage:

```gradle
dependencies {
    implementation platform('io.temporal:temporal-bom:1.21.0-SNAPSHOT')
    implementation 'io.temporal:temporal-sdk' // no version necessary
    implementation 'io.temporal:temporal-remote-data-encoder'
    testImplementation 'io.temporal:temporal-testing'
}
```

### DIFF
--- a/gradle/dependencyManagement.gradle
+++ b/gradle/dependencyManagement.gradle
@@ -1,4 +1,7 @@
 subprojects {
+    if (name == 'temporal-bom') {
+        return
+    }
     configurations.implementation {
         // add global exclusions here if needed
     }
--- a/gradle/gatherDependencies.gradle
+++ b/gradle/gatherDependencies.gradle
@@ -1,4 +1,7 @@
 subprojects {
+    if (name == 'temporal-bom') {
+        return
+    }
     // Creates build/runtimeDeps folder in each module that contains all the dependencies
     // that need to be in the runtime classpath of the module
     task gatherRuntimeDeps(type: Copy) {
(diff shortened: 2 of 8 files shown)


Quinn-With-Two-Ns (approved):
LGTM CI is failing because you branch is missing tags

Issue #1727, opened 2023-03-31
inconsistency on MaximumAttempts attribute between local activities and "normal" activities


antmendoza:
## Expected Behavior
setting setMaximumAttempts=0 (or not setting any value, default = 0) for local activities should make the activity retry forever.  

I expect the same behavior as "normal" activities, where MaximumAttempts=0  implies retry forever. 

## Actual Behavior

when MaximumAttempts is set to 0, local activities do not retry

## Steps to Reproduce the Problem
[HelloLocalActivity.java.zip](https://github.com/temporalio/sdk-java/files/11122584/HelloLocalActivity.java.zip)

## Specifications

  - Version:
  - Platform:

abhinavvrm2020:
@antmendoza is this issue still valid. If yes i'll work on this.

PR #1801, opened 2023-06-28
Enable Infinitely Retrying Local Activity

pragnareddye:
## Why?

Setting setMaximumAttempts=0 (or not setting any value, default = 0) for local activities make the activity not retry at all. 

There is no way to enable infinite retries for local activities 

This is inconsistent with non-local/normal activity retry behavior  

https://github.com/temporalio/sdk-java/issues/1727

## Checklist

1. Closes 
https://github.com/temporalio/sdk-java/issues/1727

3. How was this tested:

Tested by running java samples with this change  https://github.com/temporalio/samples-java 

4. Any docs updates needed?

Nope

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
@@ -140,6 +140,10 @@ final class LocalActivityWorker implements Startable, Shutdownable {
       throw (Error) attemptThrowable;
     }
 
+    if (isRetryPolicyNotSet(activityTask)) {
+      return new RetryDecision(RetryState.RETRY_STATE_RETRY_POLICY_NOT_SET, null);
+    }
+
     RetryOptions retryOptions = RetryOptionsUtils.toRetryOptions(activityTask.getRetryPolicy());
 
     if (RetryOptionsUtils.isNotRetryable(retryOptions, attemptThrowable)) {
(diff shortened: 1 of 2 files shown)


Quinn-With-Two-Ns:
Hi @pragnareddye , thank you for the contribution. I need to go over this more tomorrow, but it would be helpful if we also  add some test with this change to verify the local activity does retry infinitely until it times out.

Quinn-With-Two-Ns:
we should also verify 


If both [setScheduleToCloseTimeout(Duration)](https://www.javadoc.io/static/io.temporal/temporal-sdk/1.20.0/io/temporal/activity/LocalActivityOptions.Builder.html#setScheduleToCloseTimeout(java.time.Duration)) and [RetryOptions.Builder.setMaximumAttempts(int)](https://www.javadoc.io/static/io.temporal/temporal-sdk/1.20.0/io/temporal/common/RetryOptions.Builder.html#setMaximumAttempts(int)) are not set, the Activity will not be retried.

pragnareddye:
Thank you so much for taking a look and pointing to the documentation requiring [setScheduleToCloseTimeout(Duration)](https://www.javadoc.io/static/io.temporal/temporal-sdk/1.20.0/io/temporal/activity/LocalActivityOptions.Builder.html#setScheduleToCloseTimeout(java.time.Duration)) or [setStartToCloseTimeout(Duration)](https://www.javadoc.io/static/io.temporal/temporal-sdk/1.20.0/io/temporal/activity/LocalActivityOptions.Builder.html#setStartToCloseTimeout(java.time.Duration)). 

I was under the impression that the retry options set check wasn't necessary and that the expected behavior was that by default activities should retry indefinitely. This wasn't possible if `scheduleToCloseTimeout` was required. It is possible to infinitely retry with `startToCloseTimeout` set however. It's perhaps safer also and wouldn't require an update to the documentation.

I'll update the code first to fix it based on my new understanding and follow up with tests.

Issue #1430, opened 2022-09-14
Workflow.getVersion() could cause NonDeterminsticError with multithreading + timer


longquanzheng:
Hey Temporal team,
We found that Workflow.getVersion() is letting workflow thread to yield, which causes NonDeterminsticError because the thread execution ordering is changed — this has made it very hard for us to use this getVersion().
Imagine we have two threads:

ThreadA:
```
{
    executeActivity(A)
}
```
ThreadB
```
{
   Workflow.sleep()
}
```
And  thread A is invoked first.
And we got a JSON history for replay test. The history looks like this:
```
WF start
WF task scheduled
WF task started
WF task completed
activity task scheduled
timer scheduled
```

Then we are changing thread A to :
```
{
    if( Workflow.getVersion() >= X ){
        executeActivity(B)
    }
    executeActivity(A)
}
```
Then the replay test gets broken.
Because the thread A will yield at Workflow.getVersion() and thread B will then try to schedule before the activity A.
In Java SDK, it’s enforcing the eventId to be exactly matched with the history. So the replay test would run into NDE error.

Some thoughts to improve:
I think GoSDK is only forcing within a workflow task completed batch commands. This may be more reasonable because I don’t think the order really matter within a workflow task results.
Or if we could do special case in Workflow.getVersion() to let it not yield the current thread…

## To reproduce
```
public void anyWorklfowMethod(){
         Async.procedure( ()->{ 
             Workflow.sleep(1s)
         }   
         )
         activityStub.hello()      
}
```
Got a history for replay
Then change the code to:
```
public void anyWorklfowMethod(){
         Async.procedure( ()->{ 
                 Workflow.sleep(1s)
             }   
         )
         Workflow.getVersion("changeId, Workflow.DefaultVersion, 1) 
         activityStub.hello()      
}
```
And this should break the replay test

Spikhalskiy:
1. Which version of java-sdk are we talking about?
2. What is the value of X in 
```
{
    if( Workflow.getVersion() > X ){
        executeActivity(B)
    }
   
```
? Because it does look like a non-deterministic change.

Spikhalskiy:
Can you provide original histories and related pieces of code or a reproduction?

longquanzheng:
> Which version of java-sdk are we talking about? Can you provide original histories and related pieces of code or a reproduction?

These are the versions we are using:
```
io.temporal:temporal-opentracing=1.13.0
io.temporal:temporal-sdk=1.13.0
[...]

PR #1807, opened 2023-07-06
Make GetVersion more deterministic

Quinn-With-Two-Ns:
Fix a possible source of non determinism in `GetVersion` calls. Since `GetVersion` can block on a future adding a call to it in a workflow can lead to non determinism so we need to avoid doing so in a backwards compatible way.

I tried to refactor the Version state machine to avoid ever blocking, but found the work too complicated for the benefit since due to backwards compatibility we can never remove the old state machine.

closes https://github.com/temporalio/sdk-java/issues/1430

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/common/SdkFlag.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/common/SdkFlag.java
@@ -43,9 +43,9 @@ public enum SdkFlag {
   }
 
   public static SdkFlag getValue(int id) {
-    SdkFlag[] As = SdkFlag.values();
-    for (int i = 0; i < As.length; i++) {
-      if (As[i].compare(id)) return As[i];
+    SdkFlag[] as = SdkFlag.values();
+    for (int i = 0; i < as.length; i++) {
+      if (as[i].compare(id)) return as[i];
     }
     return SdkFlag.UNKNOWN;
   }
--- a/temporal-sdk/src/main/java/io/temporal/internal/common/SdkFlag.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/common/SdkFlag.java
@@ -27,7 +27,7 @@ package io.temporal.internal.common;
 public enum SdkFlag {
   UNSET(0),
   /*
-   * Changes behaviour of GetVersion to not yield if no previous call existed in history.
+   * Changes behavior of GetVersion to not yield if no previous call existed in history.
    */
   SKIP_YIELD_ON_DEFAULT_VERSION(1),
   UNKNOWN(Integer.MAX_VALUE);
@@ -38,14 +38,14 @@ public enum SdkFlag {
     this.value = value;
   }
 
-  public boolean Compare(int i) {
+  public boolean compare(int i) {
     return value == i;
   }
 
-  public static SdkFlag GetValue(int _id) {
+  public static SdkFlag getValue(int id) {
     SdkFlag[] As = SdkFlag.values();
     for (int i = 0; i < As.length; i++) {
-      if (As[i].Compare(_id)) return As[i];
+      if (As[i].compare(id)) return As[i];
     }
     return SdkFlag.UNKNOWN;
   }
--- a/temporal-sdk/src/main/java/io/temporal/internal/common/SdkFlag.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/common/SdkFlag.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright (C) 2022 Temporal Technologies, Inc. All Rights Reserved.
+ *
+ * Copyright (C) 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Modifications copyright (C) 2017 Uber Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this material except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.temporal.internal.common;
+
+/**
+ * SdkFlag represents a flag used to help version the sdk internally to make breaking changes in
+ * workflow logic.
+ */
+public enum SdkFlag {
+  UNSET(0),
+  /*
+   * Changes behaviour of GetVersion to not yield if no previous call existed in history.
+   */
+  SKIP_YIELD_ON_DEFAULT_VERSION(1),
+  UNKNOWN(Integer.MAX_VALUE);
+
+  private final int value;
+
+  SdkFlag(int value) {
+    this.value = value;
+  }
+
+  public boolean Compare(int i) {
+    return value == i;
+  }
+
+  public static SdkFlag GetValue(int _id) {
+    SdkFlag[] As = SdkFlag.values();
+    for (int i = 0; i < As.length; i++) {
+      if (As[i].Compare(_id)) return As[i];
+    }
+    return SdkFlag.UNKNOWN;
+  }
+
+  public int getValue() {
+    return value;
+  }
+}
--- a/temporal-sdk/src/main/java/io/temporal/internal/common/SdkFlags.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/common/SdkFlags.java
@@ -20,24 +20,21 @@
 
 package io.temporal.internal.common;
 
-import io.temporal.api.workflowservice.v1.GetSystemInfoResponse;
 import io.temporal.workflow.Functions;
-import java.util.ArrayList;
-import java.util.List;
+import java.util.EnumSet;
 
 /** Represents all the flags that are currently set in a workflow execution. */
 public final class SdkFlags {
-  private final GetSystemInfoResponse.Capabilities capabilities;
+  private final boolean supportSdkMetadata;
   private final Functions.Func<Boolean> replaying;
-  // Flags that have been received from the server
-  private final List<SdkFlag> sdkFlags = new ArrayList<>();
+  // Flags that have been received from the server or have not been sent yet.
+  private final EnumSet<SdkFlag> sdkFlags = EnumSet.noneOf(SdkFlag.class);
   // Flags that have been set this WFT that have not been sent to the server.
   // Keep track of them separately, so we know what to send to the server.
-  private final List<SdkFlag> unsentSdkFlags = new ArrayList<>();
+  private final EnumSet<SdkFlag> unsentSdkFlags = EnumSet.noneOf(SdkFlag.class);
 
-  public SdkFlags(
-      GetSystemInfoResponse.Capabilities capabilities, Functions.Func<Boolean> replaying) {
-    this.capabilities = capabilities;
+  public SdkFlags(boolean supportSdkMetadata, Functions.Func<Boolean> replaying) {
+    this.supportSdkMetadata = supportSdkMetadata;
     this.replaying = replaying;
   }
 
@@ -47,7 +44,7 @@ public final class SdkFlags {
    * @return True, as long as the server supports SDK flags
    */
   public boolean setSdkFlag(SdkFlag flag) {
-    if (!capabilities.getSdkMetadata()) {
+    if (!supportSdkMetadata) {
       return false;
     }
     sdkFlags.add(flag);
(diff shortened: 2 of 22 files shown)


temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
```diff
@@ -235,6 +236,13 @@ private Result createCompletedWFTRequest(
       }
       completedRequest.setStickyAttributes(attributes);
     }
+    if (result.getSdkFlags().size() > 0) {
```
cretz, line 239:
```suggestion
    if (!result.getSdkFlags().isEmpty()) {
```

(I am assuming this is a collection type

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
```diff
@@ -148,11 +151,25 @@ enum HandleEventStatus {
...
+
+  @VisibleForTesting
+  public WorkflowStateMachines(
+      StatesMachinesCallback callbacks,
+      Functions.Proc1<StateMachine> stateMachineSink,
+      GetSystemInfoResponse.Capabilities capabilities) {
+    this.callbacks = Objects.requireNonNull(callbacks);
+    this.commandSink = cancellableCommands::add;
+    this.stateMachineSink = stateMachineSink;
+    this.localActivityRequestSink = (request) -> localActivityRequests.add(request);
```
cretz, line 171:
```suggestion
    this.localActivityRequestSink = localActivityRequests::add;
```

(didn't check this)

temporal-sdk/src/main/java/io/temporal/internal/common/SdkFlags.java
```diff
@@ -0,0 +1,85 @@
...
+      return true;
+    } else {
+      return unsentSdkFlags.contains(flag) || sdkFlags.contains(flag);
+    }
+  }
+
+  /***
+   * @return All flags set since the last call to takeNewSdkFlags.
+   */
+  public List<Integer> takeNewSdkFlags() {
```
cretz, line 76:
```suggestion
  public SortedSet<Integer> takeNewSdkFlags() {
```

IMO. Or better yet, don't burden this general class with integers. That's the caller's problem that they want integers instead of flags.

temporal-sdk/src/main/java/io/temporal/internal/common/SdkFlag.java
```diff
@@ -0,0 +1,56 @@
...
+ * SdkFlag represents a flag used to help version the sdk internally to make breaking changes in
+ * workflow logic.
+ */
+public enum SdkFlag {
+  UNSET(0),
+  /*
+   * Changes behaviour of GetVersion to not yield if no previous call existed in history.
+   */
+  SKIP_YIELD_ON_DEFAULT_VERSION(1),
+  UNKNOWN(Integer.MAX_VALUE);
```
cretz, line 33: Why have an unknown flag? Just make the return from `GetValue` (sic) as `Optional`

temporal-sdk/src/main/java/io/temporal/internal/common/SdkFlag.java
```diff
@@ -0,0 +1,56 @@
...
+    SdkFlag[] As = SdkFlag.values();
+    for (int i = 0; i < As.length; i++) {
+      if (As[i].Compare(_id)) return As[i];
+    }
+    return SdkFlag.UNKNOWN;
+  }
+
+  public int getValue() {
+    return value;
+  }
```
cretz, line 55:
```suggestion
```

`.ordinal()` I think is good enough.

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
```diff
@@ -862,27 +915,31 @@ public void mutableSideEffect(
...
-        maxSupported,
-        (v, e) -> {
-          callback.apply(v, e);
-          // without this getVersion call will trigger the end of WFT,
-          // instead we want to prepare subsequent commands and unblock the execution one more
-          // time.
-          eventLoop();
-        });
+            (idKey) -> {
+              System.out.println("Creating VersionStateMachine");
```
cretz, line 927: Leftover debug code?

temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflowContext.java
```diff
@@ -913,20 +915,33 @@ private <R> R mutableSideEffectImpl(
...
+                        result.complete(v);
+                      } else {
+                        result.completeExceptionally(e);
+                      }
+                    }));
+    /*
+     * If we are replaying a workflow and encounter a getVersion call it is possible that this call did not exist
+     * on the original execution. If the call did not exist on the original execution then we cannot block on results
+     * because it can lead to non-deterministic scheduling.
+     * */
```
cretz, line 937: I am not understanding this exactly. Are you saying that before now, Java would never return default version if `getVersion` was called during replay?

temporal-sdk/src/test/java/io/temporal/workflow/versionTests/GetVersionInSignalOnReplayTest.java
```diff
@@ -65,6 +66,12 @@ public void testGetVersionInSignal() {
     assertEquals("[done]", result);
   }
 
+  @Test
+  public void testGetVersionInSignal_replay() throws Exception {
```
cretz, line 70: No underscores in Java method names

temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflowContext.java
Quinn-With-Two-Ns: No, I am not sure what your confused about

cretz: Sorry, I think I am a bit dense. So how does this code skip a yield like the flag name says?

I think a test that demonstrates this issue from scratch may be clearest to me. You can just send queries to trigger replay (or stop/start worker w/ new code version). I am not following the test to demonstrate the issue. Even just standalone code here may help if we're not willing to be able to disable flag setting during a test.

Quinn-With-Two-Ns: Sorry wasn't trying to be rude I just wasn't sure what you were confused about so I could explain.

> So how does this code skip a yield like the flag name says?

this if statement is skipping the yield since the function will eagerly return if it succeeds. If it fails it will try to call `result.get();` and that will yield back to the scheduler and introduce non determinism


There is a test that demonstrates the issue `testGetVersionDefaultMultithreading`

cretz: > Sorry wasn't trying to be rude I just wasn't sure what you were confused about so I could explain.

I totally read it as a simple request for more info, no rudeness detected, no worries

> this if statement is skipping the yield since the function will eagerly return if it succeeds

Ah, I see. What about other cases where get version yields, is that not equally concerning? Or is the yield so predictable that so long as `getVersion` is called, it is predictable regardless of what the values set in it are?

So to confirm, this works a little bit better for you than a SDK flag like `SdkFlag.GET_VERSION_AS_SYNCHRONOUS`? I understand from your issue description that it is complicated. Yeah, I just took a dive in those state machine classes and they are complicated.

Question, can you confirm that by the time you get to this line of code the promise has been set? If not, where in the Java code could it be yielding (not counting the `get` below which, after reading the code, doesn't yield unless the promise is not already completed)?

Quinn-With-Two-Ns: >Ah, I see. What about other cases where get version yields, is that not equally concerning? Or is the yield so predictable that so long as getVersion is called, it is predictable regardless of what the values set in it are?

yes, the issues isn't yielding, it is the fact that on replay it would yield, but not executing for the first time

>So to confirm, this works a little bit better for you than a SDK flag like SdkFlag.GET_VERSION_AS_SYNCHRONOUS? 

yeah, but neither solution is great

>Question, can you confirm/guarantee that by the time you get to this line of code the promise has been set? If not, where in the Java code could it be yielding (not counting the get below which, after reading the code, doesn't yield unless the promise is not already completed)?

No, I am pretty sure I can guarantee that by the time the `get()` is reached the promise is set

cretz: > No, I am pretty sure I can guarantee that by the time the get() is reached the promise is NOT set

Is that only because `runner.executeInWorkflowThread(` or is there another thing happening inside the code that may be non-synchronous. I wonder what would happen if, with an SDK flag, you just ran the code instead wrapped in `executeInWorkflowThread`.

Quinn-With-Two-Ns: its because the state machine , like almost all state machines in java, changes state based on command events.

cretz: I see, and between events is where it can yield? If so, then I'll approve this PR as is. I just want to confirm that if you removed the `executeInWorkflowThread` outer part in the callback that the promise isn't resolved synchronously (_before_ you call `get` which does yield to other async things).

Quinn-With-Two-Ns: Yes, well commands specifically.

removing the `executeInWorkflowThread` will also cause another breaking change in behaviour because it will change how the callback is scheduled

cretz: Right, I only want to test with that. Then only remove it when an SDK flag is set, so the flag would then be more general purpose for making `getVersion` sync. But if you're saying the removal of that call does not help make the promise setting synchronous, then nevermind.

Quinn-With-Two-Ns: Yeah removing it does not make the promise setting synchronous. It also shouldn't be removed because then you could end up running\ workflow code in a non workflow thread which isn't really correct.

temporal-sdk/src/main/java/io/temporal/internal/common/SdkFlag.java
Quinn-With-Two-Ns: `UNKNOWN ` is not the same as an empty optional

temporal-sdk/src/main/java/io/temporal/internal/common/SdkFlag.java
Quinn-With-Two-Ns: We don't have any `ordinal` usage in the java SDK by non generated code and I would like to keep it that way

temporal-sdk/src/main/java/io/temporal/internal/common/SdkFlags.java
```diff
@@ -0,0 +1,85 @@
...
+
+/** Represents all the flags that are currently set in a workflow execution. */
+public final class SdkFlags {
+  private final GetSystemInfoResponse.Capabilities capabilities;
+  private final Functions.Func<Boolean> replaying;
+  // Flags that have been received from the server
+  private final List<SdkFlag> sdkFlags = new ArrayList<>();
+  // Flags that have been set this WFT that have not been sent to the server.
+  // Keep track of them separately, so we know what to send to the server.
+  private final List<SdkFlag> unsentSdkFlags = new ArrayList<>();
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: Yeah good point, i'll convert to enum sets until we are setting the flags on the WFT

temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflowContext.java
cretz: :+1: Can consider this thread resolved. Will approve as other things are addressed. Thanks for explaining this all to me.

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
Quinn-With-Two-Ns: kept the code as is because this does not work

temporal-sdk/src/main/java/io/temporal/internal/common/SdkFlag.java
cretz: What I'm saying is it shouldn't be the same. You should only have enumerates for real flags IMO. There's no reason to have all but one enumerate be a real flag IMO.

cretz: Eek. I hope we at least don't in TS, Python, and .NET. All those languages have a good way of representing an absent enum without having to work around that some values of the enum are representable in history and some aren't.

cretz (approved):

temporal-sdk/src/main/java/io/temporal/internal/common/SdkFlag.java
```diff
@@ -0,0 +1,56 @@
...
+  SdkFlag(int value) {
+    this.value = value;
+  }
+
+  public boolean Compare(int i) {
+    return value == i;
+  }
+
+  public static SdkFlag GetValue(int _id) {
+    SdkFlag[] As = SdkFlag.values();
```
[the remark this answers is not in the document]
cretz: This was never fixed

temporal-sdk/src/main/java/io/temporal/internal/common/SdkFlag.java
cretz: I don't agree, but y'all's design choice here. At least it's internal.

temporal-sdk/src/main/java/io/temporal/internal/common/SdkFlag.java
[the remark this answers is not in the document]
Quinn-With-Two-Ns: oh my bad I think I mixed this up with the method being capitalized

Issue #1808, opened 2023-07-10
DataConverter interface does not provide a default implementation for `failureToException` and `exceptionToFailure`


Quinn-With-Two-Ns:
## Expected Behavior
A default implementation for `failureToException` and `exceptionToFailure` exist since adding these without one breaks the API of `DataConverter`

## Actual Behavior
no default implementation exist

See also:
https://github.com/temporalio/sdk-java/pull/1645

PR #1809, opened 2023-07-10
Add default methods for exceptionToFailure and failureToException

Quinn-With-Two-Ns:
Add default methods for `exceptionToFailure` and `failureToException`

closes:
https://github.com/temporalio/sdk-java/issues/1808

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/common/converter/DataConverter.java
+++ b/temporal-sdk/src/main/java/io/temporal/common/converter/DataConverter.java
@@ -22,10 +22,12 @@ package io.temporal.common.converter;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.base.Defaults;
+import com.google.common.base.Preconditions;
 import io.temporal.api.common.v1.Payload;
 import io.temporal.api.common.v1.Payloads;
 import io.temporal.api.failure.v1.Failure;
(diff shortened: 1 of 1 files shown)


temporal-sdk/src/main/java/io/temporal/common/converter/DataConverter.java
```diff
@@ -174,7 +176,10 @@ default Object[] fromPayloads(
    * @throws NullPointerException if failure is null
    */
   @Nonnull
-  TemporalFailure failureToException(@Nonnull Failure failure);
+  default TemporalFailure failureToException(@Nonnull Failure failure) {
+    Preconditions.checkNotNull(failure, "failure");
+    return new DefaultFailureConverter().failureToException(failure, this);
```
Quinn-With-Two-Ns, line 181: I didn't make this a interface variable because then it would be public

cretz: Could make it a singleton inside that class, e.g.

```suggestion
return DefaultFailureConverter.getInstance().failureToException(failure, this);
```

though I suspect instantiating this for every failure is mostly harmless and you couldn't make the constructor private or protected anyways for other compatibility reasons (so it'd be a singleton w/ a public constructor which is kinda weird).

cretz (approved):

PR #1818, opened 2023-07-15
fix(schedule): try catch clause and rethrow StatusRuntimeException that does not match status c…

### LINKED ISSUE #1817: RootScheduleClientInvoker.createSchedule swallows errors
## Actual Behavior

It appears, as pointed out by [this comment](https://github.com/temporalio/sdk-java/pull/1776/files#r1263855932), that the `RootScheduleClientInvoker.createSchedule` code will swallow all exceptions that don't satisfy the `ScheduleAlreadyRunningException` conditions. We should rethrow others (and change catch clause to the specific type we want to catch).

bryanmlyr:
…ode ALREADY_EXISTS

## What was changed

Should resolve #1817
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/client/RootScheduleClientInvoker.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/client/RootScheduleClientInvoker.java
@@ -105,7 +105,7 @@ public class RootScheduleClientInvoker implements ScheduleClientCallsInterceptor
       if (Status.Code.ALREADY_EXISTS.equals(e.getStatus().getCode())) {
         throw new ScheduleAlreadyRunningException(e);
       } else {
-        throw e;
+        throw new ScheduleException(e);
       }
     }
   }
--- a/temporal-sdk/src/main/java/io/temporal/internal/client/RootScheduleClientInvoker.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/client/RootScheduleClientInvoker.java
@@ -102,8 +102,8 @@ public class RootScheduleClientInvoker implements ScheduleClientCallsInterceptor
     try {
       genericClient.createSchedule(request.build());
     } catch (StatusRuntimeException e) {
-      if (Status.Code.ALREADY_EXISTS.equals(sre.getStatus().getCode())) {
-        throw new ScheduleAlreadyRunningException(sre);
+      if (Status.Code.ALREADY_EXISTS.equals(e.getStatus().getCode())) {
+        throw new ScheduleAlreadyRunningException(e);
       } else {
         throw e;
       }
--- a/temporal-sdk/src/main/java/io/temporal/internal/client/RootScheduleClientInvoker.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/client/RootScheduleClientInvoker.java
@@ -101,12 +101,11 @@ public class RootScheduleClientInvoker implements ScheduleClientCallsInterceptor
 
     try {
       genericClient.createSchedule(request.build());
-    } catch (Exception e) {
-      if (e instanceof StatusRuntimeException) {
-        StatusRuntimeException sre = (StatusRuntimeException) e;
-        if (Status.Code.ALREADY_EXISTS.equals(sre.getStatus().getCode())) {
-          throw new ScheduleAlreadyRunningException(sre);
-        }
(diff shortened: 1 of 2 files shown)


cretz (approved):
Thanks! Will let @Quinn-With-Two-Ns or @mjameswh review.

temporal-sdk/src/main/java/io/temporal/internal/client/RootScheduleClientInvoker.java
```diff
@@ -101,12 +101,11 @@ public void createSchedule(CreateScheduleInput input) {
...
-      if (e instanceof StatusRuntimeException) {
-        StatusRuntimeException sre = (StatusRuntimeException) e;
-        if (Status.Code.ALREADY_EXISTS.equals(sre.getStatus().getCode())) {
-          throw new ScheduleAlreadyRunningException(sre);
-        }
+    } catch (StatusRuntimeException e) {
+      if (Status.Code.ALREADY_EXISTS.equals(e.getStatus().getCode())) {
+        throw new ScheduleAlreadyRunningException(e);
+      } else {
+        throw e;
```
Quinn-With-Two-Ns, line 108: This should be wrapped in some subclass of `TemporalException` I am not sure if any current one fit this case, so we should probably add one

cretz: I wonder if it's a general and separate issue to wrap all non-specialized schedule client exceptions.

Quinn-With-Two-Ns: Yeah we could tackle it in separate issue, but I wouldn't release this fix until that issue is also resolved.

bryanmlyr: Changes made 😁 but not sure about the error message

temporal-sdk/src/main/java/io/temporal/internal/client/RootScheduleClientInvoker.java
```diff
@@ -101,12 +101,11 @@ public void createSchedule(CreateScheduleInput input) {
...
-      if (e instanceof StatusRuntimeException) {
-        StatusRuntimeException sre = (StatusRuntimeException) e;
-        if (Status.Code.ALREADY_EXISTS.equals(sre.getStatus().getCode())) {
-          throw new ScheduleAlreadyRunningException(sre);
-        }
+    } catch (StatusRuntimeException e) {
+      if (Status.Code.ALREADY_EXISTS.equals(e.getStatus().getCode())) {
+        throw new ScheduleAlreadyRunningException(e);
+      } else {
+        throw new ScheduleException(e);
```
cretz, line 108: Wonder if we want to catch every `Exception` then and wrap them all (seems to be what we do with other calls). I also wonder if the `ScheduleException` should extend `WorkflowServiceException` and set `workflowType` to `null` like we do on start workflow. @Quinn-With-Two-Ns - thoughts?

Quinn-With-Two-Ns: >Wonder if we want to catch every Exception then and wrap them all (seems to be what we do with other calls).

Yes we should, I just missed this because we made schedules have its own client

> I also wonder if the ScheduleException should extend WorkflowServiceException and set workflowType to null like we do on start workflow.

I didn't suggest that option because `WorkflowServiceException` extend `WorkflowException` (which is weird, but we can't change that now) and `ScheduleException` is not a `WorkflowException`

>I also wonder if we want it to be ScheduleClientException so people don't think it's for everything about a schedule (but probably not needed). 

I don't like that option because the `ScheduleHandle` should also wrap exceptions so I don't like having "client" in the name.

cretz: Works for me. Top-level `ScheduleException` that just extends the regular Temporal one as written currently LGTM.

bryanmlyr:
> @bryanmlyr looks like your branch is failing CI because it is missing tags which unfortunately some of our CI requires , when you pull the repo you need to use `--tags`

Should be good now ! 😁

PR #1819, opened 2023-07-16
Tracing support improvements

siavashkavousi:
Improve support for tracing signals, queries, and updates.

Our team encountered issues with tracing not being supported across workflow signals, queries, and updates when using temporal. I am proposing a proof of concept to address this, and if it is accepted, I will handle the necessary improvements and tests.

Note:
I have manually tested this in our test environment, and it appears to be working. However, as mentioned earlier, I will add tests after the initial review.

### DIFF
--- a/temporal-opentracing/src/main/java/io/temporal/opentracing/internal/ActionTypeAndNameSpanBuilderProvider.java
+++ b/temporal-opentracing/src/main/java/io/temporal/opentracing/internal/ActionTypeAndNameSpanBuilderProvider.java
@@ -88,8 +88,11 @@ public class ActionTypeAndNameSpanBuilderProvider implements SpanBuilderProvider
       case RUN_WORKFLOW:
       case START_ACTIVITY:
       case RUN_ACTIVITY:
+      case SIGNAL_EXTERNAL_WORKFLOW:
       case SIGNAL_WORKFLOW:
       case UPDATE_WORKFLOW:
+      case HANDLE_SIGNAL:
+      case HANDLE_UPDATE:
         String runId = context.getRunId();
         Preconditions.checkNotNull(
             runId, "runId is expected to be not null for span operation type %s", operationType);
@@ -97,6 +100,7 @@ public class ActionTypeAndNameSpanBuilderProvider implements SpanBuilderProvider
             StandardTagNames.WORKFLOW_ID, context.getWorkflowId(),
             StandardTagNames.RUN_ID, context.getRunId());
       case QUERY_WORKFLOW:
+      case HANDLE_QUERY:
         return ImmutableMap.of();
     }
     throw new IllegalArgumentException("Unknown span operation type provided");
--- a/temporal-opentracing/src/main/java/io/temporal/opentracing/internal/ActionTypeAndNameSpanBuilderProvider.java
+++ b/temporal-opentracing/src/main/java/io/temporal/opentracing/internal/ActionTypeAndNameSpanBuilderProvider.java
@@ -88,6 +88,7 @@ public class ActionTypeAndNameSpanBuilderProvider implements SpanBuilderProvider
       case RUN_WORKFLOW:
       case START_ACTIVITY:
       case RUN_ACTIVITY:
+      case SIGNAL_WORKFLOW:
         String runId = context.getRunId();
         Preconditions.checkNotNull(
             runId, "runId is expected to be not null for span operation type %s", operationType);
--- a/temporal-opentracing/src/main/java/io/temporal/opentracing/internal/ActionTypeAndNameSpanBuilderProvider.java
+++ b/temporal-opentracing/src/main/java/io/temporal/opentracing/internal/ActionTypeAndNameSpanBuilderProvider.java
@@ -89,6 +89,7 @@ public class ActionTypeAndNameSpanBuilderProvider implements SpanBuilderProvider
       case START_ACTIVITY:
       case RUN_ACTIVITY:
       case SIGNAL_WORKFLOW:
+      case UPDATE_WORKFLOW:
         String runId = context.getRunId();
         Preconditions.checkNotNull(
             runId, "runId is expected to be not null for span operation type %s", operationType);
--- a/temporal-opentracing/src/main/java/io/temporal/opentracing/internal/ActionTypeAndNameSpanBuilderProvider.java
+++ b/temporal-opentracing/src/main/java/io/temporal/opentracing/internal/ActionTypeAndNameSpanBuilderProvider.java
@@ -94,13 +94,14 @@ public class ActionTypeAndNameSpanBuilderProvider implements SpanBuilderProvider
       case QUERY_WORKFLOW:
       case HANDLE_SIGNAL:
       case HANDLE_UPDATE:
-      case HANDLE_QUERY:
         String runId = context.getRunId();
         Preconditions.checkNotNull(
             runId, "runId is expected to be not null for span operation type %s", operationType);
         return ImmutableMap.of(
             StandardTagNames.WORKFLOW_ID, context.getWorkflowId(),
             StandardTagNames.RUN_ID, context.getRunId());
+      case HANDLE_QUERY:
+        return ImmutableMap.of();
     }
     throw new IllegalArgumentException("Unknown span operation type provided");
   }
--- a/temporal-opentracing/src/main/java/io/temporal/opentracing/internal/ActionTypeAndNameSpanBuilderProvider.java
+++ b/temporal-opentracing/src/main/java/io/temporal/opentracing/internal/ActionTypeAndNameSpanBuilderProvider.java
@@ -91,17 +91,16 @@ public class ActionTypeAndNameSpanBuilderProvider implements SpanBuilderProvider
       case SIGNAL_EXTERNAL_WORKFLOW:
       case SIGNAL_WORKFLOW:
       case UPDATE_WORKFLOW:
+      case QUERY_WORKFLOW:
       case HANDLE_SIGNAL:
       case HANDLE_UPDATE:
+      case HANDLE_QUERY:
         String runId = context.getRunId();
         Preconditions.checkNotNull(
             runId, "runId is expected to be not null for span operation type %s", operationType);
         return ImmutableMap.of(
             StandardTagNames.WORKFLOW_ID, context.getWorkflowId(),
             StandardTagNames.RUN_ID, context.getRunId());
-      case QUERY_WORKFLOW:
-      case HANDLE_QUERY:
-        return ImmutableMap.of();
     }
     throw new IllegalArgumentException("Unknown span operation type provided");
   }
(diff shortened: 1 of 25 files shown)


temporal-opentracing/src/main/java/io/temporal/opentracing/internal/ActionTypeAndNameSpanBuilderProvider.java
```diff
@@ -88,12 +88,16 @@ protected Map<String, String> getSpanTags(SpanCreationContext context) {
...
+      case SIGNAL_WORKFLOW:
+      case UPDATE_WORKFLOW:
         String runId = context.getRunId();
         Preconditions.checkNotNull(
             runId, "runId is expected to be not null for span operation type %s", operationType);
         return ImmutableMap.of(
             StandardTagNames.WORKFLOW_ID, context.getWorkflowId(),
             StandardTagNames.RUN_ID, context.getRunId());
+      case QUERY_WORKFLOW:
+        return ImmutableMap.of();
```
cretz, line 100: In our other SDKs, query spans also have workflow ID and run ID.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowClientCallsInterceptor.java
```diff
@@ -58,6 +59,27 @@ public WorkflowStartOutput start(WorkflowStartInput input) {
...
+  public WorkflowSignalOutput signal(WorkflowSignalInput input) {
+    Span workflowSignalSpan =
+        contextAccessor.writeSpanContextToHeader(
+            () ->
+                spanFactory
+                    .createWorkflowSignalSpan(
+                        tracer,
+                        input.getSignalName(),
+                        input.getWorkflowExecution().getWorkflowId(),
+                        References.FOLLOWS_FROM)
```
cretz, line 72: I think the workflow inbound should make this a follows-from, but client outbound should be a child-of

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowInboundCallsInterceptor.java
```diff
@@ -82,4 +82,81 @@ public WorkflowOutput execute(WorkflowInput input) {
...
+  @Override
+  public void handleSignal(SignalInput input) {
+    Tracer tracer = options.getTracer();
+    SpanContext rootSpanContext =
+        contextAccessor.readSpanContextFromHeader(input.getHeader(), tracer);
+    Span workflowSignalSpan =
+        spanFactory
+            .createWorkflowSignalSpan(
+                tracer,
+                input.getSignalName(),
```
cretz, line 95: I recommend, to match other SDKs, that workflow inbound signal, query, and update span names be `HandleSignal:<name>`, `HandleQuery:<name>`, and `HandleQuery:<name>` instead of reusing the same outbound names

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowInboundCallsInterceptor.java
```diff
@@ -82,4 +82,81 @@ public WorkflowOutput execute(WorkflowInput input) {
...
   }
+
+  @Override
+  public void handleSignal(SignalInput input) {
+    Tracer tracer = options.getTracer();
+    SpanContext rootSpanContext =
+        contextAccessor.readSpanContextFromHeader(input.getHeader(), tracer);
+    Span workflowSignalSpan =
+        spanFactory
+            .createWorkflowSignalSpan(
```
cretz, line 93: I didn't read this closely, but what many of our other SDKs do is make this span a child-of the original inbound workflow-level span and a follows-from the inbound signal-level span. Is that what's happening here?

cretz (commented):
Looks great and seems to mostly match our other SDKs. Will let @Quinn-With-Two-Ns and otehrs review.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/ActionTypeAndNameSpanBuilderProvider.java
siavashkavousi: The problem here is that handleQuery is not executed within a workflow thread, leading to non-deterministic errors.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowInboundCallsInterceptor.java
siavashkavousi: done

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowClientCallsInterceptor.java
siavashkavousi: done

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowInboundCallsInterceptor.java
siavashkavousi: can you elaborate more on this (a link, doc or something related to other SDKs)?

cretz: When you view the trace in many SDKs, a `HandleSignal`'s parent is the same as the `RunWorkflow`'s parent. And the `HandleSignal` is just linked via follows-from to the span in the header.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/ActionTypeAndNameSpanBuilderProvider.java
siavashkavousi: Yes

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/SpanFactory.java
```diff
@@ -77,6 +77,46 @@ public Tracer.SpanBuilder createChildWorkflowStartSpan(
...
+        SpanCreationContext.newBuilder()
+            .setSpanOperationType(SpanOperationType.SIGNAL_WORKFLOW)
+            .setActionName(signalName)
+            .setWorkflowId(workflowId)
+            .setRunId(runId)
+            .build();
+    return createSpan(context, tracer, null, References.FOLLOWS_FROM);
+  }
+
+  public Tracer.SpanBuilder createWorkflowSignalSpan(
```
cretz, line 104:
```suggestion
  public Tracer.SpanBuilder createWorkflowHandleSignalSpan(
```

Can you change the names for handle-based spans instead of overloads where the caller can't know they are creating different operation types just by calling different overloads?

siavashkavousi: done

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/ActionTypeAndNameSpanBuilderProvider.java
siavashkavousi: I did change it for `QUERY_WORKFLOW` case but it is not possible for `HANDLE_QUERY` because it is not running inside a workflow thread.

Quinn-With-Two-Ns:
hm looks like `opentracing.SignalWithStartTest` test if failing, does the test need to be updated?

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/ActionTypeAndNameSpanBuilderProvider.java
Quinn-With-Two-Ns: @cretz He is  saying he cannot set them the same way as `HANDLE_SIGNAL` because in the Java SDK queries cannot call ` Workflow.getInfo()`

cretz: Can store as state in the interceptor probably, but I guess the question is, should we get the workflow info in here properly or is this good enough for now and we change later?

Quinn-With-Two-Ns: You can't currently get the workflow info here "properly" and it is a large change to enable it

cretz: I'm not sure that a span without these tags is ok, but I'll let y'all decide

Quinn-With-Two-Ns: Well the options are either no span or a span without those tags. I have no preference.

Regardless of which we go with we will need to open an issue to address this when we fix `getInfo` in query handlers , but I don't think that should block this PR.

cretz: Concur, will let you make the call on tagless span vs no span. I don't have a super-strong opinion here.

siavashkavousi: I believe it would be better to have a tagless span since the span itself is useful for tracing.

Quinn-With-Two-Ns: Works for me

Quinn-With-Two-Ns:
@siavashkavousi looks like `opentracing.SignalWithStartTest::signalWithStart` is still failing in CI, do you understand why?

siavashkavousi:
> @siavashkavousi looks like `opentracing.SignalWithStartTest::signalWithStart` is still failing in CI, do you understand why?

@Quinn-With-Two-Ns I'm sorry; I didn't have the time to investigate. I'll look into it in the coming days.

siavashkavousi:
@Quinn-With-Two-Ns I have reviewed the situation, and it appears that the issue arises from the disparity between in-memory tests and Docker integration tests. To address this, I used `SDKTestWorkflowRule.useExternalService` to accommodate distinct assertions for these two environments, thus preventing the need for further modifications. I propose that we address the in-memory test environment bug in a separate pull request. what do you suggest?

Quinn-With-Two-Ns:
@siavashkavousi That is fine by me.  The in memory test server probably does not pass headers through in some situations . For CI to pass, unfortunately some of our CI requires that when you pull the repo you need to use `--tags` can  you update your branch? Sorry about the inconvenience.

Quinn-With-Two-Ns:
@siavashkavousi  some of our CI requires that when you pull the repo you need to use `--tags` can you update your branch so we CI can pass and we can merge your contribution?

siavashkavousi:
@Quinn-With-Two-Ns I apologize for the delay. I've pulled and pushed tags. Could you please verify if I've done the right thing or not, please?

Quinn-With-Two-Ns (approved):
@siavashkavousi Thanks for the contribution, I'll leave open for a couple days if @cretz has anything to add.

cretz (approved):
LGTM (though I didn't pour over the details of the latest commits)

PR #1821, opened 2023-07-19
Block mutating workflow state in a read only context

Quinn-With-Two-Ns:
Block calling most mutating workflow functions in read only context. This includes update validators and side effects.

Note: This does not include queries because queries are not run in a workflow thread

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/ActivityInvocationHandler.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/ActivityInvocationHandler.java
@@ -25,6 +25,7 @@ import io.temporal.activity.ActivityOptions;
 import io.temporal.common.MethodRetry;
 import io.temporal.common.interceptors.WorkflowOutboundCallsInterceptor;
 import io.temporal.workflow.ActivityStub;
+import io.temporal.workflow.Functions;
 import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.Method;
 import java.util.HashMap;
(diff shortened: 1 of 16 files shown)


Sushisource:
Hmm, the way I imagined in my head this would work is rather than passing this `assertReadOnly` callback into all the different handlers, you'd instead just wrap the act of inserting into the command buffer, and complain if you're in a readonly context - that way you don't have to pass that callback all over.

Also, for setting/removing the readonly flag, maybe some RAII pattern would be safer than try/finally?

Do either/both of those feel reasonable to you?

Quinn-With-Two-Ns:
>Hmm, the way I imagined in my head this would work is rather than passing this assertReadOnly callback into all the different handlers, you'd instead just wrap the act of inserting into the command buffer, and complain if you're in a readonly context - that way you don't have to pass that callback all over.

No,  by the point your going into the command  buffer is too late, we want to  fail  before you even go through interceptors. This is how it works in Python and Go. Also some illegal calls do not insert anything into the command buffer or only conditionally do so It makes the code more confusing and harder to verify correctness IMO since you'd have checks at very  different level.

>Also, for setting/removing the readonly flag, maybe some RAII pattern would be safer than try/finally?

Java doesn't really have  RAII, the only other thing we could do it a `try-with-resources`. I don't think it would improve the  safety much since you just change what you have to remember  from "remember to set back  to false" to "remember to use this wrapper object"

Sushisource:
OK - at the end of the day you're failing the WFT still though, so does it really matter at what point you do that? I get the bit about some things not generating commands though.

Sushisource (approved):

PR #1822, opened 2023-07-21
Add unit test for concurrent WFT

Quinn-With-Two-Ns:
Add test to verify the workflow worker hold a lock when processing a WFT, including reseting event ID

### DIFF
--- a/temporal-sdk/src/test/java/io/temporal/internal/worker/WorkflowWorkerTest.java
+++ b/temporal-sdk/src/test/java/io/temporal/internal/worker/WorkflowWorkerTest.java
@@ -1,0 +1,196 @@
+/*
+ * Copyright (C) 2022 Temporal Technologies, Inc. All Rights Reserved.
+ *
+ * Copyright (C) 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Modifications copyright (C) 2017 Uber Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this material except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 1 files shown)


temporal-sdk/src/test/java/io/temporal/internal/worker/WorkflowWorkerTest.java
```diff
@@ -0,0 +1,196 @@
...
+    respondTaskLatch.await();
+    // Sleep to allow metrics to be published
+    Thread.sleep(100);
+    // No task should have the lock anymore
+    assertEquals(runLockManager.totalLocks(), 0);
+    // All slots should be available
+    reporter.assertGauge(
+        MetricsType.WORKER_TASK_SLOTS_AVAILABLE,
+        ImmutableMap.of("worker_type", "WorkflowWorker"),
+        100.0);
```
mmcshane, line 190: Nit: you could expose this as a constant from the main source.

temporal-sdk/src/test/java/io/temporal/internal/worker/WorkflowWorkerTest.java
```diff
@@ -0,0 +1,196 @@
...
+    assertTrue(worker.start());
+    // Wait until we have got all the polls
+    pollTaskQueueLatch.await();
+    // Wait until the worker handles at least one WFT
+    handleTaskLatch.await();
+    // Sleep to allow metrics to be published
+    Thread.sleep(100);
+    // Since all polls have the same runID only one should get through, the other two should be
+    // blocked
+    assertEquals(runLockManager.totalLocks(), 1);
```
mmcshane, line 174: The comments are good - you might consider including them in the assert call as Strings so that they show up in the test failure output as well.

mmcshane (approved):
Comments are nits or stylistic; no changes needed.

PR #1826, opened 2023-07-26
Set security_opt to no-new-privileges:true

jackdawm:
**What changed?**
Locking down some images running in Buildkite to prevent container processes from gaining additional privileges.

**Why?**
Responding to medium findings in Semgrep. Not sure this rises to medium, because of it being development images, but good hygiene is always nice.

**How did you test it?**
Opening this PR so Buildkite can pick it up.

**Is hotfix candidate?**
No


Sushisource (approved):

cretz (approved):
Would wait for @Quinn-With-Two-Ns or @mjameswh to merge

Quinn-With-Two-Ns (approved):
@jackdawm Can you rebase so  CI will pass, then I can merge it

PR #1827, opened 2023-07-31
Properly wrap exceptions from schedule client

Quinn-With-Two-Ns:
Wrap exceptions from the `ScheduleClient`. Add some tests to show the exception is wrapped

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/client/RootScheduleClientInvoker.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/client/RootScheduleClientInvoker.java
@@ -107,6 +107,8 @@ public class RootScheduleClientInvoker implements ScheduleClientCallsInterceptor
       } else {
         throw new ScheduleException(e);
       }
+    } catch (Exception e) {
+      throw new ScheduleException(e);
     }
   }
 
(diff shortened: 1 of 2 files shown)


cretz:
I feel like I remember us discussing this somewhere, but I forget the details. I wonder if we need to do this in other SDKs.

Quinn-With-Two-Ns:
This is more java specific because java SDK always wraps all  the exceptions it throws. I don't think other SDKs try to.

mjameswh:
> This is more java specific because java SDK always wraps all the exceptions it throws. I don't think other SDKs try to.

We do in TS.

You should test that non-schedules related errors (like [namespace not found](https://github.com/temporalio/sdk-typescript/issues/1145)) don't get wrapped as `ScheduleException`. In TS, we now parse the `grpc-status-details-bin` metadata on errors (see [PR](https://github.com/temporalio/sdk-typescript/pull/1147)), which makes proper error reporting significantly easier.

cretz (approved):

Quinn-With-Two-Ns:
@mjameswh I think typescript is unique in having a `NamespaceNotFound` exception. I don't think Java or Dotnet has it (didn't check python).

@cretz I remember on the original PR I had `ScheduleNotFound` exception but we removed it because the server did not send enough status details for us to disambiguate on the SDK side?

cretz:
> because the server did not send enough status details for us to disambiguate on the SDK side?

Right, but server gives enough information for namespace not found. We do have namespace not found in Go, but yes probably not the newer non-TS core SDKs

PR #1829, opened 2023-07-31
Adjust MaxBufferedQueryCount

Quinn-With-Two-Ns:
Fix test failures like https://buildkite.com/temporal/java-sdk-public/builds/2754

### DIFF
--- a/docker/buildkite/dynamicconfig/development.yaml
+++ b/docker/buildkite/dynamicconfig/development.yaml
@@ -12,3 +12,5 @@ frontend.workerVersioningWorkflowAPIs:
   - value: true
 frontend.workerVersioningDataAPIs:
   - value: true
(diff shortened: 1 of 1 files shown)


Quinn-With-Two-Ns:
This change is needed because of this server change https://github.com/temporalio/temporal/commit/2830f3816728a782252ab9074752a303591c0804 to intentionally make exceeding the max query buffer a non retry able error.  The other options would be to lower the intensity of the test or add retries in the test. Since the purpose of the test is to stress the SDK with lots of queries while running local activities I think increasing the server limit stresses the SDK the most

cretz:
> I think increasing the server limit stresses the SDK the most

Concur. A bit strange to stress the SDK in CI like this, but meh.

cretz (approved):

PR #1841, opened 2023-08-15
Fix bug in VersionStateMachine causing wrong version to get returned

Quinn-With-Two-Ns:
Fix a bug in `VersionStateMachine` causing `DEFAULT_VERSION` to get returned incorrectly if the same change ID was read multiple times in a Workflow. This was missed in the original [fix](https://github.com/temporalio/sdk-java/commit/6054c315e934f4a9b295f2da85c8ae003851c4d7) because we lacked some basic tests of the `VersionStateMachine`


Quinn-With-Two-Ns:
All tests were written and passed against the commit before `6054c315e934f4a9b295f2da85c8ae003851c4d7`

cretz (approved):

Issue #1847, opened 2023-08-24
Typed search attributes aren't consistent with schedules


Quinn-With-Two-Ns:
Typed search attributes in Java don't interact cleanly with Schedule Workflows in Java. 

One issue is the search attribute field on the `ScheduleOptions` still use the old approach. This was just a miss in the typed search attribute PR because it was mostly done before schedules was merged.

Another more annoying behaviour is typed search attributes on the schedule workflow action: 
* If a user uses the old, deprecated way of setting search attributes when creating a schedule, Java SDK just ignores them when creating the schedule. 
* If a users set the SA through both the deprecated and typed way we just accept the typed values and don't error or warn like we do in other places in the Java SDK. 
* On the describe path Java SDK will panic if the search attributes does not type information in their metadata. Java always writes this, but other SDKs like Go did not (we would warn in this case) and the server does not currently fill this information in like was expected.

PR #1848, opened 2023-08-24
Reconcile typed search attributes with schedules

Quinn-With-Two-Ns:
Fix typed search attributes in java schedules by:
* Adding typed search attributes to `ScheduleOptions`
* Documenting old `SearchAttributes` do not work when creating a schedule
* Allow for untyped search attributes in the `SearchAttributes` container through a special key type

closes https://github.com/temporalio/sdk-java/issues/1847

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/common/SearchAttributeKey.java
+++ b/temporal-sdk/src/main/java/io/temporal/common/SearchAttributeKey.java
@@ -21,6 +21,7 @@
 package io.temporal.common;
 
 import com.google.common.reflect.TypeToken;
+import io.temporal.api.common.v1.Payload;
 import io.temporal.api.enums.v1.IndexedValueType;
 import java.lang.reflect.Type;
 import java.time.OffsetDateTime;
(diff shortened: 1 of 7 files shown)


temporal-sdk/src/main/java/io/temporal/common/SearchAttributeKey.java
```diff
@@ -73,6 +74,17 @@ public static SearchAttributeKey<List<String>> forKeywordList(String name) {
         KEYWORD_LIST_REFLECT_TYPE);
   }
 
+  /**
+   * Create a search attribute key for an untyped attribute type.
+   *
+   * <p>This should only be used when the server can return untyped search attributes, for example,
+   * when describing a schedule workflow action.
+   */
+  public static SearchAttributeKey<Payload> forUntyped(String name) {
```
Quinn-With-Two-Ns, line 83: We could also make calling `valueSet` and `valueUnset` an error since I am not aware of any valid use case

cretz: Yeah, that would work, I think we should. We can hide this more in other langs. Arguably we could make the name of this method more unwieldy, but this is probably good.

cretz (approved):
This looks great! I am thinking how I can apply to my Python conundrum.

temporal-sdk/src/main/java/io/temporal/common/SearchAttributeKey.java
Quinn-With-Two-Ns: Done

Issue #1840, opened 2023-08-14
Spring Boot 3 native image support


zhxnlai:
**Is your feature request related to a problem? Please describe.**

During native image build, Spring Boot auto-configuration fails with the following error message:
```
[           main] ocalVariableTableParameterNameDiscoverer : Using deprecated '-debug' fallback for parameter name resolution. Compile the affected code with '-parameters' instead or avoid its introspection: io.temporal.spring.boot.autoconfigure.properties.WorkerProperties
Exception in thread "main" org.springframework.boot.context.properties.bind.MissingParametersCompilerArgumentException: Constructor binding in a native image requires compilation with -parameters but the following classes were compiled without it:
	io.temporal.spring.boot.autoconfigure.properties.WorkerProperties
	io.temporal.spring.boot.autoconfigure.properties.WorkerProperties$CapacityConfigurationProperties
	io.temporal.spring.boot.autoconfigure.properties.TemporalProperties
	io.temporal.spring.boot.autoconfigure.properties.WorkerProperties$RateLimitsConfigurationProperties
	io.temporal.spring.boot.autoconfigure.properties.NamespaceProperties$WorkflowCacheProperties

	at org.springframework.boot.context.properties.bind.BindableRuntimeHintsRegistrar.registerHints(BindableRuntimeHintsRegistrar.java:100)
	at org.springframework.boot.context.properties.ConfigurationPropertiesBeanFactoryInitializationAotProcessor$ConfigurationPropertiesReflectionHintsContribution.applyTo(ConfigurationPropertiesBeanFactoryInitializationAotProcessor.java:74)
	at org.springframework.context.aot.BeanFactoryInitializationAotContributions.applyTo(BeanFactoryInitializationAotContributions.java:78)
	at org.springframework.context.aot.ApplicationContextAotGenerator.lambda$processAheadOfTime$0(ApplicationContextAotGenerator.java:58)
	at org.springframework.context.aot.ApplicationContextAotGenerator.withCglibClassHandler(ApplicationContextAotGenerator.java:67)
	at org.springframework.context.aot.ApplicationContextAotGenerator.processAheadOfTime(ApplicationContextAotGenerator.java:53)
	at org.springframework.context.aot.ContextAotProcessor.performAotProcessing(ContextAotProcessor.java:106)
	at org.springframework.context.aot.ContextAotProcessor.doProcess(ContextAotProcessor.java:84)
[...]

PR #1877, opened 2023-09-29
Support Spring Boot 3 native image build

zhxnlai:
## What was changed

Added the `-parameters` compiler argument for several classes, addressing the `MissingParametersCompilerArgumentException` during native image compilation in projects using Temporal.

## Why?

This change is critical to enable users to compile their projects into native images without encountering `MissingParametersCompilerArgumentException`. It facilitates smoother and error-free native compilations, especially for projects relying on Temporal’s classes for constructor injection. By ensuring the parameter names are available at runtime via reflection, it mitigates runtime issues and enhances the user experience in native environments.

## Checklist

1. Closes 
#1840

2. How was this tested:

The project was built and tested locally, and the `MissingParametersCompilerArgumentException` error was no longer present. Further testing in environments using native images is recommended.

3. Any docs updates needed?

No documentation updates are needed as this change is related to the compilation process and does not alter any functionalities or features of the Temporal classes. The users consuming these classes do not require any additional steps or modifications in their usage or integration.

### DIFF
--- a/gradle/java.gradle
+++ b/gradle/java.gradle
@@ -14,7 +14,7 @@ subprojects {
 
     compileJava {
         options.encoding = 'UTF-8'
-        options.compilerArgs << '-Xlint:none' << '-Xlint:deprecation' << '-Werror'
+        options.compilerArgs << '-Xlint:none' << '-Xlint:deprecation' << '-Werror' << '-parameters'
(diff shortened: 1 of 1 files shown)


Quinn-With-Two-Ns:
Thanks for the contribution @zhxnlai

Quinn-With-Two-Ns (approved):

PR #1878, opened 2023-10-02
Fix caching in WorkflowLocal/WorkflowThreadLocal (#1876)

### LINKED ISSUE #1876: WorkflowLocal shares initial value between all Workflows
## Expected Behavior
If I create a `WorkflowLocal` via `WorkflowLocal.withInitial() -> new MyClass())`, I get a unique instance of `MyClass` the first time I call `workflowLocal.get()` in a given Workflow.

## Actual Behavior
I get the same instance of `MyClass` returned in every Workflow.

This test demonstrates the issue:

``` java
    @Test
    void test(TestWorkflowEnvironment env, WorkflowClient client, Worker worker) {
        worker.registerWorkflowImplementationFactory(ChildWf.class, () -> () -> {
            var r = ref.get();
            System.out.println(r.get());
        });
        worker.registerWorkflowImplementationFactory(ParentWf.class, () -> () -> {
            var r = ref.get();
            r.set("hi");
            Workflow.newChildWorkflowStub(ChildWf.class).executeChild();
        });
        env.start();
        client.newWorkflowStub(ParentWf.class, WorkflowOptions.newBuilder()
                        .setTaskQueue(worker.getTaskQueue())
                .validateBuildWithDefaults()).execute();
    }

    private static final WorkflowLocal<AtomicReference<String>> ref = WorkflowLocal.withInitial(() -> new AtomicReference<>("bye"));

    @WorkflowInterface
    public interface ParentWf {
        @WorkflowMethod
        void execute();
    }

    @WorkflowInterface
    public interface ChildWf {
        @WorkflowMethod
        void executeChild();
    }
```
This should print "bye" (which is the 1.19.1 behavior), but instead it prints "hi".

## Specifications
  - Version: 1.21.1

dano:
## What was changed
Reverted caching changes made to WorkflowLocal/WorkflowThreadLocal, which broke backwards compatibility and accidentally shared values between Workflows/Threads. Re-implemented caching as an optional feature, and deprecated the factory methods that created non-caching instances.

## Why?
The current implementation both broke backwards compatibility, and did not function correctly; it shared values between Workflows and Threads when it should not have. See https://github.com/temporalio/sdk-java/issues/1876 for details.

## Checklist

1. Closes #1876 

2. How was this tested:
Unit tests

3. Any docs updates needed?
I don't think `WorkflowLocal`/`WorkflowThreadLocal` are documented outside of Javadocs, so no.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/context/ContextThreadLocal.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/context/ContextThreadLocal.java
@@ -32,7 +32,7 @@ import java.util.function.Supplier;
 public class ContextThreadLocal {
 
   private static final WorkflowThreadLocal<List<ContextPropagator>> contextPropagators =
-      WorkflowThreadLocal.withInitial(
+      WorkflowThreadLocal.withCachedInitial(
           new Supplier<List<ContextPropagator>>() {
             @Override
             public List<ContextPropagator> get() {
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/RunnerLocalInternal.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/RunnerLocalInternal.java
@@ -24,24 +24,26 @@ import java.util.Optional;
 import java.util.function.Supplier;
 
 public final class RunnerLocalInternal<T> {
-  private T supplierResult = null;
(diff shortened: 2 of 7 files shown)


Quinn-With-Two-Ns (approved):
Thanks for the contribution @dano

PR #1879, opened 2023-10-03
Disable eager start by default

Quinn-With-Two-Ns:
Disable eager start by default because it does not interact with worker versioning well.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/client/WorkflowOptions.java
+++ b/temporal-sdk/src/main/java/io/temporal/client/WorkflowOptions.java
@@ -105,7 +105,7 @@ public final class WorkflowOptions {
 
     private List<ContextPropagator> contextPropagators;
 
-    private boolean disableEagerExecution;
+    private boolean disableEagerExecution = true;
 
     private Builder() {}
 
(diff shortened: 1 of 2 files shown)


temporal-sdk/src/main/java/io/temporal/client/WorkflowOptions.java
```diff
@@ -105,7 +105,7 @@ public static final class Builder {
 
     private List<ContextPropagator> contextPropagators;
 
-    private boolean disableEagerExecution;
+    private boolean disableEagerExecution = true;
```
cretz, line 108: If we were to do it over again, I suspect we'd prefer `enableEagerExecution` here and I suspect that's what it'll be in all other SDKs. Is it worth breaking compatibility to make it more clearly opt-in? If we are willing to change it, I wonder if `enableEagerStart` is a better name to match Go.

Quinn-With-Two-Ns: I agree if we were doing it from scratch I would change the name, but we aren't doing it from scratch and we need to keep in mind of compatibility guarantees so I wouldn't change the name since it's more trouble then it's worth

cretz: :+1: I think at least for any other SDKs we implement this in, we can be more Go like and just let Java be the outlier

cretz (approved):

Issue #1880, opened 2023-10-04
Speculative workflow task can cause WFT failure


Quinn-With-Two-Ns:
## Expected Behavior
Speculative workflow task don't cause WFT failure

## Actual Behavior
Speculative workflow task can cause WFT failure

## Steps to Reproduce the Problem

  1. Send a speculative update that is rejected
  2. Send a signal

PR #1881, opened 2023-10-04
Reset lastHandledEventId on speculative WFT

Quinn-With-Two-Ns:
On a rejected speculative WFT we also need to reset `lastHandledEventId` so we don't accidentally skip new workflow tasks.

```
  public void handleEvent(HistoryEvent event, boolean hasNextEvent) {
    long eventId = event.getEventId();
    if (eventId <= lastHandledEventId) {
      // already handled
      return;
    }
...
``` 

We also should delete any new state machines created by the speculative task

closes https://github.com/temporalio/sdk-java/issues/1880


cretz (commented):
I have no insight in the correctness/danger of this change and can only hope tests catch any issues. I will defer this review to others (if there are any others)

Sushisource (approved):

PR #1886, opened 2023-10-08
Only rely on next page token for pagination

Quinn-With-Two-Ns:
It is valid behaviour for the server to respond to `GetWorkflowExecutionHistory` with no events, but a next page token, and that page token may point to a page with more events for the SDK to process. Before this change the SDK would stop looking if it saw a page with no events without also checking if the next page token was populated.


cretz (approved):
Did this surface somewhere recently? (can discuss off PR if needed)

Sushisource (approved):
Yes, this was fixed in Core a while back

PR #1892, opened 2023-10-10
Release Java SDK v1.22.0

Quinn-With-Two-Ns:
Release Java SDK v1.22.0

### DIFF
--- a/releases/v1.21.2
+++ b/releases/v1.21.2
@@ -1,0 +1,12 @@
+# Highlights
+
+## Bugfixes
+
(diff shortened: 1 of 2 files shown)


releases/v1.21.2
cretz, line 1: Is this just missing from last release?

cretz (approved):

releases/v1.21.2
Quinn-With-Two-Ns: yes, because `1.21.2` was off a branch

Issue #1722, opened 2023-03-28
Signal with a local activity in the last Workflow Task hangs the workflow execution


Spikhalskiy:
"last workflow task of the workflow" in this ticket means the WFT that will complete workflow method execution and will lead to `CompleteWorkflowExecution` command. 

## Expected Behavior

If the workflow receives a signal in the last workflow task, this signal should be successfully processed first and the workflow should complete after that.

## Actual Behavior

If workflow receives a signal with the last workflow task and this signal starts a local activity and the workflow method is not explicitly waiting for the signal completion, it leads to 

```
io.grpc.StatusRuntimeException: INVALID_ARGUMENT: invalid command sequence: [CompleteWorkflowExecution, RecordMarker], command CompleteWorkflowExecution must be the last command.
```

See https://github.com/Spikhalskiy/java-sdk/commit/c0a397ec4e5e0101b70f417ff4d244df7c291c26 for a test reproducing the problem.

## Proposed solutions

One of:

1. Completion of the workflow method shouldn't lead to an immediate `CompleteWorkflowExecution` command. Instead, `CompleteWorkflowExecution` should be issued at the end of that workflow task.
2. Completion of the workflow method shouldn't lead to an immediate `CompleteWorkflowExecution` command. Instead, `CompleteWorkflowExecution` should be issued at the end of all workflow threads (including Async and Signal)
3. If the first two are hard to implement and maintain backward compatibility, CompleteWorkflowExecution` should be the last command written. All commands after it should be discarded. This is the least graceful way of handling this situation.

longquanzheng:
Hey @Spikhalskiy , probably just ciruous, is that only an issue with local activity? What about other commands like activity scheduled or childWF or upsert search attributes?

Spikhalskiy:
It's true for any command happening in a signal after a yield if the workflow method doesn't wait for the signal method completion.

Issue #1281, opened 2022-06-25
Local Activity asynchronously started at the end of the workflow implementation leads to failure


Spikhalskiy:
## Expected Behavior

```
    @Override
    public void wofklowMethod() {
      Async.procedure(activity::execute);
    }
```

An activity scheduled at the end of the workflow code asynchronously without waiting on the promise should be ignored.
Alternatively, it may be executed. 
Such a code shouldn't block workflow completion, it's not an invalid workflow implementation.

## Actual Behavior

It leads to
`io.grpc.StatusRuntimeException: NOT_FOUND: workflow execution already completed` exception on worker with workflow never completing.

## Steps to Reproduce the Problem
Unit test: https://github.com/Spikhalskiy/java-sdk/commit/718fe8e794a50188c639e79a6ee47576960ae353

bergundy:
@Spikhalskiy please add to sdk features. Not 100% sure we put it there

Spikhalskiy:
Related to #1722

Quinn-With-Two-Ns:
https://github.com/temporalio/sdk-java/commit/8b3be3bb521f7ab62b7d384ee417d053286a95d5

PR #1902, opened 2023-10-20
Don't schedule LA on completed workflow

Quinn-With-Two-Ns:
Don't schedule LA on completed workflows

closes https://github.com/temporalio/sdk-java/issues/1722
closes https://github.com/temporalio/sdk-java/issues/1281

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowRunTaskHandler.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowRunTaskHandler.java
@@ -163,7 +163,9 @@ class ReplayWorkflowRunTaskHandler implements WorkflowRunTaskHandler {
       }
 
       handleWorkflowTaskImpl(workflowTask, historyIterator);
-      processLocalActivityRequests(wftHearbeatDeadline);
+      if (!context.isWorkflowMethodCompleted()) {
+        processLocalActivityRequests(wftHearbeatDeadline);
+      }
       List<Command> commands = workflowStateMachines.takeCommands();
       List<Message> messages = workflowStateMachines.takeMessages();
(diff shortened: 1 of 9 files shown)


temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowRunTaskHandler.java
```diff
@@ -163,7 +163,9 @@ public WorkflowTaskResult handleWorkflowTask(
       }
 
       handleWorkflowTaskImpl(workflowTask, historyIterator);
-      processLocalActivityRequests(wftHearbeatDeadline);
+      if (!context.isWorkflowMethodCompleted()) {
+        processLocalActivityRequests(wftHearbeatDeadline);
```
cretz, line 167: Can you help me understand what is the exact behavior today? Do we start the LA but the server fails to accept the result marker command since workflows are done? Are there any concerns that people starting LAs at end of workflow but not waiting on them still expect that code to be reached and we are not calling it anymore?

Quinn-With-Two-Ns: > Do we start the LA but the server fails to accept the result marker command since workflows are done

Correct

>Are there any concerns that people starting LAs at end of workflow but not waiting on them still expect that code to be reached and we are not calling it anymore?

I don't have any concern here because if someone was starting LAs at end of workflow but not waiting on them their workflow could never complete

cretz (approved):
Confirmed that this is expected to only affect people whose workflow task completions were being rejected anyways

Quinn-With-Two-Ns:
Found some edge cases that the current fix does not address, going to have to make some serious modifications to the state machine

Issue #1915, opened 2023-10-28
Java SDK should verify it replays up to `StartedEventId` when processing a workflow task


Quinn-With-Two-Ns:
## Expected Behavior
Java SDK should verify it replays up to `StartedEventId` sent on the poll request when processing a workflow task

## Actual Behavior
Java SDK does not verify it replays up to `StartedEventId` sent on the poll request when processing a workflow task. This can lead to the Java SDK not replaying the full workflow history if it is fetching history using `GetWorkflowExecutionHistory` after receiving a partial history with no cached execution and hitting a stale history node.

PR #1916, opened 2023-10-28
Verify history is replayed up to StartedEventId

Quinn-With-Two-Ns:
Verify history is replayed up to StartedEventId. Previously the Java SDK did not verify it repalyed up to StartedEventId sent on the poll request when processing a workflow task. This lead to the Java SDK not replaying the full workflow history if it was fetching history using GetWorkflowExecutionHistory after receiving a partial history with no cached execution and hitting a stale history node.

Go SDK also has a similar check: https://github.com/temporalio/sdk-go/blob/3cccbdd5f47d640fffe6651e4f49c236a4803a47/internal/internal_task_handlers.go#L383

closes https://github.com/temporalio/sdk-java/issues/1915

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowRunTaskHandler.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowRunTaskHandler.java
@@ -228,20 +228,22 @@ class ReplayWorkflowRunTaskHandler implements WorkflowRunTaskHandler {
     workflowStateMachines.setWorklfowStartedEventId(workflowTask.getStartedEventId());
     workflowStateMachines.setReplaying(workflowTask.getPreviousStartedEventId() > 0);
     workflowStateMachines.setMessages(workflowTask.getMessagesList());
-    applyServerHistory(historyIterator);
+    applyServerHistory(workflowTask.getStartedEventId(), historyIterator);
   }
 
-  private void applyServerHistory(WorkflowHistoryIterator historyIterator) {
+  private void applyServerHistory(long lastEventId, WorkflowHistoryIterator historyIterator) {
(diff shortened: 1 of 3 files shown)


temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowStoreImpl.java
```diff
@@ -351,7 +348,7 @@ public void sendQueryTask(
       task.setPreviousStartedEventId(previousStaredEventId);
       // it's not a real workflow task and the server sends 0 for startedEventId for such a workflow
       // task
-      task.setStartedEventId(startedEventId);
+      task.setStartedEventId(0);
```
Quinn-With-Two-Ns, line 351: Can't say why  this was done since the comment in the code acknowledges it should be zero

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowRunTaskHandler.java
```diff
@@ -264,13 +268,29 @@ private void applyServerHistory(WorkflowHistoryIterator historyIterator) {
           timerStopped = true;
         }
       }
+      if (!replayFailed) {
```
cretz, line 271: How can `replayFailed` ever be `true` on this line?

Sushisource (approved):

cretz (approved):

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowRunTaskHandler.java
Quinn-With-Two-Ns: you're right that was an error

Issue #1862, opened 2023-09-13
Support build ID reachability API


Quinn-With-Two-Ns:
Build ID reachability API should be supported in the Java SDK

https://docs.temporal.io/workers#build-id-reachability

PR #1919, opened 2023-10-31
Add getWorkerTaskReachability API

Quinn-With-Two-Ns:
Add getWorkerTaskReachability API

closes  https://github.com/temporalio/sdk-java/issues/1862

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/client/BuildIdReachability.java
+++ b/temporal-sdk/src/main/java/io/temporal/client/BuildIdReachability.java
@@ -24,15 +24,16 @@ import io.temporal.api.enums.v1.TaskReachability;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 /** Contains information about the reachability of a specific Build ID. */
 public class BuildIdReachability {
   private final Map<String, List<TaskReachability>> taskQueueReachability;
-  private final List<String> unretrievedTaskQueues;
+  private final Set<String> unretrievedTaskQueues;
 
   public BuildIdReachability(
       Map<String, List<TaskReachability>> taskQueueReachability,
-      List<String> unretrievedTaskQueues) {
+      Set<String> unretrievedTaskQueues) {
     this.taskQueueReachability = taskQueueReachability;
     this.unretrievedTaskQueues = unretrievedTaskQueues;
   }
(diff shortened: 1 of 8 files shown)


temporal-sdk/src/main/java/io/temporal/client/BuildIdReachability.java
```diff
@@ -0,0 +1,55 @@
...
+    this.taskQueueReachability = taskQueueReachability;
+    this.unretrievedTaskQueues = unretrievedTaskQueues;
+  }
+
+  /**
+   * @return A map of Task Queue names to the reachability status of the Build ID on that queue. If
+   *     the value is an empty list, the Build ID is not reachable on that queue.
+   */
+  public Map<String, List<TaskReachability>> getTaskQueueReachability() {
+    return Collections.unmodifiableMap(taskQueueReachability);
```
cretz, line 45: I think this should be done in the constructor or by the constructor's caller (same with unmodifiable list below)

temporal-sdk/src/main/java/io/temporal/client/BuildIdReachability.java
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  /**
+   * @return a list of any Task Queues that could not be retrieved because the server limits the
+   *     number that can be queried at once.
+   */
+  public List<String> getUnretrievedTaskQueues() {
```
cretz, line 52:
```suggestion
  public Set<String> getUnretrievedTaskQueues() {
```

Not required as a set, but arguably clearer.

temporal-sdk/src/main/java/io/temporal/client/BuildIdReachability.java
Quinn-With-Two-Ns: I would prefer it to be here then in the constructor or trust the caller to have used the correct type

temporal-sdk/src/main/java/io/temporal/client/BuildIdReachability.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+/** Contains information about the reachability of a specific Build ID. */
+public class BuildIdReachability {
+  private final Map<String, List<TaskReachability>> taskQueueReachability;
+  private final List<String> unretrievedTaskQueues;
+
+  public BuildIdReachability(
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: ah good catch `WorkerTaskReachability` should also be public

Sushisource (approved):
Thanks for this! I hope this wasn't something I was supposed to go back and add and forgot about :grimacing:

temporal-sdk/src/main/java/io/temporal/client/BuildIdReachability.java
cretz: I see this is already merged, but curious why not the constructor? Why create a new object every time a getter is called if you don't need to?

PR #1923, opened 2023-11-05
Treat signal after workflow complete as NonDeterministicException

Quinn-With-Two-Ns:
Treat signal after workflow complete as NonDeterministicException. I see this exception confusing users as they think their is something wrong with the temporal service or the SDK when it is normally their workflow code.


Quinn-With-Two-Ns:
Note: not going to merge this right away, targeting v1.23.0

cretz (approved):

PR #1942, opened 2023-11-26
Allow specifying workflow id when starting the workflow

gnagy:
## What was changed

For now this is an experiment to gather feedback.

- Add a new method to `WorkflowStub`:  `WorkflowExecution startWithId(String workflowId, Object... args);`
- Add a new annotation `@WorkflowId` for a paramater on a `@WorkflowMethod`
- Extend `WorkflowInvocationHandler`: if a `@WorkflowId` is specified, start workflow with `startWithId()`
- Added tests

## Why?

Currently the `workflowId` can only be set in `WorkflowOptions`, along with other options that are basically static config for a workflow instance, whereas`workflowId` differs typically between workflow executions.

This makes injecting pre-configured workflow instances where they are to be started virtually impossible, which adds noise to the call site. With this new API, such code becomes possible (spring-like pseudo-code). The idea is that ExampleWorkflowImpl could be created by an IOC container for each request (e.g. in spring Request scope).

```
@Controller
class WorkflowController {

  @Autowired private ExampleWorkflow workflow;

  @PostMapping
  public void start(UUID idempotencyKey, Data data) {
    workflow.execute(idempotencyKey, data)
  }
}


@WorkflowInterface
public interface ExampleWorkflow {
  @WorkflowMethod
  String execute(@WorkflowId UUID idempotencyKey, Data data);
}
```

## Checklist

1. Closes: N/A

2. How was this tested: With the added junit tests

3. Any docs updates needed? Probably, as this is a public API addition.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/client/WorkflowInvocationHandler.java
+++ b/temporal-sdk/src/main/java/io/temporal/client/WorkflowInvocationHandler.java
@@ -179,14 +179,24 @@ class WorkflowInvocationHandler implements InvocationHandler {
     return Defaults.defaultValue(method.getReturnType());
   }
 
-  private static void startWorkflow(WorkflowStub untyped, Object[] args) {
+  private static void startWorkflow(
+      POJOWorkflowMethodMetadata methodMetadata, WorkflowStub untyped, Object[] args) {
     Optional<WorkflowOptions> options = untyped.getOptions();
     if (untyped.getExecution() == null
         || (options.isPresent()
             && options.get().getWorkflowIdReusePolicy()
                 == WorkflowIdReusePolicy.WORKFLOW_ID_REUSE_POLICY_ALLOW_DUPLICATE)) {
       try {
-        untyped.start(args);
+        if (methodMetadata.getWorkflowIdParameterIndex().isPresent()) {
+          Object workflowId = args[methodMetadata.getWorkflowIdParameterIndex().get()];
+          if (workflowId == null) {
+            untyped.start(args);
+          } else {
+            untyped.startWithId(workflowId.toString(), args);
+          }
+        } else {
+          untyped.start(args);
+        }
       } catch (WorkflowExecutionAlreadyStarted e) {
         // We do allow duplicated calls if policy is not AllowDuplicate. Semantic is to wait for
         // result.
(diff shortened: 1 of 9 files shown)


Quinn-With-Two-Ns:
`WorkflowStub`  is meant to represent  a signal workflow execution. I am not a fan of adding multiple ways to specify `workflowID` overwriting each other. If you do want this functionality in your own code it seems you could easily provide this in wrapper class or utility function and does not need to be part of the SDK.

I am strongly against adding a  `WorkflowId` to `WorkflowMethod` since I don't see a strong value in forcing a certain workflow type to always have the same ID. It also wouldn't work if the workflow is launched through an untyped stub, I could see that causing a lot of confusion for users.

Quinn-With-Two-Ns:
My apologize, didn't mean to immediately close

gnagy:
Thank you for the feedback. 

> I am strongly against adding a `WorkflowId` to `WorkflowMethod` since I don't see a strong value in forcing a certain workflow type to always have the same ID. 

To clarify my idea, `@WorkflowId` is optional, meaning users don't have to add it to their `@WorkflowMethod` signature. However, if specified, the ID will be taken from the method argument, so ID is not the same for a certain workflow type -- or maybe I'm misunderstanding your comment?

> I am not a fan of adding multiple ways to specify workflowID overwriting each other

I see your point. I was wondering about it too, and ended up implementing it in a way where if @WorkflowId parameter is null, the ID is taken from WorkflowOptions thus facilitating a fallback to default mechanism. So, if a caller doesn't specify a workflow id (e.g. in a http request) then it can fall back to some pre-defined value. This could be made nicer with `Optional` of course.

This is totally opt-in, if there is no parameter annotated with `@WorkflowId`, the behavior is exactly as before.

> It also wouldn't work if the workflow is launched through an untyped stub, I could see that causing a lot of confusion for users.

For untyped stubs the original start() method is left unaffected. On one hand not breaking APIs is a good thing, also it just seemed easier to implement with a separate startWithId() method.

Quinn-With-Two-Ns:
I understand  you left  `start()` unchanged, that does not mean the new API is something Temporal would want to add for the reasons I listed above.

gnagy:
Ok thanks. I'll explore solving my needs with a wrapper class / utility function instead.

Issue #1626, opened 2023-01-27
Allow WorkflowImplementationOptions to be passed in TestWorkflowExtension->setWorkflowTypes


tsurdilo:
TestWorkflowRule:

       public Builder setWorkflowTypes(
           WorkflowImplementationOptions implementationOptions, Class<?>... workflowTypes) { ... }

TestWorkflowExtention:

       public Builder setWorkflowTypes(Class<?>... workflowTypes) { ... }

Allow WorkflowImplementationOptions to also be passed in setWorkflowTypes in TestWorkflowExtention.

picpromusic:
@tsurdilo is there a workaround when working with TestWorkflowExtension?

PR #1947, opened 2023-12-04
Allow WorkflowImplementationOptions to be passed in TestWorkflowExtension (#1626)

picpromusic:
(/)

## What was changed
It is possible to configure WorkflowImplementationsOptions using TestWorkflowExtension.
https://github.com/temporalio/sdk-java/issues/1626

## Why?
In order to configure WorkflowImplementationOptions  (like default ActivityOptions) using TestWorkflowExtension two new Methods registerWorkflowImplementationTypes are implemented. I choosed registerWorkflowImplementationTypes and not setWorkflowTypes because I found the registerWorkflowImplementationTypes like the one needed to configure WorkflowImplementationOptions to a worker in "non-test" Code. 

## Checklist
1. Closes [1626](https://github.com/temporalio/sdk-java/issues/1626)

2. How was this tested:
With my tests in a different project and through 2 Unit-Tests. One (TestWorkflowImplementationOptions.java) providing WorkflowImplementationOptions and one not providing the options (which would run into end endless loop). I did not find a more direct-way to tests this. So just testing side-effects of the implementation. An because side-effects might change in future TestWorkflowImplementationOptionsViceVersa.java for testing that TestWorkflowImplementationOptions.java would brake if the side-effect changes. 

3. Any docs updates needed?
I did not find a documentation that needs to be adapted. But https://github.com/temporalio/documentation/blob/main/docs-src/java/testing-frameworks.md should be fixed, It contains a broken link to https://github.com/temporalio/samples-java/tree/main/src/test/java/io/temporal/samples which should be https://github.com/temporalio/samples-java/tree/main/core/src/main/java/io/temporal/samples. As it is a different repository. I will open up another pull-request for it.

### DIFF
--- a/temporal-testing/src/main/java/io/temporal/testing/TestWorkflowExtension.java
+++ b/temporal-testing/src/main/java/io/temporal/testing/TestWorkflowExtension.java
@@ -31,13 +31,16 @@ import io.temporal.serviceclient.WorkflowServiceStubsOptions;
 import io.temporal.worker.Worker;
 import io.temporal.worker.WorkerFactoryOptions;
 import io.temporal.worker.WorkerOptions;
+import io.temporal.worker.WorkflowImplementationOptions;
 import io.temporal.workflow.DynamicWorkflow;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Parameter;
 import java.time.Instant;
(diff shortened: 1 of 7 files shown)


picpromusic:
again wrong user

Issue #1938, opened 2023-11-19
`toProtoDuration` may throw an `ArithmeticException` if Duration is too large


Quinn-With-Two-Ns:
When the SDK converts a Java `Duration` to a `protobuf.Duration` it uses `toMillis()`, but if the Java `Duration` represents a very large value converting to milliseconds may result in an overflow. We should probably convert the seconds directly and truncate the nanoseconds to millisecond precision to maintain consistency.

```
  public static com.google.protobuf.Duration toProtoDuration(Duration d) {
    // TODO we should refactor an implicit conversion of empty values into ZERO and rename the
    // current method into toJavaDurationSafe, toJavaDurationOrDefault or something like that
    if (d == null) {
      return Durations.ZERO;
    }

    return Durations.fromMillis(d.toMillis());
  }

```

PR #1950, opened 2023-12-06
Fix ArithmeticException in toJavaDuration.

chronos-tachyon:
Fixes https://github.com/temporalio/sdk-java/issues/1938 aka SDK-1319.

## What was changed
Eliminate risk of ArithmeticException due to `long` overflow in `toJavaDuration`.

## Why?
Make the API exception-safe for extreme duration values, especially for values which are legal in the protobuf wire format but not for Java's `Duration` type.

## Checklist

1. Closes [GitHub Issue #1938](https://github.com/temporalio/sdk-java/issues/1938).

2. How was this tested: created new tests in temporal-sdk module that captured existing behavior with small values to prove that the visible behavior was unchanged for already-legal values.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/common/ProtobufTimeUtils.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/common/ProtobufTimeUtils.java
@@ -31,6 +31,12 @@ import javax.annotation.Nullable;
 
 public class ProtobufTimeUtils {
 
+  static final long MAX_SECONDS = 315_576_000_000L;
+  static final long MIN_SECONDS = -315_576_000_000L;
+  static final int MAX_NANOS = 999_999_999;
+  static final int MIN_NANOS = -999_999_999;
+  static final int MILLIS_PER_NANO = 1_000_000;
+
   /**
    * Converts a Protobuf Duration to a Java Duration with millisecond precision. Null inputs are
    * treated as zero.
@@ -40,7 +46,27 @@ public class ProtobufTimeUtils {
     if (Objects.isNull(d)) {
       return Duration.ZERO;
     }
-    return Duration.ofMillis(Durations.toMillis(d));
+
+    // NB: Durations.toMillis is tempting, but it can throw ArithmeticException if
+    // Durations.isValid would return false, e.g. if the proto contains a number of
+    // seconds outside [MIN_SECONDS,MAX_SECONDS].  We don't want to throw an exception
+    // under any circumstances, so we clip it to the range using saturating arithmetic,
+    // since that's probably the best we can do to reflect the user's intent.
+    //
+    // Additionally, it's worth noting that Durations.toMillis truncates toward zero,
+    // just as we do here.  This behavior is correctly documented in the JavaDoc, but
+    // the exact wording is confusing: one can easily misread the phrase
+    // "rounded ... to the nearest millisecond" as a reference to "round to nearest"
+    // a.k.a. grade school rounding, where a fractional value of half or more gets
+    // rounded away from zero.  However, a careful reading of the JavaDoc shows that
+    // the doc author does not mean "nearest" in that way.
+    //
+    final long rawSeconds = d.getSeconds();
+    final int rawNanos = d.getNanos();
+    final long saturatedSeconds = Math.min(MAX_SECONDS, Math.max(MIN_SECONDS, rawSeconds));
+    final int saturatedNanos = Math.min(MAX_NANOS, Math.max(MIN_NANOS, rawNanos));
+    final int roundedNanos = (saturatedNanos / MILLIS_PER_NANO) * MILLIS_PER_NANO;
+    return Duration.ofSeconds(saturatedSeconds, roundedNanos);
   }
 
   /**
@@ -52,6 +78,9 @@ public class ProtobufTimeUtils {
     if (Objects.isNull(d)) {
       return Durations.ZERO;
     }
+
+    // NB: the overflow behavior is not documented, but Duration.toMillis seems to
+    // use saturating arithmetic, which is what we want.
     return Durations.fromMillis(d.toMillis());
   }
 
--- a/temporal-sdk/src/main/java/io/temporal/internal/common/ProtobufTimeUtils.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/common/ProtobufTimeUtils.java
@@ -25,28 +25,33 @@ import com.google.protobuf.util.Durations;
 import com.google.protobuf.util.Timestamps;
 import java.time.Duration;
 import java.time.Instant;
+import java.util.Objects;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
 public class ProtobufTimeUtils {
+
+  /**
+   * Converts a Protobuf Duration to a Java Duration with millisecond precision. Null inputs are
+   * treated as zero.
+   */
   @Nonnull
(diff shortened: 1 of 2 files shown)


Quinn-With-Two-Ns:
One thing I called out in the [issue](https://github.com/temporalio/sdk-java/issues/1938), but may not have mentioned is we should maintain the precision of the conversion.

>We should probably convert the seconds directly and truncate the nanoseconds to millisecond precision to maintain consistency

changing the conversion precision might subtly break users tests using the time skipping test server.

Quinn-With-Two-Ns:
For the CI failure you just need to rebase

cretz:
I am not sure this change is backwards compatible. While I understand the reasoning for preventing overflow, the loss of precision I think is intentional. No use of proto duration by the Java SDK or the Temporal server is expected to be more precise than the millisecond level.

`Durations.toMillis` rounds. If I my workflow run timeout is 1.0005 seconds and I have this code:

```java
if (Workflow.getInfo().getWorkflowRunTimeout().toMillis() > 1000) {
    myActivities.doThing();
}
```

This PR changes that conditional. Now, I admit this is a contrived example and I'm not _that_ worried about this incompatibility, but it should be known.

I think we should stick with millisecond precision for all Temporal duration use (this doesn't apply to users' use of protos). Can you help me understand a situation where losing the precision has consequence?

chronos-tachyon:
> One thing I called out in the [issue](https://github.com/temporalio/sdk-java/issues/1938), but may not have mentioned is we should maintain the precision of the conversion.
> 
> > We should probably convert the seconds directly and truncate the nanoseconds to millisecond precision to maintain consistency
> 
> changing the conversion precision might subtly break users tests using the time skipping test server.

Whoops, I overlooked that.  Fixed.

cretz:
I still contend we should be ok with loss of precision and just fix the overflow part

temporal-sdk/src/main/java/io/temporal/internal/common/ProtobufTimeUtils.java
```diff
@@ -25,29 +25,83 @@
...
+    if (Objects.isNull(d)) {
       return Durations.ZERO;
     }
+    return toProtoDurationExact(roundToMillis(d));
+  }
 
-    return Durations.fromMillis(d.toMillis());
+  /** Converts a Protobuf Duration to a Java Duration with nanosecond precision. */
+  @Nonnull
+  public static Duration toJavaDurationExact(@Nonnull com.google.protobuf.Duration d) {
```
cretz, line 63:
```suggestion
  private static Duration toJavaDurationExact(@Nonnull com.google.protobuf.Duration d) {
```

temporal-sdk/src/main/java/io/temporal/internal/common/ProtobufTimeUtils.java
```diff
@@ -25,29 +25,83 @@
...
+    //
+    // In the Protobuf to Java direction, the two-arg overload of Duration.ofSeconds already
+    // handles all cases correctly, so we don't have to do anything else but call it.
+    //
+    return Duration.ofSeconds(d.getSeconds(), d.getNanos());
+  }
+
+  /** Converts a Java Duration to a Protobuf Duration with nanosecond precision. */
+  @Nonnull
+  public static com.google.protobuf.Duration toProtoDurationExact(@Nonnull Duration d) {
```
cretz, line 75:
```suggestion
  private static com.google.protobuf.Duration toProtoDurationExact(@Nonnull Duration d) {
```

cretz (commented):
To confirm, with the current changes, the behavior is exactly as it is today for normal positive durations that don't overflow and only handles overflow but not the loss of precision? The PR title could be confusing if we removed the fix-loss-of-precision requirement, which is fine.

May want a test or two.

temporal-sdk/src/main/java/io/temporal/internal/common/ProtobufTimeUtils.java
```diff
@@ -77,4 +131,12 @@ public static com.uber.m3.util.Duration toM3DurationSinceNow(Timestamp t) {
 
     return Timestamp.newBuilder().setSeconds(t.getEpochSecond()).setNanos(t.getNano()).build();
   }
+
+  @Nonnull
+  private static Duration roundToMillis(@Nonnull Duration exact) {
+    final long s = exact.getSeconds();
+    final int ns = exact.getNano();
+    final int ms = (ns + NANOS_PER_MILLI / 2) / NANOS_PER_MILLI;
```
Quinn-With-Two-Ns, line 139: Couldn't we take the seconds and subtract them from `exact` and call `toMillis`? `toMillis` looks like it is doing a slightly different conversion. I worry about breaking any time skipping tests comparing exact timestamps
```
millis = Math.addExact(millis, tempNanos / NANOS_PER_MILLI);
```

Can we also add some tests for the conversion?

temporal-sdk/src/test/java/io/temporal/internal/common/ProtobufTimeUtilsTest.java
```diff
@@ -0,0 +1,115 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameters;
+
+@RunWith(Parameterized.class)
+public class ProtobufTimeUtilsTest {
+
+  @Parameters
+  public static Collection<Object[]> data() {
```
Quinn-With-Two-Ns, line 39: Can we confirm all the tests that didn't throw have the same behaviour before and after your change?

chronos-tachyon: Yes, I specifically tested that in the first commit after the rebase + force-push.

chronos-tachyon: Happily, the belief that `Durations.toMillis` does round to nearest instead of truncation was mistaken, as that would have meant that there was an asymmetry in the conversion.  Both `Duration.toMillis` (Java) and `Durations.toMillis` (protobuf) truncate.

chronos-tachyon:
> To confirm, with the current changes, the behavior is exactly as it is today for normal positive durations that don't overflow and only handles overflow but not the loss of precision? The PR title could be confusing if we removed the fix-loss-of-precision requirement, which is fine.
> 
> May want a test or two.

It now has tests and I've retitled the pull request + interactive rebased the commit messages to better reflect what's going on.

Quinn-With-Two-Ns (approved):
LGTM, first commit 🎉

Quinn-With-Two-Ns (approved):

chronos-tachyon:
> `Durations.toMillis` rounds. What if my workflow run timeout is 1.0005 seconds and I have this code:
> 
> ```java
> if (Workflow.getInfo().getWorkflowRunTimeout().toMillis() > 1000) {
>     myActivities.doThing();
> }
> ```
> 
> This PR changes that conditional I think.

This turns out to be factually false, but the documentation is confusingly worded.  [The `Durations.toMillis` JavaDoc](https://protobuf.dev/reference/java/api-docs/com/google/protobuf/util/Durations.html#toMillis-com.google.protobuf.Duration-) states: "The result will be rounded towards 0 to the nearest millisecond. E.g., if the duration represents -1 nanosecond, it will be rounded to 0."  The word "nearest" is causing confusion here, I think.

PR #1957, opened 2023-12-16
Removed heardcoded 10 second timeout for an activity under test

mfateev:
## What was changed
Removed hardcoded 10-second timeout for TestActivityEnvironment.

## Why?
We want to support unit testing activities that take longer than 10 seconds.

Reported at https://community.temporal.io/t/activities-run-in-tests-using-testactivityenvironment-intermittently-fails-with-java-util-concurrent-timeoutexception/10397/3

### DIFF
--- a/temporal-testing/src/main/java/io/temporal/testing/TestActivityEnvironmentInternal.java
+++ b/temporal-testing/src/main/java/io/temporal/testing/TestActivityEnvironmentInternal.java
@@ -363,16 +363,12 @@ public final class TestActivityEnvironmentInternal implements TestActivityEnviro
                       localActivity));
 
       try {
-        // 10 seconds is just a "reasonable" wait to not make an infinite waiting
-        return activityFuture.get(10, TimeUnit.SECONDS);
+        return activityFuture.get();
       } catch (InterruptedException e) {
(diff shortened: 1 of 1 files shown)


temporal-testing/src/main/java/io/temporal/testing/TestActivityEnvironmentInternal.java
```diff
@@ -363,17 +363,13 @@ private Result executeActivity(
                       localActivity));
 
       try {
-        // 10 seconds is just a "reasonable" wait to not make an infinite waiting
-        return activityFuture.get(10, TimeUnit.SECONDS);
+        return activityFuture.get();
```
Quinn-With-Two-Ns, line 366: Shouldn't the timeout be based on the activities timeout?

mfateev: I don't think anyone is using real activity timeouts in the TestActivityEnvironment.

Quinn-With-Two-Ns (approved):

lr-josephcho:
Just wondering, when would this change be available? If it's on the next release, would it be possible to know approx. when that would be?

Issue #1959, opened 2023-12-28
Missing Retry Policy on Continue-As-New


MatthewHou:
## Expected Behavior
RetryPolicy should either be propagated or accepted as part of the `ContinueAsNewOptions` for Contine-As-New

## Actual Behavior
RetryPolicy is not set in the new workflow execution when `contineAsNew` is called. 
And we are not able to explicitly set `RetryOptions` when calling `contineAsNew`  since it is not included in the `ContinueAsNewOptions`. 

## Steps to Reproduce the Problem

  1. Start a workflow with a retry policy
  2. Workflow continued as new
  3. The retry policy is not set for the new workflow execution

## Specifications

  - Version: Java SDK 1.22.3
  - Platform: Java SDK

MatthewHou:
Retry Options and other attributes are propagated by default in Go SDK and Python/TS SDK. This should help Java SDK to align the behaviour on continue-as-new. Please see the following issues

- https://github.com/temporalio/sdk-go/issues/695 
- https://github.com/temporalio/sdk-core/issues/372

Quinn-With-Two-Ns:
closed by https://github.com/temporalio/sdk-java/pull/1961

Issue #1201, opened 2022-05-11
Revisit continue-as-new


Spikhalskiy:
# Expected Behavior

Users calling `Workflow.continueAsNew` without an explicit `ContinueAsNewOptions` expect most workflow properties to carry over to the new run.

## Actual Behavior

Right now `continueAsNew` is implemented in a way that assumes that most parameters are carried over by the server if not specified on the continueAsNew request. While in fact, it's the opposite. Server doesn't carry over most parameters are not carried over by the server.

## Additional context

This task requires a careful alignment with GoSDK and documenting which parameters are carried-over by the Server.
Also, Test server should be updated to be aligned in its behavior with the real Temporal Server.

Quinn-With-Two-Ns:
related: https://github.com/temporalio/sdk-java/issues/1424

PR #1961, opened 2023-12-29
Added workflow retry policy propagation during continue-as-new and to ContinueAsNewOptions.

mfateev:
## What was changed
After this change if a worklfow has RetryOptions configured then they are copied to the continued workflow.

Added RetryOptions to ContinueAsNewOptions to allow overriding them when calling contineu-as-new.

Added WorkflowInfo.getRetryOptions to give access to the options from the worklfow code.

## Why?
Currently there is no ability to specify retry options to the next workflow execution when calling continue-as-new.


Quinn-With-Two-Ns (approved):
Thanks for this,

Technically this is a backwards incompatible change, but we where OK with it in the Go SDK so I see no reason not to allow it here.

Quinn-With-Two-Ns:
closes https://github.com/temporalio/sdk-java/issues/1959

Quinn-With-Two-Ns:
related https://github.com/temporalio/sdk-java/issues/1201

MatthewHou:
@mfateev Thank you for the fix! I have tested out your change locally and noticed 2 things:
1.  An empty `ContinueAsNewOptions` needs to be provided when calling `continueAsNew` for the default `RetryOptions` to be copied over. This seems to be related to the null check [here](https://github.com/temporalio/sdk-java/pull/1961/files#diff-0dd6530b3cbc81275990c91ae5132c38067aea036893ae87771134ed23b60eb4R1099). I'm not sure if this the expected behaviour. 
2. The retry attempt count resets to 1 in the continued workflow. I guess this is a server side bug related to this issue? 
https://github.com/temporalio/temporal/issues/5249

mfateev:
@MatthewHou 

1. Good catch. I've updated the PR to copy RetryOptions on empty ContinueAsNewOptions. 
2. This is by design, as retries are reset on each continue-as-new.

Issue #1973, opened 2024-01-19
WorkflowUpdateException incorrectly wrapped by ExecutionException in some edge case


Quinn-With-Two-Ns:
`WorkflowUpdateException` may be wrapped in a `ExecutionException` on a sync update call if the update request could not finish in the servers long poll deadline and the SDK switched to long polling the update request. This can happen if the update took longer then the servers long poll deadline request deadline was longer then the servers long poll deadline. This is undesirable since from the user perspective they did not wait on any future so the exception being wrapped in a `ExecutionException` is unexpected.

https://github.com/temporalio/sdk-java/blob/806eab71f2b65f762cdd1c4079d3fbad0b62ae1d/temporal-sdk/src/main/java/io/temporal/client/WorkflowStubImpl.java#L307

PR #1974, opened 2024-01-20
Unwrap ExecutionException on sync update

Quinn-With-Two-Ns:
Unwrap ExecutionException on sync update. We shouldn't return the `ExecutionException` to caller since they aren't expecting an ExecutionException on a sync update.

closes https://github.com/temporalio/sdk-java/issues/1973

### DIFF
--- a/temporal-sdk/src/test/java/io/temporal/client/functional/UpdateLongPollTest.java
+++ b/temporal-sdk/src/test/java/io/temporal/client/functional/UpdateLongPollTest.java
@@ -1,0 +1,98 @@
+/*
+ * Copyright (C) 2022 Temporal Technologies, Inc. All Rights Reserved.
+ *
+ * Copyright (C) 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Modifications copyright (C) 2017 Uber Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this material except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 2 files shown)


temporal-sdk/src/test/java/io/temporal/client/functional/UpdateLongPollTest.java
```diff
@@ -0,0 +1,98 @@
...
+    workflow.close();
+    WorkflowStub.fromTyped(workflow).getResult(Void.class);
+  }
+
+  @Test(timeout = 3 * HISTORY_LONG_POLL_TIMEOUT_SECONDS * 1000)
+  public void testGetUpdateResultsFail() {
+    TestUpdatedWorkflow workflow = testWorkflowRule.newWorkflowStub(TestUpdatedWorkflow.class);
+    WorkflowClient.start(workflow::execute);
+    assertThrows(
+        WorkflowUpdateException.class,
```
cretz, line 57: Can you verify the cause here is the serialized application failure? (or confirm a test elsewhere confirms the proper thrown exception type from update)

Quinn-With-Two-Ns: Yes we have lots of other tests that verify the exception type from an update

cretz: :+1: I would have expected them to also confirm/deny `WorkflowUpdateException` was the properly thrown exception, but I am guessing they just validated the cause not the top-level type?

cretz (approved):

PR #1986, opened 2024-02-12
SpringBoot - Add registered workflow and activity impl info to workers template

tsurdilo:
Update to WorkersTemplate to expose registered workflow and activity class info per task queue. 
Allows users to build custom actuator endpoints that display this.

Sample that uses this info to create a custom actuator endpoint https://github.com/temporalio/samples-java/pull/576

### DIFF
--- a/temporal-spring-boot-autoconfigure-alpha/src/main/java/io/temporal/spring/boot/autoconfigure/template/WorkersTemplate.java
+++ b/temporal-spring-boot-autoconfigure-alpha/src/main/java/io/temporal/spring/boot/autoconfigure/template/WorkersTemplate.java
@@ -74,6 +74,7 @@ public class WorkersTemplate implements BeanFactoryAware, EnvironmentAware {
 
   private WorkerFactory workerFactory;
   private Collection<Worker> workers;
+  private final Map<String, RegisteredInfo> registeredInfo = new HashMap<>();
 
   public WorkersTemplate(
       @Nonnull TemporalProperties properties,
@@ -109,6 +110,13 @@ public class WorkersTemplate implements BeanFactoryAware, EnvironmentAware {
       this.workers = createWorkers(getWorkerFactory());
     }
     return workers;
+  }
+
+  public Map<String, RegisteredInfo> getRegisteredInfo() {
+    if (workers == null) {
+      this.workers = createWorkers(getWorkerFactory());
+    }
+    return registeredInfo;
   }
 
   WorkerFactory createWorkerFactory(WorkflowClient workflowClient) {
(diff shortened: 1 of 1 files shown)


tsurdilo:
Sample screenshot:

Screenshot 2024-02-11 at 7 30 23 PM

temporal-spring-boot-autoconfigure-alpha/src/main/java/io/temporal/spring/boot/autoconfigure/template/WorkersTemplate.java
```diff
@@ -410,6 +426,47 @@ private Worker createNewWorker(
...
+      registeredActivityInfo.add(activityClass);
+      return this;
+    }
+
+    public RegisteredInfo addWorkflowInfo(String workflowClass) {
+      registeredWorkflowInfo.add(workflowClass);
+      return this;
+    }
+
+    public List<String> getRegisteredActivityInfo() {
```
Quinn-With-Two-Ns, line 461: Do you think there is other information users may want us to expose in the future about activities or workflows other then their name?

Quinn-With-Two-Ns: May make sense to return a list of objects so we can add new fields if decided to expose more info.

tsurdilo: ok good point, will update

temporal-spring-boot-autoconfigure-alpha/src/main/java/io/temporal/spring/boot/autoconfigure/template/WorkersTemplate.java
```diff
@@ -111,6 +112,13 @@ public Collection<Worker> getWorkers() {
     return workers;
   }
 
+  public Map<String, RegisteredInfo> getRegisteredInfo() {
```
Quinn-With-Two-Ns, line 115: Does this need javadocs?

tsurdilo: done

tsurdilo:
@Quinn-With-Two-Ns made updates, let me know if ok now

Quinn-With-Two-Ns (changes requested):
Can we mark this api as `@Experimental`

tsurdilo:
Added Experimental flag to getRegisteredInfo method

Issue #1866, opened 2023-09-21
Allow to get the retry policy set for execution from client via WorkflowInfo


tsurdilo:
It's not advised to set workflow retry policy in almost all cases, but if client does set it for some reason it would be good to 
be able to check it in workflow code via Workflow.getInfo()

Quinn-With-Two-Ns:
https://github.com/temporalio/sdk-java/pull/1961

PR #1993, opened 2024-02-17
Expose the workflow execution retry policy from the client (#1866)

mohammed-ezzedine:
## What was changed
Added a getter to the WorkflowInfo interface to return the retry policy of a workflow execution

## Why?
Addressing requirement

1. Closes Issue 1866

2. How was this tested:
Manually, since there was no pre-existing testing class for WorkflowInfoImpl

3. Any docs updates needed?
No


mohammed-ezzedine:
just noticed that the method already existed

PR #1995, opened 2024-02-26
SpringBoot - add workflow and activity metadata to RegisteredInfo

tsurdilo:
Added Workflow and Activity metadata to registered info 
Tested with sample https://github.com/temporalio/samples-java/pull/576

Now users can get a lot more info about registered workflows and activities:

Screenshot 2024-02-25 at 9 40 45 PM

@Quinn-With-Two-Ns sorry should have added this in the initial pr :) your suggestion to make it object type so we can add more data to it was awesome, allowed for this


temporal-spring-boot-autoconfigure-alpha/src/main/java/io/temporal/spring/boot/autoconfigure/template/WorkersTemplate.java
```diff
@@ -484,6 +501,7 @@ public List<RegisteredWorkflowInfo> getRegisteredWorkflowInfo() {
   public static class RegisteredActivityInfo {
```
Quinn-With-Two-Ns, line 501: Can we mark these classes `RegisteredActivityInfo` and `RegisteredWorkflowInfo` as `@Experimental`? Sorry I should have suggested this in the last PR

Quinn-With-Two-Ns:
Can we also add some basic tests for this? just verify a workflow registered returns the correct metadata?

tsurdilo:
added test lets me know if ok to merge now

Issue #1996, opened 2024-02-27
SpringBoot - add "server-name" property to connection options


tsurdilo:
Feature request to add "server-name" to connection options. This can be useful for users going through PrivateLink or similar.
This would use ManagedChannelBuilder#overrideAuthority to set the server name directly rather than getting it from target.

Note this is currently possible to configure via TemporalOptionsCustomizer, but is useful for convenience.

cretz:
This is a TLS-specific option, can we make sure it is presented the same way as other TLS options?

PR #1998, opened 2024-02-27
SpringBoot - add server-name to mtls config options

tsurdilo:
adds server-name to mtls config. allows to override authority on ChannelInitializer when using mtls

closes https://github.com/temporalio/sdk-java/issues/1996


cretz:
Note, we should never have called this mTLS options. Server CA, trust manager, and SNI override are TLS options and are not specific to mTLS (which is "mutual" TLS and about client cert/key auth). Many users don't have mTLS setup but need to set things like SNI override.

PR #2000, opened 2024-02-29
Ignore history events with worker_may_ignore: true.

chronos-tachyon:
## What was changed
* Unknown event types with `worker_may_ignore: true` are skipped
* All other unknown event types continue to throw
* Added new replay tests verifying that `worker_may_ignore` is handled correctly

## Why?
The new `worker_may_ignore` flag is intended to mark events that can be handled as no-ops if the SDK doesn't know the event type.

## Checklist

1. Closes https://github.com/temporalio/sdk-java/issues/1945

2. How was this tested: added new unit tests to verify the new behavior before implementing it.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
@@ -58,7 +58,8 @@ public final class WorkflowStateMachines {
 
   /** Initial set of SDK flags that will be set on all new workflow executions. */
   private static final List<SdkFlag> initialFlags =
-      Collections.unmodifiableList(Arrays.asList(SdkFlag.SKIP_YIELD_ON_DEFAULT_VERSION));
+      Collections.unmodifiableList(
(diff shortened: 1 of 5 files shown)


cretz (approved):
LGTM, though may want @Quinn-With-Two-Ns to peek before merging.

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
```diff
@@ -430,7 +428,11 @@ private void handleSingleEvent(HistoryEvent event, boolean hasNextEvent) {
       replaying = false;
     }
 
-    Long initialCommandEventId = getInitialCommandEventId(event);
+    final long initialCommandEventId = getInitialCommandEventId(event);
```
Quinn-With-Two-Ns, line 431: Can we document `getInitialCommandEventId` and the fact that if it return < 0 the event can be ignored?

chronos-tachyon: Added a docstring, but also switched from `long` / `-1` to `OptionalLong` / `empty()` because it's less cognitive load.

Quinn-With-Two-Ns (approved):

PR #2004, opened 2024-03-07
Replace buildkite with github actions

Quinn-With-Two-Ns:
Switch from buildkite to github actions. I tried to keep the CI jobs themselves the same and leave any improvements there for follow up PRs


Sushisource (approved):

cretz (approved):
Awesome

PR #2014, opened 2024-03-22
Slot supplier interface & fixed-size implementation

Sushisource:
## What was changed
Implements the [slot management proposal](https://github.com/temporalio/proposals/pull/88)

## Why?
Users have expressed the desire for more control over when workers choose to poll for new tasks

## Checklist

1. Closes 

2. How was this tested:
Existing tests (adding some additional UTs in progress)

3. Any docs updates needed?
There will be, after this has had some time in preview

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/worker/EagerActivitySlotsReservation.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/worker/EagerActivitySlotsReservation.java
@@ -79,40 +79,27 @@ class EagerActivitySlotsReservation implements Closeable {
         activityTasksCount,
         this.reservedSlots.size());
 
-    // Release any slots that we won't be using
-    releaseSlots(this.reservedSlots.size() - activityTasksCount);
-
     for (PollActivityTaskQueueResponse act : serverResponse.getActivityTasksList()) {
-      // don't release slots here, instead the semaphore release reference is passed to the activity
-      // worker to release when the activity is done
+      // don't release slots here, instead the release function is called in the activity worker to
+      // release when the activity is done
       SlotPermit permit = this.reservedSlots.remove(0);
       this.eagerActivityDispatcher.dispatchActivity(act, permit);
     }
 
-    this.reservedSlots.clear();
+    // Release any remaining that we won't be using
+    try {
+      this.eagerActivityDispatcher.releaseActivitySlotReservations(this.reservedSlots);
+    } finally {
+      this.reservedSlots.clear();
+    }
   }
 
   @Override
   public void close() {
     if (!this.reservedSlots.isEmpty()) {
-      releaseSlots(this.reservedSlots.size());
-    }
-  }
-
-  private void releaseSlots(int slotsToRelease) {
-    if (slotsToRelease == 0) return;
-    if (slotsToRelease < 0)
-      throw new IllegalArgumentException("Trying to release a negative number of activity slots");
-    if (slotsToRelease > this.reservedSlots.size())
-      throw new IllegalStateException(
-          "Trying to release more activity slots than outstanding reservations");
-
-    List<SlotPermit> slotsToReleaseList = this.reservedSlots.subList(0, slotsToRelease);
-    try {
-      this.eagerActivityDispatcher.releaseActivitySlotReservations(slotsToReleaseList);
-    } finally {
-      // Actually remove the permits from the class-level storage
-      slotsToReleaseList.clear();
+      // Release all slots
+      this.eagerActivityDispatcher.releaseActivitySlotReservations(this.reservedSlots);
+      this.reservedSlots.clear();
     }
   }
 }
--- a/temporal-sdk/src/main/java/io/temporal/internal/worker/EagerActivitySlotsReservation.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/worker/EagerActivitySlotsReservation.java
@@ -100,7 +100,9 @@ class EagerActivitySlotsReservation implements Closeable {
   }
 
   private void releaseSlots(int slotsToRelease) {
-    if (slotsToRelease <= 0) return;
+    if (slotsToRelease == 0) return;
+    if (slotsToRelease < 0)
+      throw new IllegalArgumentException("Trying to release a negative number of activity slots");
     if (slotsToRelease > this.reservedSlots.size())
       throw new IllegalStateException(
           "Trying to release more activity slots than outstanding reservations");
--- a/temporal-sdk/src/main/java/io/temporal/internal/worker/EagerActivitySlotsReservation.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/worker/EagerActivitySlotsReservation.java
@@ -28,14 +28,18 @@ import io.temporal.api.workflowservice.v1.PollActivityTaskQueueResponse;
 import io.temporal.api.workflowservice.v1.RespondWorkflowTaskCompletedRequest;
 import io.temporal.api.workflowservice.v1.RespondWorkflowTaskCompletedResponse;
 import io.temporal.internal.Config;
+import io.temporal.worker.tuning.SlotPermit;
 import java.io.Closeable;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
 import javax.annotation.concurrent.NotThreadSafe;
 
 /** This class is not thread safe and shouldn't leave the boundaries of one activity executor */
 @NotThreadSafe
 class EagerActivitySlotsReservation implements Closeable {
   private final EagerActivityDispatcher eagerActivityDispatcher;
-  private int outstandingReservationSlotsCount = 0;
+  private final List<SlotPermit> reservedSlots = new ArrayList<>();
 
   EagerActivitySlotsReservation(EagerActivityDispatcher eagerActivityDispatcher) {
     this.eagerActivityDispatcher = eagerActivityDispatcher;
@@ -49,10 +53,14 @@ class EagerActivitySlotsReservation implements Closeable {
       ScheduleActivityTaskCommandAttributes commandAttributes =
           command.getScheduleActivityTaskCommandAttributes();
       if (!commandAttributes.getRequestEagerExecution()) continue;
+      boolean atLimit = this.reservedSlots.size() >= Config.EAGER_ACTIVITIES_LIMIT;
+      Optional<SlotPermit> permit = Optional.empty();
+      if (!atLimit) {
+        permit = this.eagerActivityDispatcher.tryReserveActivitySlot(commandAttributes);
+      }
 
-      if (this.outstandingReservationSlotsCount < Config.EAGER_ACTIVITIES_LIMIT
-          && this.eagerActivityDispatcher.tryReserveActivitySlot(commandAttributes)) {
-        this.outstandingReservationSlotsCount++;
+      if (permit.isPresent()) {
+        this.reservedSlots.add(permit.get());
       } else {
         mutableRequest.setCommands(
             i,
@@ -66,34 +74,43 @@ class EagerActivitySlotsReservation implements Closeable {
   public void handleResponse(RespondWorkflowTaskCompletedResponse serverResponse) {
     int activityTasksCount = serverResponse.getActivityTasksCount();
     Preconditions.checkArgument(
-        activityTasksCount <= this.outstandingReservationSlotsCount,
+        activityTasksCount <= this.reservedSlots.size(),
         "Unexpectedly received %s eager activities though we only requested %s",
         activityTasksCount,
-        this.outstandingReservationSlotsCount);
+        this.reservedSlots.size());
 
-    releaseSlots(this.outstandingReservationSlotsCount - activityTasksCount);
+    // Release any slots that we won't be using
+    releaseSlots(this.reservedSlots.size() - activityTasksCount);
 
     for (PollActivityTaskQueueResponse act : serverResponse.getActivityTasksList()) {
       // don't release slots here, instead the semaphore release reference is passed to the activity
       // worker to release when the activity is done
-      this.eagerActivityDispatcher.dispatchActivity(act);
+      SlotPermit permit = this.reservedSlots.remove(0);
+      this.eagerActivityDispatcher.dispatchActivity(act, permit);
     }
 
-    this.outstandingReservationSlotsCount = 0;
+    this.reservedSlots.clear();
   }
 
   @Override
   public void close() {
-    if (this.outstandingReservationSlotsCount > 0)
-      releaseSlots(this.outstandingReservationSlotsCount);
+    if (!this.reservedSlots.isEmpty()) {
+      releaseSlots(this.reservedSlots.size());
+    }
   }
 
   private void releaseSlots(int slotsToRelease) {
-    if (slotsToRelease > this.outstandingReservationSlotsCount)
+    if (slotsToRelease <= 0) return;
+    if (slotsToRelease > this.reservedSlots.size())
       throw new IllegalStateException(
           "Trying to release more activity slots than outstanding reservations");
 
-    this.eagerActivityDispatcher.releaseActivitySlotReservations(slotsToRelease);
-    this.outstandingReservationSlotsCount -= slotsToRelease;
+    List<SlotPermit> slotsToReleaseList = this.reservedSlots.subList(0, slotsToRelease);
+    try {
+      this.eagerActivityDispatcher.releaseActivitySlotReservations(slotsToReleaseList);
+    } finally {
+      // Actually remove the permits from the class-level storage
+      slotsToReleaseList.clear();
+    }
   }
 }
--- a/temporal-sdk/src/main/java/io/temporal/internal/worker/PollTaskExecutor.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/worker/PollTaskExecutor.java
@@ -21,12 +21,9 @@
 package io.temporal.internal.worker;
 
 import com.google.common.base.Preconditions;
-import com.uber.m3.tally.Scope;
 import io.temporal.internal.logging.LoggerTag;
-import io.temporal.worker.MetricsType;
 import java.util.Objects;
 import java.util.concurrent.*;
-import java.util.concurrent.atomic.AtomicInteger;
 import javax.annotation.Nonnull;
 import org.slf4j.MDC;
 
@@ -43,10 +40,8 @@ final class PollTaskExecutor<T> implements ShutdownableTaskExecutor<T> {
   private final String identity;
   private final TaskHandler<T> handler;
   private final PollerOptions pollerOptions;
-  private final Scope metricsScope;
 
   private final ThreadPoolExecutor taskExecutor;
-  private final AtomicInteger availableTaskSlots;
   private final String pollThreadNamePrefix;
 
   PollTaskExecutor(
@@ -56,14 +51,12 @@ final class PollTaskExecutor<T> implements ShutdownableTaskExecutor<T> {
       @Nonnull TaskHandler<T> handler,
       @Nonnull PollerOptions pollerOptions,
       int workerTaskSlots,
-      @Nonnull Scope metricsScope,
       boolean synchronousQueue) {
     this.namespace = Objects.requireNonNull(namespace);
     this.taskQueue = Objects.requireNonNull(taskQueue);
     this.identity = Objects.requireNonNull(identity);
     this.handler = Objects.requireNonNull(handler);
     this.pollerOptions = Objects.requireNonNull(pollerOptions);
-    this.metricsScope = Objects.requireNonNull(metricsScope);
 
     this.taskExecutor =
         new ThreadPoolExecutor(
@@ -78,9 +71,6 @@ final class PollTaskExecutor<T> implements ShutdownableTaskExecutor<T> {
             TimeUnit.SECONDS,
             synchronousQueue ? new SynchronousQueue<>() : new LinkedBlockingQueue<>());
     this.taskExecutor.allowCoreThreadTimeOut(true);
-
-    this.availableTaskSlots = new AtomicInteger(workerTaskSlots);
-    publishSlotsMetric();
 
     this.pollThreadNamePrefix =
         pollerOptions.getPollThreadNamePrefix().replaceFirst("Poller", "Executor");
@@ -97,8 +87,6 @@ final class PollTaskExecutor<T> implements ShutdownableTaskExecutor<T> {
     Preconditions.checkNotNull(task, "task");
     taskExecutor.execute(
         () -> {
-          availableTaskSlots.decrementAndGet();
-          publishSlotsMetric();
           try {
             MDC.put(LoggerTag.NAMESPACE, namespace);
             MDC.put(LoggerTag.TASK_QUEUE, taskQueue);
@@ -117,8 +105,6 @@ final class PollTaskExecutor<T> implements ShutdownableTaskExecutor<T> {
             //              throw (Error)e;
             //            }
           } finally {
-            availableTaskSlots.incrementAndGet();
-            publishSlotsMetric();
             MDC.remove(LoggerTag.NAMESPACE);
             MDC.remove(LoggerTag.TASK_QUEUE);
           }
@@ -155,11 +141,5 @@ final class PollTaskExecutor<T> implements ShutdownableTaskExecutor<T> {
     // around
     // that will simplify such kind of logging through workers.
     return String.format("PollTaskExecutor{name=%s, identity=%s}", pollThreadNamePrefix, identity);
-  }
-
-  private void publishSlotsMetric() {
-    this.metricsScope
-        .gauge(MetricsType.WORKER_TASK_SLOTS_AVAILABLE)
-        .update(availableTaskSlots.get());
   }
 }
--- a/temporal-sdk/src/main/java/io/temporal/internal/worker/TrackingSlotSupplier.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/worker/TrackingSlotSupplier.java
@@ -96,6 +96,10 @@ public class TrackingSlotSupplier<SI> {
     this.metricsScope = metricsScope;
   }
 
+  Map<SlotPermit, SI> getUsedSlots() {
+    return usedSlots;
+  }
+
   private void publishSlotsMetric() {
     this.metricsScope
         .gauge(MetricsType.WORKER_TASK_SLOTS_AVAILABLE)
--- a/temporal-sdk/src/main/java/io/temporal/internal/worker/TrackingSlotSupplier.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/worker/TrackingSlotSupplier.java
@@ -1,0 +1,136 @@
+/*
+ * Copyright (C) 2022 Temporal Technologies, Inc. All Rights Reserved.
+ *
+ * Copyright (C) 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Modifications copyright (C) 2017 Uber Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this material except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.temporal.internal.worker;
+
+import com.uber.m3.tally.NoopScope;
+import com.uber.m3.tally.Scope;
+import io.temporal.worker.MetricsType;
+import io.temporal.worker.tuning.SlotPermit;
+import io.temporal.worker.tuning.SlotReleaseReason;
+import io.temporal.worker.tuning.SlotReservationContext;
+import io.temporal.worker.tuning.SlotSupplier;
+import java.util.Collections;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Wraps a slot supplier and supplements it with additional tracking information that is useful to
+ * provide to all implementations. This type is used internally rather than {@link SlotSupplier}
+ * directly.
+ *
+ * @param <SI> The slot info type
+ */
+public class TrackingSlotSupplier<SI> {
+  private final SlotSupplier<SI> inner;
+  private final AtomicInteger issuedSlots = new AtomicInteger();
+  private final Map<SlotPermit, SI> usedSlots = new ConcurrentHashMap<>();
+  private Scope metricsScope;
+
+  public TrackingSlotSupplier(SlotSupplier<SI> inner) {
+    this.inner = inner;
+    metricsScope = new NoopScope();
+  }
+
+  public SlotPermit reserveSlot(SlotReservationData dat) throws InterruptedException {
+    SlotPermit p = inner.reserveSlot(createCtx(dat));
+    issuedSlots.incrementAndGet();
+    return p;
+  }
+
+  public Optional<SlotPermit> tryReserveSlot(SlotReservationData dat) {
+    Optional<SlotPermit> p = inner.tryReserveSlot(createCtx(dat));
+    if (p.isPresent()) {
+      issuedSlots.incrementAndGet();
+    }
+    return p;
+  }
+
+  public void markSlotUsed(SI slotInfo, SlotPermit permit) {
+    if (permit == null) {
+      return;
+    }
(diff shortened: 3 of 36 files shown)


temporal-sdk/src/main/java/io/temporal/internal/worker/EagerActivitySlotsReservation.java
```diff
@@ -66,34 +74,44 @@ public void applyToRequest(RespondWorkflowTaskCompletedRequest.Builder mutableRe
   public void handleResponse(RespondWorkflowTaskCompletedResponse serverResponse) {
     int activityTasksCount = serverResponse.getActivityTasksCount();
     Preconditions.checkArgument(
-        activityTasksCount <= this.outstandingReservationSlotsCount,
+        // TODO: This check seems backwards?
```
Quinn-With-Two-Ns, line 77: Why is it backwards?

temporal-sdk/src/main/java/io/temporal/internal/worker/PollTaskExecutor.java
```diff
@@ -97,6 +97,9 @@ public void process(@Nonnull T task) {
     Preconditions.checkNotNull(task, "task");
     taskExecutor.execute(
         () -> {
+          // TODO: Unclear if we want to keep this metric here, since users might define
```
Quinn-With-Two-Ns, line 100: Though we said the metric should be handled in the slot supplier?

temporal-sdk/src/main/java/io/temporal/worker/slotsupplier/SlotSupplier.java
```diff
@@ -0,0 +1,45 @@
...
+
+  void markSlotUsed(SlotInfo info, SlotPermit permit);
+
+  void releaseSlot(SlotReleaseReason reason, SlotPermit permit);
+
+  /**
+   * Because we currently use thread pools to execute tasks, there must be *some* defined
+   * upper-limit on the size of the thread pool for each kind of task.
+   *
+   * <p>This method may be deprecated and removed later if and when the SDK supports virtual
```
Quinn-With-Two-Ns, line 39: Just to note, this wouldn't be deprecated/removed since virtual threads are only in Java 21 and we still support Java 8

temporal-sdk/src/main/java/io/temporal/worker/slotsupplier/SlotSupplier.java
```diff
@@ -0,0 +1,45 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.temporal.worker.slotsupplier;
+
+import java.util.Optional;
+
+public interface SlotSupplier<SlotInfo> {
+  // TODO: Needs to be cancellable
```
Quinn-With-Two-Ns, line 26: Is interruption not enough?

temporal-sdk/src/main/java/io/temporal/worker/slotsupplier/TrackingSlotSupplier.java
```diff
@@ -0,0 +1,75 @@
...
+
+package io.temporal.worker.slotsupplier;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.atomic.AtomicInteger;
+
+// TODO: Also make pauseable?
+public class TrackingSlotSupplier<SlotInfo> implements SlotSupplier<SlotInfo> {
```
Quinn-With-Two-Ns, line 29: Should this class be internal?

temporal-sdk/src/main/java/io/temporal/worker/slotsupplier/TrackingSlotSupplier.java
```diff
@@ -0,0 +1,75 @@
...
+  private final SlotSupplier<SlotInfo> inner;
+  private final AtomicInteger issuedSlots = new AtomicInteger();
+  private final Map<SlotPermit, SlotInfo> usedSlots = new HashMap<>();
```
Quinn-With-Two-Ns, line 32: What is the use of  `usedSlots `?

temporal-sdk/src/main/java/io/temporal/worker/slotsupplier/SlotSupplier.java
```diff
@@ -0,0 +1,45 @@
...
+  SlotPermit reserveSlot(SlotReservationContext ctx) throws InterruptedException;
+
+  Optional<SlotPermit> tryReserveSlot(SlotReservationContext ctx);
+
+  void markSlotUsed(SlotInfo info, SlotPermit permit);
+
+  void releaseSlot(SlotReleaseReason reason, SlotPermit permit);
```
Quinn-With-Two-Ns, line 33: Can `releaseSlots` be blocking? If we had some sort of remote slot provider it seems like it could be?

temporal-sdk/src/main/java/io/temporal/internal/worker/PollTaskExecutor.java
Sushisource: Yep, this is why the draft said "don't review" :wink:

temporal-sdk/src/main/java/io/temporal/internal/worker/EagerActivitySlotsReservation.java
Sushisource: Ugh, nevermind, I'm used to the check being "if this condition is met, it's a problem" but I spaced and forgot `Preconditions` is "if this check _isn't_ met, it's a problem".

temporal-sdk/src/main/java/io/temporal/worker/slotsupplier/SlotSupplier.java
Sushisource: I don't think we can allow it to be blocking. The code has no expectation now that it might, and allowing it to could I think gum up stuff in unexpected ways. I mentioned in the proposal that users can delegate any blocking work to a background thread or to calls to reserve slot, and I think that's what they ought to do.

temporal-sdk/src/main/java/io/temporal/internal/worker/EagerActivitySlotsReservation.java
```diff
@@ -66,34 +74,43 @@ public void applyToRequest(RespondWorkflowTaskCompletedRequest.Builder mutableRe
...
-        activityTasksCount <= this.outstandingReservationSlotsCount,
+        activityTasksCount <= this.reservedSlots.size(),
         "Unexpectedly received %s eager activities though we only requested %s",
         activityTasksCount,
-        this.outstandingReservationSlotsCount);
+        this.reservedSlots.size());
 
-    releaseSlots(this.outstandingReservationSlotsCount - activityTasksCount);
+    // Release any slots that we won't be using
+    releaseSlots(this.reservedSlots.size() - activityTasksCount);
```
Quinn-With-Two-Ns, line 83: Since the slots are reserved based on the task queue don't we also need to keep track of that when using the slot and releasing the slot?

temporal-sdk/src/main/java/io/temporal/internal/worker/EagerActivitySlotsReservation.java
```diff
@@ -66,34 +74,43 @@ public void applyToRequest(RespondWorkflowTaskCompletedRequest.Builder mutableRe
...
-    if (this.outstandingReservationSlotsCount > 0)
-      releaseSlots(this.outstandingReservationSlotsCount);
+    if (!this.reservedSlots.isEmpty()) {
+      releaseSlots(this.reservedSlots.size());
+    }
   }
 
   private void releaseSlots(int slotsToRelease) {
-    if (slotsToRelease > this.outstandingReservationSlotsCount)
+    if (slotsToRelease <= 0) return;
```
Quinn-With-Two-Ns, line 103: If we throw on `slotsToRelease ` being to large shouldn't we throw on it being negative as well?

temporal-sdk/src/main/java/io/temporal/worker/WorkerOptions.java
```diff
@@ -393,14 +430,6 @@ public WorkerOptions build() {
     public WorkerOptions validateAndBuildWithDefaults() {
       Preconditions.checkState(
           maxWorkerActivitiesPerSecond >= 0, "negative maxActivitiesPerSecond");
-      Preconditions.checkState(
```
Quinn-With-Two-Ns, line 396: Why are these checks no longer needed?

temporal-sdk/src/main/java/io/temporal/worker/tuning/SlotSupplier.java
```diff
@@ -0,0 +1,89 @@
...
+  void releaseSlot(SlotReleaseReason reason, SlotPermit permit);
+
+  /**
+   * Because we currently use thread pools to execute tasks, there must be *some* defined
+   * upper-limit on the size of the thread pool for each kind of task, and your implementation must
+   * provide this number. You must not hand out more permits than this number.
+   *
+   * @return the maximum number of slots that can ever be in use at one type for this slot type.
+   */
+  int maximumSlots();
```
Quinn-With-Two-Ns, line 88: Hm I wonder now if we really need this? It is totally valid to have an unbounded threadpool in Java by setting `maximumPoolSize` to `Integer.MAX_VALUE`.

https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ThreadPoolExecutor.html

temporal-sdk/src/main/java/io/temporal/worker/tuning/SlotSupplier.java
```diff
@@ -0,0 +1,89 @@
...
+
+  /**
+   * This function is called once a permit is no longer needed. This could be because the task has
+   * finished, whether successfully or not, or because the slot was no longer needed (ex: the number
+   * of active pollers decreased). This call should be non-blocking.
+   *
+   * @param reason The reason the slot is being released.
+   * @param permit The permit that was given out when the slot was reserved.
+   */
+  void releaseSlot(SlotReleaseReason reason, SlotPermit permit);
```
cretz, line 79:
```suggestion
  void releaseSlot(SlotReleaseContext<SI> ctx);
```

Same reason as before, this allows us to grow the information w/out altering signature. Also, the slot info should be provided to releasers.

Also would consider not providing a reason until needed, but harmless of course.

temporal-sdk/src/main/java/io/temporal/worker/tuning/SlotReleaseReason.java
```diff
@@ -0,0 +1,76 @@
...
+
+  public boolean isError() {
+    return false;
+  }
+
+  /**
+   * @return the exception that caused the slot to be released, if this is a reason of type {@link
+   *     Error}.
+   */
+  public Exception getException() {
```
cretz, line 45: Consider making this optional or using annotations

temporal-sdk/src/main/java/io/temporal/worker/tuning/SlotReleaseReason.java
```diff
@@ -0,0 +1,76 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.temporal.worker.tuning;
+
+public abstract class SlotReleaseReason {
```
cretz, line 23: Add a package private constructor here

temporal-sdk/src/main/java/io/temporal/worker/tuning/WorkflowSlotInfo.java
```diff
@@ -0,0 +1,124 @@
...
+public class WorkflowSlotInfo {
+  private final String workflowType;
+  private final String taskQueue;
+  private final String workflowId;
+  private final String runId;
+  private final String workerIdentity;
+  private final String workerBuildId;
+  private final boolean fromStickyQueue;
+
+  public WorkflowSlotInfo(
```
cretz, line 39: Would suggest hiding this detail (can consider making slot infos abstract classes or interfaces or something and move construction to internal if wanted, or just change the constructor for every field in here and potentially add javadoc on this one and activity one saying not to rely on stability of constructor signature).

temporal-sdk/src/main/java/io/temporal/internal/worker/TrackingSlotSupplier.java
```diff
@@ -0,0 +1,132 @@
...
+    this.metricsScope
+        .gauge(MetricsType.WORKER_TASK_SLOTS_AVAILABLE)
+        .update(maximumSlots() - usedSlots.size());
+  }
+
+  private SlotReservationContext<SI> createCtx(SlotReservationData dat) {
+    return new SlotResCtx(dat.getTaskQueue(), Collections.unmodifiableMap(usedSlots));
+  }
+
+  private class SlotResCtx implements SlotReservationContext<SI> {
```
cretz, line 113: Java doesn't usually abbreviate like this (even for private stuff despite the annoyance to the author)

temporal-sdk/src/main/java/io/temporal/worker/tuning/SlotSupplier.java
Sushisource: Yeah I'm just a little worried that could be a big footgun? They can always return a huge number here and accomplish the same thing if they want to?

temporal-sdk/src/main/java/io/temporal/worker/tuning/SlotSupplier.java
```diff
@@ -0,0 +1,89 @@
...
+   * This function is called before polling for new tasks. Your implementation should block until a
+   * slot is available, then return a permit to use that slot.
+   *
+   * @param ctx The context for slot reservation.
+   * @return A permit to use the slot which may be populated with your own data.
+   * @throws InterruptedException The worker may choose to interrupt the thread in order to cancel
+   *     the reservation, or during shutdown. You may perform cleanup, and then should rethrow the
+   *     exception.
+   */
+  SlotPermit reserveSlot(SlotReservationContext<SI> ctx) throws InterruptedException;
```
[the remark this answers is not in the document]
Sushisource: I like having the internal ID because it ensures equality & hashing always work properly, regardless of what the user ends up doing with their data (yeah, objects have an ID, but they could end up implementing eq/hash at the top level in a way that potentially breaks things - but to your point we could get rid of the internal ID and still have that work seemingly), and it also makes for a nice, clear type at the top-level. The type name alone feels worth it to me.

temporal-sdk/src/main/java/io/temporal/worker/tuning/SlotSupplier.java
Quinn-With-Two-Ns: >Yeah I'm just a little worried that could be a big footgun?

What is the footgun? 

>They can always return a huge number here and accomplish the same thing if they want to?

True, but why isn't that just the default then?

Sushisource: The footgun is possibly way over-subscribing the CPU in the case of activity executions, for example. You can easily end up running way too many concurrently. Buuuuut, I suppose that's definitely not happening if they use our fixed size supplier, since it'll do the right thing, and if they implement their own, well, then it should be fairly clear to them not to do that, or at least is their responsibility.

So, yeah, I can see allowing it to default to something big.

temporal-sdk/src/main/java/io/temporal/worker/tuning/SlotPermit.java
```diff
@@ -0,0 +1,59 @@
...
+  public final Object userData;
+
+  private static final AtomicLong nextId = new AtomicLong(0);
+
+  public SlotPermit() {
+    id = nextId.getAndIncrement();
+    this.userData = null;
+  }
+
+  public SlotPermit(Object userData) {
```
Quinn-With-Two-Ns, line 42: I suspect slot providers will often be implemented by chaining  or wrapping other slot providers, where each slot provider may want to attach data to the permit. Maybe an API like a Go `context`?

Sushisource: Are you thinking something like `permit.withData(obj)`?

temporal-sdk/src/main/java/io/temporal/internal/worker/EagerActivitySlotsReservation.java
Sushisource: Interesting point. I wouldn't say they're _based on_ so much as that information is provided when making the reservation (implementations may or may not care). It does make sense, of course, that we might also want to provide that information when releasing.

Arguably, users could stuff that in the permit data, but maybe we want to make it first-class. In that case, I'd probably just stick it as a field in the permit. However, you could also argue that implementations that want to balance across multiple queues maybe don't want to associate a permit with a specific queue, in which case it makes more sense on some `ReleaseContext` object to provide as another arg when releasing.

The latter maybe makes a bit more sense to me?

temporal-sdk/src/main/java/io/temporal/worker/tuning/SlotSupplier.java
Quinn-With-Two-Ns: >The footgun is possibly way over-subscribing the CPU in the case of activity executions, for example.

That is a problem of handing out to many slots. Basically I don't think there is any technical reason we need this in Java so if we add it here  we should add it to all SDKs.

temporal-sdk/src/main/java/io/temporal/worker/tuning/SlotSupplier.java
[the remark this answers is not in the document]
cretz: > yeah, objects have an ID, but they could end up implementing eq/hash at the top level in a way that potentially breaks things - but to your point we could get rid of the internal ID and still have that work seemingly

I'd have to understand this a bit better. Equals/hashCode will work for `new Object()` directly and if someone overrides those for a specific object, I would expect them to work too. I don't think we should ever invoke equals/hashCode on this object for anything though.

temporal-sdk/src/main/java/io/temporal/worker/tuning/SlotPermit.java
cretz: This is where allowing the Java author to return the type of their choice (even if they were forced to extend `SlotPermit` though not as much of a fan there) gives more freedom than attaching an object to an existing object.

temporal-sdk/src/main/java/io/temporal/worker/WorkerOptions.java
```diff
@@ -90,9 +91,9 @@ private Builder(WorkerOptions o) {
         return;
       }
       this.maxWorkerActivitiesPerSecond = o.maxWorkerActivitiesPerSecond;
-      this.maxConcurrentActivityExecutionSize = o.maxConcurrentActivityExecutionSize;
-      this.maxConcurrentWorkflowTaskExecutionSize = o.maxConcurrentWorkflowTaskExecutionSize;
-      this.maxConcurrentLocalActivityExecutionSize = o.maxConcurrentLocalActivityExecutionSize;
+      this.workflowSlotSupplier = o.workflowSlotSupplier;
+      this.activitySlotSupplier = o.activitySlotSupplier;
+      this.localActivitySlotSupplier = o.localActivitySlotSupplier;
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: I agree with 1. , I  don't have a strong opinion on 2 , but I slightly prefer 3 separate options for each `SlotSupplier` since it fits with a builder pattern better and handle a mix of `SlotSupplier`  and legacy config.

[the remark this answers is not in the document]
cretz: Think about it from a user POV. What is the simplest configuration for a user to choose "resource limited tuning"? I am happy we didn't make a user set 3 different interceptor fields for use of OTel interceptors. It's not that the 3 things can still be provided, it's just from a user and default impl, people shouldn't have to set three things if they don't need to.

temporal-sdk/src/main/java/io/temporal/worker/tuning/SlotPermit.java
Quinn-With-Two-Ns: I would not want users to be able to return an object of their choice. I would like to maintain control over the object and provide an interface to attach and get metadata.

temporal-sdk/src/main/java/io/temporal/worker/tuning/WorkflowSlotInfo.java
Sushisource: Yeah, the lack of granular enough visibility specifiers is annoying. I'll use a doc comment for now and we can potentially change it later if/when we change more here.

temporal-sdk/src/main/java/io/temporal/worker/tuning/SlotSupplier.java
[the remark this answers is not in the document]
Sushisource: The hash/eq is used because we use it in the hash map in the tracking supplier. So, sure, we expect it to work just fine - but our implementation will always be correct (and likely cheaper) than a user-supplied one.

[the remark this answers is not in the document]
cretz: I think you should then make `SlotPermit` `final` (so nobody extends it and messes up this expectation) and you should rely on the default equals/hashCode which is identity based (i.e. don't override it).

temporal-sdk/src/main/java/io/temporal/worker/tuning/SlotSupplier.java
```diff
@@ -0,0 +1,93 @@
...
+   * upper-limit on the size of the thread pool for each kind of task. You must not hand out more
+   * permits than this number. If unspecified, the default is {@link Integer#MAX_VALUE}. Be aware
+   * that if your implementation hands out unreasonable numbers of permits, you could easily
+   * oversubscribe the worker, and cause it to run out of resources.
+   *
+   * <p>This value should never change during the lifetime of the supplier.
+   *
+   * @return the maximum number of slots that can ever be in use at one type for this slot type.
+   */
+  default int maximumSlots() {
```
cretz, line 90:
```suggestion
  default int getMaximumSlots() {
```

Pedantic and don't have strong opinion here, just saw https://github.com/temporalio/sdk-java/pull/2014#discussion_r1555820733 marked resolved

temporal-sdk/src/main/java/io/temporal/worker/tuning/SlotPermit.java
cretz: :+1: Ok, no prob. Still not a big fan of the ID in here, but no prob there too. Mentioned in other comment, but we should probably make this class final, consider removing the equals/hashCode override (i.e. default to identity-based), and document on the context interfaces that the permits are the _exact same instances_ as the ones returned.

cretz (commented):
Mostly there, things left IMO:

* At least make `SlotPermit` final and document we use same instances (should also consider removing unnecessarily ID and removing unnecessary equals/hashCode)
* Figure out the worker options (would like to leave existing ones alone and just make mutually exclusive at build time, and would like a single option to set a tuner instead of 3)
* Add `@Experimental` on all types in tuning package and on the new worker option(s)

temporal-sdk/src/main/java/io/temporal/worker/tuning/SlotPermit.java
Sushisource: Fixed to do that

temporal-sdk/src/main/java/io/temporal/worker/tuning/ActivitySlotInfo.java
```diff
@@ -0,0 +1,78 @@
...
+ * limitations under the License.
+ */
+
+package io.temporal.worker.tuning;
+
+import io.temporal.activity.ActivityInfo;
+import java.util.Objects;
+
+/** Contains information about a slot that is being used to execute an activity task. */
+public class ActivitySlotInfo extends SlotInfo {
```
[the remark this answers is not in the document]
Sushisource: The worker options are marked experimental as is the interface class, which seems sufficient to me, but sure I'll mark them all that way.

temporal-sdk/src/main/java/io/temporal/worker/WorkerOptions.java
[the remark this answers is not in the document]
Sushisource: I think it's dramatically easier for the user to provide separate suppliers for separate kinds, which is I think more likely the more common case than a shared implementation across different kinds - and it makes that case no harder. Setting two more options fields is effectively zero effort. 

If it was one option, we either have to have something like `SlotSupplier.activity(a).workflow(b).localActivity(c)` which is now the same thing, but with one _extra_ step of putting that whole thing in the worker config. _Or_ we have to change the whole interface to pass the type on every reservation call which I think is much messier and conceptually more difficult for most implementations.

I'll change the options to keep the old values, though.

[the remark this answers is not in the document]
Sushisource: Plus, since we now already have mutually exclusive options, there's zero reason we can do `builder().setResourceLimitedTuner(options)` and under the hood it sets all there.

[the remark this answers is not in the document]
Sushisource: I'm not totally sure 1. makes sense to me either, honestly, as it requires establishing precedence rules for these rather than last-call-wins, but you both seem to agree so I'll make the change.

temporal-sdk/src/main/java/io/temporal/internal/worker/EagerActivitySlotsReservation.java
Sushisource: Added to release context

temporal-sdk/src/main/java/io/temporal/worker/tuning/SlotSupplier.java
Sushisource: Fixed, must have marked it by accident.

temporal-sdk/src/main/java/io/temporal/worker/tuning/SlotPermit.java
```diff
@@ -0,0 +1,44 @@
...
+ * This class is handed out by implementations of {@link SlotSupplier}. Permits are held until the
+ * tasks they are associated with (if any) are finished processing, or if the reservation is no
+ * longer needed. Your supplier implementation may store additional data in the permit, if desired.
+ *
+ * <p>When {@link SlotSupplier#releaseSlot(SlotReleaseContext)} is called, the exact same instance
+ * of the permit is passed back to the supplier.
+ */
+@Experimental
+public final class SlotPermit {
+  public final Object userData;
```
cretz, line 35: Even though it's final, it's a bit of a faux pas to have public fields. Would prefer getter.

cretz (approved):

temporal-sdk/src/main/java/io/temporal/worker/tuning/SlotPermit.java
Sushisource: Fixed

temporal-sdk/src/main/java/io/temporal/internal/worker/EagerActivitySlotsReservation.java
Quinn-With-Two-Ns: Sorry to clarify what I meant is your slot reservations are not fungible since they are tied to the task queue they are requested for. So you can't just blindly release first x slots you need to release the slots based on task queue.

Quinn-With-Two-Ns:
Can we make issue (if we don't already) for:
* Pausable slot provider
* Resource based slot provider
* Springboot integration for slot supplier

temporal-sdk/src/main/java/io/temporal/worker/tuning/SlotPermit.java
Quinn-With-Two-Ns: To clarify I was taking about having a map of userdata like a Go `context` does so different providers wrapping each other can add their own metadata without impacting each other.

temporal-sdk/src/main/java/io/temporal/worker/WorkerOptions.java
```diff
@@ -519,15 +595,15 @@ public double getMaxWorkerActivitiesPerSecond() {
...
   }
 
   public int getMaxConcurrentWorkflowTaskExecutionSize() {
-    return maxConcurrentWorkflowTaskExecutionSize;
+    return workflowSlotSupplier.getMaximumSlots();
   }
 
   public int getMaxConcurrentLocalActivityExecutionSize() {
-    return maxConcurrentLocalActivityExecutionSize;
+    return localActivitySlotSupplier.getMaximumSlots();
```
Quinn-With-Two-Ns, line 606: These 3 `getMaxConcurrent*ExecutionSize()` should be zero if the corresponding `maxConcurrent*ExecutionSize` is not set.

Quinn-With-Two-Ns: I think a buidlers setter should only effect one getter

Sushisource: Ah, yeah forgot to change those ones.

temporal-sdk/src/main/java/io/temporal/worker/tuning/SlotPermit.java
Sushisource: Something like a typemap of <T -> Data> would be useful. But, perhaps we add that later if requested

temporal-sdk/src/main/java/io/temporal/worker/WorkerOptions.java
Sushisource: Yeah... I was hoping not to retain the property in the _built_ workflow options since having mutually exclusive stuff represented in the built type is less than good from a type design perspective. It also means the logic for the threadpool size has to get pushed out of the options, but, that's not so bad.

temporal-sdk/src/main/java/io/temporal/internal/worker/EagerActivitySlotsReservation.java
Sushisource: Yeah... I was hoping not to retain the property in the _built_ workflow options since having mutually exclusive stuff represented in the built type is less than good from a type design perspective. It also means the logic for the threadpool size has to get pushed out of the options, but, that's not so bad.

Sushisource: Ah, yeah that makes sense. I don't know that it necessarily must work that way, we could only make the guarantee that we'll return permits since they all came from the same supplier regardless of queue, but we can do it fairly easily and it's probably helpful so I'll make that change.

temporal-sdk/src/main/java/io/temporal/worker/WorkerOptions.java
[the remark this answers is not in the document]
cretz: > If it was one option, we either have to have something like `WorkerTuner.activity(a).workflow(b).localActivity(c)`

It'd be more like `builder().setWorkerTuner(new TheTuner(a, b, c))` which is nice because people can extend that one or get one from a third part lib or whatever

> Plus, since we now already have mutually exclusive options, there's zero reason we can't do `builder().setResourceLimitedTuner(options)` and under the hood it sets all three.

That means only we get the single-option approach because we are Temporal. Everyone should be allowed to have users use _their_ tuner via a single option too. In the meantime, they may have to provide a static `applyToWorkerOptionsBuilder` to make sure all three suppliers are set right.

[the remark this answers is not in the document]
Sushisource: I don't think that's an apples-to-apples comparison at all though. Your tracing interceptor isn't making decisions about what happens where the first and most important criteria is the type of slot - rather it is dealing with application-wide traces, and may or may not care about the specifics of the thing being traced. Here, you _always_ care about the type of slot.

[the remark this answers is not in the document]
cretz: The implementation always cares about the type of thing it's intercepting or the slot, the one setting the worker option does not. Well, they may in some cases (e.g. fixed size things they may provide three options to this tuner), but they may not in other cases (e.g. they only want to provide resource limits to tuner or just leave defaults). But it should be up to the tuner implementation to decide whether the configurer cares about separate things or not (the implementation itself of course will always care about separate things).

Even in cases where you may want to mix (e.g. fixed and resource-limited), I think still requiring an impl that returns different slot suppliers for those specific cases is better UX than making _everyone_ care about those separate slot suppliers in _every_ case.

I think it may be worth seeing how a user would provide their own implementation of a resource tuner that needs to work across the three (we shouldn't be special in that we're the only ones that get single-option config).

Sushisource:
> Can we make issue (if we don't already) for:
> 
>     * Pausable slot provider
> 
>     * Resource based slot provider
> 
>     * Springboot integration for slot supplier

Added https://github.com/temporalio/sdk-java/issues/2030 and https://github.com/temporalio/sdk-java/issues/2031

Won't add the resource one until we're done with the PoC, and then I'll make items for that everywhere.

Quinn-With-Two-Ns (approved):

PR #2019, opened 2024-03-26
Add StickyTaskQueueDrainTimeout

Quinn-With-Two-Ns:
Add an option to drain sticky task queue during graceful shutdown. Worker will stop polling on the normal task queue and only poll on the stick task queue to complete workflows while the worker is shutting down.

There is a lot  of follow up work we could do around here based on user demand and if the server adds some APIs. For know this is the minimal implementation  to avoid sticky workflow timeout for short workflow during graceful shutdown.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/worker/ShutdownManager.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/worker/ShutdownManager.java
@@ -78,6 +78,23 @@ public class ShutdownManager implements Closeable {
   }
 
   /**
+   * waitForStickyQueueBalancer -&gt; disableNormalPoll -&gt; timed wait for graceful completion of
+   * sticky workflows
+   */
+  public CompletableFuture<Void> waitForStickyQueueBalancer(
+      StickyQueueBalancer balancer, Duration timeout) {
+    CompletableFuture<Void> future = new CompletableFuture<>();
+    balancer.disableNormalPoll();
+    scheduledExecutorService.schedule(
+        () -> {
+          future.complete(null);
+        },
+        timeout.toMillis(),
+        TimeUnit.MILLISECONDS);
+    return future;
+  }
+
+  /**
    * Wait for {@code executorToShutdown} to terminate. Only completes the returned CompletableFuture
    * when the executor is terminated.
    */
--- a/temporal-sdk/src/main/java/io/temporal/worker/WorkerOptions.java
+++ b/temporal-sdk/src/main/java/io/temporal/worker/WorkerOptions.java
@@ -24,6 +24,7 @@ import static java.lang.Double.compare;
 
 import com.google.common.base.Preconditions;
 import io.temporal.common.Experimental;
+import io.temporal.serviceclient.WorkflowServiceStubsOptions;
 import java.time.Duration;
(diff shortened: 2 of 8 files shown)


temporal-sdk/src/main/java/io/temporal/internal/worker/ShutdownManager.java
```diff
@@ -77,6 +77,23 @@ public CompletableFuture<Void> waitForSemaphorePermitsReleaseUntimed(
     return future;
   }
 
+  /**
+   * waitForStickyQueueBalancer -&gt; disableNormalPoll -&gt; timed wait for graceful completion of
```
Sushisource, line 81: Fucking Javadoc lol

temporal-sdk/src/main/java/io/temporal/worker/WorkerOptions.java
```diff
@@ -349,6 +354,22 @@ public Builder setBuildId(String buildId) {
       return this;
     }
 
+    /**
+     * During graceful shutdown, like calling {@link WorkerFactory#shutdown()}, if "sticky worker"
+     * is enabled, this timeout controls how long to wait for the sticky task queue to drain before
+     * shutting down the worker. If set the worker will stop making new poll requests on the normal
+     * task queue, but will continue to poll the sticky task queue until the timeout is reached.
+     * This value should always be greater than clients rpc long poll timeout, which can be set via
+     * {@link WorkflowServiceStubsOptions.Builder#setRpcLongPollTimeout(Duration)}.
```
Sushisource, line 363:
```suggestion
     * During graceful shutdown, as when calling {@link WorkerFactory#shutdown()}, if the workflow cache
     * is enabled, this timeout controls how long to wait for the sticky task queue to drain before
     * shutting down the worker. If set the worker will stop making new poll requests on the normal
     * task queue, but will continue to poll the sticky task queue until the timeout is reached.
     * This value should always be greater than clients rpc long poll timeout, which can be set via
     * {@link WorkflowServiceStubsOptions.Builder#setRpcLongPollTimeout(Duration)}.
```

Slightly more clear I think

Sushisource (approved):

temporal-sdk/src/main/java/io/temporal/worker/WorkerOptions.java
cretz, line 363: Now I wonder if it'd make more sense to make this a `boolean` for `waitForStickyEmptyPoll`, because what if you have a time race? For instance, what if my long poll timeout and my sticky drain timeout is 1m. And as soon as the poll comes back empty I haven't hit sticky drain timeout, so I make another sticky poll for just some nanoseconds before sticky drani timeout does hit? Maybe that's ok though.

Quinn-With-Two-Ns: That poll would be cancelled, there is a possible race, but the race also exists for normal shutdown as well and I don't think we should tackle it here.

cretz: :+1: Would like to see us not kill poll calls anywhere, but yes can be tackled elsewhere

PR #2021, opened 2024-03-28
Fix exception in GrpcRetryer.

chronos-tachyon:
## What was changed
* Ensure that `ChannelManager.getServerCapabilities` follows the standard gRPC retry policy, even though it is called by `GrpcRetryer`.

## Why?
In the course of writing new gRPC retry tests for the [features](https://temporalio/features) repo, it was discovered that any gRPC failures returned by the server in response to `WorkflowService.getSystemInfo` will cause an uncaught `io.grpc.StatusRuntimeException` that happens before the actual gRPC retry code gets the chance to run.

## Checklist

1. Closes internal JIRA issue SDK-1888.  (Sorry, I should have opened it as a GitHub issue and let it sync to JIRA instead, I didn't learn this until after the fact.)

2. How was this tested: using the new features repo tests in https://github.com/temporalio/features/pull/435 .

### DIFF
--- a/temporal-serviceclient/src/main/java/io/temporal/serviceclient/ChannelManager.java
+++ b/temporal-serviceclient/src/main/java/io/temporal/serviceclient/ChannelManager.java
@@ -311,22 +311,24 @@ final class ChannelManager {
    */
   public HealthCheckResponse healthCheck(
       String healthCheckServiceName, @Nullable Duration timeout) {
-    HealthGrpc.HealthBlockingStub stub;
-    if (timeout != null) {
-      stub =
-          this.healthBlockingStub.withDeadline(
-              Deadline.after(
-                  options.getHealthCheckAttemptTimeout().toMillis(), TimeUnit.MILLISECONDS));
-    } else {
-      stub = this.healthBlockingStub;
+    if (timeout == null) {
+      timeout = options.getHealthCheckAttemptTimeout();
     }
-    return stub.check(HealthCheckRequest.newBuilder().setService(healthCheckServiceName).build());
+    return this.healthBlockingStub
+        .withDeadline(deadlineFrom(timeout))
+        .check(HealthCheckRequest.newBuilder().setService(healthCheckServiceName).build());
   }
 
   public Supplier<Capabilities> getServerCapabilities() {
     return () ->
         SystemInfoInterceptor.getServerCapabilitiesWithRetryOrThrow(
-            serverCapabilitiesFuture, interceptedChannel, null);
+            serverCapabilitiesFuture,
+            interceptedChannel,
+            deadlineFrom(options.getHealthCheckAttemptTimeout()));
+  }
+
+  private static Deadline deadlineFrom(Duration duration) {
+    return Deadline.after(duration.toMillis(), TimeUnit.MILLISECONDS);
   }
 
   public void shutdown() {
--- a/temporal-serviceclient/src/main/java/io/temporal/serviceclient/ChannelManager.java
+++ b/temporal-serviceclient/src/main/java/io/temporal/serviceclient/ChannelManager.java
@@ -27,8 +27,9 @@ import io.grpc.health.v1.HealthCheckResponse;
 import io.grpc.health.v1.HealthGrpc;
 import io.grpc.netty.shaded.io.grpc.netty.NettyChannelBuilder;
 import io.grpc.stub.MetadataUtils;
-import io.temporal.api.workflowservice.v1.GetSystemInfoResponse;
+import io.temporal.api.workflowservice.v1.GetSystemInfoResponse.Capabilities;
 import io.temporal.internal.retryer.GrpcRetryer;
+import io.temporal.internal.retryer.GrpcRetryer.GrpcRetryerOptions;
 import java.time.Duration;
 import java.util.Collection;
 import java.util.List;
@@ -87,7 +88,7 @@ final class ChannelManager {
   private final Channel interceptedChannel;
   private final HealthGrpc.HealthBlockingStub healthBlockingStub;
 
-  private final CompletableFuture<GetSystemInfoResponse.Capabilities> serverCapabilitiesFuture =
+  private final CompletableFuture<Capabilities> serverCapabilitiesFuture =
       new CompletableFuture<>();
 
   public ChannelManager(
@@ -289,8 +290,8 @@ final class ChannelManager {
     if (timeout == null) {
       timeout = options.getRpcTimeout();
     }
-    GrpcRetryer.GrpcRetryerOptions grpcRetryerOptions =
-        new GrpcRetryer.GrpcRetryerOptions(
+    GrpcRetryerOptions grpcRetryerOptions =
+        new GrpcRetryerOptions(
             RpcRetryOptions.newBuilder().setExpiration(timeout).validateBuildWithDefaults(), null);
 
     new GrpcRetryer(getServerCapabilities())
@@ -322,18 +323,10 @@ final class ChannelManager {
     return stub.check(HealthCheckRequest.newBuilder().setService(healthCheckServiceName).build());
   }
 
-  public Supplier<GetSystemInfoResponse.Capabilities> getServerCapabilities() {
-    return () -> {
-      synchronized (serverCapabilitiesFuture) {
-        GetSystemInfoResponse.Capabilities capabilities = serverCapabilitiesFuture.getNow(null);
-        if (capabilities == null) {
-          serverCapabilitiesFuture.complete(
-              SystemInfoInterceptor.getServerCapabilitiesOrThrow(interceptedChannel, null));
-          capabilities = serverCapabilitiesFuture.getNow(null);
-        }
-        return capabilities;
-      }
-    };
+  public Supplier<Capabilities> getServerCapabilities() {
+    return () ->
+        SystemInfoInterceptor.getServerCapabilitiesWithRetryOrThrow(
+            serverCapabilitiesFuture, interceptedChannel, null);
   }
 
   public void shutdown() {
--- a/temporal-serviceclient/src/main/java/io/temporal/serviceclient/SystemInfoInterceptor.java
+++ b/temporal-serviceclient/src/main/java/io/temporal/serviceclient/SystemInfoInterceptor.java
@@ -21,7 +21,6 @@
 package io.temporal.serviceclient;
 
 import io.grpc.*;
-import io.grpc.health.v1.HealthGrpc;
 import io.temporal.api.workflowservice.v1.GetSystemInfoRequest;
 import io.temporal.api.workflowservice.v1.GetSystemInfoResponse;
 import io.temporal.api.workflowservice.v1.GetSystemInfoResponse.Capabilities;
@@ -75,8 +74,6 @@ public class SystemInfoInterceptor implements ClientInterceptor {
                     super.onClose(status, trailers);
                   }
                 };
-          } else if (method == HealthGrpc.getCheckMethod()) {
-            // do nothing
           } else {
             // Need to reach system capabilities, so make a getSystemInfo call in a blocking manner.
             // We don't try to squash into one and optimize the several getSystemInfo calls that may
@@ -85,8 +82,8 @@ public class SystemInfoInterceptor implements ClientInterceptor {
             // If a server is able to take the load of the requests, it should be able to serve some
             // additional lightweight static getSystemInfo calls that are serialized with the actual
             // calls.
-            getServerCapabilitiesWithRetryOrThrow(
-                serverCapabilitiesFuture, next, callOptions.getDeadline());
+            serverCapabilitiesFuture.complete(
+                getServerCapabilitiesOrThrow(next, callOptions.getDeadline()));
           }
         }
 
--- a/temporal-serviceclient/src/main/java/io/temporal/serviceclient/SystemInfoInterceptor.java
+++ b/temporal-serviceclient/src/main/java/io/temporal/serviceclient/SystemInfoInterceptor.java
@@ -21,6 +21,7 @@
 package io.temporal.serviceclient;
 
 import io.grpc.*;
+import io.grpc.health.v1.HealthGrpc;
 import io.temporal.api.workflowservice.v1.GetSystemInfoRequest;
 import io.temporal.api.workflowservice.v1.GetSystemInfoResponse;
 import io.temporal.api.workflowservice.v1.GetSystemInfoResponse.Capabilities;
@@ -74,6 +75,8 @@ public class SystemInfoInterceptor implements ClientInterceptor {
                     super.onClose(status, trailers);
                   }
                 };
+          } else if (method == HealthGrpc.getCheckMethod()) {
+            // do nothing
           } else {
             // Need to reach system capabilities, so make a getSystemInfo call in a blocking manner.
             // We don't try to squash into one and optimize the several getSystemInfo calls that may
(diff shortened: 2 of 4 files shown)


temporal-serviceclient/src/main/java/io/temporal/internal/retryer/GrpcAsyncRetryer.java
```diff
@@ -40,25 +35,34 @@ class GrpcAsyncRetryer<R> {
   private static final Logger log = LoggerFactory.getLogger(GrpcRetryer.class);
 
   private final ScheduledExecutorService executor;
-  private final GrpcRetryer.GrpcRetryerOptions options;
-  private final GetSystemInfoResponse.Capabilities serverCapabilities;
+  private final Supplier<GetSystemInfoResponse.Capabilities> serverCapabilities;
```
cretz, line 38: Will let @Quinn-With-Two-Ns review as he is more familiar w/ the Java side, but I have a couple of questions:

* Can you confirm (sorry I'm lazy) that this server capabilities getter is memoized? Meaning after the first time it is fetched, it never fetches again?
* Go does this similar approach for lazy server capabilities (which is also our first-call health check), but in order to validate the connection on worker start, we eagerly make sure the this capabilities call is made on worker start. Can we make sure Java does the same? Unsure of behavior today, but in other SDKs we've seen lazy clients w/ bad connection options just fail inside the worker poller never to have user-facing error.

chronos-tachyon: Regarding point 1, the Supplier comes from [`ChannelManager.getServerCapabilities()`](https://github.com/temporalio/sdk-java/blob/master/temporal-serviceclient/src/main/java/io/temporal/serviceclient/ChannelManager.java#L325) which seems to be using `j.u.c.CompletableFuture` as the world's strangest memoization cache.

cretz: :+1: Now just need to confirm we're doing some form of eager connection and capability lookup on worker start. I would guess we are.

temporal-serviceclient/src/main/java/io/temporal/internal/retryer/GrpcRetryerUtils.java
```diff
@@ -67,8 +66,14 @@ class GrpcRetryerUtils {
         // never retry these codes
         return currentException;
       case INTERNAL:
+        GetSystemInfoResponse.Capabilities capabilities;
+        try {
+          capabilities = serverCapabilities.get();
```
Quinn-With-Two-Ns, line 71: Sorry I don't understand this change, why do we want to catch every exception from `serverCapabilities.get()`? and if we do why only in this one place?

chronos-tachyon: `serverCapabilities.get()` performs a memoized `WorkflowService.getSystemInfo` gRPC call.  That can throw on the first call, or if all previous calls also threw.  If it throws, it will not be caught and retried, but instead will fail the entire operation (for which `WorkflowService.getSystemInfo` is almost irrelevant except for one error-handling branch).

temporal-serviceclient/src/main/java/io/temporal/internal/retryer/GrpcAsyncRetryer.java
chronos-tachyon: Okay, I've refactored a bunch of stuff to use eager `Capabilities` messages.  I also swapped out `CompletableFuture` for an `AtomicReference`, which it probably should have been using all along.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/ChannelManager.java
```diff
@@ -293,7 +287,7 @@ public void connect(String healthCheckServiceName, @Nullable Duration timeout) {
         new GrpcRetryer.GrpcRetryerOptions(
             RpcRetryOptions.newBuilder().setExpiration(timeout).validateBuildWithDefaults(), null);
 
-    new GrpcRetryer(getServerCapabilities())
+    new GrpcRetryer(getServerCapabilitiesNow())
```
cretz, line 290: Does this change turn server capabilities from lazy to eager?

temporal-serviceclient/src/main/java/io/temporal/internal/retryer/GrpcAsyncRetryer.java
cretz: Hrmm, I can see that for the final answer, but completable future can help making sure an invocation is lazy.

chronos-tachyon: I'm not sure I understand what you mean?  1) I thought making invocations lazy is exactly what you're asking us *not* to do, and 2) the future isn't exposed to the caller in the existing code, we just call `getNow()` on it when we want to access it.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/ChannelManager.java
chronos-tachyon: Yes, as you requested.

temporal-serviceclient/src/main/java/io/temporal/internal/retryer/GrpcAsyncRetryer.java
cretz: > I thought making invocations lazy is exactly what you're asking us not to do

Only at worker start, not in general. Sorry if I wasn't clear. Today's Java devs (unfortunately) expect creation of the client to not make any outbound calls IIUC.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/ChannelManager.java
cretz: I don't think we should change standard client behavior, I just wanted to confirm we fetched them on worker start (sorry if I was unclear). I figure we probably do since some of those capabilities are needed for worker, but just wanted to confirm.

cretz (commented):
After internal discussion, it seems everyone agrees that no when-to-make-first-call behavior should change in the Java SDK (that would be an incompatible change), but of course fixing retry behavior and other similar ones makes sense. Will defer to @Quinn-With-Two-Ns on final approval.

temporal-serviceclient/src/main/java/io/temporal/internal/retryer/GrpcRetryerUtils.java
Quinn-With-Two-Ns: Yeah I think failing the entire operation because `WorkflowService.getSystemInfo` failed is wrong, totally agree with you. What I am not sure is if this is the right way to solve the problem. We call `getSystemInfo` in other places, like starting a worker, do we need to do anything to those callers?

I can check how Go handle this, @Sushisource How does core handle `getSystemInfo` and specifically failures on `WorkflowService.getSystemInfo`? Does it retry it like any other gRPC request?

Sushisource: Huh, interesting. I would've said it retries, but, after looking it up it doesn't. Trivial to change though (and probably should be). You can see here we make the call before putting the client inside the retry wrapper: https://github.com/temporalio/sdk-core/blob/bcda5d9f73b082ddf6816921e15db5e68a8ce47e/client/src/lib.rs#L430

Quinn-With-Two-Ns: How would you handle it in core we should do the same thing in Java IMO. What I would expect is `get_system_info` is retried by the normal retry with a set deadline (5s?) and the retry has logic to avoid the recursion of `get_system_info` being needed by the retrier?

Sushisource: Yes. If all I do is just move that call to happen from the `RetryClient`, that will all work (though the default deadline is 30s). I could change the deadline to be shorter specifically for that call.

temporal-serviceclient/src/main/java/io/temporal/internal/retryer/GrpcAsyncRetryer.java
chronos-tachyon: Ah.  So it turns out this *does* currently get called on connect, since `ChannelManager.connect` calls `ChannelManager.healthCheck` in a `retryWithResult`.  (Unfortunately, `healthCheck` is doing a gRPC `Health.check` and not a `WorkflowService.getSystemInfo`, but that's a different problem.)

I'm currently looking into ways I can call WS.gSI from ChannelManager.connect to populate the `AtomicReference`, which would allow `getServerCapabilities` to be implemented as `return capabilitiesAtomicReference::get`.

temporal-serviceclient/src/main/java/io/temporal/internal/retryer/GrpcRetryer.java
```diff
@@ -32,41 +32,69 @@
...
+    }
   }
 
+  @FunctionalInterface
   public interface RetryableFunc<R, E extends Throwable> {
     R apply() throws E;
   }
 
-  public GrpcRetryer(Supplier<GetSystemInfoResponse.Capabilities> serverCapabilities) {
+  public static <T extends Throwable> void retry(
```
Quinn-With-Two-Ns, line 52: I'd prefer not to change this if there is not a user benefit.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/ChannelManager.java
```diff
@@ -322,18 +323,43 @@ public HealthCheckResponse healthCheck(
     return stub.check(HealthCheckRequest.newBuilder().setService(healthCheckServiceName).build());
   }
 
-  public Supplier<GetSystemInfoResponse.Capabilities> getServerCapabilities() {
-    return () -> {
-      synchronized (serverCapabilitiesFuture) {
-        GetSystemInfoResponse.Capabilities capabilities = serverCapabilitiesFuture.getNow(null);
+  public Capabilities getServerCapabilities() {
```
Quinn-With-Two-Ns, line 326: If  this is public we should not  change the signature.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/SystemInfoInterceptor.java
```diff
@@ -1,104 +0,0 @@
...
-
-        super.start(responseListener, headers);
-      }
-    };
-  }
-
-  public static GetSystemInfoResponse.Capabilities getServerCapabilitiesOrThrow(
-      Channel channel, @Nullable Deadline deadline) {
-    try {
-      return WorkflowServiceGrpc.newBlockingStub(channel)
```
Quinn-With-Two-Ns, line 93: I think we just need to add a retried with a deadline here no?

temporal-serviceclient/src/main/java/io/temporal/serviceclient/ChannelManager.java
chronos-tachyon: It is not.  It's a package-private class.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/SystemInfoInterceptor.java
chronos-tachyon: ???

I thought we'd discussed the fact that no other language SDK has an interceptor for this purpose.

In particular, after studying it I don't believe this interceptor ever does anything, ever, since all `CompletableFuture.complete` calls after the first will be ignored.

temporal-serviceclient/src/main/java/io/temporal/internal/retryer/GrpcRetryer.java
chronos-tachyon: I realize the benefit is tiny, but creating less garbage on the GC heap is a user benefit.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/SystemInfoInterceptor.java
chronos-tachyon: On second look, are you speaking specifically of this method?  I moved it to ChannelManager and renamed it to `getServerCapabilitiesRaw`, then wrapped it with a retry.  I already did what you're asking for?

Quinn-With-Two-Ns: My suggestion was to keep to simple and just retry this call and don't modify anything else.

Quinn-With-Two-Ns: >I thought we'd discussed the fact that no other language SDK has an interceptor for this purpose.

`getServerCapabilities` is supposed to be called before any RPC, an interceptor is a way to accomplish that and I do believe it does that.

temporal-serviceclient/src/main/java/io/temporal/internal/retryer/GrpcRetryer.java
Quinn-With-Two-Ns: Not worth it in my opinion, if we every do need to add any state to the retrier will have to revert this change anyway.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/SystemInfoInterceptor.java
chronos-tachyon: > > I thought we'd discussed the fact that no other language SDK has an interceptor for this purpose.
> 
> `getServerCapabilities` is supposed to be called before any RPC, an interceptor is a way to accomplish that and I do believe it does that.

Um, what, no?  It hooks into existing calls to `WorkflowService.getSystemInfo` to capture the outputs.  It does not provoke additional calls to `WorkflowService.getSystemInfo`.  Have you ever read the code?

temporal-serviceclient/src/main/java/io/temporal/serviceclient/ChannelManager.java
Quinn-With-Two-Ns: Ah yeah I saw it being used in other modules, but they are under the same package. My bad

temporal-serviceclient/src/main/java/io/temporal/serviceclient/SystemInfoInterceptor.java
Quinn-With-Two-Ns: But there is an if condition that makes a call no?
```
if (method == WorkflowServiceGrpc.getGetSystemInfoMethod()) {
...
} else {
            // Need to reach system capabilities, so make a getSystemInfo call in a blocking manner.
            // We don't try to squash into one and optimize the several getSystemInfo calls that may
            // be initiated by several client calls here. Doing so it will require tricky
            // implementation to ensure proper deadlines that may be different between calls.
            // If a server is able to take the load of the requests, it should be able to serve some
            // additional lightweight static getSystemInfo calls that are serialized with the actual
            // calls.
            serverCapabilitiesFuture.complete(
                getServerCapabilitiesOrThrow(next, callOptions.getDeadline()));
          }
}
```

If the call is not `getGetSystemInfoMethod` won't we call `getServerCapabilitiesOrThrow()` which does make an an additional call to `WorkflowService.getSystemInfo`?

chronos-tachyon: That code path is only executed if `!serverCapabilitiesFuture.isDone()`.  It is done when `ChannelManager.connect` calls `new GrpcRetryer(getServerCapabilities()).retryWithResult`, which calls `Supplier.get` before entering the retry loop, which completes the future in the lambda body of `ChannelManager.getServerCapabilities`.  I guarantee to you that this code path will never execute if you put a logging statement in it.

Quinn-With-Two-Ns:
I don't think the CI failure is a result of any of your changes, I opened a separate issue to address this flake https://github.com/temporalio/sdk-java/issues/2038

Quinn-With-Two-Ns:
As part of the PR can we please also add some unit tests to verify the changes address the issue? I understand the integration tests will live, so maybe just a targeted test of `getServerCapabilitiesWithRetryOrThrow` since that takes a `channel` and you can create a in processes server like this:

https://github.com/grpc/grpc-java/blob/master/examples/src/test/java/io/grpc/examples/helloworld/HelloWorldClientTest.java

PR #2025, opened 2024-04-05
Make StickyQueueBalancer synchronized

Quinn-With-Two-Ns:
Make StickyQueueBalancer synchronized to avoid potential race where all initial calls to `makePoll()` all get assigned the same task queue.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/worker/StickyQueueBalancer.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/worker/StickyQueueBalancer.java
@@ -21,19 +21,16 @@
 package io.temporal.internal.worker;
 
 import io.temporal.api.enums.v1.TaskQueueKind;
-import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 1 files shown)


Sushisource (approved):

temporal-sdk/src/main/java/io/temporal/internal/worker/StickyQueueBalancer.java
```diff
@@ -43,35 +40,35 @@ public StickyQueueBalancer(int pollersCount, boolean stickyQueueEnabled) {
   /**
    * @return task queue kind that should be used for the next poll
    */
-  public TaskQueueKind makePoll() {
+  public synchronized TaskQueueKind makePoll() {
```
cretz, line 43: Usually linters will tell you not to add this keyword but instead make a private final object that you wrap in `synchronized (theObject) {` because it can be unclear to users of this class that they can cause deadlocks if they try to also synchronize on this instance. Probably fine in internal code though.

cretz (approved):

PR #2039, opened 2024-04-15
Update proto API to 1.32.0

Quinn-With-Two-Ns:
Update Proto API to 1.32.0

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,3 @@
 [submodule "temporal-serviceclient/src/main/proto"]
 	path = temporal-serviceclient/src/main/proto
 	url = https://github.com/temporalio/api.git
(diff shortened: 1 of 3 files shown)


.gitmodules
```diff
@@ -1,4 +1,3 @@
 [submodule "temporal-serviceclient/src/main/proto"]
 	path = temporal-serviceclient/src/main/proto
 	url = https://github.com/temporalio/api.git
-	hexsha = ddf07ab9933e8230309850e3c579e1ff34b03f53
```
Quinn-With-Two-Ns, line 4: This does not mean anything AFAIK 

https://git-scm.com/docs/gitsubmodules

cretz (approved):

PR #2041, opened 2024-04-17
Add support for update admitted event

Quinn-With-Two-Ns:
Add support for update admitted event. Required some modification to how the Java SDK batches event during replay because when the SDK processed an update admitted event while replaying it needs to know if it was accepted or not. Preciously the SDK batches events like [WFT started, WFT completed, Command events] and now it batches events like [Non command events, WFT scheduled, WFT started, WFT completed, Command events]
.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/statemachines/WFTBuffer.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/statemachines/WFTBuffer.java
@@ -20,13 +20,13 @@
 
 package io.temporal.internal.statemachines;
 
-import com.google.common.base.Preconditions;
 import io.temporal.api.enums.v1.EventType;
 import io.temporal.api.history.v1.HistoryEvent;
 import io.temporal.internal.common.WorkflowExecutionUtils;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Optional;
 
 /**
  * This class buffers events between WorkflowTaskStarted events and return them in one chunk so any
@@ -42,9 +42,29 @@ public class WFTBuffer {
     Closed,
   }
 
+  public static class EventBatch {
+    private final List<HistoryEvent> events;
+    private final Optional<HistoryEvent> workflowTaskCompletedEvent;
+
+    public EventBatch(
+        Optional<HistoryEvent> workflowTaskCompletedEvent, List<HistoryEvent> events) {
+      this.workflowTaskCompletedEvent = workflowTaskCompletedEvent;
+      this.events = events;
+    }
+
+    public List<HistoryEvent> getEvents() {
+      return events;
+    }
+
+    public Optional<HistoryEvent> getWorkflowTaskCompletedEvent() {
+      return workflowTaskCompletedEvent;
+    }
+  }
+
   private WFTState wftSequenceState = WFTState.None;
 
   private final List<HistoryEvent> wftBuffer = new ArrayList<>();
+  private Optional<HistoryEvent> workflowTaskCompletedEvent = Optional.empty();
   private final List<HistoryEvent> readyToFetch = new ArrayList<>();
 
   /**
@@ -73,9 +93,7 @@ public class WFTBuffer {
     // This is the only way to enter into the WFT sequence -
     // any WorkflowTaskStarted event that it not the last in the history
     if (EventType.EVENT_TYPE_WORKFLOW_TASK_STARTED.equals(event.getEventType())) {
-      // if there is something in wftBuffer, let's flush it
-      flushBuffer();
-      // and init a new sequence
+      // Init a new sequence
       wftSequenceState = WFTState.Started;
       addToBuffer(event);
       return;
@@ -83,7 +101,12 @@ public class WFTBuffer {
 
     if (WFTState.Started.equals(wftSequenceState)
         && WorkflowExecutionUtils.isWorkflowTaskClosedEvent(event)) {
-      wftSequenceState = WFTState.Closed;
+      if (event.getEventType().equals(EventType.EVENT_TYPE_WORKFLOW_TASK_COMPLETED)) {
+        workflowTaskCompletedEvent = Optional.of(event);
+        wftSequenceState = WFTState.Closed;
+      } else {
+        wftSequenceState = WFTState.None;
+      }
       addToBuffer(event);
       return;
     }
@@ -94,14 +117,19 @@ public class WFTBuffer {
       // If event is WFT_STARTED or any of the Closing events, it's handled by if statements
       // earlier, so it's safe to switch to None here, we are not inside WFT sequence
       wftSequenceState = WFTState.None;
-      // no open WFT sequence, can't add to buffer, it's ok to add directly to readyToFetch, this
-      // event can't be EVENT_TYPE_WORKFLOW_TASK_STARTED because we checked it above.
-      readyToFetch.add(event);
+
+      addToBuffer(event);
+      return;
+    }
+    if (WFTState.Closed.equals(wftSequenceState) && WorkflowExecutionUtils.isCommandEvent(event)) {
+      // we are inside a closed WFT sequence, we can add to buffer
+      addToBuffer(event);
       return;
     }
 
-    if (WFTState.None.equals(wftSequenceState)) {
-      // we should be returning the events one by one, we are not inside a WFT sequence
+    if (WorkflowExecutionUtils.isCommandEvent(event)
+        || WorkflowExecutionUtils.isWorkflowTaskClosedEvent(event)) {
+      flushBuffer();
       readyToFetch.add(event);
     } else {
       addToBuffer(event);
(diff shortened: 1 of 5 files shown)


temporal-sdk/src/main/java/io/temporal/internal/statemachines/WFTBuffer.java
```diff
@@ -114,21 +142,22 @@ private void flushBuffer() {
...
       readyToFetch.clear();
-      return Collections.singletonList(event);
+      return new EventBatch(wftStarted, Collections.singletonList(event));
     } else {
       List<HistoryEvent> result = new ArrayList<>(readyToFetch);
+      Optional<HistoryEvent> wftStarted = workflowTaskCompletedEvent;
+      workflowTaskCompletedEvent = Optional.empty();
       readyToFetch.clear();
-      return result;
+      return new EventBatch(wftStarted, result);
```
Sushisource, line 160: GH won't let me suggest here because it's a whiny baby about deleted lines, but looks like this could be:

```java
    List<HistoryEvent> res;
    if (readyToFetch.size() == 1) {
      HistoryEvent event = readyToFetch.get(0);
      res = Collections.singletonList(event));
    } else {
      res = new ArrayList<>(readyToFetch);
    }
    Optional<HistoryEvent> wftStarted = workflowTaskCompletedEvent;
    workflowTaskCompletedEvent = Optional.empty();
    readyToFetch.clear();
    return new EventBatch(wftStarted, res);
```

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
```diff
@@ -304,7 +309,10 @@ private void handleEventsBatch(List<HistoryEvent> events, boolean hasNextEvent)
       }
 
       try {
-        handleSingleEvent(event, iterator.hasNext() || hasNextEvent);
+        handleSingleEvent(
+            event,
+            !hasNextEvent && !eventBatch.getWorkflowTaskCompletedEvent().isPresent(),
+            iterator.hasNext() || hasNextEvent);
```
Sushisource, line 315: I'd suggest lifting these bool operations up into named bindings to make it clear what they're for

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
```diff
@@ -352,6 +367,10 @@ private void handleSingleEventLookahead(HistoryEvent event) {
           }
           flags.setSdkFlag(sdkFlag);
         }
+        // Remove any finished update protocol state machines. We can't remove them on an event like
+        // other state machines
+        // because a rejected update produces no event in history.
```
Sushisource, line 372: nit: weird line wrapping

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
```diff
@@ -416,16 +435,28 @@ private void handleSingleMessage(Message message) {
...
 
-  private void handleSingleEvent(HistoryEvent event, boolean hasNextEvent) {
+  private void handleSingleEvent(HistoryEvent event, boolean lastTask, boolean hasNextEvent) {
     if (isCommandEvent(event)) {
       handleCommandEvent(event);
       return;
     }
 
+    // This definition of replaying here is that we are no longer replaying as soon as we
+    // see new events that have never been seen or produced by the SDK.
```
Sushisource, line 445: Nice :+1: glad we're consistent now

Sushisource (approved):
Makes sense

temporal-sdk/src/test/java/io/temporal/internal/statemachines/UpdateProtocolStateMachineTest.java
```diff
@@ -396,4 +400,486 @@ protected void update(UpdateMessage message, AsyncWorkflowBuilder<Void> builder)
...
+    {
+      TestEntityManagerListenerBase listener = new TestUpdateListener();
+      stateMachines = newStateMachines(listener);
+      List<Command> commands = h.handleWorkflowTaskTakeCommands(stateMachines);
+      assertEquals(0, commands.size());
+    }
+  }
+
+  @Test
+  public void testUpdateSignalSandwichAdmittedMessage() {
```
Quinn-With-Two-Ns, line 732: @Sushisource @dandavison This is something we should test in Core as well, previously due to server limitations, it wasn't possible to have an `[update, signal, update]` in history, but with update admitted it now is. We should check the handlers get called in the correct order.

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
Quinn-With-Two-Ns: Yeah I just took the comment straight from core

Sushisource: Hehe I noticed :)

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
```diff
@@ -705,6 +735,20 @@ private void handleNonStatefulEvent(HistoryEvent event, boolean hasNextEvent) {
...
+            event.getWorkflowExecutionUpdateAdmittedEventAttributes();
+        Message msg =
+            Message.newBuilder()
+                .setId(admittedEvent.getRequest().getMeta().getUpdateId() + "/request")
+                .setProtocolInstanceId(admittedEvent.getRequest().getMeta().getUpdateId())
+                .setEventId(event.getEventId())
+                .setBody(Any.pack(admittedEvent.getRequest()))
+                .build();
+        if (replaying && acceptedUpdates.remove(msg.getProtocolInstanceId()) || !replaying) {
+          messages.add(msg);
```
dandavison, line 749: (Note to check I'm understanding; correct me if not!) OK so here, you've already peeked forwards, and hence on encountering `Admitted` know whether it's going to be `Accepted` in the future. In contrast, in the `sdk-core` implementation, we do not peek forwards for this, and instead hold the durably `Admitted` info for use when we encounter an `Accepted`.

dandavison (approved):
Nice, I really like the new consistency between core and Java regarding `isReplaying`.

Not specific to Java but one thing that's at the back of mind and worries me slightly is: I've made server code not send protocol messages for durably admitted updates. But it's a fragile contract. What would happen if a server engineer accidentally changed that? Do we have tests that would fail in a fairly understandable way?

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
```diff
@@ -160,6 +161,9 @@ enum HandleEventStatus {
 
   private List<Message> messages = new ArrayList<>();
 
+  /** Set of accepted admitted updates by update id */
+  private final Set<String> acceptedUpdates = new HashSet<>();
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: I am going to refactor this a bit so this comment won't apply anymore

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
Quinn-With-Two-Ns: Hm thinking about it more i think I can make Javas implementation align more with Core and Go's. going to resolve all the comments related how admitted is handled since it will all change slightly.

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
```diff
@@ -705,6 +726,34 @@ private void handleNonStatefulEvent(HistoryEvent event, boolean hasNextEvent) {
...
+                      m -> {
+                        if (m.getProtocolInstanceId().equals(msg.getProtocolInstanceId())) {
+                          // Check if we need to replace the update message synthesize from an
+                          // accepted event with the update message synthesize from an admitted
+                          // event
+                          return m.toBuilder().setBody(msg.getBody()).build();
+                        }
+                        return m;
+                      })
+                  .collect(Collectors.toList());
```
dandavison, line 752: Technically this is quadratic in the number of events -- does this worry you at all from a performance point of view? For example, suppose someone did a batch reset of many workflows that were accepting a high volume of updates (some sort of entity workflow pattern). That could send a post-reset spike of `UpdateAdmitted` events. Could a scenario like that result in a spike in worker CPU activity worth worrying about?

Quinn-With-Two-Ns: Why is it quadratic and why over the number of events? A map is a linear operation and a collect is as well? very similar to your PR in core https://github.com/temporalio/sdk-core/pull/700/files#diff-fcb2b6f2d5782438cd1f138a0d39cc349ddfaf7e3f4049ad8a7499631cacc3f1R729

dandavison: This map/collect over messages is inside a for loop over events starting at [WorkflowStateMachines.java#L310](https://github.com/temporalio/sdk-java/blob/master/temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java#L310), and the number of messages grows with the number of events, since we synthesize them from events. (I haven't looked at core, I believe you that my question could apply there too!)

Quinn-With-Two-Ns: Oh the number of Update Admitted Events yes that makes sense, sorry I thought you meant over the total number of history events my bad. Yeah I could switch to storing messages in a map, hmm

Quinn-With-Two-Ns: OK going to go back to the previous approach , not sure how big a problem issue this would really be, but neither solution is clearly better then the other.

dandavison: > very similar to your PR in core https://github.com/temporalio/sdk-core/pull/700/files#diff-fcb2b6f2d5782438cd1f138a0d39cc349ddfaf7e3f4049ad8a7499631cacc3f1R729

Right, my PR in core is using a hash map, so that seems reasonable to me if you did want to change it.

PR #2050, opened 2024-05-02
Add ScheduleClientInterceptor APIs (fixes #2048)

### LINKED ISSUE #2048: ScheduleClient is missing an option to add interceptors
## Expected Behavior
There should be an API on ScheduleClientOptions to add ScheduleClientCallsInterceptor (or maybe ScheduleClientInterceptor, if the pattern used by WorkflowClient is followed) instances to the ScheduleClient.

## Actual Behavior
There is no such API.

dano:
## What was changed
This adds APIs for users to specify interceptors when creating a `ScheduleClient`. Previously there was interceptor infrastructure built into the internals of the client, but no public API to allow users to actually add them.

## Why?
So that users can add interceptors as needed.

## Checklist

1. Closes #2048 

2. How was this tested:
Unit tests

3. Any docs updates needed?
Not sure. The current docs don't talk about how `ScheduleClient` is created right now, so probably not?

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/client/schedules/ScheduleClientOptions.java
+++ b/temporal-sdk/src/main/java/io/temporal/client/schedules/ScheduleClientOptions.java
@@ -23,6 +23,7 @@ package io.temporal.client.schedules;
 import io.temporal.common.context.ContextPropagator;
 import io.temporal.common.converter.DataConverter;
 import io.temporal.common.converter.GlobalDataConverter;
+import io.temporal.common.interceptors.ScheduleClientInterceptor;
 import java.lang.management.ManagementFactory;
 import java.util.Collections;
 import java.util.List;
@@ -56,11 +57,14 @@ public final class ScheduleClientOptions {
     private static final String DEFAULT_NAMESPACE = "default";
     private static final List<ContextPropagator> EMPTY_CONTEXT_PROPAGATORS =
         Collections.emptyList();
+    private static final List<ScheduleClientInterceptor> EMPTY_INTERCEPTORS =
+        Collections.emptyList();
 
     private String namespace;
     private DataConverter dataConverter;
     private String identity;
     private List<ContextPropagator> contextPropagators;
+    private List<ScheduleClientInterceptor> interceptors;
 
     private Builder() {}
 
@@ -72,6 +76,7 @@ public final class ScheduleClientOptions {
       dataConverter = options.dataConverter;
       identity = options.identity;
       contextPropagators = options.contextPropagators;
+      interceptors = options.interceptors;
     }
 
     /** Set the namespace this client will operate on. */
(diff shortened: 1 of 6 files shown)


temporal-sdk/src/main/java/io/temporal/client/schedules/ScheduleClientOptions.java
```diff
@@ -106,30 +111,44 @@ public Builder setContextPropagators(List<ContextPropagator> contextPropagators)
       return this;
     }
 
+    /**
+     * Set the interceptors for this client.
+     *
+     * @param interceptors specifies the list of interceptors to use with the client.
+     */
+    public Builder setInterceptors(List<ScheduleClientInterceptor> interceptors) {
```
Quinn-With-Two-Ns, line 119: Was going to comment that we need to update the `toString` etc, but `ScheduleClientOptions ` doesn't implement those. Will make a separate issue to address this

temporal-sdk/src/main/java/io/temporal/common/interceptors/ScheduleClientInterceptor.java
```diff
@@ -0,0 +1,42 @@
...
+import io.temporal.client.schedules.ScheduleClient;
+import io.temporal.client.schedules.ScheduleHandle;
+import io.temporal.common.Experimental;
+
+/**
+ * Intercepts calls to the {@link ScheduleClient} and {@link ScheduleHandle} related to the
+ * lifecycle of a Schedule.
+ */
+@Experimental
+public interface ScheduleClientInterceptor {
```
Quinn-With-Two-Ns, line 32: I think we should also create a base interceptor and recommend users only override the methods they need,  like we have `WorkflowClientInterceptorBase` for `WorkflowClientInterceptor`

Quinn-With-Two-Ns:
Thanks for the contribution! I think we just need a base implementation like other interceptors and we are good to merge

temporal-sdk/src/main/java/io/temporal/common/interceptors/ScheduleClientInterceptor.java
dano: Sure, done. I also noticed the CI pipeline failed, I think because one of the assertions in my unit test assumed that no other schedules would exist on the external Temporal instance. Hopefully I've fixed that now as well.

dano: Ah, from looking at other tests, I think I just need to insert a sleep after I create the schedule to ensure it appears when you subsequently call `listSchedules()`. Giving that a shot now.

dano:
Unit test issue is fixed now.

Quinn-With-Two-Ns (approved):

Issue #2033, opened 2024-04-11
Support for WorkflowIdConflictPolicy


josh-berry:
See https://github.com/temporalio/features/issues/437

PR #2055, opened 2024-05-07
Add setWorkflowIdConflictPolicy

Quinn-With-Two-Ns:
Add setWorkflowIdConflictPolicy

closes https://github.com/temporalio/sdk-java/issues/2033

### DIFF
--- a/temporal-sdk/src/test/java/io/temporal/client/functional/WorkflowIdConflictPolicyTest.java
+++ b/temporal-sdk/src/test/java/io/temporal/client/functional/WorkflowIdConflictPolicyTest.java
@@ -1,0 +1,196 @@
+/*
+ * Copyright (C) 2022 Temporal Technologies, Inc. All Rights Reserved.
+ *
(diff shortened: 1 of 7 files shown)


temporal-sdk/src/test/java/io/temporal/client/functional/WorkflowIdConflictPolicyTest.java
```diff
@@ -0,0 +1,196 @@
...
+            testWorkflowRule
+                .getWorkflowClient()
+                .newUntypedWorkflowStub(
+                    workflowExecution1,
+                    Optional.of(TestWorkflows.TestSignaledWorkflow.class.toString()))
+                .getResult(String.class));
+  }
+
+  @Test
+  public void policyUseExisting() {
```
cretz, line 87: Looks like this test fails against server (as it should). We should probably just hold up this PR until the code is in server.

cretz (approved):

Issue #2045, opened 2024-04-25
Make startUpdate users aware that it's synchronous w/ worker


drewhoskins-temporal:
**Is your feature request related to a problem? Please describe.**
https://github.com/temporalio/features/issues/469

**Describe the solution you'd like**
Related to https://github.com/temporalio/sdk-java/blob/ed211fa611112288b576a2c979be9284e17fec89/temporal-sdk/src/main/java/io/temporal/client/WorkflowStub.java#L94-L106
Developers should be informed about and opt in to behavior that `startWorkflow` waits for acceptance

**Describe alternatives you've considered**

**Additional context**

PR #2066, opened 2024-05-17
Don't return update handles until desired stage reached

Sushisource:
## What was changed
* Add admitted stage to wait policy
* Don't return update handle from startUpdate until the update is complete if the user specified that stage (not sure why they would, though, when using start)
* Retry submitting update to server until seen accepted

## Why?
See https://github.com/temporalio/features/issues/432

## Checklist

1. Closes https://github.com/temporalio/sdk-java/issues/2002

2. How was this tested:
Existing / new tests

3. Any docs updates needed?

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/client/LazyUpdateHandleImpl.java
+++ b/temporal-sdk/src/main/java/io/temporal/client/LazyUpdateHandleImpl.java
@@ -72,7 +72,7 @@ final class LazyUpdateHandleImpl<T> implements UpdateHandle<T> {
 
   @Override
   public CompletableFuture<T> getResultAsync(long timeout, TimeUnit unit) {
-    WorkflowClientCallsInterceptor.PollWorkflowUpdateOutput output =
+    WorkflowClientCallsInterceptor.PollWorkflowUpdateOutput<T> output =
         workflowClientInvoker.pollWorkflowUpdate(
             new WorkflowClientCallsInterceptor.PollWorkflowUpdateInput<>(
                 execution, updateName, id, resultClass, resultType, timeout, unit));
--- a/temporal-sdk/src/main/java/io/temporal/client/LazyUpdateHandleImpl.java
+++ b/temporal-sdk/src/main/java/io/temporal/client/LazyUpdateHandleImpl.java
@@ -27,7 +27,6 @@ import io.temporal.common.Experimental;
 import io.temporal.common.interceptors.WorkflowClientCallsInterceptor;
 import io.temporal.serviceclient.CheckedExceptionWrapper;
 import java.lang.reflect.Type;
-import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
 import java.util.concurrent.TimeUnit;
--- a/temporal-sdk/src/main/java/io/temporal/client/LazyUpdateHandleImpl.java
+++ b/temporal-sdk/src/main/java/io/temporal/client/LazyUpdateHandleImpl.java
@@ -42,7 +42,7 @@ final class LazyUpdateHandleImpl<T> implements UpdateHandle<T> {
   private final WorkflowExecution execution;
   private final Class<T> resultClass;
   private final Type resultType;
-  private WorkflowClientCallsInterceptor.PollWorkflowUpdateOutput<T> previousPollCall;
+  private WorkflowClientCallsInterceptor.PollWorkflowUpdateOutput<T> waitCompletedPollCall;
 
   LazyUpdateHandleImpl(
       WorkflowClientCallsInterceptor workflowClientInvoker,
@@ -73,13 +73,21 @@ final class LazyUpdateHandleImpl<T> implements UpdateHandle<T> {
 
   @Override
   public CompletableFuture<T> getResultAsync(long timeout, TimeUnit unit) {
-    WorkflowClientCallsInterceptor.PollWorkflowUpdateOutput<T> pollCall;
-    if (previousPollCall == null) {
-      pollUntilComplete(timeout, unit);
+
+    WorkflowClientCallsInterceptor.PollWorkflowUpdateOutput<T> pollCall = null;
+    boolean setFromWaitCompleted = false;
+
+    // If waitCompleted was called, use the result from that call.
+    synchronized (this) {
+      if (waitCompletedPollCall != null) {
+        pollCall = waitCompletedPollCall;
+        waitCompletedPollCall = null;
+      }
     }
-    pollCall = previousPollCall;
-    // Get result may be called more than once for non-complete wait policies, so reset it here.
-    previousPollCall = null;
+
+    if (!setFromWaitCompleted) {
+      pollCall = pollUntilComplete(timeout, unit);
+    }
 
     return pollCall
         .getResult()
@@ -114,10 +122,16 @@ final class LazyUpdateHandleImpl<T> implements UpdateHandle<T> {
     return this.getResultAsync(Long.MAX_VALUE, TimeUnit.MILLISECONDS);
   }
 
-  void pollUntilComplete(long timeout, TimeUnit unit) {
-    previousPollCall =
-        workflowClientInvoker.pollWorkflowUpdate(
-            new WorkflowClientCallsInterceptor.PollWorkflowUpdateInput<>(
-                execution, updateName, id, resultClass, resultType, timeout, unit));
+  // Can be called immediately after initialization to wait for the update to be completed, but
+  // still have the result be returned by getResultAsync.
+  void waitCompleted() {
+    waitCompletedPollCall = pollUntilComplete(Long.MAX_VALUE, TimeUnit.MILLISECONDS);
+  }
+
+  private WorkflowClientCallsInterceptor.PollWorkflowUpdateOutput<T> pollUntilComplete(
+      long timeout, TimeUnit unit) {
+    return workflowClientInvoker.pollWorkflowUpdate(
+        new WorkflowClientCallsInterceptor.PollWorkflowUpdateInput<>(
+            execution, updateName, id, resultClass, resultType, timeout, unit));
   }
 }
--- a/temporal-sdk/src/main/java/io/temporal/client/LazyUpdateHandleImpl.java
+++ b/temporal-sdk/src/main/java/io/temporal/client/LazyUpdateHandleImpl.java
@@ -27,6 +27,7 @@ import io.temporal.common.Experimental;
 import io.temporal.common.interceptors.WorkflowClientCallsInterceptor;
 import io.temporal.serviceclient.CheckedExceptionWrapper;
 import java.lang.reflect.Type;
+import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
 import java.util.concurrent.TimeUnit;
@@ -42,6 +43,7 @@ final class LazyUpdateHandleImpl<T> implements UpdateHandle<T> {
   private final WorkflowExecution execution;
   private final Class<T> resultClass;
   private final Type resultType;
+  private WorkflowClientCallsInterceptor.PollWorkflowUpdateOutput<T> previousPollCall;
 
   LazyUpdateHandleImpl(
       WorkflowClientCallsInterceptor workflowClientInvoker,
@@ -72,12 +74,11 @@ final class LazyUpdateHandleImpl<T> implements UpdateHandle<T> {
 
   @Override
   public CompletableFuture<T> getResultAsync(long timeout, TimeUnit unit) {
-    WorkflowClientCallsInterceptor.PollWorkflowUpdateOutput<T> output =
-        workflowClientInvoker.pollWorkflowUpdate(
-            new WorkflowClientCallsInterceptor.PollWorkflowUpdateInput<>(
-                execution, updateName, id, resultClass, resultType, timeout, unit));
+    if (previousPollCall == null) {
+      pollUntilComplete(timeout, unit);
+    }
 
-    return output
+    return previousPollCall
         .getResult()
         .exceptionally(
             failure -> {
@@ -108,5 +109,12 @@ final class LazyUpdateHandleImpl<T> implements UpdateHandle<T> {
   @Override
   public CompletableFuture<T> getResultAsync() {
     return this.getResultAsync(Long.MAX_VALUE, TimeUnit.MILLISECONDS);
+  }
+
+  void pollUntilComplete(long timeout, TimeUnit unit) {
+    previousPollCall =
+        workflowClientInvoker.pollWorkflowUpdate(
+            new WorkflowClientCallsInterceptor.PollWorkflowUpdateInput<>(
+                execution, updateName, id, resultClass, resultType, timeout, unit));
   }
 }
--- a/temporal-sdk/src/main/java/io/temporal/client/LazyUpdateHandleImpl.java
+++ b/temporal-sdk/src/main/java/io/temporal/client/LazyUpdateHandleImpl.java
@@ -73,11 +73,15 @@ final class LazyUpdateHandleImpl<T> implements UpdateHandle<T> {
 
   @Override
   public CompletableFuture<T> getResultAsync(long timeout, TimeUnit unit) {
+    WorkflowClientCallsInterceptor.PollWorkflowUpdateOutput<T> pollCall;
     if (previousPollCall == null) {
       pollUntilComplete(timeout, unit);
     }
+    pollCall = previousPollCall;
+    // Get result may be called more than once for non-complete wait policies, so reset it here.
+    previousPollCall = null;
 
-    return previousPollCall
+    return pollCall
         .getResult()
         .exceptionally(
             failure -> {
--- a/temporal-sdk/src/main/java/io/temporal/client/UpdateWaitPolicy.java
+++ b/temporal-sdk/src/main/java/io/temporal/client/UpdateWaitPolicy.java
@@ -23,7 +23,19 @@ package io.temporal.client;
 import io.temporal.api.enums.v1.UpdateWorkflowExecutionLifecycleStage;
 
 public enum UpdateWaitPolicy {
-  /** Update request waits for the update to be accepted by the workflow */
+  /**
+   * Update request waits for the update to be until the update request has been admitted by the
+   * server - it may be the case that due to a considerations like load or resource limits that an
+   * update is made to wait before the server will indicate that it has been received and will be
+   * processed. This value does not wait for any sort of acknowledgement from a worker.
+   */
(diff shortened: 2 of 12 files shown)


temporal-sdk/src/test/java/io/temporal/workflow/shared/TestWorkflows.java
```diff
@@ -177,16 +177,6 @@ public interface QueryableWorkflow {
     void mySignal(String value);
   }
 
-  @WorkflowInterface
-  public interface SimpleWorkflowWithUpdate {
```
Sushisource, line 181: This was unused

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
```diff
@@ -334,8 +334,17 @@ public <R> StartUpdateOutput<R> startUpdate(StartUpdateInput<R> input) {
...
     Deadline pollTimeoutDeadline = Deadline.after(POLL_UPDATE_TIMEOUT_S, TimeUnit.SECONDS);
-    UpdateWorkflowExecutionResponse result =
-        genericClient.update(updateRequest, pollTimeoutDeadline);
+
+    // Re-attempt the update until it is at least accepted, or passes the lifecycle stage specified
+    // by the user.
+    UpdateWorkflowExecutionResponse result;
+    do {
+      result = genericClient.update(updateRequest, pollTimeoutDeadline);
+    } while (result.getStage().getNumber() < input.getWaitPolicy().getLifecycleStage().getNumber()
```
Quinn-With-Two-Ns, line 343: Do we set a default for `input.getWaitPolicy()`?

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
```diff
@@ -334,8 +334,17 @@ public <R> StartUpdateOutput<R> startUpdate(StartUpdateInput<R> input) {
             .setRequest(request)
             .build();
     Deadline pollTimeoutDeadline = Deadline.after(POLL_UPDATE_TIMEOUT_S, TimeUnit.SECONDS);
```
Quinn-With-Two-Ns, line 336: Shouldn't the deadline be in the loop?

temporal-sdk/src/main/java/io/temporal/client/LazyUpdateHandleImpl.java
```diff
@@ -72,12 +73,15 @@ public String getId() {
 
   @Override
   public CompletableFuture<T> getResultAsync(long timeout, TimeUnit unit) {
-    WorkflowClientCallsInterceptor.PollWorkflowUpdateOutput output =
-        workflowClientInvoker.pollWorkflowUpdate(
-            new WorkflowClientCallsInterceptor.PollWorkflowUpdateInput<>(
-                execution, updateName, id, resultClass, resultType, timeout, unit));
+    WorkflowClientCallsInterceptor.PollWorkflowUpdateOutput<T> pollCall;
```
Quinn-With-Two-Ns, line 76: Sorry, i have read this a few times and I am not sure logic trying to accomplish?

Quinn-With-Two-Ns:
While we are refactoring this area could we also do https://github.com/temporalio/sdk-java/issues/2045? I believe @cretz did this in python by requiring a wait stage be passed? If not I'll do it in a follow up PR.

temporal-sdk/src/main/java/io/temporal/client/UpdateWaitPolicy.java
```diff
@@ -23,7 +23,19 @@
 import io.temporal.api.enums.v1.UpdateWorkflowExecutionLifecycleStage;
 
 public enum UpdateWaitPolicy {
```
Quinn-With-Two-Ns, line 25: Hmm in python looks like we changed the name to `WorkflowUpdateStage` I think we should do the same here because this enum will also be used when describing an updates stage.

cretz: 👍 And the docs here about what each enum means only apply to starting a workflow and should probably move to there.

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
cretz: Per @drewhoskins-temporal's latest requirements, we want wait-for-stage to be a required field for start. Also, we should call it "wait-for-stage" IMO to match Python and future SDKs (or if we don't like that term, we should call it something else and be consistent across SDKs with what it is called).

cretz, line 343: I think the latest requirements for start were to, if the wait stage is COMPLETED, after ACCEPTED you switched to polling for response _before_ returning from the start call. Can you confirm at least from the user perspective that occurs?

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
```diff
@@ -334,8 +334,17 @@ public <R> StartUpdateOutput<R> startUpdate(StartUpdateInput<R> input) {
...
-    UpdateWorkflowExecutionResponse result =
-        genericClient.update(updateRequest, pollTimeoutDeadline);
+
+    // Re-attempt the update until it is at least accepted, or passes the lifecycle stage specified
+    // by the user.
+    UpdateWorkflowExecutionResponse result;
+    do {
+      result = genericClient.update(updateRequest, pollTimeoutDeadline);
```
cretz, line 342: Can you make sure down below in `pollWorkflowUpdateHelper` that you remove the logic that retries on gRPC deadline exceeded error? That should no longer occur, we should just be bubbling all errors out

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
cretz: Arguably it doesn't need to be set at all

temporal-sdk/src/main/java/io/temporal/client/LazyUpdateHandleImpl.java
Sushisource: Since before the handle is returned from start when user says complete, there might be a result from polling already and if there is we want to use that, otherwise try it - but then we need to wipe that result in case getResult gets called again

temporal-sdk/src/main/java/io/temporal/client/UpdateWaitPolicy.java
Sushisource: K, I'm down to change the name

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
Sushisource: Yes that happens here: https://github.com/temporalio/sdk-java/pull/2066/files#diff-64e98c2533a94b01fbda951826a90567fb17dd80b70af7f6f44c5c6dd1759c69R364

temporal-sdk/src/main/java/io/temporal/client/UpdateWaitPolicy.java
Sushisource: Docs wise, weirdly, even the proto APIs mention nothing about what the stages mean beyond as input to requests. That would be good to fix.

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
Sushisource: Wait, isn't the Python one doing that when it passes `retry=True` to the service client? Or, if that doesn't retry timeouts, then where is that happening? Because https://github.com/temporalio/sdk-python/blob/1a2acd59634a3b1d694937b8a8433c0014247370/temporalio/client.py#L4303 says it will, but there's no explicit handling of timeouts here: https://github.com/temporalio/sdk-python/blob/1a2acd59634a3b1d694937b8a8433c0014247370/temporalio/client.py#L4359

Sushisource: It's easy to change Java to not do this and just default to max timeout for getResult calls, but, not sure that's the right thing to do.

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
Sushisource: I unset this in the most recent commit - but, I'm not sure having a super long timeout by default is what we want to do? OTOH I don't have a firm reason why not I suppose.

temporal-sdk/src/main/java/io/temporal/client/LazyUpdateHandleImpl.java
Quinn-With-Two-Ns: hm this seems racy, if you have two concurrent calls isn't is possible for `pollCall=null ` if two threads interleave in the right way?

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
cretz: I have no strong opinion so long as it's always longer than server's by enough to let server return an empty response on its timeout

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
cretz: :+1: Sorry I missed that

temporal-sdk/src/main/java/io/temporal/client/LazyUpdateHandleImpl.java
Sushisource: Ah yes it is, I didn't think about this being called concurrently, too used to Rust.

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
Sushisource: Yeah, that's done now. All it's doing is just interpreting the failure code into the right exception type which makes sense to me.

temporal-sdk/src/main/java/io/temporal/client/LazyUpdateHandleImpl.java
Sushisource: I've just sync'd the whole thing since arguably it doesn't make sense to make multiple concurrent calls to server in that case anyway

Sushisource: Simplified this

temporal-sdk/src/main/java/io/temporal/client/LazyUpdateHandleImpl.java
```diff
@@ -73,18 +73,18 @@ public String getId() {
...
-    previousPollCall = null;
+
+    WorkflowClientCallsInterceptor.PollWorkflowUpdateOutput<T> pollCall =
+        pollUntilComplete(timeout, unit);
 
     return pollCall
         .getResult()
         .exceptionally(
             failure -> {
+              // If the poll didn't find the completion successfully, reset the previous poll call
```
Quinn-With-Two-Ns, line 84: Just because the poll call failed does not mean it shouldn't be cached right? A update rejection would also complete the poll future exceptional If I recall correctly. I would probably drop caching since we don't cache workflow result either and if we get user feedback address all these functions with a consistent strategy.

temporal-sdk/src/main/java/io/temporal/client/LazyUpdateHandleImpl.java
```diff
@@ -109,4 +113,18 @@ public CompletableFuture<T> getResultAsync(long timeout, TimeUnit unit) {
   public CompletableFuture<T> getResultAsync() {
     return this.getResultAsync(Long.MAX_VALUE, TimeUnit.MILLISECONDS);
   }
+
+  WorkflowClientCallsInterceptor.PollWorkflowUpdateOutput<T> pollUntilComplete(
+      long timeout, TimeUnit unit) {
+    synchronized (this) {
```
Quinn-With-Two-Ns, line 119: Using synchronized like this can be really problematic with virtual threads because the virtual thread will be pinned while executing `pollWorkflowUpdate` https://mikemybytes.com/2024/02/28/curiosities-of-java-virtual-threads-pinning-with-synchronized/

temporal-sdk/src/main/java/io/temporal/client/LazyUpdateHandleImpl.java
Sushisource: Sure, I could just delete it every time still. I need something like it for the don't-return-handle-until-completed case and I figured why not cache it for the success case since that's definitely not going to change.

temporal-sdk/src/main/java/io/temporal/client/LazyUpdateHandleImpl.java
Sushisource: Ah, interesting didn't realize that.

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
Quinn-With-Two-Ns: We should treat start update as a long poll, hence the long timeout

temporal-sdk/src/main/java/io/temporal/client/LazyUpdateHandleImpl.java
Quinn-With-Two-Ns: I think for the don't-return-handle-until-completed case you can take the result and put it in the `CompletedUpdateHandleImpl`

temporal-sdk/src/main/java/io/temporal/client/LazyUpdateHandleImpl.java
Quinn-With-Two-Ns: Yeah it is a very unfortunate issue with the current virtual thread limitation. My current stance  is avoid IO in synchronized  blocks and remove any I see

temporal-sdk/src/main/java/io/temporal/client/LazyUpdateHandleImpl.java
Sushisource: Sure, that works for me

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
```diff
@@ -434,31 +441,23 @@ private void pollWorkflowUpdateHelper(
...
-        .pollUpdateAsync(request, pollTimeoutDeadline)
+        .pollUpdateAsync(request, deadline)
         .whenComplete(
             (r, e) -> {
               if ((e instanceof StatusRuntimeException
                       && ((StatusRuntimeException) e).getStatus().getCode()
                           == Status.Code.DEADLINE_EXCEEDED)
-                  || pollTimeoutDeadline.isExpired()
+                  || deadline.isExpired()
                   || (e == null && !r.hasOutcome())) {
```
cretz, line 452: I think this is supposed to recuse in this situation

temporal-sdk/src/main/java/io/temporal/client/LazyUpdateHandleImpl.java
Sushisource: K, I've done a much more targeted version of this.

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
Sushisource: Right, OK, I keep getting confused about the situations where server could return no outcome, but, it's like long polling on tasks

Quinn-With-Two-Ns:
Don't we also need to update the test server to return the updates  current lifecycle state? or does the test server never actually  return an empty response?

temporal-sdk/src/main/java/io/temporal/client/LazyUpdateHandleImpl.java
Quinn-With-Two-Ns: We still do want to avoid caching most exceptions though, just the ones from the update outcome should be cached

Sushisource:
> Don't we also need to update the test server to return the updates current lifecycle state? or does the test server never actually return an empty response?

I have changed it to do so

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
```diff
@@ -434,31 +441,26 @@ private void pollWorkflowUpdateHelper(
...
-                          "WorkflowId="
-                              + request.getUpdateRef().getWorkflowExecution().getWorkflowId()
-                              + ", runId="
-                              + request.getUpdateRef().getWorkflowExecution().getRunId()
-                              + ", updateId="
-                              + request.getUpdateRef().getUpdateId()));
-                }
+                  || deadline.isExpired()) {
+                resultCF.completeExceptionally(
+                    new TimeoutException(
```
cretz, line 457: Note this may be changing shortly with #2069

cretz (approved):

Quinn-With-Two-Ns (approved):

temporal-sdk/src/main/java/io/temporal/client/LazyUpdateHandleImpl.java
```diff
@@ -72,12 +73,23 @@ public String getId() {
...
 
-    return output
+    WorkflowClientCallsInterceptor.PollWorkflowUpdateOutput<T> pollCall = null;
+    boolean setFromWaitCompleted = false;
+
+    // If waitCompleted was called, use the result from that call.
+    synchronized (this) {
+      if (waitCompletedPollCall != null) {
+        pollCall = waitCompletedPollCall;
+        waitCompletedPollCall = null;
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: yeah this looks like a bug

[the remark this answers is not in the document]
Quinn-With-Two-Ns: The Java `StartUpdate` code path is different from the other SDKs in a few subtle ways as well i'll try to align it with other SDKs as well

PR #2067, opened 2024-05-17
Send unset catchupWindow in schedules if not specified

### LINKED ISSUE #2065: It's impossible to send unset duration
## Expected Behavior

When `java.util.Duration` is `null` in an options class like `SchedulePolicy` it is treated as an unset value in a protobuf message _in some cases_

## Actual Behavior

`null` values are sent as zero duration. There was a TODO in the appropriate method about refactoring this behaviour, but was removed in #1950

https://github.com/temporalio/sdk-java/blob/ec18bc86d52f23add2dc97682d94c629e6282bf1/temporal-sdk/src/main/java/io/temporal/internal/common/ProtobufTimeUtils.java#L43-L48

## A problem 

I'm creating schedules using the following code:
```java
...
var policy = SchedulePolicy.newBuilder()
                .setOverlap(ScheduleOverlapPolicy.SCHEDULE_OVERLAP_POLICY_BUFFER_ONE)
                .build();
var schedule Schedule.newBuilder()
                .setAction(actionStartWorkflow)
                .setSpec(spec)
                .setPolicy(policy)
                .build();
```
Which causes sending `0` in the `catchupWindow` property of `SchedulePolicy`. Hence, server side treats it as `0` and uses `MinCatchupWindow` (=10 sec) instead of expected `DefaultCatchupWindow` (=1yr)

https://github.com/temporalio/temporal/blob/aca0d911a908b2584f4cffcc73425c5119dbe722/service/worker/scheduler/workflow.go#L750-L759

  1. Create a schedule using the code above
  2. Check its catchupWindow

Duzhinsky:
## What was changed

If the catchupWindow property is not set in SchedulePolicy, it is sent as a zero duration in the appropriate protobuf message. However, the server expects an unset value here. The previous behavior led to setting the actual catchupWindow on the server to the minimal value (10 seconds) instead of the default (1 year).

## Why?
Schedules are created with wrong `catchupWindow`. That leads to missed runs

## Checklist

1. Closes  #2065

2. How was this tested:
Tested in existing `ScheduleTest`

3. Any docs updates needed?
Nope

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/client/ScheduleProtoUtil.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/client/ScheduleProtoUtil.java
@@ -161,11 +161,13 @@ public class ScheduleProtoUtil {
   }
 
   public SchedulePolicies policyToProto(SchedulePolicy policy) {
-    return SchedulePolicies.newBuilder()
-        .setCatchupWindow(ProtobufTimeUtils.toProtoDuration(policy.getCatchupWindow()))
-        .setPauseOnFailure(policy.isPauseOnFailure())
-        .setOverlapPolicy(policy.getOverlap())
-        .build();
+    SchedulePolicies.Builder builder = SchedulePolicies.newBuilder();
+    if (policy.getCatchupWindow() != null) {
+      builder.setCatchupWindow(ProtobufTimeUtils.toProtoDuration(policy.getCatchupWindow()));
+    }
+    builder.setPauseOnFailure(policy.isPauseOnFailure());
+    builder.setOverlapPolicy(policy.getOverlap());
+    return builder.build();
   }
 
   public List<Range> scheduleRangeToProto(List<ScheduleRange> scheduleRanges) {


Quinn-With-Two-Ns:
Looks like the feature tests failed because the `tags` are missing on your branch that we use to find the version of the SDK. You can specify `--tags ` when you fetch

Duzhinsky:
> Looks like the feature tests failed because the `tags` are missing on your branch that we use to find the version of the SDK. You can specify `--tags ` when you fetch

Done

Quinn-With-Two-Ns:
You''ll need to push your update branch with tags as well

Duzhinsky:
I've done `git fetch --tags` and then `git push --tags` on my branch.  Now I'm a little confused about following actions

Quinn-With-Two-Ns:
Ah my apologizes, I didn't realize you had pushed since github didn't display any indication

Quinn-With-Two-Ns (approved):

PR #2072, opened 2024-05-20
Try builder first, then fallback to deprecated constructor

### LINKED ISSUE #2071: KotlinObjectMapperFactory is not forwards compatible
## Expected Behavior
KotlinObjectMapperFactory works

## Actual Behavior
KotlinObjectMapperFactory crashes with MethodNotFoundError

  1. Use temporal kotlin
  1. Upgrade jackson kotlin library
  1. Crash

## Specifications

  - Version:
  - Platform:

Instead of using a deprecated method, it would have been best to try using the builder first, then fall back to the deprecated KotlinModule constructor. Alternatively, we should just use the builder as we can't even pick which KotlinModule constructor we compile against.

Sineaggi:
## What was changed
Tries the new kotlin jackson module builder first, then falls back to the deprecated constructor.

## Why?
Newest jackson (2.17.0) broke abi.

## Checklist

1. Closes #2071

2. How was this tested:
Tested the code manually against jackson 2.11, 2.14 and 2.17.

3. Any docs updates needed?


Sineaggi:
The upstream issue was fixed here https://github.com/FasterXML/jackson-module-kotlin/pull/799 but using the builder would have avoided this issue.

temporal-kotlin/src/main/kotlin/io/temporal/common/converter/KotlinObjectMapperFactory.kt
```diff
@@ -29,9 +29,14 @@ class KotlinObjectMapperFactory {
     fun new(): ObjectMapper {
       val mapper = JacksonJsonPayloadConverter.newDefaultObjectMapper()
 
-      // use deprecated constructor instead of builder to maintain compatibility with old jackson versions
-      @Suppress("deprecation")
-      val km = KotlinModule()
+      val km = try {
+        KotlinModule.Builder()
+          .build()
```
Quinn-With-Two-Ns, line 34: Have we confirmed this works with jackson `v2.9.0`?

Sineaggi: The existing code doesn't work with jackson `2.9.0`

Sineaggi:
There is another way to simplify this further, and that is to call the `registerKotlinModule` function instead of trying to register a KotlinModule ourselves.

temporal-kotlin/src/main/kotlin/io/temporal/common/converter/KotlinObjectMapperFactory.kt
Quinn-With-Two-Ns: Not sure I understand, the Java SDK currently does build and all test pass with jackson `2.9.0`

Sineaggi: It currently builds with jackson 1.14.2 https://github.com/temporalio/sdk-java/blob/5ccb85927acf01db619eea4eefae0fb41c5ad471/build.gradle#L34

Sineaggi: It fails with jackson 2.9.0
```
'void com.fasterxml.jackson.module.kotlin.KotlinModule.<init>(int, boolean, boolean, boolean, com.fasterxml.jackson.module.kotlin.SingletonSupport, boolean, int, kotlin.jvm.internal.DefaultConstructorMarker)'
java.lang.NoSuchMethodError: 'void com.fasterxml.jackson.module.kotlin.KotlinModule.<init>(int, boolean, boolean, boolean, com.fasterxml.jackson.module.kotlin.SingletonSupport, boolean, int, kotlin.jvm.internal.DefaultConstructorMarker)'
```

Quinn-With-Two-Ns: Note the comment indicating our support for `2.9.0`

Sineaggi: Unless I'm doing something wrong, the code as-is won't work with jackson 2.9.0 https://github.com/temporalio/sdk-java/compare/master...Sineaggi:sdk-java:jackson-version-test
If you run `./gradlew :temporal-kotlin:test`

Quinn-With-Two-Ns: If we need to drop support to maintain forward compatibility I would 100% be in favor, but if we can support older versions without harming forward compatibility that would be the best for all SDK users.

Sineaggi: I've gone the `registerKotlinModule` route, which should allow for maximum backwards and forwards compatibility.

Quinn-With-Two-Ns:
Ah CI is failing because your branch does not have tags , which our CI unfortunately makes use of to detect the version of the SDK, you can use `git fetch --tags` and then you can push you branch with tags. Sorry for the inconvenience.

Sineaggi:
> Ah CI is failing because your branch does not have tags

Right, that happened during local testing as well. I've pushed tags now.

temporal-kotlin/build.gradle
```diff
@@ -42,3 +43,27 @@ task registerNamespace(type: JavaExec) {
 
 test.dependsOn 'registerNamespace'
 
+def jacksonVersions = ['2.9.0', "$jacksonVersion".toString()]
+
+testing {
+    suites {
+        jacksonVersions.forEach { jacksonVersion ->
```
Quinn-With-Two-Ns, line 50: These tests are not needed since we already have a setup to test multiple versions of dependencies

Sineaggi: How are these tests compiled/run? As it stands, it is impossible to call `KotlinObjectMapperFactory.new()` in a project that forcefully downgrades to jackson 2.9.0. This test shows that that is now possible.

Quinn-With-Two-Ns: Hm yes looking at more carefully your test is doing something slightly different, we test building and running with different version of our dependencies, but here your testing building with one version, but running with another. Finding some breaking ABI change between versions most likely.

Quinn-With-Two-Ns (approved):

temporal-kotlin/build.gradle
Quinn-With-Two-Ns: This test, or the inclusion of `org.gradle.api.plugins.jvm.JvmTestSuite` appears to be causing some issues with our [feature test](https://github.com/temporalio/features) I haven't had a chance to deep dive as to why. I am fine merging this without this test for now and I'll open an issue to create more thorough testing strategy for breaking catching breaking ABI changes. We should probably do tests like this for more then just `jackson` versions

akdoan1:
@Quinn-With-Two-Ns can you rerun the checks? It seems as though only maintainers can do so. I think the `features-test / test` job failed and the logs were removed due to limited retention.

cretz:
(since @Quinn-With-Two-Ns is out, after a quick peek at the PR, I went ahead and triggered CI again)

akdoan1:
@cretz , @Quinn-With-Two-Ns - is there anything else needed for us to merge this change?

cretz:
I merged master into your branch, but actually I would like @Quinn-With-Two-Ns to re-review just to be sure. Everything is good from your end, but may be a bit from ours since he is currently unavailable for a bit. Thanks for your patience.

PR #2074, opened 2024-05-21
Send original update request back in accept/reject response

Quinn-With-Two-Ns:
Send original update request back in accept/reject response

close https://github.com/temporalio/sdk-java/issues/2073

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/statemachines/UpdateProtocolStateMachine.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/statemachines/UpdateProtocolStateMachine.java
@@ -79,6 +79,7 @@ final class UpdateProtocolStateMachine
   private String requestMsgId;
   private long requestSeqID;
   private Meta meta;
+  private Optional<Request> originalRequest = Optional.empty();
   private String messageId;
 
   public static final StateMachineDefinition<State, ExplicitEvent, UpdateProtocolStateMachine>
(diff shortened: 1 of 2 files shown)


temporal-sdk/src/main/java/io/temporal/internal/statemachines/UpdateProtocolStateMachine.java
```diff
@@ -199,8 +201,10 @@ public void accept() {
         Acceptance.newBuilder()
             .setAcceptedRequestMessageId(requestMsgId)
             .setAcceptedRequestSequencingEventId(requestSeqID)
+            .setAcceptedRequest(originalRequest.get())
             .build();
-
+    // Clear the original request to allow GC to reclaim the memory.
+    originalRequest = Optional.empty();
```
cretz, line 207: Should do this on reject too correct? Or is there no reason because reject removes this whole state machine?

cretz (approved):

Sushisource (approved):

temporal-sdk/src/main/java/io/temporal/internal/statemachines/UpdateProtocolStateMachine.java
Quinn-With-Two-Ns: reject is a terminal state so the whole state machine will be cleaned up

PR #2079, opened 2024-05-28
Ensure thread safety of AllOfPromise

pivovarit:
## What was changed
Replaced the internal `AllOfPromise` `int notReadyCount` counter with a thread-safe `final AtomicInteger`

## Why?
`AllOfPromise` registers multiple `Promise` instances that were created outside of its scope and uses the `handle()` method to schedule counter decrements. When the counter reaches zero, it means that all futures completed.

However, `handle()` method can be executed by different threads, and the internal counter in the current form:
- doesn't guarantee that new value will be immediately visible for other threads due to the lack of `volatile`
- doesn't guarantee atomicity of the increment/decrement operations, which can result in lost updates

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/AllOfPromise.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/AllOfPromise.java
@@ -26,17 +26,18 @@ import io.temporal.workflow.Promise;
 import io.temporal.workflow.Workflow;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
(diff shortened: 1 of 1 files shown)


Quinn-With-Two-Ns:
@pivovarit all threads in a workflow are managed and coordinated by the SDK, no external threads not managed by the SDK are allowed to be created inside a workflow, so no two threads are ever running concurrently so this should not be necessary .

pivovarit:
Does it mean that those consequent `handle()` callbacks are effectively executed serially by the same thread?

Quinn-With-Two-Ns:
They  are  executed serially, potentially by different threads, all with all threads using a common lock for serialization

PR #2080, opened 2024-05-29
Add identity to WorkerOptions

### LINKED ISSUE #2078: Add the ability to specify worker identity for each worker
It is assumed that in a single application, there are multiple workers that operate for different queues. Due to internal integrations in our infrastructure, I must specify a queue in the identity property. The worker is created using a `WorkerFactory`, which is linked to a `WorkflowClient`. Currently, there is only the ability to specify identity for the client.

I propose the introduction of an option to assign an identity within `WorkerOptions`, and for this identity to be treated as a priority when `SingleWorkerOptions` are being configured.

https://github.com/temporalio/sdk-java/blob/5e5cf0bf2db695cba46d1855959e02b4cdfa3cff/temporal-sdk/src/main/java/io/temporal/worker/Worker.java#L586-L608

A possible workaround could involve creating a distinct `WorkflowClient` for every queue, which in turn would necessitate a separate `WorkerFactory` for each. However, employing a single client is fully sufficient for the application's needs.

Duzhinsky:
Provides the ability to override worker identity specified in a WorkflowClient

## Checklist

1. Closes #2078 

2. How was this tested: manual tests

3. Any docs updates needed? 
It would be great to have a docs section about identities in the sdk docs, but there is no such yet

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/worker/WorkerOptions.java
+++ b/temporal-sdk/src/main/java/io/temporal/worker/WorkerOptions.java
@@ -86,6 +86,7 @@ public final class WorkerOptions {
     private SlotSupplier<WorkflowSlotInfo> workflowSlotSupplier;
     private SlotSupplier<ActivitySlotInfo> activitySlotSupplier;
     private SlotSupplier<LocalActivitySlotInfo> localActivitySlotSupplier;
+    private String identity;
 
     private Builder() {}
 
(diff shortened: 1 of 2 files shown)


temporal-sdk/src/main/java/io/temporal/worker/WorkerOptions.java
```diff
@@ -683,6 +695,10 @@ public SlotSupplier<LocalActivitySlotInfo> getLocalActivitySlotSupplier() {
     return localActivitySlotSupplier;
   }
 
+  public String getIdentity() {
```
Quinn-With-Two-Ns, line 698: Can we please annotate this may be null?

Quinn-With-Two-Ns (approved):

PR #2082, opened 2024-05-29
Fix schedule workflow action retry policy

Quinn-With-Two-Ns:
Fix schedule workflow action retry policy. Originally it was not being properly propagated from the action to the workflow options.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/client/ScheduleProtoUtil.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/client/ScheduleProtoUtil.java
@@ -130,7 +130,7 @@ public class ScheduleProtoUtil {
         workflowRequest.setInput(inputArgs.get());
       }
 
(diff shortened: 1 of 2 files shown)


temporal-sdk/src/main/java/io/temporal/internal/client/ScheduleProtoUtil.java
```diff
@@ -128,6 +130,11 @@ public ScheduleAction actionToProto(io.temporal.client.schedules.ScheduleAction
         workflowRequest.setInput(inputArgs.get());
       }
 
+      RetryOptions retryOptions = startWorkflowAction.getOptions().getRetryOptions();
```
cretz, line 133:
```suggestion
      RetryOptions retryOptions = wfOptions.getRetryOptions();
```

(pedantic, can ignore)

cretz (approved):

PR #2088, opened 2024-06-03
Require WaitForStage in StartUpdate

Quinn-With-Two-Ns:
Require `WaitForStage` in `StartUpdate`

closes https://github.com/temporalio/sdk-java/issues/2045


cretz (approved):
I assume the work that repeatedly calls start is/was done elsewhere

Quinn-With-Two-Ns:
Yes it was done here https://github.com/temporalio/sdk-java/pull/2066

Issue #2094, opened 2024-06-04
Java SDK `StartUpdate` waits outside of intercepted call


Quinn-With-Two-Ns:
Currently when the Java SDK executes a start update call and the update request does not reach the required waitstage the SDK will block the call waiting for the update request to reach the required stage. Currently we do this after the interceptor, but in all other SDKs we do this inside the intercepted call. This makes it impossible to run logic in the interceptor after the start update call has completed. We need to align the Java SDK with other SDKs here.

PR #2098, opened 2024-06-06
Fix caching in LazyUpdateHandleImpl

Quinn-With-Two-Ns:
Fix caching in `LazyUpdateHandleImpl`. `setFromWaitCompleted` was not being used.

Note: Very likely I will refactor all these handlers as part of https://github.com/temporalio/sdk-java/issues/2094


cretz (approved):

Sushisource (approved):
Thank ye

PR #2099, opened 2024-06-06
Allow SDK to handle speculative workflow task with command events

Quinn-With-Two-Ns:
Allow SDK to handle speculative workflow task with command events

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
@@ -217,21 +217,21 @@ public final class WorkflowStateMachines {
    *     and triggered an execution. Used in {@link WorkflowTaskStateMachine} only to understand
    *     that this workflow task will not have a matching closing event and needs to be executed.
    */
-  public void setWorklfowStartedEventId(long workflowTaskStartedEventId) {
+  public void setWorkflowStartedEventId(long workflowTaskStartedEventId) {
     this.workflowTaskStartedEventId = workflowTaskStartedEventId;
   }
 
-  public void setLastWFTStartedEventId(long eventId) {
+  public void resetStartedEvenId(long eventId) {
+    // When we reset the event ID on a speculative WFT we need to move this counter back
+    // to account for the speculative WFT.
+    long resetLastHandledEventId = this.lastHandledEventId - 2;
     // We have to drop any state machines (which should only be one workflow task machine)
     // created when handling the speculative workflow task
-    for (long i = this.lastHandledEventId; i > eventId; i--) {
+    for (long i = this.lastHandledEventId; i > resetLastHandledEventId; i--) {
       stateMachines.remove(i);
     }
     this.lastWFTStartedEventId = eventId;
-    // When we reset the event ID on a speculative WFT we need to move this counter back
-    // to the last WFT completed to allow new tasks to be processed. Assume the WFT complete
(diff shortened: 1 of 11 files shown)


temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
```diff
@@ -217,21 +217,21 @@ public WorkflowStateMachines(
...
-  public void setWorklfowStartedEventId(long workflowTaskStartedEventId) {
+  public void setWorkflowStartedEventId(long workflowTaskStartedEventId) {
     this.workflowTaskStartedEventId = workflowTaskStartedEventId;
   }
 
-  public void setLastWFTStartedEventId(long eventId) {
+  public void resetStartedEvenId(long eventId) {
+    // When we reset the event ID on a speculative WFT we need to move this counter back
+    // to account for the speculative WFT.
+    long resetLastHandledEventId = this.lastHandledEventId - 2;
```
Sushisource, line 227: Best to say why it's `- 2`

temporal-sdk/src/test/java/io/temporal/internal/statemachines/UpdateProtocolStateMachineTest.java
```diff
@@ -451,6 +451,130 @@ protected void update(UpdateMessage message, AsyncWorkflowBuilder<Void> builder)
...
+                  })));
+      List<Command> commands = h.handleWorkflowTaskTakeCommands(stateMachines);
+      assertEquals(0, commands.size());
+      List<Message> messages = stateMachines.takeMessages();
+      assertEquals(1, messages.size());
+      Rejection rejection = messages.get(0).getBody().unpack(Rejection.class);
+      assertNotNull(rejection);
+      assertEquals(request, rejection.getRejectedRequest());
+      // Simulate the server request to reset the workflow event ID
+      stateMachines.resetStartedEvenId(3);
```
Sushisource, line 547: I think it'd be better to test this by actually putting it into the workflow task - that's how the test in Core works. Otherwise there's nothing verifying that the value is extracted from the task.

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
Quinn-With-Two-Ns: That is what the comment  above is meant to explain

Sushisource: Right, it says move back but not why by two. Equivalent comment in core: https://github.com/temporalio/sdk-core/blob/90db5b485f8805902834f48c5ed87f6de5cdaac9/core/src/worker/workflow/machines/workflow_machines.rs#L477

Quinn-With-Two-Ns: I'll probably just plagiarize this comment :)

temporal-sdk/src/test/java/io/temporal/internal/statemachines/UpdateProtocolStateMachineTest.java
Quinn-With-Two-Ns: We already have other tests integration tests with the server verifying workflow reset, but I added a unit test verifying the reset ID is passed from the task to the handler.

Sushisource (approved):

PR #2104, opened 2024-06-10
Add integration test for reset with update

Quinn-With-Two-Ns:
Add integration test for reset with update now that the server supports it.


cretz (approved):

cretz:
I wonder if we need a general feature-repo test for this too

Quinn-With-Two-Ns:
> I wonder if we need a general feature-repo test for this too
Yeah I agree, i'll make an issue

Issue #2107, opened 2024-06-12
Concurrent Modification Exception when executing workflow with local activities


Quinn-With-Two-Ns:
## Issue

When executing workflow tasks with local activities `LocalActivityMeteringHelper` may attempt to make illegal modifications to its underlying data structure when calling `newWFTStarting` causing a `ConcurrentModificationException`

## Example: 
```
java.util.ConcurrentModificationException: null
	at java.base/java.util.HashMap$HashIterator.nextNode(HashMap.java:1605)
	at java.base/java.util.HashMap$KeyIterator.next(HashMap.java:1628)
	at io.temporal.internal.replay.ReplayWorkflowRunTaskHandler$LocalActivityMeteringHelper.newWFTStarting(ReplayWorkflowRunTaskHandler.java:431)
	at io.temporal.internal.replay.ReplayWorkflowRunTaskHandler$LocalActivityMeteringHelper.access$200(ReplayWorkflowRunTaskHandler.java:425)
	at io.temporal.internal.replay.ReplayWorkflowRunTaskHandler.handleWorkflowTask(ReplayWorkflowRunTaskHandler.java:143)
	at io.temporal.internal.replay.ReplayWorkflowTaskHandler.handleWorkflowTaskWithQuery(ReplayWorkflowTaskHandler.java:133)
	at io.temporal.internal.replay.ReplayWorkflowTaskHandler.handleWorkflowTask(ReplayWorkflowTaskHandler.java:98)
	at io.temporal.internal.worker.WorkflowWorker$TaskHandlerImpl.handleTask(WorkflowWorker.java:413)
	at io.temporal.internal.worker.WorkflowWorker$TaskHandlerImpl.handle(WorkflowWorker.java:320)
	at io.temporal.internal.worker.WorkflowWorker$TaskHandlerImpl.handle(WorkflowWorker.java:261)
	at io.temporal.internal.worker.PollTaskExecutor.lambda$process$0(PollTaskExecutor.java:105)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
	at java.base/java.lang.Thread.run(Thread.java:1583)
```

https://community.temporal.io/t/concurrent-modification-exception-when-executing-workflow/12479/1

## Specifications

  - Platform: 1.23.1

PR #2108, opened 2024-06-12
Fix ConcurrentModificationException in LocalActivityMeteringHelper

Quinn-With-Two-Ns:
Fix `ConcurrentModificationException` in `LocalActivityMeteringHelper`. `newWFTStarting` was calling `remove` on `firstWftActivities ` in a for loop iterating over `firstWftActivities` which is not allowed.

closes https://github.com/temporalio/sdk-java/issues/2107


Sushisource (approved):
Cool, that works fine.

The thing I don't quite get is how `newWFTStarting` was getting called concurrently?

PR #2110, opened 2024-06-13
Resource based tuner

Sushisource:
## What was changed
Add a resource-based worker tuner a-la the one created in Core

## Why?
Helpful for users who want an easy and automated approach to worker tuning

## Checklist

1. Closes 

2. How was this tested:
Added some tests, load testing in Omes

3. Any docs updates needed?
Yes, eventually once we leave prerelease

### DIFF
--- a/temporal-sdk/build.gradle
+++ b/temporal-sdk/build.gradle
@@ -44,5 +44,6 @@ test {
 task testResourceIndependent(type: Test) {
     useJUnit {
         includeCategories 'io.temporal.worker.IndependentResourceBasedTests'
+        maxParallelForks = 1
     }
 }
--- a/temporal-sdk/build.gradle
+++ b/temporal-sdk/build.gradle
@@ -34,3 +34,15 @@ task registerNamespace(type: JavaExec) {
 }
 
 test.dependsOn 'registerNamespace'
+
+test {
+    useJUnit {
+        excludeCategories 'io.temporal.worker.IndependentResourceBasedTests'
+    }
+}
+
+task testResourceIndependent(type: Test) {
+    useJUnit {
+        includeCategories 'io.temporal.worker.IndependentResourceBasedTests'
+    }
+}
--- a/temporal-sdk/src/main/java/io/temporal/worker/tuning/JVMSystemResourceInfo.java
+++ b/temporal-sdk/src/main/java/io/temporal/worker/tuning/JVMSystemResourceInfo.java
@@ -23,6 +23,8 @@ package io.temporal.worker.tuning;
 import com.sun.management.OperatingSystemMXBean;
 import io.temporal.common.Experimental;
 import java.lang.management.ManagementFactory;
+import java.time.Instant;
+import java.util.concurrent.locks.ReentrantLock;
 
 /** {@link SystemResourceInfo} implementation that uses JVM-specific APIs to get resource usage. */
 @Experimental
@@ -33,27 +35,55 @@ public class JVMSystemResourceInfo implements SystemResourceInfo {
   // have been backported to.
   OperatingSystemMXBean osBean = ManagementFactory.getPlatformMXBean(OperatingSystemMXBean.class);
 
+  private final ReentrantLock refreshLock = new ReentrantLock();
+  volatile Instant lastRefresh = Instant.EPOCH;
+  volatile double lastMemUsage = Double.NaN;
+  volatile double lastCpuUsage = Double.NaN;
+
   @Override
-  @SuppressWarnings("deprecation") // deprecated APIs needed since replacements are for Java 14+
   public double getCPUUsagePercent() {
-    return osBean.getSystemCpuLoad();
+    refresh();
+    return lastCpuUsage;
   }
 
   @Override
-  @SuppressWarnings("deprecation") // deprecated APIs needed since replacements are for Java 14+
   public double getMemoryUsagePercent() {
-    Runtime runtime = Runtime.getRuntime();
-    long jvmUsedMemory = runtime.totalMemory() - runtime.freeMemory();
-    long jvmMaxMemory = runtime.maxMemory();
-    long systemTotalMemory = osBean.getTotalPhysicalMemorySize();
-    long systemUsedMemory = systemTotalMemory - osBean.getFreePhysicalMemorySize();
+    refresh();
+    return lastMemUsage;
+  }
 
-    double jvmMemoryUsagePercent = ((double) jvmUsedMemory / jvmMaxMemory);
-    double systemMemoryUsagePercent = ((double) systemUsedMemory / systemTotalMemory);
+  @SuppressWarnings("deprecation") // deprecated APIs needed since replacements are for Java 14+
+  private void refresh() {
+    if (Instant.now().isBefore(lastRefresh.plusMillis(100))) {
+      return;
+    }
+    refreshLock.lock();
 
-    // We want to return either the JVM memory usage or the system memory usage, whichever is
-    // higher, since we want to avoid OOMing either the JVM or the system.
-    // TODO: Slightly worried this could thrash back and forth
-    return Math.max(jvmMemoryUsagePercent, systemMemoryUsagePercent);
+    try {
+      lastCpuUsage = osBean.getSystemCpuLoad();
+
+      Runtime runtime = Runtime.getRuntime();
+      long jvmUsedMemory = runtime.totalMemory() - runtime.freeMemory();
+      long jvmMaxMemory = runtime.maxMemory();
+      long systemTotalMemory = osBean.getTotalPhysicalMemorySize();
+      long systemUsedMemory = systemTotalMemory - osBean.getFreePhysicalMemorySize();
+
+      double jvmMemoryUsagePercent = ((double) jvmUsedMemory / jvmMaxMemory);
+      double systemMemoryUsagePercent = ((double) systemUsedMemory / systemTotalMemory);
+
+      System.out.println(
+          "System used memory: "
+              + systemUsedMemory
+              + " / System total memory: "
+              + systemTotalMemory);
+
+      // We want to return either the JVM memory usage or the system memory usage, whichever is
+      // higher, since we want to avoid OOMing either the JVM or the system.
+      // TODO: Slightly worried this could thrash back and forth
+      lastMemUsage = Math.max(jvmMemoryUsagePercent, systemMemoryUsagePercent);
+      lastRefresh = Instant.now();
+    } finally {
+      refreshLock.unlock();
+    }
   }
 }
--- a/temporal-sdk/src/main/java/io/temporal/worker/tuning/JVMSystemResourceInfo.java
+++ b/temporal-sdk/src/main/java/io/temporal/worker/tuning/JVMSystemResourceInfo.java
@@ -1,0 +1,36 @@
+package io.temporal.worker.tuning;
+
+import com.sun.management.OperatingSystemMXBean;
+import java.lang.management.ManagementFactory;
+import java.lang.management.MemoryMXBean;
+
+public class JVMSystemResourceInfo implements SystemResourceInfo {
+  // As of relatively recent Java versions (including backports), this class will properly deal with
+  // containerized environments as well as running on bare metal.
+  // See https://bugs.openjdk.org/browse/JDK-8226575 for more details on which versions the fixes
+  // have been backported to.
+  OperatingSystemMXBean osBean = ManagementFactory.getPlatformMXBean(OperatingSystemMXBean.class);
+  MemoryMXBean memoryBean = ManagementFactory.getMemoryMXBean();
+
+  @Override
+  public double getCpuUsagePercent() {
+    return osBean.getSystemCpuLoad();
+  }
+
+  @Override
+  public double getMemoryUsagePercent() {
+    long jvmUsedMemory =
+        memoryBean.getHeapMemoryUsage().getUsed() + memoryBean.getNonHeapMemoryUsage().getUsed();
+    long jvmMaxMemory = Runtime.getRuntime().maxMemory();
+    long systemTotalMemory = osBean.getTotalPhysicalMemorySize();
+    long systemUsedMemory = systemTotalMemory - osBean.getFreePhysicalMemorySize();
+
+    double jvmMemoryUsagePercent = ((double) jvmUsedMemory / jvmMaxMemory);
+    double systemMemoryUsagePercent = ((double) systemUsedMemory / systemTotalMemory);
+
+    // We want to return either the JVM memory usage or the system memory usage, whichever is
+    // higher, since we want to avoid OOMing either the JVM or the system.
+    // TODO: Slightly worried this could thrash back and forth
+    return Math.max(jvmMemoryUsagePercent, systemMemoryUsagePercent);
+  }
+}
--- a/temporal-sdk/src/main/java/io/temporal/worker/tuning/JVMSystemResourceInfo.java
+++ b/temporal-sdk/src/main/java/io/temporal/worker/tuning/JVMSystemResourceInfo.java
@@ -23,7 +23,6 @@ package io.temporal.worker.tuning;
 import com.sun.management.OperatingSystemMXBean;
 import io.temporal.common.Experimental;
 import java.lang.management.ManagementFactory;
-import java.lang.management.MemoryMXBean;
 
 /** {@link SystemResourceInfo} implementation that uses JVM-specific APIs to get resource usage. */
 @Experimental
--- a/temporal-sdk/src/main/java/io/temporal/worker/tuning/JVMSystemResourceInfo.java
+++ b/temporal-sdk/src/main/java/io/temporal/worker/tuning/JVMSystemResourceInfo.java
@@ -33,7 +33,6 @@ public class JVMSystemResourceInfo implements SystemResourceInfo {
   // See https://bugs.openjdk.org/browse/JDK-8226575 for more details on which versions the fixes
   // have been backported to.
   OperatingSystemMXBean osBean = ManagementFactory.getPlatformMXBean(OperatingSystemMXBean.class);
-  MemoryMXBean memoryBean = ManagementFactory.getMemoryMXBean();
 
   @Override
   @SuppressWarnings("deprecation") // deprecated APIs needed since replacements are for Java 14+
@@ -44,9 +43,9 @@ public class JVMSystemResourceInfo implements SystemResourceInfo {
   @Override
   @SuppressWarnings("deprecation") // deprecated APIs needed since replacements are for Java 14+
   public double getMemoryUsagePercent() {
-    long jvmUsedMemory =
-        memoryBean.getHeapMemoryUsage().getUsed() + memoryBean.getNonHeapMemoryUsage().getUsed();
-    long jvmMaxMemory = Runtime.getRuntime().maxMemory();
+    Runtime runtime = Runtime.getRuntime();
+    long jvmUsedMemory = runtime.totalMemory() - runtime.freeMemory();
+    long jvmMaxMemory = runtime.maxMemory();
     long systemTotalMemory = osBean.getTotalPhysicalMemorySize();
     long systemUsedMemory = systemTotalMemory - osBean.getFreePhysicalMemorySize();
 
--- a/temporal-sdk/src/main/java/io/temporal/worker/tuning/JVMSystemResourceInfo.java
+++ b/temporal-sdk/src/main/java/io/temporal/worker/tuning/JVMSystemResourceInfo.java
@@ -60,27 +60,17 @@ public class JVMSystemResourceInfo implements SystemResourceInfo {
     refreshLock.lock();
 
     try {
+      // This can return NaN seemingly when usage is very low
       lastCpuUsage = osBean.getSystemCpuLoad();
+      if (lastCpuUsage < 0 || Double.isNaN(lastCpuUsage)) {
+        lastCpuUsage = 0;
+      }
 
       Runtime runtime = Runtime.getRuntime();
       long jvmUsedMemory = runtime.totalMemory() - runtime.freeMemory();
       long jvmMaxMemory = runtime.maxMemory();
-      long systemTotalMemory = osBean.getTotalPhysicalMemorySize();
-      long systemUsedMemory = systemTotalMemory - osBean.getFreePhysicalMemorySize();
 
-      double jvmMemoryUsagePercent = ((double) jvmUsedMemory / jvmMaxMemory);
-      double systemMemoryUsagePercent = ((double) systemUsedMemory / systemTotalMemory);
-
-      System.out.println(
-          "System used memory: "
-              + systemUsedMemory
-              + " / System total memory: "
-              + systemTotalMemory);
-
-      // We want to return either the JVM memory usage or the system memory usage, whichever is
-      // higher, since we want to avoid OOMing either the JVM or the system.
-      // TODO: Slightly worried this could thrash back and forth
-      lastMemUsage = Math.max(jvmMemoryUsagePercent, systemMemoryUsagePercent);
+      lastMemUsage = ((double) jvmUsedMemory / jvmMaxMemory);
       lastRefresh = Instant.now();
     } finally {
       refreshLock.unlock();
--- a/temporal-sdk/src/main/java/io/temporal/worker/tuning/JVMSystemResourceInfo.java
+++ b/temporal-sdk/src/main/java/io/temporal/worker/tuning/JVMSystemResourceInfo.java
@@ -35,7 +35,7 @@ public class JVMSystemResourceInfo implements SystemResourceInfo {
 
   @Override
   @SuppressWarnings("deprecation") // deprecated APIs needed since replacements are for Java 14+
-  public double getCpuUsagePercent() {
+  public double getCPUUsagePercent() {
     return osBean.getSystemCpuLoad();
   }
 
--- a/temporal-sdk/src/main/java/io/temporal/worker/tuning/JVMSystemResourceInfo.java
+++ b/temporal-sdk/src/main/java/io/temporal/worker/tuning/JVMSystemResourceInfo.java
@@ -24,6 +24,7 @@ import com.sun.management.OperatingSystemMXBean;
 import io.temporal.common.Experimental;
 import java.lang.management.ManagementFactory;
 import java.time.Instant;
+import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 
 /** {@link SystemResourceInfo} implementation that uses JVM-specific APIs to get resource usage. */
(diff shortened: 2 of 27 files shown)


temporal-sdk/src/main/java/io/temporal/worker/tuning/JVMSystemResourceInfo.java
```diff
@@ -0,0 +1,61 @@
...
+  @Override
+  @SuppressWarnings("deprecation") // deprecated APIs needed since replacements are for Java 14+
+  public double getCpuUsagePercent() {
+    return osBean.getSystemCpuLoad();
+  }
+
+  @Override
+  @SuppressWarnings("deprecation") // deprecated APIs needed since replacements are for Java 14+
+  public double getMemoryUsagePercent() {
+    long jvmUsedMemory =
```
Quinn-With-Two-Ns, line 47: Is there a reason you aren't doing `Runtime.getRuntime().totalMemory() - Runtime.getRuntime().freeMemory()`?

Sushisource: Nah, not really. That probably makes sense to do instead.

temporal-sdk/src/main/java/io/temporal/worker/tuning/JVMSystemResourceInfo.java
```diff
@@ -0,0 +1,59 @@
...
+    long jvmUsedMemory = runtime.totalMemory() - runtime.freeMemory();
+    long jvmMaxMemory = runtime.maxMemory();
+    long systemTotalMemory = osBean.getTotalPhysicalMemorySize();
+    long systemUsedMemory = systemTotalMemory - osBean.getFreePhysicalMemorySize();
+
+    double jvmMemoryUsagePercent = ((double) jvmUsedMemory / jvmMaxMemory);
+    double systemMemoryUsagePercent = ((double) systemUsedMemory / systemTotalMemory);
+
+    // We want to return either the JVM memory usage or the system memory usage, whichever is
+    // higher, since we want to avoid OOMing either the JVM or the system.
```
cretz, line 55: I would think with how the JVM works that we can only care about it and not the system and let the JVMs interaction with the system (whether it swaps or what) be unrelated to us. But I think this approach is probably harmless except for the flapping back and forth you mentioned.

temporal-sdk/src/main/java/io/temporal/worker/tuning/TunerHolder.java
```diff
@@ -0,0 +1,83 @@
...
+import io.temporal.common.Experimental;
+import java.util.Objects;
+import javax.annotation.Nonnull;
+
+/**
+ * Can be used to create a {@link WorkerTuner} which uses specific {@link SlotSupplier}s for each
+ * type of slot.
+ */
+@Experimental
+public class TunerHolder implements WorkerTuner {
```
cretz, line 32: I think `SimpleWorkerTuner` or `BasicWorkerTuner` or something is clearer

temporal-sdk/src/main/java/io/temporal/worker/tuning/ResourceController.java
```diff
@@ -0,0 +1,96 @@
...
+import java.util.concurrent.locks.ReentrantLock;
+
+/**
+ * Is used by {@link ResourceBasedSlotSupplier} and {@link ResourceBasedTuner} to make decisions
+ * about whether slots should be handed out based on system resource usage.
+ *
+ * @param <RI> How this controller obtains system resource information
+ */
+@Experimental
+public class ResourceController<RI extends SystemResourceInfo> {
```
cretz, line 34: I think `ResourceBasedController` is a clearer name

temporal-sdk/src/main/java/io/temporal/worker/tuning/SystemResourceInfo.java
```diff
@@ -0,0 +1,39 @@
...
+
+import io.temporal.common.Experimental;
+
+/** Implementors determine how resource usage is measured. */
+@Experimental
+interface SystemResourceInfo {
+  /**
+   * @return System-wide CPU usage as a percentage [0.0, 1.0]
+   */
+  double getCpuUsagePercent();
```
cretz, line 31: You are inconsistent with whether you capitalize `CPU`

temporal-sdk/src/main/java/io/temporal/worker/tuning/ResourceBasedControllerOptions.java
```diff
@@ -0,0 +1,172 @@
...
+    private double cpuIGain = 0;
+    private double cpuDGain = 1;
+    private double cpuOutputThreshold = 0.05;
+
+    public Builder setTargetMemoryUsage(double targetMemoryUsage) {
+      this.targetMemoryUsage = targetMemoryUsage;
+      return this;
+    }
+
+    public Builder setTargetCPUUsage(double targetCPUUsage) {
```
cretz, line 54: You are inconsistent with whether you capitalize CPU

temporal-sdk/src/main/java/io/temporal/worker/tuning/ResourceBasedTuner.java
```diff
@@ -0,0 +1,102 @@
...
+    this.controller = ResourceController.newSystemInfoController(controllerOptions);
+  }
+
+  /**
+   * Set the slot options for workflow tasks. Has no effect after the worker using this tuner
+   * starts.
+   *
+   * <p>Defaults to minimum 2 slots, maximum 500 slots, and no ramp throttle.
+   */
+  public void setWorkflowSlotOptions(ResourceBasedSlotOptions workflowSlotOptions) {
```
cretz, line 53: I think these need to be constructor values. It is confusing to be able to set these after they are in use. Is there precedent in the Java SDK to have builders on builders? If so, a `ResourceBasedTunerOptions` that had `ResourceBasedControllerOptions` and these 3 options may work.

temporal-sdk/src/main/java/io/temporal/worker/tuning/JVMSystemResourceInfo.java
```diff
@@ -0,0 +1,59 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.temporal.worker.tuning;
+
+import com.sun.management.OperatingSystemMXBean;
```
cretz, line 23: This might pose a problem for some module-based users. I think they may have to "add opens" and such to their module setups. We may need a Java SDK sample or at least confirmation that module-based usage of the Java SDK does not have issues here. Worst case scenario, resource-based tuning could move to a separate JAR.

temporal-sdk/src/main/java/io/temporal/worker/tuning/JVMSystemResourceInfo.java
```diff
@@ -0,0 +1,59 @@
...
+public class JVMSystemResourceInfo implements SystemResourceInfo {
+  // As of relatively recent Java versions (including backports), this class will properly deal with
+  // containerized environments as well as running on bare metal.
+  // See https://bugs.openjdk.org/browse/JDK-8226575 for more details on which versions the fixes
+  // have been backported to.
+  OperatingSystemMXBean osBean = ManagementFactory.getPlatformMXBean(OperatingSystemMXBean.class);
+
+  @Override
+  @SuppressWarnings("deprecation") // deprecated APIs needed since replacements are for Java 14+
+  public double getCpuUsagePercent() {
```
cretz, line 38: It looks like this and `getMemoryUsagePercent()` are called on every attempted slot reservation. Can you confirm that these JMX calls and Runtime calls are cheap and/or whether we should memoize for a short bit?

temporal-sdk/src/main/java/io/temporal/worker/tuning/ResourceBasedSlotOptions.java
```diff
@@ -0,0 +1,57 @@
...
+ */
+
+package io.temporal.worker.tuning;
+
+import io.temporal.common.Experimental;
+import java.time.Duration;
+
+/** Options resource-based slot suppliers */
+@Experimental
+public class ResourceBasedSlotOptions {
```
cretz, line 28: There is a bit of a mix between builder patterns and POJOs for options that is a bit inconsistent. Maybe it's ok, but just threw me off a bit.

temporal-sdk/src/main/java/io/temporal/worker/tuning/ResourceController.java
cretz, line 34: I don't think there's any value in a generic here. You only care about the interface, so just reference `SystemResourceInfo` everywhere.

temporal-sdk/src/main/java/io/temporal/worker/tuning/TunerHolder.java
Sushisource: I don't really like either of these - if anything this is a bit of a more advanced option since it allows you to compose different kinds of suppliers. This is the same name as in TS too

temporal-sdk/src/main/java/io/temporal/worker/tuning/ResourceController.java
Sushisource: Yeah, good point

Sushisource: Sure

temporal-sdk/src/main/java/io/temporal/worker/tuning/JVMSystemResourceInfo.java
```diff
@@ -0,0 +1,59 @@
...
+
+package io.temporal.worker.tuning;
+
+import com.sun.management.OperatingSystemMXBean;
+import io.temporal.common.Experimental;
+import java.lang.management.ManagementFactory;
+
+/** {@link SystemResourceInfo} implementation that uses JVM-specific APIs to get resource usage. */
+@Experimental
+public class JVMSystemResourceInfo implements SystemResourceInfo {
```
Quinn-With-Two-Ns, line 29: Why is the class public? Shouldn't it be in internal?

temporal-sdk/src/main/java/io/temporal/worker/tuning/ResourceBasedTuner.java
Sushisource: I think it's fine to have this take a builder. I'll do that.

temporal-sdk/src/main/java/io/temporal/worker/tuning/JVMSystemResourceInfo.java
Sushisource: Good point. I believe this should be accessible since the javadocs include it: https://docs.oracle.com/en/java/javase/17/docs/api/jdk.management/com/sun/management/OperatingSystemMXBean.html and you'd think that means it's part of the open API, but, I can try to confirm that.

temporal-sdk/src/main/java/io/temporal/worker/tuning/JVMSystemResourceInfo.java
Sushisource: So there is throttling at the slot supplier level, but yeah that could still potentially make for some rapid calls from different suppliers. I will investigate how expensive they are.

temporal-sdk/src/main/java/io/temporal/worker/tuning/JVMSystemResourceInfo.java
cretz: We can't OOM before the JVM does, we run inside the JVM, so the JVM is the thing that would OOM if they have configured it to happen. But I guess there are some people that don't configure the JVM and let it OOM.

temporal-sdk/src/main/java/io/temporal/worker/tuning/TunerHolder.java
cretz: For me a certain type of `WorkerTuner` should also be called a tuner. I don't mind the specific pieces used, but I think it should be clear it's a tuner from the name

temporal-sdk/src/main/java/io/temporal/worker/tuning/JVMSystemResourceInfo.java
cretz: I have found using JVM modules in modern Java versions that if you access `com.sun` things these days it poses a big problem and is a pain (you have to "add opens" and all this other mess).

temporal-sdk/src/main/java/io/temporal/worker/tuning/JVMSystemResourceInfo.java
Sushisource: I think it's kind of a nice option that people can, in the event that they have some specific requirements around resourcing, but don't want to have to reimplement the whole resource-based tuner.

temporal-sdk/src/main/java/io/temporal/worker/tuning/JVMSystemResourceInfo.java
```diff
@@ -0,0 +1,89 @@
...
+      long systemUsedMemory = systemTotalMemory - osBean.getFreePhysicalMemorySize();
+
+      double jvmMemoryUsagePercent = ((double) jvmUsedMemory / jvmMaxMemory);
+      double systemMemoryUsagePercent = ((double) systemUsedMemory / systemTotalMemory);
+
+      System.out.println(
+          "System used memory: "
+              + systemUsedMemory
+              + " / System total memory: "
+              + systemTotalMemory);
```
Sushisource, line 78: This (and other) printlns left in for now because these numbers are just straight up wrong, at least on my macOS / JVM combo.

temporal-sdk/src/main/java/io/temporal/worker/tuning/JVMSystemResourceInfo.java
Sushisource: Yes, that's exactly what I mean, and it would still be nice to save them. But, looks like I'll have to do JVM only because there's no good non-native way to reliably get system memory from what I can see.

temporal-sdk/src/main/java/io/temporal/worker/tuning/TunerHolder.java
Sushisource: Then I'd do `CompositeTuner`

temporal-sdk/src/main/java/io/temporal/worker/tuning/JVMSystemResourceInfo.java
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  @Override
+  public double getMemoryUsagePercent() {
+    refresh();
+    return lastMemUsage;
+  }
+
+  @SuppressWarnings("deprecation") // deprecated APIs needed since replacements are for Java 14+
+  private void refresh() {
```
cretz, line 56: Usually I'd just put this behind `synchronized`, I don't think a reentrant lock is what is needed (not sure how same thread would call back in here). And I think you should update everything all at once to avoid cases where two at the same time cause two refreshes (because you are not checking staleness inside the lock only outside). Can simplify to:

```java
private SystemInfo lastSystemInfo;

@Override
public double getCPUUsagePercent() {
    getSystemInfo().cpuUsagePercent;
}

@Override
public double getMemoryUsagePercent() {
    getSystemInfo().memoryUsagePercent;
}

private synchronized SystemInfo getSystemInfo() {
    if lastSystemInfo == null || Instant.now().isAfter(refreshed.plusMillis(100)) {     
        // Do logic
    }
    return lastSystemInfo;
}

private static class SystemInfo {
    private final Instant refreshed;
    private final double cpuUsagePercent;
    private final double memoryUsagePercent;

    private SystemInfo(Instant refreshed, double cpuUsagePercent, double memoryUsagePercent) {
        this.refreshed = refreshed;
        this.cpuUsagePercent = cpuUsagePercent;
        this.memoryUsagePercent= memoryUsagePercent;
    }
}
```

Just reads a bit more thread safe to me

temporal-sdk/src/main/java/io/temporal/worker/tuning/ResourceBasedController.java
```diff
@@ -0,0 +1,139 @@
...
+  private final PIDController cpuController;
+  private final SystemResourceInfo systemInfoSupplier;
+  private Instant lastPidRefresh = Instant.now();
+
+  private Scope metricsScope = new NoopScope();
+  private boolean metricsScopeSet = false;
+  private volatile double last_mem_usage = 0;
+  private volatile double last_cpu_usage = 0;
+  private volatile double last_mem_pid_out = 0;
+  private volatile double last_cpu_pid_out = 0;
```
cretz, line 50: No underscores in Java (and don't need to initialize these values, they are the default.

temporal-sdk/src/main/java/io/temporal/worker/tuning/ResourceBasedController.java
```diff
@@ -0,0 +1,139 @@
...
+  /** Visible for internal usage. Can only be set once. */
+  public void setMetricsScope(Scope metricsScope) {
+    synchronized (this) {
+      if (metricsScopeSet) {
+        return;
+      }
+      this.metricsScope = metricsScope;
+      metricsScopeSet = true;
+      // Launch a task to periodically emit the metrics
+      ForkJoinPool.commonPool().execute(this::emitMetrics);
```
cretz, line 121: Would rather not eat a thread here. Since Tally doesn't expose a pull model and I'm guessing you are unable to set the metrics scope at construction for whatever reason, I would recommend once this setter is called, you an `AtomicReference<Gauges>` where `Gauges` is a static inner class of the 4 gauges, and then `.get()` and if not null set them in the `pidDecision()` call when you get them. This should be cheap and doesn't eat a thread;.

temporal-sdk/src/main/java/io/temporal/worker/tuning/TunerHolder.java
cretz: To me, that would make it seem like it's a composite of many tuners.

Another approach since this is just 3 simple getters is to not expose this class to users. Rather you can just make this package private and have a `WorkerTuner.create` or something that accepts these three values and returns a `WorkerTuner` instance (but it's really an impl of this class). There's not a ton of value in exposing this class I don't think since there's not a lot of inheritance value for how simple it is.

temporal-sdk/src/main/java/io/temporal/worker/tuning/JVMSystemResourceInfo.java
Sushisource: My concern is that I'm not sure what is going on within the JVM calls and I didn't want to encounter a thread pinning issue like Quinn had brought up before because of something happening inside of those, hence wanting to avoid `sychronized`. Reentrant lock just because that's seemingly used everywhere in the codebase regardless of need, not because I actually care about reentrancy.

The little info class makes sense to me though.

I had a check inside the lock, too, to avoid double-refreshing and oddly it was still causing me issues, but I'll make sure that works after using the class which ought to help.

temporal-sdk/src/main/java/io/temporal/worker/tuning/ResourceBasedController.java
Sushisource: Weird how easy it is for me to forget that. Fixed.

temporal-sdk/src/main/java/io/temporal/worker/tuning/ResourceBasedController.java
Sushisource: Sounds good. I had some concern about calling them overly frequently, but I suppose pid decision is already appropriately throttled anyway.

temporal-sdk/src/main/java/io/temporal/worker/tuning/JVMSystemResourceInfo.java
cretz: If You'll get the same issue with `ReentrantLock` as you will with `synchronized` for this case. If you're not using any of the more advanced methods on `ReentrantLock` and you are not calling it recursively, it is the same as `synchronized`.

Sushisource: This is the pinning thing I'm referring to: https://mikemybytes.com/2024/02/28/curiosities-of-java-virtual-threads-pinning-with-synchronized/ which was the reason to use it over `synchronized`. I agree that the locking semantics aren't any different.

cretz: Ah virtual threads, yes (though I am guessing the internals of the getting of the system info stuff doesn't yield to other threads, but unsure).

Quinn-With-Two-Ns: Yeah we should avoid performing IO under `synchronized ` , if it doesn't yield it doesn't matter

Sushisource: I sort of doubt it would too, but, regardless may as well be safe

cretz (approved):
Marking as approved since this is experimental anyways. Only thing I can think of outstanding is maybe hiding `TunerHolder` and exposing a `WorkerTuner.create` method that accepts the three slot providers (there's no magic in `TunerHolder` specifically that a user needs).

temporal-sdk/build.gradle
```diff
@@ -34,3 +34,15 @@ task registerNamespace(type: JavaExec) {
...
+
+test {
+    useJUnit {
+        excludeCategories 'io.temporal.worker.IndependentResourceBasedTests'
+    }
+}
+
+task testResourceIndependent(type: Test) {
+    useJUnit {
+        includeCategories 'io.temporal.worker.IndependentResourceBasedTests'
```
Quinn-With-Two-Ns, line 46: I wonder if you want  to disable concurrent tests as well  here in case you add more tests in the future

temporal-sdk/src/main/java/io/temporal/worker/tuning/ResourceBasedController.java
```diff
@@ -0,0 +1,140 @@
...
+  private final ReentrantLock decisionLock = new ReentrantLock();
+  private final PIDController memoryController;
+  private final PIDController cpuController;
+  private final SystemResourceInfo systemInfoSupplier;
+  private Instant lastPidRefresh = Instant.now();
+
+  private AtomicReference<Metrics> metrics = new AtomicReference<>();
+  private boolean metricsScopeSet;
+  private volatile double lastMemUsage;
+  private volatile double lastCpuUsage;
```
Quinn-With-Two-Ns, line 48: Why are you using volatile and a lock. The lock should be enough to prevent a race.

Also it seems like these values are written to, but never read?

temporal-sdk/src/main/java/io/temporal/worker/tuning/ResourceBasedTuner.java
```diff
@@ -0,0 +1,129 @@
...
+    public Builder setControllerOptions(ResourceBasedControllerOptions controllerOptions) {
+      this.controllerOptions = controllerOptions;
+      return this;
+    }
+
+    /**
+     * Set the slot options for workflow tasks. Has no effect after the worker using this tuner
+     * starts.
+     *
+     * <p>Defaults to minimum 2 slots, maximum 500 slots, and no ramp throttle.
```
Quinn-With-Two-Ns, line 62: These defaults don't align with the default poller count which is 5

temporal-sdk/src/main/java/io/temporal/worker/tuning/ResourceBasedSlotOptions.java
```diff
@@ -0,0 +1,57 @@
...
+  private final int minimumSlots;
+  private final int maximumSlots;
+  private final Duration rampThrottle;
+
+  /**
+   * @param minimumSlots minimum number of slots that will be issued without any resource checks
+   * @param maximumSlots maximum number of slots that will ever be issued
+   * @param rampThrottle time to wait between slot issuance. This value matters because how many
+   *     resources a task will use cannot be determined ahead of time, and thus the system should
+   *     ait to see how much resources are used before issuing more slots.
```
Quinn-With-Two-Ns, line 38:
```suggestion
   *     wait to see how much resources are used before issuing more slots.
```

temporal-sdk/src/main/java/io/temporal/worker/tuning/ResourceBasedSlotOptions.java
Quinn-With-Two-Ns: Yeah I think all classes called `*Options`  in the Java SDK follow a builder  pattern

temporal-sdk/src/main/java/io/temporal/worker/tuning/ResourceBasedController.java
Sushisource: Ah, yeah they're not any more. Missed that when I changed the metrics stuff.

They needed to be volatile before because they were *read* outside the lock, but now I'll just get rid of them.

temporal-sdk/build.gradle
Sushisource: Ya good call

temporal-sdk/src/main/java/io/temporal/worker/tuning/ResourceBasedSlotSupplier.java
```diff
@@ -0,0 +1,100 @@
...
+        }
+        if (mustWaitFor.compareTo(Duration.ZERO) > 0) {
+          Thread.sleep(mustWaitFor.toMillis());
+        }
+
+        Optional<SlotPermit> permit = tryReserveSlot(ctx);
+        if (permit.isPresent()) {
+          return permit.get();
+        } else {
+          Thread.sleep(10);
```
Quinn-With-Two-Ns, line 68: Seems like we if we are out of resources for new slots we could potentially have a lot of threads spinning in this loop with a very short wait, is that a concern? What is the overhead of getting system info?  When I tested in Go it was no trivial.  Also can we confirmed this slot supplier handles cancellation gracefully?

temporal-sdk/src/main/java/io/temporal/worker/tuning/ResourceBasedTuner.java
Sushisource: Changed to 5

temporal-sdk/src/main/java/io/temporal/worker/tuning/ResourceBasedSlotSupplier.java
Sushisource: 10ms is more than enough to make sure they're not going to be burning CPU - and the actual calls for getting system info are serialized and have their own throttle, so this ought to be fine. This wait is shorter than the wait for getting info, because once that refreshes, it's nice to have some lower latency there where it might free up multiple waiters in succession.

temporal-sdk/src/main/java/io/temporal/worker/tuning/ResourceBasedSlotOptions.java
Quinn-With-Two-Ns: I think we should make this a builder and add all the boilerplate helpers like all the other options classes have

temporal-sdk/src/test/java/io/temporal/worker/IndependentResourceBasedTests.java
```diff
@@ -0,0 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.temporal.worker;
+
+public interface IndependentResourceBasedTests {}
```
Quinn-With-Two-Ns, line 23: Does this need to be public?

Quinn-With-Two-Ns (approved):
LGTM, I would like us to switch `ResourceBasedSlotOptions` to align with all are other option classes.

temporal-sdk/src/test/java/io/temporal/worker/IndependentResourceBasedTests.java
Sushisource: Apparently not

PR #2115, opened 2024-06-18
Remove Spring Boot alpha tag

Quinn-With-Two-Ns:
Remove Spring Boot integration alpha tag


tlalfano (approved):
Changes and readme look good 👍

cretz (approved):

PR #2133, opened 2024-07-03
Workflow-friendly concurrency primitives

Quinn-With-Two-Ns:
Add Workflow-friendly concurrency primitives. This PR adds a workflow safe `Mutex` and `Semaphore` based on the Java interfaces

* https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/locks/Lock.html
* https://docs.oracle.com/javase%2F8%2Fdocs%2Fapi%2F%2F/java/util/concurrent/Semaphore.html

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowLockImpl.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowLockImpl.java
@@ -1,3 +1,23 @@
+/*
+ * Copyright (C) 2022 Temporal Technologies, Inc. All Rights Reserved.
+ *
+ * Copyright (C) 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Modifications copyright (C) 2017 Uber Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this material except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package io.temporal.internal.sync;
 
 import static io.temporal.internal.sync.WorkflowInternal.assertNotReadOnly;
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowLockImpl.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowLockImpl.java
@@ -1,0 +1,57 @@
+package io.temporal.internal.sync;
+
+import static io.temporal.internal.sync.WorkflowInternal.assertNotReadOnly;
+
+import com.google.common.base.Preconditions;
+import io.temporal.workflow.CancellationScope;
+import io.temporal.workflow.WorkflowLock;
+import java.time.Duration;
+
+class WorkflowLockImpl implements WorkflowLock {
+  private boolean locked = false;
+
+  @Override
+  public void lock() {
+    WorkflowInternal.await(
+        "WorkflowLock.lock",
+        () -> {
+          CancellationScope.throwCanceled();
+          return !locked;
+        });
+    locked = true;
+  }
+
+  @Override
+  public boolean tryLock() {
+    assertNotReadOnly("WorkflowLock.tryLock");
+    if (!locked) {
+      locked = true;
+      return true;
+    }
+    return false;
+  }
+
+  @Override
+  public boolean tryLock(Duration timeout) {
+    boolean unlocked =
+        WorkflowInternal.await(
+            timeout,
+            "WorkflowLock.tryLock",
+            () -> {
+              CancellationScope.throwCanceled();
+              return !locked;
+            });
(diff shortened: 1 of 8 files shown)


temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowLockImpl.java
```diff
@@ -0,0 +1,77 @@
...
+
+class WorkflowLockImpl implements WorkflowLock {
+  private boolean locked = false;
+
+  @Override
+  public void lock() {
+    WorkflowInternal.await(
+        "WorkflowLock.lock",
+        () -> {
+          CancellationScope.throwCanceled();
```
cretz, line 38: Looking at javadoc for `Workflow.await` it should automatically cancel the await and throw if outer scoped is cancelled right?

temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowSemaphoreImpl.java
```diff
@@ -0,0 +1,107 @@
...
+
+  @Override
+  public void acquire() {
+    acquire(1);
+  }
+
+  @Override
+  public void acquire(int permits) {
+    Preconditions.checkArgument(
+        permits > 0, "WorkflowSemaphore.acquire called with negative permits");
```
cretz, line 45:
```suggestion
        permits > 0, "WorkflowSemaphore.acquire called with non-positive permits");
```

Pedantic though

temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowSemaphoreImpl.java
```diff
@@ -0,0 +1,107 @@
...
+  public void release() {
+    release(1);
+  }
+
+  @Override
+  public void release(int permits) {
+    assertNotReadOnly("WorkflowSemaphore.release");
+    Preconditions.checkArgument(
+        permits > 0, "WorkflowSemaphore.release called with negative permits");
+    currentPermits += permits;
```
cretz, line 105: Not a big deal, but could add a final field for the number initially given and throw if this exceeds that number (but again not a big deal and maybe existing Java semaphore doesn't)

cretz (commented):
LGTM, just one minor check on the cancellation throw inside the await callback

temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowLockImpl.java
Quinn-With-Two-Ns: Yes, but this is calling `WorkflowInternal.await` which does not cancel by default. I use `WorkflowInternal.await` to override the wait message.

temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowSemaphoreImpl.java
Quinn-With-Two-Ns: I didn't add that to match Java's semaphore. I also documented and tested this part of the API.

temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowSemaphoreImpl.java
Quinn-With-Two-Ns: Checking Java's semaphore, the exception message was correct the check should be `>=`

temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowLockImpl.java
cretz: I wonder then if this should just return if cancelled or lock available, and then issue the actual throw outside of the lambda. But so long as the exception propagates to the user, works for me.

temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowSemaphoreImpl.java
cretz: To me it's a bit strange to allow 0 permits (I can't think of a a use case), but it's harmless.

cretz (approved):

temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowLockImpl.java
Quinn-With-Two-Ns: Yeah it will be propagated, this is how `Workflow.await` is implemented by the java SDK

temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowSemaphoreImpl.java
Quinn-With-Two-Ns: Yeah just copying Java's Semaphore here

temporal-sdk/src/main/java/io/temporal/workflow/WorkflowLock.java
```diff
@@ -0,0 +1,67 @@
...
+
+  /** Releases the lock. */
+  void unlock();
+
+  /**
+   * Checks if a lock is held.
+   *
+   * @return true if the lock is held and false otherwise.
+   */
+  boolean isHeld();
```
Quinn-With-Two-Ns, line 66: @cretz I added this API since you reviewed, I added it to support checking the mutex in a validator.

cretz: LGTM. Arguably semaphore should also provide its currently held/acquired count to users.

cretz (approved):

PR #2135, opened 2024-07-04
Change build_native_images mac runner to macos-13

Quinn-With-Two-Ns:
Change build_native_images mac runner to macos-13. `macos-latest` now points to `aarch64`. Funny enough this doesn't cause an error since out `graal` plugin we use only checks for `64` so still downloads the `x86_64` toolchain.

### DIFF
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -125,12 +125,13 @@ jobs:
     needs: create_draft_release
     if: github.event.inputs.do_build_native_images == 'true'
     strategy:
+      fail-fast: false
       matrix:
         include:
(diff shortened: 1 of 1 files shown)


cretz (approved):

cretz:
Arguably we should still test on mac arm, but understood it's a larger change to separate testing from building. Also curious if anything has changed with whatever was blocking our ability to have mac arm native executables.

Quinn-With-Two-Ns:
>Arguably we should still test on mac arm, but understood it's a larger change to separate testing from building

To be clear we were not really testing on arm mac, we were testing x86 stuff through rosetta 

> Also curious if anything has changed with whatever was blocking our ability to have mac arm native executables

No unfortunately

Issue #2103, opened 2024-06-10
Java SDK is not reporting workflow_failures from the code path via the FailWorkflowExceptionTypes


jlacefie:
## Expected Behavior
Workflow Failed Cloud Metrics and Java SDK Metrics should "closely" match each other

## Actual Behavior
Workflow Failed metrics differ between Temporal Cloud and SDK metrics. 

## Steps to Reproduce the Problem
- NDE should typically get retried through WFT, I'm not sure what exactly happened to there WFs, but I'm aware we can force the Workflow exec failure by setting WorkflowImplementationOptions with .setFailWorkflowExceptionTypes(NonDeterministicException.class)
- This allowed me to reproduce the metric discrepancy between SDK's [workflow_failed](https://docs.temporal.io/references/sdk-metrics#workflow_failed) vs server/cloud's temporal_cloud_v0_workflow_failed_count, which seems to be the main concern.
- it appears Java SDK is not reporting workflow_failures from the code path via the FailWorkflowExceptionTypes
- The only place that SDK reports workflow_failures is from: https://github.com/temporalio/sdk-java/blob/v1.23.2/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowExecutor.java#L97, which is not reachable when a workflow fails with WorkflowExecutionException from here: [https://github.com/temporalio/sdk-java/blob/v1.23.2/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowRunTaskHandler.java#L2[…]60](https://github.com/temporalio/sdk-java/blob/v1.23.2/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowRunTaskHandler.java#L256-L260)
## Specifications

  - Version:
  - Platform:
Temporal Cloud

Quinn-With-Two-Ns:
Related to https://github.com/temporalio/sdk-java/issues/1590

Quinn-With-Two-Ns:
Investigating this more the Java SDK does report `workflow_failures` correctly if the exception is throw in the user workflow code. The miss is when the exception is thrown by the SDK itself, which is often where `NonDeterministicException` are thrown.

Issue #1590, opened 2023-01-08
workflow_completed counter counts successful completions of workflow method instead of workflow executions


Spikhalskiy:
## Expected Behavior

workflow_completed metric counter is expected to count new completed executions.

## Actual Behavior

workflow_completed counter counts successful completion of workflow methods. Including replays for example.

cretz:
I think we need to fix this in multiple SDKs. May be worth an sdk-features issue.

PR #2141, opened 2024-07-08
Make sure workflow_failed is incremented on NonDeterministicException

Quinn-With-Two-Ns:
Make sure `workflow_failed` is incremented on `NonDeterministicException`

closes https://github.com/temporalio/sdk-java/issues/2103

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowRunTaskHandler.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowRunTaskHandler.java
@@ -255,9 +255,13 @@ class ReplayWorkflowRunTaskHandler implements WorkflowRunTaskHandler {
               implementationOptions.getFailWorkflowExceptionTypes();
           for (Class<? extends Throwable> failType : failTypes) {
             if (failType.isAssignableFrom(e.getClass())) {
+              metricsScope.counter(MetricsType.WORKFLOW_FAILED_COUNTER).inc(1);
               throw new WorkflowExecutionException(
                   workflow.getWorkflowContext().mapWorkflowExceptionToFailure(e));
             }
+          }
+          if (e instanceof WorkflowExecutionException) {
+            metricsScope.counter(MetricsType.WORKFLOW_FAILED_COUNTER).inc(1);
           }
           throw wrap(e);
         }
--- a/temporal-sdk/src/test/java/io/temporal/internal/worker/WorkflowFailedMetricsTests.java
+++ b/temporal-sdk/src/test/java/io/temporal/internal/worker/WorkflowFailedMetricsTests.java
@@ -30,6 +30,7 @@ import io.temporal.client.WorkflowClient;
 import io.temporal.client.WorkflowFailedException;
 import io.temporal.client.WorkflowOptions;
 import io.temporal.common.reporter.TestStatsReporter;
+import io.temporal.failure.ApplicationFailure;
 import io.temporal.testing.internal.SDKTestWorkflowRule;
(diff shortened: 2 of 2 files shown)


temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowRunTaskHandler.java
```diff
@@ -255,10 +255,14 @@ private void applyServerHistory(long lastEventId, WorkflowHistoryIterator histor
               implementationOptions.getFailWorkflowExceptionTypes();
           for (Class<? extends Throwable> failType : failTypes) {
             if (failType.isAssignableFrom(e.getClass())) {
+              metricsScope.counter(MetricsType.WORKFLOW_FAILED_COUNTER).inc(1);
```
cretz, line 258: I see `completeWorkflow` increments this metric. Do these exceptions not make it to `completeWorkflow`? Is there a common place that `completeWorkflow`-based exceptions _and_ these are turned to workflow failures that may be a better place to move the metric, e.g. `WorkflowStateMachines.failWorkflow`? Or even if we don't want to change `completeWorkflow`, is there a common place where `WorkflowExecutionException` is caught and converted?

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowRunTaskHandler.java
```diff
@@ -255,10 +255,14 @@ private void applyServerHistory(long lastEventId, WorkflowHistoryIterator histor
...
               throw new WorkflowExecutionException(
                   workflow.getWorkflowContext().mapWorkflowExceptionToFailure(e));
             }
           }
+          if (e instanceof WorkflowExecutionException) {
```
cretz, line 263: Curious, what is the situation where `e` is a `WorkflowExecutionException` here?

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowRunTaskHandler.java
Quinn-With-Two-Ns: >Do these exceptions not make it to completeWorkflow

Correct, this code catches exceptions from the state machine, not the workflow code itself

>Is there a common place that completeWorkflow-based exceptions and these are turned to workflow failures that may be a better place to move the metric, e.g. WorkflowStateMachines.failWorkflow

No

>is there a common place where WorkflowExecutionException is caught and converted?

The only place would be when the SDK responds to the server, but that would require changing where all completion metrics are emitted. That work is captured here https://github.com/temporalio/sdk-java/issues/1590. I was hopping to defer that work till later.

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowRunTaskHandler.java
Quinn-With-Two-Ns: I don't think it can happen today, but if `workflowStateMachines.handleEvent(event, hasNext);` did throw a `WorkflowExecutionException` it would lead to a workflow failure. So it is more future proofing, it also may be possible with some user interceptor could throw a `WorkflowExecutionException`.

cretz: Works for me.

cretz (approved):
LGTM. I feel better knowing we are asserting _exactly_ 1 and so we will make sure dupe fail counts don't creep in if/when we refactor to a common place.

PR #2143, opened 2024-07-10
Reintroduce slot supplier & add many tests

Sushisource:
## What was changed
Re-introduce slot supplier framework reverted in https://github.com/temporalio/sdk-java/pull/2134

## Why?
Bring back feature with fix for leaking local activity slots

## Checklist

1. Closes 

2. How was this tested:
Added many tests

3. Any docs updates needed?


temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
```diff
@@ -256,51 +263,70 @@ public boolean dispatch(
...
-          }
-          if (!accepted) {
-            log.warn(
-                "LocalActivity queue is full and submitting timed out for activity {} with acceptanceTimeoutMs: {}",
-                activityTask.getActivityId(),
-                acceptanceTimeoutMs);
+          try {
+            TimeLimiter timeLimiter = SimpleTimeLimiter.create(timeoutThreadPool);
+            permit =
+                timeLimiter.callWithTimeout(
```
Sushisource, line 290: This is the actual fix. The source of the bug was not assigning the permit when reservation was called behind this `callWithTimeout` function.

temporal-sdk/src/test/java/io/temporal/internal/worker/WorkflowSlotsSmallSizeTests.java
```diff
@@ -0,0 +1,285 @@
...
+  public void setup() {
+    reporter.flush();
+    parallelSemRunning = new Semaphore(0);
+    parallelSemBlocked = new Semaphore(0);
+  }
+
+  @After
+  public void tearDown() {
+    testWorkflowRule.getTestEnvironment().close();
+    assertEquals(
```
Sushisource, line 109: There is some dupe among these tests with this bit, and asserting on the metrics, etc.

Not sure if it's actually worth it or not to dedupe it, though. Welcome any thoughts on that.

temporal-sdk/src/test/java/io/temporal/internal/worker/WorkflowSlotTests.java
```diff
@@ -49,9 +54,16 @@ public class WorkflowSlotTests {
   private final int MAX_CONCURRENT_WORKFLOW_TASK_EXECUTION_SIZE = 100;
   private final int MAX_CONCURRENT_ACTIVITY_EXECUTION_SIZE = 1000;
   private final int MAX_CONCURRENT_LOCAL_ACTIVITY_EXECUTION_SIZE = 10000;
+  private final CountingSlotSupplier<WorkflowSlotInfo> workflowTaskSlotSupplier =
```
Quinn-With-Two-Ns, line 57: I'm not sure how I feel about this because now you are not longer testing the default worker setting.

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
```diff
@@ -98,13 +98,13 @@ public WorkflowWorker(
     this.pollerOptions = getPollerOptions(options);
     this.workerMetricsScope =
         MetricsTag.tagged(options.getMetricsScope(), WorkerMetricsTag.WorkerType.WORKFLOW_WORKER);
+    slotSupplier.setMetricsScope(workerMetricsScope);
```
Quinn-With-Two-Ns, line 101: Could we not take a more generic slot supplier in the constructor of `WorkflowWorker` and construct the `TrackingSlotSupplier` here where we have the metric scope.

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
```diff
@@ -256,51 +263,70 @@ public boolean dispatch(
...
+                    () -> slotSupplier.reserveSlot(reservationCtx),
```
Quinn-With-Two-Ns, line 291: I think this is showing `slotSupplier` needs to have a method to try to reserve a slot with a timeout. A lot of slot suppliers can be more efficient by taking the timeout directly without needing to spin up a bunch of threads .

temporal-sdk/src/main/java/io/temporal/internal/worker/TrackingSlotSupplier.java
```diff
@@ -0,0 +1,219 @@
...
+  }
+
+  Map<SlotPermit, SI> getUsedSlots() {
+    return usedSlots;
+  }
+
+  private void publishSlotsMetric() {
+    this.metricsScope
+        .gauge(MetricsType.WORKER_TASK_SLOTS_AVAILABLE)
+        .update(maximumSlots() - usedSlots.size());
```
Quinn-With-Two-Ns, line 115: If `inner` does not implement `maximumSlots()` then I don't think we should be emitting this metric. May need to modify the definition of `maximumSlots` to help indicate if the underlying slot supplier actually has a fixed size

temporal-sdk/src/test/java/io/temporal/internal/worker/WorkflowSlotTests.java
Sushisource: It's a fairly thin wrapper and the only reasonable way I had to actually test the slot mechanism directly instead of via metrics without exposing a bunch more stuff publicly. 

If you have another suggestion I'm open for sure, but, this was seemingly the best way

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
Sushisource: Yeah probably, let me check that out

temporal-sdk/src/main/java/io/temporal/internal/worker/TrackingSlotSupplier.java
Sushisource: Yeah, I thought about this and can't remember why I decided it was still worth doing, but I'll make that change.

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
Sushisource: I don't like really like how that asks more of implementers, but I do get why the additional thread usage would be potentially problematic with tons of LAs. It'd be nice if there was some alternative to forcing it on the user but, I don't really see an obvious one.

The other option that seems to come to mind is to change this to a call to `tryReserveSlot`, polled on some interval, and then one thread can handle polling for all LAs, but, that could potentially have some weird fairness side-effects.

temporal-sdk/src/test/java/io/temporal/internal/worker/WorkflowSlotTests.java
Quinn-With-Two-Ns: Yeah I was thinking you could have a few tests that schedule more activities/workflows then you have slots so you know some had to be freed. Thoughts on a test like that? You can also keep a static count in the activity to cross check with.

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
Quinn-With-Two-Ns: Your not really asking more, just something different. The current implementation ask for them to make their code interruption aware, which isn't always the default. It also provides no way for the implementer to know the deadline of the request. I wouldn't do polling for fairness reasons and it still doesn't give the interface all the information. It is very normal in Java for interfaces like this to have overloads for blocking get, non blocking get, blocking with timeout, see [lock](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/locks/Lock.html) for example.

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
```diff
@@ -449,6 +474,20 @@ public void handle(LocalActivityAttemptTask attemptTask) throws Exception {
       MDC.put(LoggerTag.WORKFLOW_ID, activityTask.getWorkflowExecution().getWorkflowId());
       MDC.put(LoggerTag.WORKFLOW_TYPE, activityTask.getWorkflowType().getName());
       MDC.put(LoggerTag.RUN_ID, activityTask.getWorkflowExecution().getRunId());
+
+      // At this point the permit should definitely be set
+      if (executionContext.getPermit() == null) {
+        throw new IllegalStateException("Permit is expected to be set at this point");
```
Quinn-With-Two-Ns, line 480: Is there a reason this check can't just be in the `TrackingSlotSupplier`?  I think its' current behaviour to silently do nothing on `null` permits contributed to us not noticing this issue sooner.

temporal-sdk/src/test/java/io/temporal/worker/ResourceBasedTunerTests.java
```diff
@@ -0,0 +1,146 @@
...
+import io.temporal.workflow.*;
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+
+public class ResourceBasedTunerTests {
```
Quinn-With-Two-Ns, line 42: Can we include local activities in this test? I guess the resource based tuner does not emit any metrics about how many slots are currently alllocated?

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
Sushisource: OK I will add the timeout to the interface

temporal-sdk/src/test/java/io/temporal/internal/worker/WorkflowSlotTests.java
Sushisource: Sure I can add something like that

temporal-sdk/src/test/java/io/temporal/worker/ResourceBasedTunerTests.java
Sushisource: I actually need to add the `worker_task_slots_used` metric that exists in Core anyway, I have added it in my pending Go changes but didn't add it here so I'll use it from this tests and do LAs.

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
Sushisource: Seems like it could be in `markUsed`. I will verify and move it there.

temporal-sdk/src/main/java/io/temporal/internal/worker/TrackingSlotSupplier.java
```diff
@@ -64,7 +64,8 @@ public Optional<SlotPermit> tryReserveSlot(SlotReservationData dat) {
 
   public void markSlotUsed(SI slotInfo, SlotPermit permit) {
     if (permit == null) {
-      return;
+      throw new IllegalStateException(
```
Quinn-With-Two-Ns, line 67: Nit: `IllegalArgumentException`

Quinn-With-Two-Ns: Also doesn't this same check also apply to release as well?

Sushisource: I picked state because a user seeing this never specified the argument and doesn't care. It's illegal state in the sense that the SDK is in an illegal state. Yeah, will add to release too.

Quinn-With-Two-Ns: So `IllegalStateException` is if a method is invoked at an inappropriate time, `IllegalArgumentException` is if the argument was invalid, here the argument was invalid

temporal-sdk/src/main/java/io/temporal/worker/MetricsType.java
```diff
@@ -135,6 +135,8 @@ private MetricsType() {}
   // gauge
   public static final String WORKER_TASK_SLOTS_AVAILABLE =
       TEMPORAL_METRICS_PREFIX + "worker_task_slots_available";
+  public static final String WORKER_TASK_SLOTS_USED =
```
Quinn-With-Two-Ns, line 138: Nit: these are public, which is weird but it is what it is, so should be marked experimental.

temporal-sdk/src/main/java/io/temporal/internal/worker/TrackingSlotSupplier.java
Quinn-With-Two-Ns: Like I understand your point, but these exceptions have pretty standard meanings and I'd prefer to stick to them.

Sushisource: Yeah I just think it's very debatable because the user _has no control_ over the argument. I'll change it since it's really not a big deal either way.

temporal-sdk/src/main/java/io/temporal/internal/worker/TrackingSlotSupplier.java
```diff
@@ -0,0 +1,212 @@
...
+  private final Map<SlotPermit, SI> usedSlots = new ConcurrentHashMap<>();
+  private final Scope metricsScope;
+
+  public TrackingSlotSupplier(SlotSupplier<SI> inner, Scope metricsScope) {
+    this.inner = inner;
+    this.metricsScope = metricsScope;
+    publishSlotsMetric();
+  }
+
+  public SlotPermit reserveSlot(SlotReservationData dat, long timeout, TimeUnit timeUnit)
```
Quinn-With-Two-Ns, line 53: Sorry this is my fault I should have clarified I think you need 3 over loads here `reserveSlot ` that takes no timeout, `tryReserveSlot` with not timeout that is non blocking, and `tryReserveSlot ` that takes a timeout, just like https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/locks/Lock.html or other interfaces in Java

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
```diff
@@ -416,14 +437,13 @@ private AttemptTaskHandlerImpl(ActivityTaskHandler handler) {
 
     @Override
     public void handle(LocalActivityAttemptTask attemptTask) throws Exception {
```
Quinn-With-Two-Ns, line 439: I would need to double check the Java SDK, but is `handle` here called per attempt of a local activity? Is there a risk we `markSlotUsed` multiple times for the same activity?

temporal-sdk/src/main/java/io/temporal/internal/worker/TrackingSlotSupplier.java
Sushisource: Yeah, kinda realized that after the fact

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
Sushisource: Good call. Added a test & fixed.

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityExecutionContext.java
```diff
@@ -154,6 +161,14 @@ public boolean callback(LocalActivityResult result) {
     if (scheduleToCloseFuture != null) {
       scheduleToCloseFuture.cancel(false);
     }
+    SlotReleaseReason reason = SlotReleaseReason.taskComplete();
```
Quinn-With-Two-Ns, line 164: Is it possible that the slot was never used?

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
Quinn-With-Two-Ns: Looking at the original code it seems like before the local activity slot metric was not considered "used" while the local activity was backing off. Now I think the slot will be considered "used" if the activity is executing or not. Do you think I am correct? If so I think that is fine since the slot is not available.

temporal-sdk/src/main/java/io/temporal/internal/worker/TrackingSlotSupplier.java
```diff
@@ -0,0 +1,223 @@
...
+    return p;
+  }
+
+  public void markSlotUsed(SI slotInfo, SlotPermit permit) {
+    if (permit == null) {
+      throw new IllegalArgumentException(
+          "Permit cannot be null when marking slot as used. This is an SDK bug.");
+    }
+    // Avoid marking the same slot as used twice. EX: LA retries might call this multiple times.
+    if (usedSlots.put(permit, slotInfo) == null) {
```
Quinn-With-Two-Ns, line 83: Hm, but `slotInfo` has information about the current attempt as well right? With this check the information is not being updated. How did you deal with this in Core? I think the "correct" fix is to add a `markUnused` separate from  `releaseSlot` , but I can understand not wanting to add a method for just this case?

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
Quinn-With-Two-Ns: Actually looking at the code more it appears there is a more significant change here. Before the local activity concurrency was limited by the number of threads in the thread pool, and more then the max local activity task slot could be scheduled. Now we are limiting the max number of local activities being scheduled to the number of local activity slots. That could have implications to users who use local activities for certain use cases like polling.

temporal-sdk/src/main/java/io/temporal/internal/worker/TrackingSlotSupplier.java
Sushisource: In Core different LA attempts actually have to re-aquire new permits, and the old one is released. That solves the issue with different information and possibly different release reasons (though there are no release reasons in Core right now).

However it does potentially mean LA retries could be blocked by new LAs. That can be fixed by always prioritizing retries for permit acquisition though.

I don't really want to change that much behavior here though, so my inclination would be to allow mark-used to be called twice here (which matches core semantics, just a bit of a different mechanism). I don't think we need `markUnused` if the docs for `markUsed` say "may be called with the same permit for new local activity attempts".

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityExecutionContext.java
Sushisource: No - it's used in the sense that we tried to run the LA even if it timed out, which counts as `taskComplete` in this case. We might need finer-grained reasons at some point if people ask for them

temporal-sdk/src/main/java/io/temporal/internal/worker/TrackingSlotSupplier.java
Quinn-With-Two-Ns: I think the current Java SDK behaviour aligns with core and your changes have made it different. See [here](https://github.com/temporalio/sdk-java/pull/2143#discussion_r1683106822)  where I explain my concern.

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
Sushisource: Discussed on call:

* Will keep the ratelimiter which is 2x max size, or for implementations w/o max size, 2x number of currently running activities (only to be triggered if tryReserve fails).
* Release slots for LAs which are backing off, prioritize retries and re-acquire then

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivitySlotSupplierQueue.java
```diff
@@ -0,0 +1,117 @@
...
+  private final Thread queueThread;
+  private static final Logger log =
+      LoggerFactory.getLogger(LocalActivitySlotSupplierQueue.class.getName());
+  private volatile boolean running = true;
+
+  LocalActivitySlotSupplierQueue(
+      TrackingSlotSupplier<LocalActivitySlotInfo> slotSupplier,
+      Functions.Proc1<LocalActivityAttemptTask> afterReservedCallback) {
+    this.afterReservedCallback = afterReservedCallback;
+    // TODO: See if I can adjust this for dynamic ones based on current rate
```
Sushisource, line 39: This is kinda the last thing. I don't think there's really much sensible that can be done here, but maybe we want to make it an option?

temporal-sdk/src/main/java/io/temporal/internal/worker/TrackingSlotSupplier.java
Sushisource: Resolving per our talk / can use other thread

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityExecutionContext.java
```diff
@@ -154,6 +161,14 @@ public boolean callback(LocalActivityResult result) {
...
+    if (result.getProcessingError() != null) {
+      reason = SlotReleaseReason.error(new Exception(result.getProcessingError().getThrowable()));
+    }
+    // Permit can be null in the event of a timeout while waiting on a permit
+    if (permit != null) {
+      slotSupplier.releaseSlot(reason, permit);
```
Quinn-With-Two-Ns, line 170: Didn't we say we would move the release out of here?

Sushisource: Yes, good point. And move it instead to `handle`, which is more like the normal task types?

Sushisource: I guess the main issue is that this gets referenced from like a bajillion places, so it's not obvious that centralizing it is going to catch all those... but I can see what happens.

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
```diff
@@ -584,6 +554,8 @@ private void handleResult(
               executionContext, activityTask, activityHandlerResult.getTaskFailed().getFailure());
 
       if (retryDecision.doNextAttempt()) {
+        // Release slot before scheduling the next attempt
+        slotSupplier.releaseSlot(SlotReleaseReason.willRetry(), executionContext.getPermit());
```
Quinn-With-Two-Ns, line 558: Shouldn't you  pass the failure here like you do for normal activities?

Sushisource: Normal activities don't actually set an error - it's more like when _processing_ failed rather than the task failed that the `error` variant is used. So I think this is consistent.

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityExecutionContext.java
Quinn-With-Two-Ns: It also looks like your release the slot again  in `handleResult`? Maybe it is being release twice.  The slot acquire/release. `handle` is the right spot IMO, currently the SDK takes and releases slots after handle.

Quinn-With-Two-Ns: At a high level the slot  acquire/release should be tied to the processing of the task, not the reporting of the result.

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivitySlotSupplierQueue.java
Quinn-With-Two-Ns: Yeah  I would open an issue to revisit the backoff logic. My  main goal was to  just maintain the current behavior for a fixed slot suppliers without some more thought.

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityExecutionContext.java
Sushisource: :+1: done

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
```diff
@@ -416,40 +407,49 @@ private AttemptTaskHandlerImpl(ActivityTaskHandler handler) {
...
-      MDC.put(LoggerTag.WORKFLOW_TYPE, activityTask.getWorkflowType().getName());
-      MDC.put(LoggerTag.RUN_ID, activityTask.getWorkflowExecution().getRunId());
       try {
+        // if an activity was already completed by any mean like scheduleToClose or scheduleToStart,
+        // discard this attempt, this execution is completed.
+        // The scheduleToStartFired check here is a bit overkill, but allows to catch an edge case
+        // where scheduleToStart is already fired, but didn't report a completion yet.
+        boolean shouldDiscardTheAttempt = scheduleToStartFired || executionContext.isCompleted();
+        if (shouldDiscardTheAttempt) {
+          return;
```
Quinn-With-Two-Ns, line 428: Here shouldn't the task slot release reason be not used?

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
```diff
@@ -497,6 +500,10 @@ public void handle(LocalActivityAttemptTask attemptTask) throws Exception {
             processingFailed(activityTask.getActivityId(), activityTask.getAttempt(), ex));
         throw ex;
       } finally {
+        // Permit can be null in the event of a timeout while waiting on a permit
+        if (executionContext.getPermit() != null) {
```
Quinn-With-Two-Ns, line 504: Really? I am suprised by this comment because how did we end up submitting the `attemptTask` without a permit?

Quinn-With-Two-Ns:
Did we add a test that shows retrying local activities do not hold the slot while in a back-off stage?

temporal-sdk/src/main/java/io/temporal/internal/worker/TrackingSlotSupplier.java
```diff
@@ -0,0 +1,211 @@
...
+          "Permit cannot be null when marking slot as used. This is an SDK bug.");
+    }
+    if (usedSlots.put(permit, slotInfo) != null) {
+      throw new IllegalStateException("Slot is being marked used twice. This is an SDK bug.");
+    }
+    inner.markSlotUsed(new SlotMarkUsedContextImpl(slotInfo, permit));
+    publishSlotsMetric();
+  }
+
+  public void releaseSlot(SlotReleaseReason reason, SlotPermit permit) {
```
Quinn-With-Two-Ns, line 77: What happens on a double release? I think right now it will silently pass?

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
Sushisource: Not quite. This is a timeout, and the timeouts right now are also taskComplete. Not used is right now just "didn't even get a task"

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
Sushisource: Yeah, I guess that's no longer true after the change to use the queue

temporal-sdk/src/main/java/io/temporal/internal/worker/TrackingSlotSupplier.java
Sushisource: It will. There's not a great way to avoid that that I see, at least not without making some kind of publicly accessible `isReleased` field on the permit, since that's in worker rather than internal. Could be done though. Alternatively it can just be tested for with `CountingSlotSupplier` in the tests.

Sushisource:
> Did we add a test that shows retrying local activities do not hold the slot while in a back-off stage?

This is implicitly covered in `TestLocalActivityFailsThenPasses` since the acquires & releases have to line up to the number that would be expected with retrying, but I can try to add something more explicit.

temporal-sdk/src/main/java/io/temporal/internal/worker/TrackingSlotSupplier.java
Quinn-With-Two-Ns: Yeah I was thinking it was easier to check then maybe it actually is since it probably comes down to the route slot supplier. Maybe just testing `CountingSlotSupplier` is enough

Quinn-With-Two-Ns (approved):
LGTM, I know @cretz reviewed the original, may want to double check if he has any thoughts, not required though.

PR #2144, opened 2024-07-10
Warn on dangling handlers and add method to help await on all handlers.

Quinn-With-Two-Ns:
Warn on dangling handlers and add method to help await on all handlers.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowExecutor.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowExecutor.java
@@ -219,8 +219,8 @@ final class ReplayWorkflowExecutor {
       Input input = update.getInput();
       Optional<Payloads> args = Optional.ofNullable(input.getArgs());
       this.workflow.handleUpdate(
-          update.getMeta().getUpdateId(),
           input.getName(),
+          update.getMeta().getUpdateId(),
           args,
           protocolMessage.getEventId(),
           input.getHeader(),
--- a/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowExecutor.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowExecutor.java
@@ -20,6 +20,7 @@
 
 package io.temporal.internal.replay;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.protobuf.InvalidProtocolBufferException;
 import com.google.protobuf.util.Timestamps;
 import com.uber.m3.tally.Scope;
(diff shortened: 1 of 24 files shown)


temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowExecutor.java
```diff
@@ -89,6 +114,18 @@ public void eventLoop() {
   }
 
   private void completeWorkflow(@Nullable WorkflowExecutionException failure) {
+    Map<Long, SignalHandlerInfo> runningSignalHandlers =
+        workflow.getWorkflowContext().getRunningSignalHandlers();
+    if (!runningSignalHandlers.isEmpty()) {
+      log.warn(unfinishedSignalHandlesWarnMessage);
```
cretz, line 120: In other SDKs we are showing the unfinished handlers. I think we should here too. May want to use a `LinkedHashMap` so we can preserve insertion order, or maybe even `LinkedHashSet` returning the info objects themselves to caller for later removal (will be identity equality on those info objects).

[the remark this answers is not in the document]
Quinn-With-Two-Ns: Yes it will this logic is not done yet

Quinn-With-Two-Ns: Agree, as noted in the PR description this logic is not done yet. I am trying to find other java libraries the log a list of things as a MDC value.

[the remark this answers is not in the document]
cretz: :+1: Let me know when to re-review

cretz: :+1: Let me know when to re-review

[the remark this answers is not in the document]
Quinn-With-Two-Ns: Fixed

Quinn-With-Two-Ns: Currently I have it logging the list as an ordered string, one open question I have is if it should be the default `ToString` or if I should format it as json?

cretz (commented):
LGTM, minor note on the `is` method

temporal-sdk/src/main/java/io/temporal/internal/sync/SignalDispatcher.java
```diff
@@ -52,6 +48,8 @@ class SignalDispatcher {
   /** Buffers signals which don't have a registered listener. */
   private final Queue<SignalData> signalBuffer = new ArrayDeque<>();
 
+  private Map<Long, SignalHandlerInfo> runningSignalHandlers = new TreeMap<>();
```
cretz, line 51: A linked hash map would also work here, but this is fine too. May want to make LHS be `SortedMap` for readers though, but no biggie.

temporal-sdk/src/main/java/io/temporal/workflow/Workflow.java
```diff
@@ -1222,6 +1222,22 @@ public static void applyLocalActivityOptions(
...
+   * <p>Consider waiting on this condition before workflow return or continue-as-new, to prevent
+   * interruption of in-progress handlers by workflow return:
+   *
+   * <pre><code>
+   *    Workflow.await(() -&gt; Workflow.isAllHandlersFinished());
+   *  </pre></code> will eventually have search attributes as:
+   *
+   * @return true if all handlers are finished, false otherwise.
+   */
+  public static boolean isAllHandlersFinished() {
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: Yeah good idea i'll do `isEveryHandlerFinished`

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowExecutor.java
```diff
@@ -36,13 +37,45 @@
...
+          + "handler annotations: `@UpdateMethod(unfinishedPolicy = HandlerUnfinishedPolicy.ABANDON)`.";
+
+  @VisibleForTesting
+  public static final String unfinishedSignalHandlesWarnMessage =
+      "[TMPRL1102] Workflow finished while signal handlers are still running. This may "
+          + "have interrupted work that the update handler was doing. You can wait for all update and "
+          + "signal handlers to complete by using `await workflow.Await(() -> workflow.isHandlersFinished())`. "
+          + "Alternatively, if both you  are okay with "
+          + "interrupting running handlers when the workflow finishes, and causing "
+          + "clients to receive errors, then you can disable this warning via the update "
```
dandavison, line 75:
```suggestion
          + "interrupting running handlers when the workflow finishes, "
          + "then you can disable this warning via the update "
```

I realized today that this part of the message doesn't apply to signals so have been going through removing it from each lang.

dandavison: It looks like "and the and the clients sending the signal" is missing:

> `if both you  are okay with`

cretz:
May need to wait on official stance on whether we're warning on failure and cancel (looks like no for failure and yes for cancel). We have to go update other SDKs once this stance is official too (cc @dandavison).

Quinn-With-Two-Ns:
@cretz spoke with Dan and the decision is to just not warn on failure.

cretz (approved):

PR #2146, opened 2024-07-17
Experimental cloud operations client

### LINKED ISSUE #2059: Cloud Operations API Client

See https://github.com/temporalio/features/issues/440 and https://github.com/temporalio/sdk-go/pull/1426. We need a separate cloud client using the cloud protos (they are separate today). The client should be marked experimental.

cretz:
## What was changed

* Added api-cloud protos as new submodule and included in proto gen
* Added `io.temporal.serviceclient.CloudServiceStubs` and impl and options
* Added `io.temporal.client.CloudOperationsClient` as (currently) just a simple client that returns the raw service for use
* Added CI test

## Checklist

1. Closes #2059

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,8 +83,6 @@ jobs:
     name: Unit test with cloud
     runs-on: ubuntu-latest
     timeout-minutes: 30
-    # Only supported in non-fork runs, since secrets are not available in forks
-    if: ${{ github.event.pull_request.head.repo.full_name == '' || github.event.pull_request.head.repo.full_name == 'temporalio/sdk-java' }}
     steps:
     - name: Checkout repo
       uses: actions/checkout@v4
@@ -103,6 +101,10 @@ jobs:
       uses: gradle/actions/setup-gradle@v3
     
     - name: Run cloud test
+      # Only supported in non-fork runs, since secrets are not available in forks. We intentionally
+      # are only doing this check on the step instead of the job so we require job passing in CI
+      # even for those that can't run this step.
+      if: ${{ github.event.pull_request.head.repo.full_name == '' || github.event.pull_request.head.repo.full_name == 'temporalio/sdk-java' }}
       env:
         USER: unittest
         TEMPORAL_CLIENT_CLOUD_NAMESPACE: sdk-ci.a2dd6
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,6 +79,43 @@ jobs:
       with:
         report_paths: '**/build/test-results/test/TEST-*.xml'
 
+  unit_test_cloud:
+    name: Unit test with cloud
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    # Only supported in non-fork runs, since secrets are not available in forks
+    if: ${{ github.event.pull_request.head.repo.full_name == '' || github.event.pull_request.head.repo.full_name == 'temporalio/sdk-java' }}
+    steps:
+    - name: Checkout repo
+      uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+        submodules: recursive
+        ref: ${{ github.event.pull_request.head.sha }}
+
+    - name: Set up Java
+      uses: actions/setup-java@v4
+      with:
+        java-version: "11"
+        distribution: "temurin"
+
+    - name: Set up Gradle
+      uses: gradle/actions/setup-gradle@v3
+    
+    - name: Run cloud test
+      env:
+        USER: unittest
+        TEMPORAL_CLIENT_CLOUD_NAMESPACE: sdk-ci.a2dd6
+        TEMPORAL_CLIENT_CLOUD_API_KEY: ${{ secrets.TEMPORAL_CLIENT_CLOUD_API_KEY }}
+        TEMPORAL_CLIENT_CLOUD_API_VERSION: 2024-05-13-00
+      run: ./gradlew --no-daemon :temporal-sdk:test --tests '*CloudOperationsClientTest'
+
+    - name: Publish Test Report
+      uses: mikepenz/action-junit-report@v4
+      if: success() || failure() # always run even if the previous step fails
+      with:
+        report_paths: '**/build/test-results/test/TEST-*.xml'
+
   copyright:
     name:  Copyright and code format
     runs-on: ubuntu-latest
--- a/temporal-serviceclient/src/main/java/io/temporal/serviceclient/CloudServiceStubsOptions.java
+++ b/temporal-serviceclient/src/main/java/io/temporal/serviceclient/CloudServiceStubsOptions.java
@@ -1,0 +1,130 @@
+/*
+ * Copyright (C) 2022 Temporal Technologies, Inc. All Rights Reserved.
+ *
+ * Copyright (C) 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Modifications copyright (C) 2017 Uber Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this material except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.temporal.serviceclient;
+
+import io.grpc.ManagedChannel;
+import java.util.Objects;
+
+/**
+ * Options for cloud service.
+ *
+ * <p>WARNING: The cloud service is currently experimental.
+ */
+public final class CloudServiceStubsOptions extends ServiceStubsOptions {
+  public static final String DEFAULT_CLOUD_TARGET = "saas-api.tmprl.cloud:443";
+
+  private static final CloudServiceStubsOptions DEFAULT_INSTANCE =
+      newBuilder().validateAndBuildWithDefaults();
+
+  /** Version header if any. */
+  private final String version;
+
+  public static Builder newBuilder() {
+    return new Builder();
+  }
+
+  public static Builder newBuilder(CloudServiceStubsOptions options) {
+    // We intentionally only accept our options and not the base class of
+    // options to ensure our defaults were originally applied at some point
+    // when the options class was first created.
+    return new Builder(options);
+  }
+
+  public static CloudServiceStubsOptions getDefaultInstance() {
+    return DEFAULT_INSTANCE;
+  }
+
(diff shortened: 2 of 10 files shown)


temporal-serviceclient/src/main/java/io/temporal/serviceclient/CloudServiceStubsOptions.java
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  /** Builder is the builder for ClientOptions. */
+  public static class Builder extends ServiceStubsOptions.Builder<Builder> {
+    private String version;
+
+    private Builder() {
+      // Set defaults only in this constructor
+      setTarget(DEFAULT_CLOUD_TARGET);
+      setEnableHttps(true);
```
cretz, line 88: I tried a few different ways of doing defaults for these, but there are struggles in each other way (e.g. making enable HTTPs a `Boolean` to determine if unset, etc). Doing it here and forcing the `newBuilder` overload to only accept a full `CloudServiceStubsOption` to copy from was the best way I could devise with limited changes.

.github/workflows/ci.yml
```diff
@@ -79,6 +79,43 @@ jobs:
       with:
         report_paths: '**/build/test-results/test/TEST-*.xml'
 
+  unit_test_cloud:
```
cretz, line 82: I wonder if this should just be a step in one of the other jobs, and if so, which one? Having this as a separate job means even if it fails it's not going to block a merge because we can't require it because it only works on non-forks.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/CloudServiceStubsOptions.java
```diff
@@ -0,0 +1,130 @@
...
+
+import io.grpc.ManagedChannel;
+import java.util.Objects;
+
+/**
+ * Options for cloud service.
+ *
+ * <p>WARNING: The cloud service is currently experimental.
+ */
+public final class CloudServiceStubsOptions extends ServiceStubsOptions {
```
Quinn-With-Two-Ns, line 31: Could we add a `toString`? users have complained in the past that *Options didn't implement `toString`

temporal-serviceclient/src/main/java/io/temporal/serviceclient/CloudServiceStubsOptions.java
```diff
@@ -0,0 +1,130 @@
...
+    /** Default is {@link #DEFAULT_CLOUD_TARGET}. See inherited method for more details. */
+    @Override
+    public Builder setTarget(String target) {
+      return super.setTarget(target);
+    }
+
+    @Override
+    public Builder setChannel(ManagedChannel channel) {
+      // Unset our defaults
+      setEnableHttps(false);
```
Quinn-With-Two-Ns, line 111: Why do we need this? `setChannel ` should be mutually exclusive with `setEnableHttps `

temporal-serviceclient/src/main/java/io/temporal/serviceclient/CloudServiceStubsOptions.java
cretz: I specifically excluded this because it's not in `OperatorServiceStubsOptions` and I'd argue it was improperly implemented on `WorkflowServiceStubsOptions` because it doesn't even include all the settings from the inherited class. But if I do need to make it just a string with the one `version` option I can.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/CloudServiceStubsOptions.java
cretz: There is this check in `ServiceStubsOptions.Builder.validateAndBuildWithDefaults`:

```
      if (this.enableHttps && this.channel != null) {
        throw new IllegalStateException(
            "Only one of the 'enableHttps' or 'channel' options can be set at a time");
      }
```

So if someone wants to set a channel, they don't know they have to go unset `enableHttps`. Earlier incarnations didn't do defaulting `enableHttps` to `true` eagerly this way, but they were all ugly.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/CloudServiceStubsOptions.java
Quinn-With-Two-Ns: If you think the `WorkflowServiceStubsOptions ` is wrong can you open an issue? I am fine to defer adding a `toString` till that issue since then there will be a lot more options included.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/CloudServiceStubsOptions.java
Quinn-With-Two-Ns: Probably not a big deal, but in principle I don't like overriding options user may have set. Why do you need to set the target and https in the builder constructor? couldn't you set it in `validateAndBuildWithDefaults` if a channel is not set?  Or just set them for `DEFAULT_INSTANCE`?

temporal-serviceclient/src/main/java/io/temporal/serviceclient/CloudServiceStubsOptions.java
cretz: Opened https://github.com/temporalio/sdk-java/issues/2148

.github/workflows/ci.yml
Quinn-With-Two-Ns: Could you just move the `if` to ` Run cloud test`?

temporal-serviceclient/src/main/java/io/temporal/serviceclient/CloudServiceStubsOptions.java
cretz: > couldn't you set it in `validateAndBuildWithDefaults` if a channel is not set?

This is what I had originally but I struggled with the logic. How can you tell whether someone set `enableHttps` to know if it should be overridden, it's a primitive boolean. If you change it to `Boolean` in the builder to maintain a null/unset state that doesn't help if you are using the primitive in the actual options because most people use `build` and then we internally use the `newBuilder` override that accepts options to copy then validate and set defaults. Or do you want `Boolean` even in the options and just change the `getEnableHttps` to `return enableHttps != null && enableHttps`? I can do that if we want.

.github/workflows/ci.yml
cretz: Yes, though there's may be a lot of extra stuff for no reason that is tested elsewhere. But I don't mind moving it.

Quinn-With-Two-Ns: What extra tests would be run?

cretz: No extra tests, extra stuff. The job has 5 steps, but if I put the if around just 1, 4 other steps are running unnecessarily. But I can do that if we want that. Or I can move "Run cloud test" under one of the other jobs.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/CloudServiceStubsOptions.java
Quinn-With-Two-Ns: Wouldn't you leave `enableHttps` false in the constructor so it is only true if a user enabled it.

cretz: No, we want cloud client to default to HTTPS enabled. There's no reasonable situation where you'd use it without TLS (and even if there is, it's by no means a good default).

Quinn-With-Two-Ns: I don't know if ew do. If someone is choosing to not use `newCloudServiceStubs` then I would not set `HTTPS` by default. It isn't consistent with our other services and if a user is not using  `newCloudServiceStubs` it's hard to say how they are connecting.

cretz: They have to set other options almost always (e.g. API key). Arguably `newCloudServiceStubs` shouldn't even exist. Here's how the test creates a client and how users most likely will too (we hope the version requirement will be removed in the future):

```java
    CloudOperationsClient client =
        CloudOperationsClient.newInstance(
            CloudServiceStubs.newServiceStubs(
                CloudServiceStubsOptions.newBuilder()
                    .addApiKey(() -> apiKey)
                    .setVersion(apiVersion)
                    .build()));
```

Unlike connecting to workflow/operator service, here we know the https setting that will be used like 99% of the time.

.github/workflows/ci.yml
Quinn-With-Two-Ns: I think it is fine so we can make it block on merge. The other steps are very quick.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/CloudServiceStubsOptions.java
cretz: Only API key at this time (and maybe that will always be the case, unsure). Yes, we could _technically_ have overloads that accept params and use the builder for you (same for workflow service too where we know common things people need).

.github/workflows/ci.yml
cretz: :+1: Will do that (I'm assuming you're talking about keeping this as a separate job as opposed to just a separate step of an existing job).

cretz: Done

temporal-serviceclient/src/main/java/io/temporal/serviceclient/CloudServiceStubsOptions.java
Quinn-With-Two-Ns: My vote is `newCloudServiceStubs` has a required parameter for the `API` key and sets all the parameters and that would be the encouraged way to create the cloud service stub.

Quinn-With-Two-Ns (approved):

temporal-serviceclient/src/main/java/io/temporal/serviceclient/CloudServiceStubsOptions.java
cretz: We are not requiring API key in any cloud ops client in any SDK. That the cloud API currently requires an API key in its experimental form doesn't mean it always will. We are trying to have users set API keys the same way for clients since they are expected to be the same API keys. So if you use a certain way to create a workflow client in a language, you can apply that knowledge to creating a cloud client.

Quinn-With-Two-Ns: >We are not requiring API key in any cloud ops client in any SDK

I just think it should be required for `newCloudServiceStubs ` not the `CloudServiceStubIOptions`. I would just like the expected case, connecting to our cloud with an API key, to be as easy as possible. If the requirements change we can change the SDK API to reflect that. I don't think it is a big deal, hence I approved.

cretz: > I would just like the expected case, connecting to our cloud with an API key, to be as easy as possible.

This is going to start being the expected case with workflow service I think too (though there is self-hosted). I think we should consider making API keys as easy to use for everyone at the same time (even if it's used more often in one place than another).

PR #2151, opened 2024-07-24
Ensure identity copied to Builder from source WorkerOptions

dustin10:
## What was changed
Fixes a bug where the `identity` property was not being copied to the worker options `Builder` when it is created from an existing `WorkerOptions` instance.

## Why?
Could not properly set the identity of the worker when using the `WorkerOptions`, instead the identity falls back to the one from the workflow client options.

## Checklist

1. N/A
2. Tested via unit test.
3. N/A


Quinn-With-Two-Ns:
Looks like the feature tests failed because the tags are missing on your branch that we use to find the version of the SDK. You can specify --tags  when you fetch

dustin10:
@Quinn-With-Two-Ns tags should be updated now. Will add a test here in a bit.

dustin10:
Updated with simple test of `WorkerOptions.Builder` creating a new `WorkerOptions` from a source `WorkerOptions` with no modifications.

Quinn-With-Two-Ns (approved):

PR #2154, opened 2024-07-25
Miscellaneous docstring and comment edits

dandavison:
I only did this because I spotted one `s/signal/update/` typo, but then I felt I should do a bit more editing to make the PR a more respectable size.


cretz (approved):

Quinn-With-Two-Ns (approved):

Quinn-With-Two-Ns:
For CI to pass you need to rebase. Your code is to far behind the github actions

Issue #2155, opened 2024-07-27
WorkflowTask failure because of an invalid state transition in LocalActivity state machine


Quinn-With-Two-Ns:
The following test case should pass

```
public class LocalActivityAfterCancelTest {
  private final TestActivitiesImpl activitiesImpl = new TestActivitiesImpl();

  @Rule
  public SDKTestWorkflowRule testWorkflowRule =
      SDKTestWorkflowRule.newBuilder()
          .setUseExternalService(true)
          .setWorkflowTypes(TestLocalActivityRetry.class, BlockingWorkflow.class)
          .setActivityImplementations(activitiesImpl)
          .build();

  @Test
  public void localActivityAfterChildWorkflowCanceled() {
    TestWorkflow1 workflowStub =
        testWorkflowRule.newWorkflowStubTimeoutOptions(TestWorkflow1.class);
    WorkflowClient.execute(workflowStub::execute, "input");
    WorkflowStub.fromTyped(workflowStub).cancel();
    WorkflowFailedException exception =
        Assert.assertThrows(WorkflowFailedException.class, () -> workflowStub.execute("input"));
    Assert.assertEquals(
        EventType.EVENT_TYPE_WORKFLOW_EXECUTION_CANCELED, exception.getWorkflowCloseEventType());
  }

  @Test
  public void testLocalActivityAfterChildWorkflowCanceledReplay() {
    assertThrows(
        RuntimeException.class,
        () ->
            WorkflowReplayer.replayWorkflowExecutionFromResource(
                "testLocalActivityAfterCancelTest.json",
                LocalActivityAfterCancelTest.TestLocalActivityRetry.class));
  }

  @WorkflowInterface
  public static class BlockingWorkflow implements TestWorkflows.TestWorkflowReturnString {
    @Override
    public String execute() {
      Workflow.await(() -> false);
      return "";
    }
  }

  public static class TestLocalActivityRetry implements TestWorkflow1 {

    @Override
    public String execute(String taskQueue) {
      try {
        ChildWorkflowOptions childOptions =
            ChildWorkflowOptions.newBuilder()
                .setWorkflowId(Workflow.getInfo().getWorkflowId() + "-child1")
                .setCancellationType(ChildWorkflowCancellationType.WAIT_CANCELLATION_REQUESTED)
                .setParentClosePolicy(ParentClosePolicy.PARENT_CLOSE_POLICY_REQUEST_CANCEL)
                .validateAndBuildWithDefaults();
[...]

PR #2156, opened 2024-07-29
Fix transition in LA when handling canceled child wf

Quinn-With-Two-Ns:
Fix transition in LA when handling canceled child wf/

closes https://github.com/temporalio/sdk-java/issues/2155


temporal-sdk/src/main/java/io/temporal/internal/statemachines/LocalActivityStateMachine.java
```diff
@@ -121,6 +121,13 @@ enum State {
                   State.REQUEST_PREPARED,
                   LocalActivityStateMachine::sendRequest)
               .add(State.REQUEST_PREPARED, ExplicitEvent.MARK_AS_SENT, State.REQUEST_SENT)
+              // This is to cover am edge case where the event loop is
```
Sushisource, line 124:
```suggestion
              // This is to cover an edge case where the event loop is
```

temporal-sdk/src/test/java/io/temporal/internal/statemachines/LocalActivityStateMachineTest.java
```diff
@@ -343,4 +348,99 @@ protected void buildWorkflow(AsyncWorkflowBuilder<Void> builder) {
...
+                    .setActivityId("id1")
+                    .setActivityType(ActivityType.newBuilder().setName("activity1")),
+                null,
+                System.currentTimeMillis(),
+                null,
+                false,
+                null);
+        // TODO: This is a workaround for the lack of support for child workflow in the test
+        // framework.
+        // The test framework has not support for state machines with multiple callbacks.
```
Sushisource, line 373:
```suggestion
        // The test framework has no support for state machines with multiple callbacks.
```

Sushisource (approved):

PR #2159, opened 2024-07-30
Move workflow update polling inside of interceptor

Sushisource:
## What was changed
Moved update polling to be inside of the interceptor a-la other SDKs

## Why?
Consistent, allows interceptor to change things post-polling

## Checklist

1. Closes https://github.com/temporalio/sdk-java/issues/2094

2. How was this tested:
Added an integ test using the interceptor

3. Any docs updates needed?

### DIFF
--- a/temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowClientCallsInterceptor.java
+++ b/temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowClientCallsInterceptor.java
@@ -23,6 +23,7 @@ package io.temporal.opentracing.internal;
 import io.opentracing.Scope;
 import io.opentracing.Span;
 import io.opentracing.Tracer;
+import io.temporal.client.UpdateHandle;
 import io.temporal.common.interceptors.WorkflowClientCallsInterceptor;
 import io.temporal.common.interceptors.WorkflowClientCallsInterceptorBase;
 import io.temporal.opentracing.OpenTracingOptions;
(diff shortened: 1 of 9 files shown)


temporal-sdk/src/test/java/io/temporal/workflow/updateTest/UpdateTest.java
```diff
@@ -103,6 +107,55 @@ public void testUpdate() {
...
   }
 
+  private static class FakesResultUpdateInterceptor extends WorkflowClientInterceptorBase {
+    @Override
+    public WorkflowClientCallsInterceptor workflowClientCallsInterceptor(
+        WorkflowClientCallsInterceptor next) {
+      return new WorkflowClientCallsInterceptorBase(next) {
+        @Override
+        public <R> UpdateHandle<R> startUpdate(StartUpdateInput<R> input) {
+          super.startUpdate(input);
```
Sushisource, line 117: Side note: Why don't we have something like `InputType.fromExisting(input)` and then some way to relatively easily modify the input? It's a bit annoying to have to re-pass every field through to a new constructor if you want to modify inputs.

I get why they're not directly mutable, but, might be nice to have a convenient way to do that.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowClientCallsInterceptor.java
```diff
@@ -119,7 +120,7 @@ public <R> QueryOutput<R> query(QueryInput<R> input) {
   }
 
   @Override
-  public <R> StartUpdateOutput<R> startUpdate(StartUpdateInput<R> input) {
+  public <R> UpdateHandle<R> startUpdate(StartUpdateInput<R> input) {
```
Quinn-With-Two-Ns, line 123: Hm I wonder if we would ever need to return more then `UpdateHandle` here? I think other SDKs just return `UpdateHandle` so if we do will have to address in all SDKs

Quinn-With-Two-Ns (approved):
LGTM, thanks!

temporal-sdk/src/test/java/io/temporal/workflow/updateTest/UpdateTest.java
Quinn-With-Two-Ns: Modify inputs is not really an encouraged pattern. Not that some users don't but I it is a advanced path to take.

Issue #2068, opened 2024-05-17
Disallow continue as new in update handlers


cretz:
**Describe the solution you'd like**

Previously at https://github.com/temporalio/features/issues/379 it was decided to allow continue-as-new in update handlers. This decision has been reverted but Java was the only one that implemented it.  Should probably fail if the user tries to call it, but up to implementer here.

PR #2167, opened 2024-08-04
Disallow continue as new in update handlers

Quinn-With-Two-Ns:
closes https://github.com/temporalio/sdk-java/issues/2068


cretz (approved):

temporal-sdk/src/main/java/io/temporal/internal/statemachines/UnsupportedContinueAsNewRequest.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+package io.temporal.internal.statemachines;
+
+/**
+ * Thrown when {@link io.temporal.workflow.Workflow#continueAsNew} is called from a location that is
+ * not supported.
+ *
+ * <p>The reason this class extends Error is for application workflow code to not catch it by
+ * mistake. The default behavior of the SDK is to block workflow execution while Error is thrown.
```
Sushisource, line 28:
```suggestion
 * Thrown when {@link io.temporal.workflow.Workflow#continueAsNew} is called from an unsupported
 * location.
 *
 * <p>This class extends Error so application workflow code won't catch it by
 * mistake. The default behavior of the SDK is to block workflow execution while Error is thrown.
```

Sushisource (approved):

PR #2172, opened 2024-08-06
Wrap GRPC::CANCELED and DEADLINE_EXCEEDED in new exception type

Quinn-With-Two-Ns:
Wrap GRPC::CANCELED and DEADLINE_EXCEEDED in new exception type `WorkflowUpdateTimeoutOrCancelledException`

closes https://github.com/temporalio/sdk-java/issues/2069

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/client/WorkflowUpdateTimeoutOrCancelledException.java
+++ b/temporal-sdk/src/main/java/io/temporal/client/WorkflowUpdateTimeoutOrCancelledException.java
@@ -22,6 +22,12 @@ package io.temporal.client;
 
 import io.temporal.api.common.v1.WorkflowExecution;
 
+/**
+ *  Error that occurs when an update call times out or is cancelled.
+ *  
+ *  <p>Note, this is not related to any general concept of timing out or cancelling
+ *  a running update, this is only related to the client call itself.
+ */
 public class WorkflowUpdateTimeoutOrCancelledException extends WorkflowServiceException {
   public WorkflowUpdateTimeoutOrCancelledException(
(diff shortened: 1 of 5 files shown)


temporal-sdk/src/main/java/io/temporal/client/WorkflowUpdateTimeoutOrCancelledException.java
```diff
@@ -0,0 +1,36 @@
...
+import io.temporal.api.common.v1.WorkflowExecution;
+
+/**
+ * Error that occurs when an update call times out or is cancelled.
+ *
+ * <p>Note, this is not related to any general concept of timing out or cancelling a running update,
+ * this is only related to the client call itself.
+ */
+public class WorkflowUpdateTimeoutOrCancelledException extends WorkflowServiceException {
+  public WorkflowUpdateTimeoutOrCancelledException(
```
cretz, line 32: May be worth docs here stating this constructor is for internal use only and that the parameters may not be stable, but maybe not if we never expect to change or we can overload

cretz (approved):

temporal-sdk/src/main/java/io/temporal/client/WorkflowUpdateTimeoutOrCancelledException.java
Quinn-With-Two-Ns: I think it is fine as is, we do the same thing for all the other children of `WorkflowServiceException`. A user may construct it for mocking/testsing

cretz: Yeah, my only concern is that we can't ever change it, only overload it. But no big deal.

PR #2186, opened 2024-08-15
Activity slot test flake

Sushisource:
Don't use metrics here and instead rely on wrapper for counts

### DIFF
--- a/temporal-sdk/src/test/java/io/temporal/internal/worker/WorkflowSlotsSmallSizeTests.java
+++ b/temporal-sdk/src/test/java/io/temporal/internal/worker/WorkflowSlotsSmallSizeTests.java
@@ -228,7 +228,6 @@ public class WorkflowSlotsSmallSizeTests {
     workflow.unblock();
     for (int i = 0; i < 5; i++) {
       parallelSemRunning.acquire(2);
-      System.out.println("!!!!!!!! CHECKING");
       assertIntraWFTSlotCount(2);
       parallelSemBlocked.release(2);
     }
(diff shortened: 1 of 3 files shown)


temporal-sdk/src/test/java/io/temporal/internal/worker/WorkflowSlotsSmallSizeTests.java
```diff
@@ -124,16 +123,9 @@ private void assertWorkerSlotCount(int worker, int activity, int localActivity)
...
-        MetricsType.WORKER_TASK_SLOTS_AVAILABLE, getWorkerTags("WorkflowWorker"), worker);
-    // All slots should be available
-    reporter.assertGauge(
-        MetricsType.WORKER_TASK_SLOTS_AVAILABLE, getWorkerTags("ActivityWorker"), activity);
-    // All slots should be available
-    reporter.assertGauge(
-        MetricsType.WORKER_TASK_SLOTS_AVAILABLE,
-        getWorkerTags("LocalActivityWorker"),
-        localActivity);
+    assertEquals(worker, workflowTaskSlotSupplier.currentUsedCount.get());
```
Quinn-With-Two-Ns, line 126: Can we get rid of the sleep above?

Sushisource: :facepalm:

temporal-sdk/src/test/java/io/temporal/internal/worker/WorkflowSlotsSmallSizeTests.java
```diff
@@ -116,24 +106,9 @@ public void tearDown() {
...
-    // All slots should be available
-    reporter.assertGauge(
-        MetricsType.WORKER_TASK_SLOTS_AVAILABLE, getWorkerTags("ActivityWorker"), activity);
-    // All slots should be available
-    reporter.assertGauge(
-        MetricsType.WORKER_TASK_SLOTS_AVAILABLE,
-        getWorkerTags("LocalActivityWorker"),
-        localActivity);
+  private void assertCurrentUsedCount(int activity, int localActivity) {
+    assertEquals(activity, activityTaskSlotSupplier.currentUsedSet.size());
```
Sushisource, line 110: I removed the workflow slot check because when it was released could be pretty timing dependent, and that's better covered by other tests anyway.

Quinn-With-Two-Ns (approved):

PR #2190, opened 2024-08-15
Fix workflow start delay docs to say signal does not interrupt delay

Quinn-With-Two-Ns:
closes https://github.com/temporalio/sdk-java/issues/2184

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/client/WorkflowOptions.java
+++ b/temporal-sdk/src/main/java/io/temporal/client/WorkflowOptions.java
@@ -374,9 +374,8 @@ public final class WorkflowOptions {
     }
 
     /**
-     * Time to wait before dispatching the first workflow task.
-     *  A signal from signal with start will not trigger a workflow task. Cannot be set the same time
-     * as a CronSchedule.
+     * Time to wait before dispatching the first workflow task. A signal from signal with start will
+     * not trigger a workflow task. Cannot be set the same time as a CronSchedule.
      */
     public Builder setStartDelay(Duration startDelay) {
       this.startDelay = startDelay;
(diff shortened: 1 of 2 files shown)


temporal-sdk/src/test/java/io/temporal/internal/worker/WorkflowWorkerTest.java
```diff
@@ -182,11 +182,6 @@ public void concurrentPollRequestLockTest() throws Exception {
                 });
 
     assertTrue(worker.start());
-    // All slots should be available
```
Quinn-With-Two-Ns, line 185: This assert is not always true since it can race with the first poll request

temporal-sdk/src/main/java/io/temporal/client/WorkflowOptions.java
```diff
@@ -374,10 +374,8 @@ public Builder setDisableEagerExecution(boolean disableEagerExecution) {
     }
 
     /**
-     * Time to wait before dispatching the first workflow task. If the workflow gets a signal before
-     * the delay, a workflow task will be dispatched and the rest of the delay will be ignored. A
-     * signal from signal with start will not trigger a workflow task. Cannot be set the same time
-     * as a CronSchedule.
+     * Time to wait before dispatching the first workflow task. A signal from signal with start will
+     * not trigger a workflow task. Cannot be set the same time as a CronSchedule.
```
Sushisource, line 378:
```suggestion
     * Time to wait before dispatching the first workflow task. A signal included in signal with start will
     * not trigger a workflow task. Cannot be set the same time as a CronSchedule.
```

Sushisource (approved):
Nitpicky clarification need not be taken

temporal-sdk/src/main/java/io/temporal/client/WorkflowOptions.java
Quinn-With-Two-Ns: Going to not take, only because I think we will be removing this whole line about signal with start

Issue #2192, opened 2024-08-18
Missing Attempt value on MDC for logging purposes


javiercanillas:
**Is your feature request related to a problem? Please describe.**
Workflow or activity execution logs miss the Attempt number, just the same way they have the WorkflowId, ActivityId and so on.

**Describe the solution you'd like**
Add a new LoggerTag ATTEMPT and introduce its value on MDC like the others.

**Describe alternatives you've considered**
None besides the first one.

**Additional context**
None

javiercanillas:
Just make https://github.com/temporalio/sdk-java/pull/2193 about this.

PR #2193, opened 2024-08-18
Adding attempt number into MDC context for better diagnosing

javiercanillas:
## What was changed
Added a new LoggerTag for the Attempt number for LocalActivityWorker, ActivityWorker and WorkflowWorker.

## Why?
When diagnosing logs it's quite useful to have this value, specially when executions may overlap each other.

## Checklist

1. Closes https://github.com/temporalio/sdk-java/issues/2192

2. How was this tested:
Run tests locally, change is quite simple, I didn't do further testing.

3. Any docs updates needed?
I didn't find it required. Am I missing something?

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/logging/LoggerTag.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/logging/LoggerTag.java
@@ -37,4 +37,5 @@ public final class LoggerTag {
   public static final String WORKER_TYPE = "WorkerType";
   public static final String SIDE_EFFECT_ID = "SideEffectId";
   public static final String CHILD_WORKFLOW_ID = "ChildWorkflowId";
+  public static final String ATTEMPT = "Attempt";
 }
--- a/temporal-sdk/src/main/java/io/temporal/internal/worker/ActivityWorker.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/worker/ActivityWorker.java
@@ -247,6 +247,7 @@ final class ActivityWorker implements SuspendableWorker {
       MDC.put(LoggerTag.WORKFLOW_ID, pollResponse.getWorkflowExecution().getWorkflowId());
       MDC.put(LoggerTag.WORKFLOW_TYPE, pollResponse.getWorkflowType().getName());
       MDC.put(LoggerTag.RUN_ID, pollResponse.getWorkflowExecution().getRunId());
+      MDC.put(LoggerTag.ATTEMPT, Integer.toString(pollResponse.getAttempt()));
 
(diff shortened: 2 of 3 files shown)


temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
```diff
@@ -304,6 +304,7 @@ public void handle(WorkflowTask task) throws Exception {
       MDC.put(LoggerTag.WORKFLOW_ID, workflowExecution.getWorkflowId());
       MDC.put(LoggerTag.WORKFLOW_TYPE, workflowType);
       MDC.put(LoggerTag.RUN_ID, runId);
+      MDC.put(LoggerTag.ATTEMPT, Integer.toString(workflowTaskResponse.getAttempt()));
```
Quinn-With-Two-Ns, line 307: We should not include this tag on the workflow task attempt for `ATTEMPT` as it has a different meaning then the ATTEMPT for local or normal activities

javiercanillas: Ok, I will remove it. But what's the meaning of attempt when dealing with workflows? Do you have any documentation regarding this topic?

Sushisource (approved):
Might make sense to add to workflow task too

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
Quinn-With-Two-Ns: attempt here is the workflow task attempt not the attempt of the workflow

javiercanillas: in case of replay, that attempt will increase? Or it is related more to retries, like on Activities?

Quinn-With-Two-Ns: Yes, it is not like retries on activities

Quinn-With-Two-Ns:
Feature tests are failing because your branch does not include git tags. Please push them to your branch and let me know so I can re run CI.

javiercanillas:
@Quinn-With-Two-Ns done, created a `none` tag for this matter.

Quinn-With-Two-Ns:
@javiercanillas You need to ensure your branch has the tags of `master`. You can specify --tags when you fetch

javiercanillas:
@Quinn-With-Two-Ns this PR is not from a branch, it's from a forked repository. If you want I can try to open a branch on this repository (I don't think I have the privileges).

PR #2195, opened 2024-08-20
LocalActivityWorker#shutdown must shutdown slotQueue #2194

### LINKED ISSUE #2194: Thread `LocalActivitySlotSupplierQueue` prevent JVM from termination
# Steps to reproduce 

* use `spring.temporal.start-workers = false` or just don't start worker
* terminate application (`io.temporal.worker.WorkerFactory#shutdown`)  - `LocalActivitySlotSupplierQueue` remains working and blocks jvm from exiting

Reproduced on last available temporal version 1.25.0

# Possible explanation

* `io.temporal.internal.worker.LocalActivitySlotSupplierQueue#LocalActivitySlotSupplierQueue` created during initialization of `io.temporal.internal.worker.LocalActivityWorker#LocalActivityWorker`
* `io.temporal.internal.worker.LocalActivityWorker#activityAttemptTaskExecutor` became null until `start` called - we don't do this
* `io.temporal.internal.worker.LocalActivityWorker#shutdown` expect `activityAttemptTaskExecutor` to be not null and doesn't call  `slotQueue.shutdown()`

# Possible fix
* Since this thread created in constructor we should probably move `slotQueue.shutdown()` one line higher to stop this thread indepenendly of worker start call.
* Or initialize it in start, not in the constructor.

# Use cases 

We want to have support of "management commands" inside our applications. This is separate run mode, which requires disable of all external services (kafka, Temporal, grpc, etc).

nikoncode:
## What was changed
LocalActivityWorker init slotQueue in constructor (it creates new thread), but doesn't shutdown it.

## Why?
It blocks JVM from exiting.

## Checklist

1. Closes #2194

2. How was this tested:

3. Any docs updates needed?
No

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
@@ -697,8 +697,8 @@ final class LocalActivityWorker implements Startable, Shutdownable {
 
   @Override
   public CompletableFuture<Void> shutdown(ShutdownManager shutdownManager, boolean interruptTasks) {
+    slotQueue.shutdown();
     if (activityAttemptTaskExecutor != null && !activityAttemptTaskExecutor.isShutdown()) {
-      slotQueue.shutdown();
       return activityAttemptTaskExecutor
           .shutdown(shutdownManager, interruptTasks)
           .thenCompose(


temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
```diff
@@ -697,8 +697,8 @@ public boolean start() {
 
   @Override
   public CompletableFuture<Void> shutdown(ShutdownManager shutdownManager, boolean interruptTasks) {
+    slotQueue.shutdown();
```
Quinn-With-Two-Ns, line 700: I think the error here is the `slotQueue` is starting a thread in the constructor unconditionally. The `slotQueue` should only be started in the workers `start` if the local activity worker has any handles. 

cc @Sushisource

nikoncode: Pay attention that `awaitTermination` call this unconditionally.

nikoncode: I mention your suggestion in original issue as one of the resoutions.

But I more worry about other usages of `slotQueue` (`submitAttempt`, `submitANewExecution`). 

For me it's not so easy to make sure that they can't be called without `start`, because if they can we will introduce NPE.

Quinn-With-Two-Ns: `submitAttempt ` and `submitANewExecution ` cannot be called without start. Also `slotQueue` should still be created in the constructor, I am just saying the thread it creates should not be started until the worker is started.

nikoncode: Then, I think we can introduce multistep initialization here.
* constructor will initialize object, without starting a thread
* introduce start method which will start listen thread and will be called from worker start


WDYT?

Quinn-With-Two-Ns: Yes exactly

Sushisource: I'll add a test to make sure whatever we end up with has the fix. Thanks for your effort on this @nikoncode !

Sushisource:
Closing this in favor of https://github.com/temporalio/sdk-java/pull/2196

Thank you again!

PR #2196, opened 2024-08-20
Fix shutdown behavior of unstarted LA slot queue

Sushisource:
## What was changed
Fixed a possible hang when shutting down a never-started worker, as well as normalized the shutdown of this slot queue to look more like other things.

## Why?
Bugfix

## Checklist

1. Closes https://github.com/temporalio/sdk-java/issues/2194

2. How was this tested:
Added tests, and there are a number of existing tests covering shutdown with running LAs

3. Any docs updates needed?

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivitySlotSupplierQueue.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivitySlotSupplierQueue.java
@@ -23,14 +23,12 @@ package io.temporal.internal.worker;
 import io.temporal.worker.tuning.LocalActivitySlotInfo;
 import io.temporal.worker.tuning.SlotPermit;
 import io.temporal.workflow.Functions;
-import java.util.concurrent.PriorityBlockingQueue;
-import java.util.concurrent.Semaphore;
-import java.util.concurrent.TimeUnit;
+import java.util.concurrent.*;
 import javax.annotation.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-class LocalActivitySlotSupplierQueue {
+class LocalActivitySlotSupplierQueue implements Shutdownable {
   static final class QueuedLARequest {
     final boolean isRetry;
     final SlotReservationData data;
@@ -47,10 +45,11 @@ class LocalActivitySlotSupplierQueue {
   private final Semaphore newExecutionsBackpressureSemaphore;
   private final TrackingSlotSupplier<LocalActivitySlotInfo> slotSupplier;
   private final Functions.Proc1<LocalActivityAttemptTask> afterReservedCallback;
-  private final Thread queueThread;
+  private final ExecutorService queueThreadService;
   private static final Logger log =
       LoggerFactory.getLogger(LocalActivitySlotSupplierQueue.class.getName());
   private volatile boolean running = true;
+  private volatile boolean wasEverStarted = false;
 
   LocalActivitySlotSupplierQueue(
       TrackingSlotSupplier<LocalActivitySlotInfo> slotSupplier,
@@ -73,13 +72,13 @@ class LocalActivitySlotSupplierQueue {
               return 0;
             });
     this.slotSupplier = slotSupplier;
-    this.queueThread = new Thread(this::processQueue, "LocalActivitySlotSupplierQueue");
-    this.queueThread.start();
+    this.queueThreadService =
+        Executors.newSingleThreadExecutor(r -> new Thread(r, "LocalActivitySlotSupplierQueue"));
   }
 
   private void processQueue() {
     try {
-      while (running) {
+      while (running || !requestQueue.isEmpty()) {
         QueuedLARequest request = requestQueue.take();
         SlotPermit slotPermit;
         try {
@@ -102,9 +101,9 @@ class LocalActivitySlotSupplierQueue {
     }
   }
 
-  void shutdown() {
-    running = false;
-    queueThread.interrupt();
+  void start() {
+    wasEverStarted = true;
+    this.queueThreadService.submit(this::processQueue);
   }
 
   boolean waitOnBackpressure(@Nullable Long acceptanceTimeoutMs) throws InterruptedException {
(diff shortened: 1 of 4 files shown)


temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivitySlotSupplierQueue.java
```diff
@@ -134,4 +133,40 @@ void submitAttempt(SlotReservationData data, boolean isRetry, LocalActivityAttem
...
+  }
+
+  @Override
+  public boolean isTerminated() {
+    return queueThreadService.isTerminated();
+  }
+
+  @Override
+  public CompletableFuture<Void> shutdown(ShutdownManager shutdownManager, boolean interruptTasks) {
+    running = false;
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: Actually I retract this at the top level shutdown is synchronized so it should be fine.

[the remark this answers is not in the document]
Sushisource: It's a volatile so that part of things ought to be fine. But yeah good point I could catch the interrupt and then just drain the queue.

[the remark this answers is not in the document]
Quinn-With-Two-Ns: I still think it may be simpler to always interrupt

Quinn-With-Two-Ns (approved):

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivitySlotSupplierQueue.java
```diff
@@ -73,13 +72,13 @@ static final class QueuedLARequest {
               return 0;
             });
     this.slotSupplier = slotSupplier;
-    this.queueThread = new Thread(this::processQueue, "LocalActivitySlotSupplierQueue");
-    this.queueThread.start();
+    this.queueThreadService =
```
Quinn-With-Two-Ns, line 75: Note I believe this will always create a thread even if users are not using local activities. To be more proper we would only create the executor if the worker is started. Originally I said that wasn't needed with the thread impl., but with a Service it is different.

temporal-sdk/src/test/java/io/temporal/worker/LocalActivityWorkerNotStartedTest.java
```diff
@@ -0,0 +1,75 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+
+public class LocalActivityWorkerNotStartedTest {
+
+  @Rule
+  public SDKTestWorkflowRule testWorkflowRule =
+      SDKTestWorkflowRule.newBuilder()
+          .setWorkflowTypes(NothingWorkflowImpl.class)
+          .setWorkerOptions(WorkerOptions.newBuilder().setLocalActivityWorkerOnly(true).build())
```
Quinn-With-Two-Ns, line 41: Don't you have to register some activity types or else the local activity worker will never start?

Quinn-With-Two-Ns (changes requested):

temporal-sdk/src/test/java/io/temporal/worker/LocalActivityWorkerNotStartedTest.java
Quinn-With-Two-Ns: Or do you think we have good coverage of the started case from existing tests?

Quinn-With-Two-Ns: There are different scenarios right:
* No workers are started
* The workflow worker is started, but the local activity worker is not started

Sushisource: The whole point is he doesn't actually start - the test did repro the behavior before the change.

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivitySlotSupplierQueue.java
Sushisource: It does not create a thread until you submit the first task: https://stackoverflow.com/questions/43569407/when-does-executor-really-create-new-thread

temporal-sdk/src/test/java/io/temporal/worker/LocalActivityWorkerNotStartedTest.java
Sushisource: Other tests adequately cover the actual started case I believe

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivitySlotSupplierQueue.java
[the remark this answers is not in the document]
Sushisource: Actually you kinda want to keep this since otherwise you need a different way to know that you should _keep_ looping while alive, which ends up being effectively the same thing as storing this bool.

temporal-sdk/src/test/java/io/temporal/worker/LocalActivityWorkerNotStartedTest.java
Quinn-With-Two-Ns: Right I was saying there are two ways you can create a worker that doesn't start the local activity worker

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivitySlotSupplierQueue.java
Quinn-With-Two-Ns: Hmm OK that goes against my reading of the java docs, but source code is king

Quinn-With-Two-Ns (approved):

temporal-sdk/src/test/java/io/temporal/worker/LocalActivityWorkerNotStartedTest.java
Sushisource: I will add one with an activity registered too :+1:

nikoncode:
Thanks, I started thinking about Shutdownable interface too.

Only one question left when it will be released, at least approximately?

@Sushisource @Quinn-With-Two-Ns

Quinn-With-Two-Ns:
@nikoncode This patch is included in `1.25.1`

PR #2176, opened 2024-08-09
Implement test server support for sync Nexus operation commands

pdoerner:
## What was changed

Added test server support for Nexus operation commands, including sync completion, timeouts, failures, and retries. Added most of the stubs for implementing asynchronous operations in followup PRs.

The test server will only support worker targets, not external, so there is no added logic for making HTTP requests.
The implementation largely follows the same design as ActivityTasks.

PR #2198, opened 2024-08-22
Test server support for async Nexus operations

pdoerner:
## What was changed

Added test server support for async Nexus operations
Refactored cancellation logic
Revised timeout / error handling
New functional tests

See also first implementation PR: https://github.com/temporalio/sdk-java/pull/2176

### DIFF
--- a/temporal-test-server/src/main/java/io/temporal/internal/testservice/NexusOperationRef.java
+++ b/temporal-test-server/src/main/java/io/temporal/internal/testservice/NexusOperationRef.java
@@ -27,7 +27,7 @@ import java.io.*;
 import java.util.Objects;
 import javax.annotation.Nonnull;
 
-class NexusOperationRef {
+public class NexusOperationRef {
 
   @Nonnull private final ExecutionId executionId;
   private final long scheduledEventId;
@@ -94,7 +94,7 @@ class NexusOperationRef {
     }
   }
 
-  static NexusOperationRef fromBytes(ByteString serialized) {
+  public static NexusOperationRef fromBytes(ByteString serialized) {
     ByteArrayInputStream bin = new ByteArrayInputStream(serialized.toByteArray());
     DataInputStream in = new DataInputStream(bin);
     try {
--- a/temporal-test-server/src/main/java/io/temporal/internal/testservice/NexusOperationRef.java
+++ b/temporal-test-server/src/main/java/io/temporal/internal/testservice/NexusOperationRef.java
@@ -1,23 +1,3 @@
-/*
- * Copyright (C) 2022 Temporal Technologies, Inc. All Rights Reserved.
- *
- * Copyright (C) 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
- *
- * Modifications copyright (C) 2017 Uber Technologies, Inc.
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this material except in compliance with the License.
- * You may obtain a copy of the License at
- *
- *   http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
-
 package io.temporal.internal.testservice;
 
 import com.google.protobuf.ByteString;
@@ -31,46 +11,31 @@ public class NexusOperationRef {
 
   @Nonnull private final ExecutionId executionId;
   private final long scheduledEventId;
-  private final int attempt;
-  private final boolean isCancel;
 
   NexusOperationRef(
-      @Nonnull String namespace,
-      @Nonnull WorkflowExecution execution,
-      long scheduledEventId,
-      int attempt,
-      boolean isCancel) {
+      @Nonnull String namespace, @Nonnull WorkflowExecution execution, long scheduledEventId) {
     this(
         new ExecutionId(Objects.requireNonNull(namespace), Objects.requireNonNull(execution)),
-        scheduledEventId,
-        attempt,
-        isCancel);
+        scheduledEventId);
   }
 
   NexusOperationRef(
       @Nonnull String namespace,
       @Nonnull String workflowId,
       @Nonnull String runId,
-      long scheduledEventId,
-      int attempt,
-      boolean isCancel) {
+      long scheduledEventId) {
     this(
         namespace,
         WorkflowExecution.newBuilder()
             .setWorkflowId(Objects.requireNonNull(workflowId))
             .setRunId(Objects.requireNonNull(runId))
             .build(),
-        scheduledEventId,
-        attempt,
-        isCancel);
+        scheduledEventId);
   }
 
-  public NexusOperationRef(
-      @Nonnull ExecutionId executionId, long scheduledEventId, int attempt, boolean isCancel) {
+  public NexusOperationRef(@Nonnull ExecutionId executionId, long scheduledEventId) {
     this.executionId = Objects.requireNonNull(executionId);
     this.scheduledEventId = scheduledEventId;
-    this.attempt = attempt;
-    this.isCancel = isCancel;
   }
 
   public ExecutionId getExecutionId() {
@@ -81,15 +46,6 @@ public class NexusOperationRef {
     return scheduledEventId;
   }
 
-  public long getAttempt() {
-    return attempt;
-  }
-
-  public boolean isCancel() {
-    return isCancel;
-  }
-
-  /** Used for task tokens. */
   public ByteString toBytes() {
     try (ByteArrayOutputStream bout = new ByteArrayOutputStream();
         DataOutputStream out = new DataOutputStream(bout)) {
@@ -98,8 +54,6 @@ public class NexusOperationRef {
       out.writeUTF(execution.getWorkflowId());
       out.writeUTF(execution.getRunId());
       out.writeLong(scheduledEventId);
-      out.writeInt(attempt);
-      out.writeBoolean(isCancel);
       return ByteString.copyFrom(bout.toByteArray());
     } catch (IOException e) {
       throw Status.INTERNAL.withCause(e).withDescription(e.getMessage()).asRuntimeException();
@@ -107,17 +61,18 @@ public class NexusOperationRef {
   }
 
   public static NexusOperationRef fromBytes(ByteString serialized) {
-    ByteArrayInputStream bin = new ByteArrayInputStream(serialized.toByteArray());
+    return fromBytes(serialized.toByteArray());
+  }
+
+  public static NexusOperationRef fromBytes(byte[] serialized) {
+    ByteArrayInputStream bin = new ByteArrayInputStream(serialized);
     DataInputStream in = new DataInputStream(bin);
     try {
       String namespace = in.readUTF();
       String workflowId = in.readUTF();
       String runId = in.readUTF();
       long scheduledEventId = in.readLong();
-      int attempt = in.readInt();
-      boolean isCancel = in.readBoolean();
-      return new NexusOperationRef(
-          namespace, workflowId, runId, scheduledEventId, attempt, isCancel);
+      return new NexusOperationRef(namespace, workflowId, runId, scheduledEventId);
     } catch (IOException e) {
       throw Status.INVALID_ARGUMENT
           .withCause(e)
--- a/temporal-test-server/src/main/java/io/temporal/internal/testservice/NexusOperationRef.java
+++ b/temporal-test-server/src/main/java/io/temporal/internal/testservice/NexusOperationRef.java
@@ -27,13 +27,13 @@ import java.io.*;
 import java.util.Objects;
 import javax.annotation.Nonnull;
 
-class NexusTaskToken {
+class NexusOperationRef {
 
   @Nonnull private final ExecutionId executionId;
   private final long scheduledEventId;
   private final int attempt;
 
-  NexusTaskToken(
+  NexusOperationRef(
       @Nonnull String namespace,
       @Nonnull WorkflowExecution execution,
       long scheduledEventId,
@@ -44,7 +44,7 @@ class NexusTaskToken {
         attempt);
   }
 
-  NexusTaskToken(
+  NexusOperationRef(
       @Nonnull String namespace,
       @Nonnull String workflowId,
       @Nonnull String runId,
@@ -60,7 +60,7 @@ class NexusTaskToken {
         attempt);
   }
 
-  NexusTaskToken(@Nonnull ExecutionId executionId, long scheduledEventId, int attempt) {
+  NexusOperationRef(@Nonnull ExecutionId executionId, long scheduledEventId, int attempt) {
     this.executionId = Objects.requireNonNull(executionId);
     this.scheduledEventId = scheduledEventId;
     this.attempt = attempt;
@@ -94,7 +94,7 @@ class NexusTaskToken {
     }
   }
 
-  static NexusTaskToken fromBytes(ByteString serialized) {
+  static NexusOperationRef fromBytes(ByteString serialized) {
     ByteArrayInputStream bin = new ByteArrayInputStream(serialized.toByteArray());
     DataInputStream in = new DataInputStream(bin);
     try {
@@ -103,7 +103,7 @@ class NexusTaskToken {
       String runId = in.readUTF();
       long scheduledEventId = in.readLong();
       int attempt = in.readInt();
-      return new NexusTaskToken(namespace, workflowId, runId, scheduledEventId, attempt);
+      return new NexusOperationRef(namespace, workflowId, runId, scheduledEventId, attempt);
     } catch (IOException e) {
       throw Status.INVALID_ARGUMENT
           .withCause(e)
--- a/temporal-test-server/src/main/java/io/temporal/internal/testservice/NexusOperationRef.java
+++ b/temporal-test-server/src/main/java/io/temporal/internal/testservice/NexusOperationRef.java
@@ -1,3 +1,23 @@
+/*
+ * Copyright (C) 2022 Temporal Technologies, Inc. All Rights Reserved.
+ *
+ * Copyright (C) 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Modifications copyright (C) 2017 Uber Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this material except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package io.temporal.internal.testservice;
 
 import com.google.protobuf.ByteString;
--- a/temporal-test-server/src/main/java/io/temporal/internal/testservice/NexusOperationRef.java
+++ b/temporal-test-server/src/main/java/io/temporal/internal/testservice/NexusOperationRef.java
@@ -32,16 +32,19 @@ public class NexusOperationRef {
   @Nonnull private final ExecutionId executionId;
   private final long scheduledEventId;
   private final int attempt;
+  private final boolean isCancel;
 
   NexusOperationRef(
       @Nonnull String namespace,
       @Nonnull WorkflowExecution execution,
       long scheduledEventId,
-      int attempt) {
+      int attempt,
+      boolean isCancel) {
     this(
         new ExecutionId(Objects.requireNonNull(namespace), Objects.requireNonNull(execution)),
         scheduledEventId,
-        attempt);
+        attempt,
+        isCancel);
   }
 
   NexusOperationRef(
@@ -49,7 +52,8 @@ public class NexusOperationRef {
       @Nonnull String workflowId,
       @Nonnull String runId,
       long scheduledEventId,
-      int attempt) {
+      int attempt,
+      boolean isCancel) {
     this(
         namespace,
         WorkflowExecution.newBuilder()
@@ -57,13 +61,16 @@ public class NexusOperationRef {
             .setRunId(Objects.requireNonNull(runId))
             .build(),
         scheduledEventId,
-        attempt);
+        attempt,
+        isCancel);
   }
 
-  NexusOperationRef(@Nonnull ExecutionId executionId, long scheduledEventId, int attempt) {
+  public NexusOperationRef(
+      @Nonnull ExecutionId executionId, long scheduledEventId, int attempt, boolean isCancel) {
     this.executionId = Objects.requireNonNull(executionId);
     this.scheduledEventId = scheduledEventId;
     this.attempt = attempt;
+    this.isCancel = isCancel;
   }
 
   public ExecutionId getExecutionId() {
@@ -78,6 +85,10 @@ public class NexusOperationRef {
     return attempt;
   }
 
+  public boolean isCancel() {
+    return isCancel;
+  }
+
   /** Used for task tokens. */
   public ByteString toBytes() {
     try (ByteArrayOutputStream bout = new ByteArrayOutputStream();
@@ -88,6 +99,7 @@ public class NexusOperationRef {
       out.writeUTF(execution.getRunId());
       out.writeLong(scheduledEventId);
       out.writeInt(attempt);
+      out.writeBoolean(isCancel);
       return ByteString.copyFrom(bout.toByteArray());
     } catch (IOException e) {
       throw Status.INTERNAL.withCause(e).withDescription(e.getMessage()).asRuntimeException();
@@ -103,7 +115,9 @@ public class NexusOperationRef {
       String runId = in.readUTF();
       long scheduledEventId = in.readLong();
       int attempt = in.readInt();
-      return new NexusOperationRef(namespace, workflowId, runId, scheduledEventId, attempt);
+      boolean isCancel = in.readBoolean();
+      return new NexusOperationRef(
+          namespace, workflowId, runId, scheduledEventId, attempt, isCancel);
     } catch (IOException e) {
       throw Status.INVALID_ARGUMENT
           .withCause(e)
--- a/temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
+++ b/temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
@@ -44,6 +44,7 @@ import static io.temporal.internal.testservice.StateMachines.State.TERMINATED;
 import static io.temporal.internal.testservice.StateMachines.State.TIMED_OUT;
 
 import com.google.common.base.Preconditions;
+import com.google.common.base.Strings;
 import com.google.protobuf.*;
 import com.google.protobuf.util.Durations;
 import com.google.protobuf.util.Timestamps;
@@ -343,14 +344,13 @@ class StateMachines {
     // Timeout for an individual Start or Cancel Operation request.
     final java.time.Duration requestTimeout = java.time.Duration.ofSeconds(10);
 
-    String operationId;
+    String operationId = "";
     Endpoint endpoint;
     NexusOperationScheduledEventAttributes scheduledEvent;
     TestWorkflowStore.NexusTask nexusTask;
     RetryPolicy retryPolicy = defaultNexusRetryPolicy();
 
     long scheduledEventId = NO_EVENT_ID;
-    boolean isStarted = false;
 
     TestServiceRetryState retryState;
     long lastAttemptCompleteTime;
@@ -359,7 +359,6 @@ class StateMachines {
     String identity;
 
     public NexusOperationData(Endpoint endpoint) {
-      this.operationId = UUID.randomUUID().toString();
       this.endpoint = endpoint;
     }
 
@@ -736,11 +735,7 @@ class StateMachines {
                     .setScheduledEventId(data.scheduledEventId)
                     .setRequestId(data.scheduledEvent.getRequestId()))
             .build());
-    ctx.onCommit(
-        historySize -> {
-          data.operationId = resp.getOperationId();
-          data.isStarted = true;
-        });
+    ctx.onCommit(historySize -> data.operationId = resp.getOperationId());
   }
 
   private static void completeNexusOperation(
@@ -795,7 +790,7 @@ class StateMachines {
       RequestContext ctx, NexusOperationData data, Failure failure, long notUsed) {
     RetryState retryState = attemptNexusOperationRetry(ctx, Optional.of(failure), data);
     if (retryState == RetryState.RETRY_STATE_IN_PROGRESS) {
-      return (data.isStarted) ? STARTED : INITIATED;
+      return (Strings.isNullOrEmpty(data.operationId)) ? INITIATED : STARTED;
     }
 
     Failure wrapped =
--- a/temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
+++ b/temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
@@ -60,6 +60,7 @@ import io.temporal.api.failure.v1.ApplicationFailureInfo;
 import io.temporal.api.failure.v1.Failure;
 import io.temporal.api.failure.v1.TimeoutFailureInfo;
 import io.temporal.api.history.v1.*;
+import io.temporal.api.nexus.v1.CancelOperationRequest;
 import io.temporal.api.nexus.v1.Endpoint;
 import io.temporal.api.nexus.v1.StartOperationRequest;
 import io.temporal.api.nexus.v1.StartOperationResponse;
@@ -345,8 +346,6 @@ class StateMachines {
     TestWorkflowStore.NexusTask nexusTask;
 
     long scheduledEventId = NO_EVENT_ID;
-    // TODO(pj): consider refactoring cancellation into its own state machine as part of async work
-    boolean cancelRequested = false;
 
     TestServiceRetryState retryState;
     long lastAttemptCompleteTime;
@@ -608,16 +607,21 @@ class StateMachines {
     return new StateMachine<>(new NexusOperationData(endpoint))
         .add(NONE, INITIATE, INITIATED, StateMachines::scheduleNexusOperation)
         .add(INITIATED, START, STARTED, StateMachines::startNexusOperation)
-        .add(INITIATED, TIME_OUT, TIMED_OUT, StateMachines::timeoutNexusOperation)
+        // Transitions to initiated in case of a retry
+        .add(
+            INITIATED,
+            TIME_OUT,
+            new State[] {TIMED_OUT, INITIATED},
+            StateMachines::timeoutNexusOperation)
+        // Transitions directly to canceled if operation has not been started
         .add(
             INITIATED,
             REQUEST_CANCELLATION,
-            CANCELLATION_REQUESTED,
-            StateMachines::requestCancelNexusOperation)
+            CANCELED,
+            StateMachines::reportNexusOperationCancellation)
         // Transitions directly from INITIATED to COMPLETE for sync completions
         .add(INITIATED, COMPLETE, COMPLETED, StateMachines::completeNexusOperation)
         .add(STARTED, COMPLETE, COMPLETED, StateMachines::completeNexusOperation)
-        .add(CANCELLATION_REQUESTED, COMPLETE, COMPLETED, StateMachines::completeNexusOperation)
         // Transitions to initiated in case of a retry
         .add(STARTED, FAIL, new State[] {FAILED, INITIATED}, StateMachines::failNexusOperation)
         // Transitions to initiated in case of a retry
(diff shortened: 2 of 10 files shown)


temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
```diff
@@ -948,6 +879,38 @@ private static void requestCancelNexusOperation(
                     .setScheduledEventId(attr.getScheduledEventId())
                     .setWorkflowTaskCompletedEventId(workflowTaskCompletedId))
             .build());
+
+    TestServiceRetryState cancelRetryState =
+        new TestServiceRetryState(data.taskRetryPolicy, data.retryState.getExpirationTime());
+    NexusOperationRef ref =
+        new NexusOperationRef(
+            ctx.getExecutionId(), attr.getScheduledEventId(), cancelRetryState.getAttempt());
```
bergundy, line 887: Add something on the token to distinguish this task from a start task.

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -712,20 +711,28 @@ private void processScheduleNexusOperation(
...
-    StateMachine<StateMachines.NexusOperationData> operationStateMachine =
-        newNexusOperation(endpoint);
+    StateMachine<StateMachines.NexusOperationData> operation = newNexusOperation(endpoint);
     long scheduleEventId = ctx.getNextEventId();
-    nexusOperations.put(scheduleEventId, operationStateMachine);
+    nexusOperations.put(scheduleEventId, operation);
 
-    operationStateMachine.action(Action.INITIATE, ctx, attr, workflowTaskCompletedId);
+    operation.action(Action.INITIATE, ctx, attr, workflowTaskCompletedId);
+    // Use TIMEOUT_TYPE_SCHEDULE_TO_START for timing out an individual start request
```
bergundy, line 719: I see now why you're using schedule to start, that's a bit confusing since we'll add schedule to start support for the operation that's intended to time out from operation schedule time, not attempt schedule time.

bergundy: I'm not super familiar with the Java test server but the best way to timeout dispatching a task would be to put the timeout on the task itself but I don't know how that would interact with time skipping.

If you must use `ctx.addTimer`, consider using a `failNexusOperationAttempt` method to avoid overloading schedule to start.

bergundy: Make sure you also set the timeout on the task so the SDK can expose the task timeout to the operation handler.

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -712,20 +711,28 @@ private void processScheduleNexusOperation(
...
+        () ->
+            timeoutNexusOperation(
+                scheduleEventId,
+                TimeoutType.TIMEOUT_TYPE_SCHEDULE_TO_START,
+                operation.getData().getAttempt()),
+        "ScheduleNexusOperation ScheduleToStart");
     ctx.addTimer(
         ProtobufTimeUtils.toJavaDuration(
-            operationStateMachine.getData().scheduledEvent.getScheduleToCloseTimeout()),
+            operation.getData().scheduledEvent.getScheduleToCloseTimeout()),
```
bergundy, line 730: This isn't required, make sure you don't set a timer unless this timeout is set.

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -712,20 +711,28 @@ private void processScheduleNexusOperation(
...
+    StateMachine<StateMachines.NexusOperationData> operation = newNexusOperation(endpoint);
     long scheduleEventId = ctx.getNextEventId();
-    nexusOperations.put(scheduleEventId, operationStateMachine);
+    nexusOperations.put(scheduleEventId, operation);
 
-    operationStateMachine.action(Action.INITIATE, ctx, attr, workflowTaskCompletedId);
+    operation.action(Action.INITIATE, ctx, attr, workflowTaskCompletedId);
+    // Use TIMEOUT_TYPE_SCHEDULE_TO_START for timing out an individual start request
+    ctx.addTimer(
+        ProtobufTimeUtils.toJavaDuration(operation.getData().taskRetryPolicy.getInitialInterval()),
```
bergundy, line 721: The attempt timeout isn't related to the retry policy. In the real server we default to 10 seconds.

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -1580,6 +1600,22 @@ private void processContinueAsNewWorkflowExecution(
         parentChildInitiatedEventId);
   }
 
+  private void processWorkflowCompletionCallbacks(RequestContext ctx) {
```
bergundy, line 1603: Can you please add tests that confirm this is triggered for every terminal workflow state and that the callbacks will be carried over CaN and workflow retries?

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -2049,48 +2085,62 @@ private void timeoutActivity(long scheduledEventId, TimeoutType timeoutType, int
...
-          if (request.getResponse().hasCancelOperation()) {
-            operation.action(Action.CANCEL, ctx, request, 0);
-          } else {
-            operation.action(StateMachines.Action.COMPLETE, ctx, request, 0);
-          }
-          // nexusOperations.remove(scheduledEventId); // TODO(pj): server currently does not delete
+          StateMachine<NexusOperationData> operation =
+              getPendingNexusOperation(ref.getScheduledEventId());
+          throwIfOperationRefDoesntMatch(ref.toBytes(), operation.getData());
+          throwIfOperationNotInFlight(operation.getState());
```
bergundy, line 2106: You should ignore this instead of throwing (unless you're catching this somewhere else) but this is expected due to ordering issues and timeouts.

(Apply this here and everywhere else where you put these assertions)

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
```diff
@@ -798,16 +817,92 @@ public void respondNexusTaskFailed(
...
+        target.failNexusOperation(ref, timedOut);
+        break;
+      default:
+        throw Status.INTERNAL
+            .withDescription("invalid workflow execution status: " + completionEvent.getEventType())
+            .asRuntimeException();
+    }
+  }
+
+  private static Failure nexusFailureToApplicationFailure(
```
bergundy, line 884: Make sure that all failures recorded on events are nexus operation failures.
Sometimes you'll want application failures embedded in those but there should always be a nexus operation failure at the top level:

https://github.com/temporalio/temporal/blob/b383ffffcbbeacdfce2fe021c30f093bab64b5d9/components/nexusoperations/executors.go#L669

temporal-test-server/src/test/java/io/temporal/testserver/functional/NexusWorkflowTest.java
```diff
@@ -54,432 +54,617 @@
...
+                  completeNexusTask(
+                      task.getTaskToken(),
+                      Response.newBuilder()
+                          .setCancelOperation(CancelOperationResponse.getDefaultInstance())
+                          .build()))
+          .get();
+
+      // Poll to verify cancellation is recorded and triggers workflow progress.
+      pollResp = pollWorkflowTask();
+      testWorkflowRule.assertHistoryEvent(
```
bergundy, line 238: Please validate the failure here has the same structure as the one in the real server.

temporal-test-server/src/test/java/io/temporal/testserver/functional/NexusWorkflowTest.java
```diff
@@ -54,432 +54,617 @@
...
+      // Poll and verify started event is recorded and triggers workflow progress
+      pollResp = pollWorkflowTask();
+      testWorkflowRule.assertHistoryEvent(
+          execution.getWorkflowId(), EventType.EVENT_TYPE_NEXUS_OPERATION_STARTED);
+      completeWorkflowTask(pollResp.getTaskToken());
+
+      // Poll to wait for new task after operation times out
+      pollResp = pollWorkflowTask();
+      testWorkflowRule.assertHistoryEvent(
+          execution.getWorkflowId(), EventType.EVENT_TYPE_NEXUS_OPERATION_TIMED_OUT);
```
bergundy, line 349: Same here, please validate the failure is set as expected.

bergundy (commented):
You should add links in a follow up PR for completeness.

temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
```diff
@@ -611,37 +610,34 @@ public static StateMachine<NexusOperationData> newNexusOperation(Endpoint endpoi
     return new StateMachine<>(new NexusOperationData(endpoint))
         .add(NONE, INITIATE, INITIATED, StateMachines::scheduleNexusOperation)
         .add(INITIATED, START, STARTED, StateMachines::startNexusOperation)
-        .add(INITIATED, TIME_OUT, TIMED_OUT, StateMachines::timeoutNexusOperation)
+        // Transitions to INITIATED in case of a retry
+        .add(
+            INITIATED,
+            TIME_OUT,
+            new State[] {TIMED_OUT, INITIATED},
```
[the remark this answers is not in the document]
pdoerner: This is how retries are represented for all the other state machines in the test server, so I'd rather not break that pattern without a good reason.

temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
```diff
@@ -733,112 +729,60 @@ private static void scheduleNexusOperation(
...
-    }
-
-    ctx.addEvent(event);
-  }
-
   private static State timeoutNexusOperation(
       RequestContext ctx, NexusOperationData data, TimeoutType timeoutType, long notUsed) {
-    if (timeoutType != TimeoutType.TIMEOUT_TYPE_SCHEDULE_TO_CLOSE) {
+    if (timeoutType != TimeoutType.TIMEOUT_TYPE_SCHEDULE_TO_CLOSE
+        && timeoutType != TimeoutType.TIMEOUT_TYPE_SCHEDULE_TO_START) {
```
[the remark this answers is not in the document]
pdoerner: Yeah, was using it to timeout an individual attempt, but will refactor this a bit since we are planning to add schedule to start support in the future.

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
pdoerner: Yeah I don't think we can put the timer on the task itself. I will refactor this a bit to avoid overloading schedule to start.

Quinn-With-Two-Ns: Isn't the deadline set on a header https://github.com/temporalio/temporal/blob/c86712f795c168ebe7e73cca0a68849afa33d5ae/service/matching/matching_engine.go#L1759?

temporal-test-server/src/test/java/io/temporal/testserver/functional/NexusWorkflowTest.java
Quinn-With-Two-Ns: +1

temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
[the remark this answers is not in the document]
pdoerner: I ended up making `TIME_OUT` a terminal operation as you suggested. Attempt timeouts instead use a `FAIL` transition in order to schedule a retry.

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
pdoerner: I am still using `ctx.addTimer` but I am reusing the `failNexusOperation` to handle attempt retries instead of having a separate method.

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
pdoerner: I set the default to 5 seconds in the test server, mostly just to make the tests run a bit faster. If we think that's a bad experience, I am happy to switch it back to 10s.

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
pdoerner: I did add the tests for terminal states and CaN. The test server doesn't support workflow retries AFAICT, so I skipped that. I also skipped an explicit test for cron runs since it uses the same code path as CaN but I can always add something if you think it's necessary.

temporal-test-server/src/main/java/io/temporal/internal/testservice/NexusOperationRef.java
```diff
@@ -27,43 +27,50 @@
 import java.util.Objects;
 import javax.annotation.Nonnull;
 
-class NexusTaskToken {
+public class NexusOperationRef {
```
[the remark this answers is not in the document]
pdoerner: I was thinking it would contain more operation information, but didn't end up adding that so I could undo the rename.

temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
```diff
@@ -341,15 +340,17 @@ public int getAttempt() {
   }
 
   static final class NexusOperationData {
+    // Timeout for an individual Start or Cancel Operation request.
+    final java.time.Duration requestTimeout = java.time.Duration.ofSeconds(5);
```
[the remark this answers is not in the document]
pdoerner: Yeah I mentioned in another comment that this is mostly just so the tests will run faster since there is no way to configure it since we don't have dynamic configs. I will switch it back to 10s and just increase the test timeout to make it consistent.

temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
```diff
@@ -612,36 +613,23 @@ public static StateMachine<NexusOperationData> newNexusOperation(Endpoint endpoi
...
         .add(INITIATED, TIME_OUT, TIMED_OUT, StateMachines::timeoutNexusOperation)
+        // Transitions directly to CANCELED if operation has not been started
         .add(
             INITIATED,
             REQUEST_CANCELLATION,
-            CANCELLATION_REQUESTED,
-            StateMachines::requestCancelNexusOperation)
+            CANCELED,
+            StateMachines::reportNexusOperationCancellation)
+        .add(INITIATED, CANCEL, CANCELED, StateMachines::reportNexusOperationCancellation)
```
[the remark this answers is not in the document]
pdoerner: It doesn't at the moment. It probably wouldn't be too hard to add an overload for that but I think it's outside the scope of this PR.

temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
```diff
@@ -612,36 +613,23 @@ public static StateMachine<NexusOperationData> newNexusOperation(Endpoint endpoi
...
-        .add(
-            CANCELLATION_REQUESTED,
-            CANCEL,
-            CANCELED,
-            StateMachines::reportNexusOperationCancellation)
-        .add(CANCELLATION_REQUESTED, COMPLETE, COMPLETED, StateMachines::completeNexusOperation)
-        .add(CANCELLATION_REQUESTED, TIME_OUT, TIMED_OUT, StateMachines::timeoutNexusOperation)
-        .add(CANCELLATION_REQUESTED, FAIL, FAILED, StateMachines::failNexusOperation);
+        // Transitions back to STARTED in case of a retry
+        .add(STARTED, FAIL, new State[] {FAILED, STARTED}, StateMachines::failNexusOperation)
```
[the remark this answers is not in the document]
pdoerner: I prefer not to add too much to this state machine framework, especially if it is only used by Nexus.

temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
```diff
@@ -686,6 +676,7 @@ private static void scheduleNexusOperation(
             .setInput(attr.getInput())
             .setScheduleToCloseTimeout(attr.getScheduleToCloseTimeout())
             .putAllNexusHeader(attr.getNexusHeaderMap())
+            .putNexusHeader("Request-Timeout", String.valueOf(data.requestTimeout.toMillis()))
```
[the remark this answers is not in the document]
pdoerner: Yeah, `data.requestTimeout` is just the duration, it's not a timer. I put it there because it was convenient for the constant, but I could also put it somewhere else.

temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
```diff
@@ -733,75 +724,102 @@ private static void scheduleNexusOperation(
...
+            .setNexusOperationStartedEventAttributes(
+                NexusOperationStartedEventAttributes.newBuilder()
+                    .setOperationId(resp.getOperationId())
+                    .setScheduledEventId(data.scheduledEventId)
+                    .setRequestId(data.scheduledEvent.getRequestId()))
+            .build());
+    ctx.onCommit(
+        historySize -> {
+          data.operationId = resp.getOperationId();
+          data.isStarted = true;
```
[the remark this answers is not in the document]
pdoerner: `operationId` is generated when the state machine is created, but the operation hasn't been started until a worker picks up the nexus task and responds with an async success.

temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
```diff
@@ -733,75 +724,102 @@ private static void scheduleNexusOperation(
...
 
-  private static void handleSyncStartOperation(
-      RequestContext ctx, NexusOperationData data, StartOperationResponse response) {
-    if (response.hasSyncSuccess()) {
-      handleSyncSuccess(ctx, data, response.getSyncSuccess());
-    } else if (response.hasOperationError()) {
-      handleUnsuccessfulOperationError(ctx, data, response.getOperationError());
-    } else {
+  private static void timeoutNexusOperation(
+      RequestContext ctx, NexusOperationData data, TimeoutType timeoutType, long notUsed) {
```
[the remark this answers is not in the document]
pdoerner: Yeah, future proofing since we're already planning on at least StartToClose, no reason not to parameterize it now.

temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
```diff
@@ -733,75 +724,102 @@ private static void scheduleNexusOperation(
...
+    if (timeoutType != TimeoutType.TIMEOUT_TYPE_SCHEDULE_TO_CLOSE) {
       throw new IllegalArgumentException(
-          "Unable to process StartOperationResponse. Expected SyncSuccess or OperationError.");
+          "Timeout type not supported for Nexus operations: " + timeoutType);
     }
-  }
 
-  private static void handleSyncSuccess(
-      RequestContext ctx, NexusOperationData data, StartOperationResponse.Sync response) {
+    Failure failure =
```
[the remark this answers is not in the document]
pdoerner: I did check against the server and added test assertions for the expected fields, not sure if you're looking for some other validation or not. We don't support an external server yet, so it's tricky to compare against an actual server response.

temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
```diff
@@ -733,75 +724,102 @@ private static void scheduleNexusOperation(
...
                     .setOperation(data.scheduledEvent.getOperation())
                     .setOperationId(data.operationId)
                     .setScheduledEventId(data.scheduledEventId))
-            .setCause(nexusFailureToApplicationFailure(err.getFailure()))
+            .setCause(failure)
             .build();
 
     HistoryEvent event;
-    if (data.cancelRequested) {
+    if (data.nexusTask.getTask().getRequest().hasCancelOperation()) {
```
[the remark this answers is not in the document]
pdoerner: This is a check for whether the operation was requested to be canceled but hasn't been yet, not just for if it was canceled. If it was already in the `CANCELED` state we would've gotten an invalid transition error before this.

I was trying to emulate the logic from here: https://github.com/temporalio/temporal/blob/main/components/nexusoperations/completion.go#L101

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -2108,12 +2142,49 @@ private void timeoutNexusOperation(
...
+
+  private void timeoutNexusRequest(long scheduledEventId, String requestMethod, int attempt) {
+    boolean unlockTimer = true;
+    try {
+      update(
+          ctx -> {
+            StateMachine<NexusOperationData> operation = getPendingNexusOperation(scheduledEventId);
+            if (attempt != operation.getData().getAttempt()
+                || isTerminalState(operation.getState())) {
+              throw Status.NOT_FOUND.withDescription("Timer fired earlier").asRuntimeException();
```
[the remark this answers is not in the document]
pdoerner: NOT_FOUND errors are ignored: https://github.com/temporalio/temporal/blob/main/components/nexusoperations/completion.go#L101

temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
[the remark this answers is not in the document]
Quinn-With-Two-Ns: If the test time is proving to be a big problem we can add some concept of dynamic config to the test server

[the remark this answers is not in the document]
pdoerner: It's a little tricky to work with because when writing tests for timeouts, I have to work around the fact that long polls have a timeout of 10s and now so do the tasks, so I had to use sleeps to avoid a race condition where the poll would timeout first and fail the test.

If we are adding more such tests in the future, having some notion of dynamic config would be handy to keep the overall run time of the functional tests down, but I think it's okay for now.

temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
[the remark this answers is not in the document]
bergundy: Not following... operation ID is only set after the operation is started.

temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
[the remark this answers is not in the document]
bergundy: I'll do a pass on all failures that go into the history events.

temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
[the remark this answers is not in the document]
pdoerner: Ah I see the confusion, I was setting the operation ID internally on creation and then overwriting it for async operations. Fixed now so the operation ID is initially empty.

temporal-test-server/src/main/java/io/temporal/internal/testservice/NexusOperationRef.java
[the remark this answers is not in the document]
pdoerner: That being said, it is being used as a reference for completion callbacks which isn't exactly a task token. So I guess I will leave it as is unless there are strong feelings to the contrary.

temporal-test-server/src/test/java/io/temporal/testserver/functional/NexusWorkflowTest.java
```diff
@@ -54,432 +54,1039 @@
...
   @Before
-  public void checkExternal() {
+  public void setup() {
     // TODO: remove this skip once 1.25.0 is officially released and
     // https://github.com/temporalio/sdk-java/issues/2165 is resolved
     assumeFalse(
         "Nexus APIs are not supported for server versions < 1.25.0",
         testWorkflowRule.isUseExternalService());
+
+    testEndpoint = createEndpoint("nexus-workflow-test-endpoint");
```
Quinn-With-Two-Ns, line 76: Nit: make this a class constant

Do you clean up this endpoint?

temporal-test-server/src/test/java/io/temporal/testserver/functional/NexusWorkflowTest.java
```diff
@@ -54,432 +54,1039 @@
...
+  private Endpoint testEndpoint;
+  private final String testService = "test-service";
+  private final String testOperation = "test-operation";
 
   @Before
-  public void checkExternal() {
+  public void setup() {
     // TODO: remove this skip once 1.25.0 is officially released and
     // https://github.com/temporalio/sdk-java/issues/2165 is resolved
     assumeFalse(
```
Quinn-With-Two-Ns, line 72: If you remove this and set the test rules to use an external server. Assuming your server version supports nexus do these tests pass?

pdoerner: I don't think the async ones will pass because of how the callback is currently handled. For the test server, I'm just encoding the operation reference; for an external server we'd need to generate the correct HTTP URL.

Quinn-With-Two-Ns: Doesn't the URL come from the Server originally anyway ?

pdoerner: It comes from the Nexus handler. But because there is no Nexus client, for the tests I have to start the handler workflow manually and specify the callback in that start request.

temporal-test-server/src/main/java/io/temporal/internal/testservice/NexusOperationRef.java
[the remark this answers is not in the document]
bergundy: The callback token shouldn't include the number of attempts though

temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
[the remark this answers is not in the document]
bergundy: The important part is that the request timeout gives the actual request timeout, it should be in sync with the time the server is willing to wait until it times out the request.
There's two options: 
- If you start a timer to track this timeout (via `timeoutNexusRequest`) when the request is scheduled and a worker takes 4 seconds to pick up the request, the header should have a value of 6.
- If you `timeoutNexusRequest` when the worker picks it up, using 10 seconds for the header value is okay.

AFAICT, this is not the case. And I don't see where you reset the nexus request timeout on retry but I didn't read the code very thoroughly.

temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
```diff
@@ -712,7 +703,7 @@ private static void scheduleNexusOperation(
                             .setService(attr.getService())
                             .setOperation(attr.getOperation())
                             .setPayload(attr.getInput())
-                            .setCallback(""))); // TODO(pj): support async operations
+                            .setCallback(ref.toBytes().toString())));
```
bergundy, line 706: Can you please make this a URL and put data in the callback header? This is user visible and should look roughly like what we have in the real server.

In the Go test framework (which operates on a single workflow), I used this:

```go
					Callback: "http://test-env/operations",
					CallbackHeader: map[string]string{
						// The test client uses this to call resolveNexusOperation.
						"operation-sequence": strconv.FormatInt(h.seq, 10),
					},
```

temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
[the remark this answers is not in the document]
bergundy: That's not what the server code does. It checks if the operation completed as canceled.

temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
```diff
@@ -948,6 +885,33 @@ private static void requestCancelNexusOperation(
...
+    TestWorkflowStore.NexusTask cancelTask =
+        new TestWorkflowStore.NexusTask(taskQueueId, pollResponse);
+
+    // Test server only supports worker targets, so just push directly to Nexus task queue without
+    // invoking Nexus client.
+    ctx.addNexusTask(cancelTask);
+    ctx.onCommit(historySize -> data.nexusTask = cancelTask);
   }
 
   private static void reportNexusOperationCancellation(
```
bergundy, line 917: You should be setting a failure on the event here.

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -1585,6 +1585,22 @@ private void processContinueAsNewWorkflowExecution(
...
+  private void processWorkflowCompletionCallbacks(RequestContext ctx) {
+    Optional<HistoryEvent> completionEvent = getCompletionEvent(ctx.getEvents());
+    if (!completionEvent.isPresent()) {
+      return;
+    }
+
+    for (Callback cb : startRequest.getCompletionCallbacksList()) {
+      if (!cb.hasNexus()) {
+        // test server only supports nexus callbacks currently
+        log.warn("skipping non-nexus completion callback");
```
bergundy, line 1597: This should just be an invalid argument instead of this log that will get lost and leave users confused.

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -2074,48 +2090,68 @@ private void timeoutActivity(long scheduledEventId, TimeoutType timeoutType, int
...
-  public void failNexusTask(long scheduledEventId, RespondNexusTaskFailedRequest request) {
+  public void completeNexusOperation(NexusOperationRef ref, Payload result) {
     update(
         ctx -> {
-          StateMachine<NexusOperationData> operation = getPendingNexusOperation(scheduledEventId);
-          throwIfOperationTokenDoesntMatch(request.getTaskToken(), operation.getData());
-          operation.action(StateMachines.Action.FAIL, ctx, request, 0);
+          StateMachine<NexusOperationData> operation =
+              getPendingNexusOperation(ref.getScheduledEventId());
+          if (!operationRefMatches(ref, operation.getData())) {
```
bergundy, line 2127: When completing an operation asynchronously, the completion token shouldn't include an attempt or cancelation info. It's just a reference to the operation state machine (namespace, workflow ID, run ID, scheduled event ID).
I would consider splitting the concept of a task token and an operation reference because those have different semantics.
A task token is valid for a single attempt, while the reference is valid for the lifetime of the operation.

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -2157,17 +2230,15 @@ private void timeoutNexusOperation(
...
 
-  private void addNexusOperationRetryTimer(
-      RequestContext ctx, StateMachine<NexusOperationData> operation) {
+  private void retryNexusTask(RequestContext ctx, StateMachine<NexusOperationData> operation) {
+    State prevState = operation.getState();
     NexusOperationData data = operation.getData();
     int attempt = data.getAttempt();
-    Duration nextDelay = ProtobufTimeUtils.toJavaDuration(data.nextBackoffInterval);
-    data.nextAttemptScheduleTime = clock.getAsLong() + nextDelay.toMillis();
     ctx.addTimer(
```
bergundy, line 2237: I see there's a TODO below, can you address it?

bergundy: Also not seeing where the request timeout is enforced on the retry here (mentioned elsewhere).

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
```diff
@@ -777,14 +782,30 @@ public void respondNexusTaskCompleted(
...
-        mutableState.startNexusTask(taskToken.getScheduledEventId(), request);
+      NexusOperationRef ref = NexusOperationRef.fromBytes(request.getTaskToken());
+      TestWorkflowMutableState mutableState = getMutableState(ref.getExecutionId());
+      if (request.getResponse().hasCancelOperation()) {
+        mutableState.cancelNexusOperation(ref);
+      } else if (request.getResponse().hasStartOperation()) {
+        StartOperationResponse startResp = request.getResponse().getStartOperation();
+        if (startResp.hasOperationError()) {
+          Failure failure = unsuccessfulOperationErrorToFailure(startResp.getOperationError());
+          mutableState.failNexusOperation(ref, failure);
```
bergundy, line 793: If the operation completed as canceled the machine should transition to CANCELED state not FAILED.

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
```diff
@@ -777,14 +782,30 @@ public void respondNexusTaskCompleted(
...
+        } else if (startResp.hasAsyncSuccess()) {
+          // Start event is only recorded for async success
+          mutableState.startNexusOperation(
+              ref.getScheduledEventId(), request.getIdentity(), startResp.getAsyncSuccess());
+        } else if (startResp.hasSyncSuccess()) {
+          mutableState.completeNexusOperation(ref, startResp.getSyncSuccess().getPayload());
```
bergundy, line 799: Here you want to use the attempts and cancel state to verify the task token (which you do, just clarifying that this is different from async completion).

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
```diff
@@ -798,16 +819,106 @@ public void respondNexusTaskFailed(
...
+        // rest of the payloads. Additionally, even if a workflow could return more than a single
+        // value,
+        // Nexus does not support it.
+        Payload p =
+            (result.getPayloadsCount() > 0) ? result.getPayloads(0) : Payload.getDefaultInstance();
+        target.completeNexusOperation(ref, p);
+        break;
+      case EVENT_TYPE_WORKFLOW_EXECUTION_FAILED:
+        target.failNexusOperation(
+            ref, completionEvent.getWorkflowExecutionFailedEventAttributes().getFailure());
```
bergundy, line 856: In the real server, we translate this failure to a minimal representation, stripping out the causes.
This may change in the real server but I would stay consistent for now.

bergundy (commented):
Didn't have that much to add. Still not ready to merge even though it's fairly close.

temporal-test-server/src/test/java/io/temporal/testserver/functional/NexusWorkflowTest.java
pdoerner: It depends on the namespace + task queue getting set up for the suite so I was getting an NPE when trying to make it a constant.

I don't clean it up but I didn't think that would matter for in-memory tests?

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
pdoerner: I added validation on the start request to throw invalid argument if there are any non-nexus callbacks. I like leaving this as a warning in case someone still somehow gets a non-nexus callback in so that we don't fail the workflow.

temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
```diff
@@ -882,20 +867,18 @@ private static void requestCancelNexusOperation(
...
                 NexusOperationCancelRequestedEventAttributes.newBuilder()
-                    .setScheduledEventId(attr.getScheduledEventId())
+                    .setScheduledEventId(data.scheduledEventId)
                     .setWorkflowTaskCompletedEventId(workflowTaskCompletedId))
             .build());
 
-    NexusOperationRef ref =
-        new NexusOperationRef(
-            ctx.getExecutionId(), attr.getScheduledEventId(), data.getAttempt(), true);
+    NexusTaskToken ref =
```
bergundy, line 874: call this `token`?

bergundy (approved):

temporal-test-server/src/test/java/io/temporal/testserver/functional/NexusWorkflowTest.java
Quinn-With-Two-Ns: Sorry I meant make `"nexus-workflow-test-endpoint"` a constant

Well when we run on the real server https://temporalio.atlassian.net/browse/SDK-2560 it will matter right, or is the plan to address clean up as part of https://temporalio.atlassian.net/browse/SDK-2560.

Quinn-With-Two-Ns (approved):

PR #2199, opened 2024-08-27
Update-with-Start operation

stephanos:
## What was changed

Adds support for Update-with-Start, using the MultiOperation API (https://github.com/temporalio/api/pull/367).

## Why?

To allow Update-with-Start feature.


temporal-sdk/src/main/java/io/temporal/client/WorkflowStartOperationUpdate.java
```diff
@@ -0,0 +1,255 @@
...
+  }
+
+  public UpdateHandle<R> get() throws ExecutionException, InterruptedException {
+    return handle.get();
+  }
+
+  // TODO: hide from user?
+  public void complete(UpdateHandle<R> updateHandle) {
+    this.handle.complete(updateHandle);
+  }
```
stephanos, line 199: I don't like that this is accessible to the user, but I haven't found a way to hide it from them yet.

temporal-sdk/src/main/java/io/temporal/client/WorkflowStartOperationUpdate.java
```diff
@@ -0,0 +1,254 @@
...
+  public static <T, A1, A2, A3, A4, A5> Builder<Void> newBuilder(
+      Functions.Proc6<T, A1, A2, A3, A4, A5> update, A1 arg1, A2 arg2, A3 arg3, A4 arg4, A5 arg5) {
+    return WorkflowStartOperationUpdate.createBuilder(
+        update, new Object[] {arg1, arg2, arg3, arg4, arg5});
+  }
+
+  public static <R> Builder<R> newBuilder(String update, Class<R> resultType, Object[] args) {
+    return new Builder<>(
+        UpdateOptions.<R>newBuilder().setResultClass(resultType).setUpdateName(update), null, args);
+  }
```
stephanos, line 105: This is the untyped API.

temporal-sdk/src/main/java/io/temporal/client/WorkflowStartOperationUpdate.java
```diff
@@ -0,0 +1,254 @@
...
+import io.temporal.workflow.UpdateMethod;
+import java.lang.invoke.MethodHandleInfo;
+import java.lang.invoke.SerializedLambda;
+import java.lang.reflect.Method;
+import java.util.Arrays;
+import java.util.Objects;
+import java.util.concurrent.*;
+
+public class WorkflowStartOperationUpdate<R>
+    implements StartWorkflowOperation, Future<UpdateHandle<R>> {
```
stephanos, line 34: Implements `Future` directly to avoid an extra method call.

temporal-sdk/src/main/java/io/temporal/client/WorkflowStartOperationUpdate.java
```diff
@@ -0,0 +1,254 @@
...
+      }
+    } catch (ClassNotFoundException e) {
+      throw new IllegalArgumentException("Invalid method reference", e);
+    }
+
+    return new Builder<>(
+        UpdateOptions.<R>newBuilder().setUpdateName(updateName).setResultClass(resultClass),
+        workflowType,
+        args);
+  }
```
stephanos, line 147: This was born out of necessity. I could not find a way to do this where the typed stub is available; as that has no access to the workflow options.

temporal-sdk/src/test/java/io/temporal/workflow/updateTest/UpdateWithStartTest.java
```diff
@@ -0,0 +1,313 @@
...
+
+  private static void ensureNoWorkflowStarted(WorkflowClient workflowClient, String workflowId) {
+    try {
+      workflowClient.fetchHistory(workflowId);
+      fail("unreachable");
+    } catch (StatusRuntimeException e) {
+      assertEquals(e.getStatus().getCode(), Status.NOT_FOUND.getCode());
+    }
+  }
+
```
stephanos, line 268: I wanted to write a test for a failed server-side validation of the Update, but couldn't find any validation in the test server to leverage.

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
```diff
@@ -86,22 +92,50 @@ public WorkflowStartOutput start(WorkflowStartInput input) {
...
+
+        if (!updateOperation.getWorkflowType().isEmpty()
+            && !updateOperation.getWorkflowType().equals(input.getWorkflowType())) {
+          throw new IllegalArgumentException(
+              "Update Operation does not belong to '"
+                  + input.getWorkflowType()
+                  + "' but '"
+                  + updateOperation.getWorkflowType()
+                  + "'");
+        }
```
stephanos, line 118: This is not 100% bullet proof but might be good enough. I couldn't get to a place where I had the Update and the Start's full Java class available for comparison. Using the workflow type seemed like the next best thing.

temporal-sdk/src/test/java/io/temporal/client/functional/UpdateTest.java
```diff
@@ -167,6 +169,64 @@ public void updateWorkflowDuplicateId() throws ExecutionException, InterruptedEx
...
+    assertEquals(execution1.getRunId(), execution2.getRunId());
+
+    UpdateHandle<String> updHandle2 = update2.get();
+    assertEquals(workflowId, updHandle2.getExecution().getWorkflowId());
+    assertEquals(execution2.getRunId(), updHandle2.getExecution().getRunId());
+    assertEquals("Hello Update 2", updHandle2.getResultAsync().get());
+
+    workflowStub2.update("complete", void.class);
+    assertEquals("complete", workflowStub2.getResult(String.class));
+  }
```
stephanos, line 228: ^ This is testing the untyped API.

temporal-sdk/src/main/java/io/temporal/client/WorkflowStartOperationUpdate.java
```diff
@@ -0,0 +1,258 @@
...
```
stephanos, line 147: ^ This was born out of necessity. I could not find a way to do this where the typed stub is available; as that has no access to the workflow options.

temporal-sdk/src/main/java/io/temporal/client/WorkflowStartOperationUpdate.java
```diff
@@ -0,0 +1,258 @@
...
+    String updateName = null;
+    String workflowType = null;
+    Class<R> resultClass = null;
+    String updateMethodName = meta.getImplMethodName();
+    try {
+      Class<?> workflowClass = Class.forName(meta.getImplClass().replace('/', '.'));
+      workflowType = workflowClass.getSimpleName();
+      Method[] allMethods = workflowClass.getDeclaredMethods();
+      for (Method method : allMethods) {
+        if (method.getName().equals(updateMethodName)) {
```
cretz, line 122: Can you confirm in the SDK that we disallow overloads for update methods? Meaning can I have:

```java
@UpdateMethod
void foo(String bar);

@UpdateMethod(name="some-other-name")
void foo(Integer bar);
```

Otherwise you may need a more advanced signature check (or maybe this code is used elsewhere in Java SDK and you just copied it at which point nevermind)

temporal-sdk/src/main/java/io/temporal/client/WorkflowStartOperationUpdate.java
```diff
@@ -0,0 +1,258 @@
...
+      throw new IllegalArgumentException("Invalid method reference: " + updateFunc);
+    }
+
+    String updateName = null;
+    String workflowType = null;
+    Class<R> resultClass = null;
+    String updateMethodName = meta.getImplMethodName();
+    try {
+      Class<?> workflowClass = Class.forName(meta.getImplClass().replace('/', '.'));
+      workflowType = workflowClass.getSimpleName();
```
cretz, line 119: Mentioned in later comment how simple class name check is not acceptable for this use case

temporal-sdk/src/main/java/io/temporal/client/WorkflowStartOperationUpdate.java
```diff
@@ -0,0 +1,258 @@
...
+    return WorkflowStartOperationUpdate.createBuilder(
+        update, new Object[] {arg1, arg2, arg3, arg4, arg5});
+  }
+
+  public static <R> Builder<R> newBuilder(String update, Class<R> resultType, Object[] args) {
+    return new Builder<>(
+        UpdateOptions.<R>newBuilder().setResultClass(resultType).setUpdateName(update), null, args);
+  }
+
+  private static <R> Builder<R> createBuilder(Object updateFunc, Object[] args) {
```
cretz, line 107: I'm not sure this logic should be in builder creation but rather use. Most (all?) of our builders allow you to recreate the builder from the thing being built, but this builder does not support that. There are a few ways this can be done if we want to pursue this.

temporal-sdk/src/main/java/io/temporal/client/WorkflowStartOperationUpdate.java
```diff
@@ -0,0 +1,258 @@
...
+        + options
+        + ", args="
+        + Arrays.toString(args)
+        + ", handle="
+        + handle
+        + '}';
+  }
+
+  public static final class Builder<R> {
+    private final UpdateOptions.Builder<R> options;
```
cretz, line 234: The current builder setup prevents users from writing something like:

```java
<R> void applyDefaultUpdateOptions(UpdateOptions.Builder<R> builder) { ... }
```

That they can use for this update and other updates. This is probably not a big deal, but at least in Go they can write this helper. I just want to make sure we acknowledge that, hopefully unlike our other SDKs, there is no shared options building reuse in Java, and we'll have to make sure every future option we add to the `UpdateOptions` that applies to update with start gets an associated setter here.

temporal-sdk/src/main/java/io/temporal/client/WorkflowStartOperationUpdate.java
```diff
@@ -0,0 +1,258 @@
...
+        update, new Object[] {arg1, arg2, arg3, arg4});
+  }
+
+  public static <T, A1, A2, A3, A4, A5> Builder<Void> newBuilder(
+      Functions.Proc6<T, A1, A2, A3, A4, A5> update, A1 arg1, A2 arg2, A3 arg3, A4 arg4, A5 arg5) {
+    return WorkflowStartOperationUpdate.createBuilder(
+        update, new Object[] {arg1, arg2, arg3, arg4, arg5});
+  }
+
+  public static <R> Builder<R> newBuilder(String update, Class<R> resultType, Object[] args) {
```
cretz, line 102: Other builders allow you to create an empty builder and set everything later. There are some ways we can do this.

For the most part builders accept simple values without much logic. So a builder for such a thing here might accept an update name + result type tuple (really just two fields with mutual exclusivity checks) or an update method reference `Object`.

For example:

```java
public static class Builder<R> {
  // Can only either have the first two or the third
  private String updateName;
  private Class<R> updateResultClass;
  private Object updateMethodReference;

  private Object[] updateArgs;
  private String updateId;
  private WorkflowUpdateStage waitForStage;

  private Builder() { }

  private Builder(WorkflowStartOperationUpdate existing) {
    // Copy all 6 fields because they are on that update object verbatim
  }

  public Builder<R> setUpdate(String updateName, Class<R> updateResultClass) {
    this.updateName = updateName;
    this.updateResultClass = updateResultClass;
    updateMethodReference = null;
    return this;
  }

  public Builder<R> setUpdate(Object updateMethodReference) {
    this.updateMethodReference = updateMethodReference;
    updateName = null;
    updateResultClass = null;
  }

  // Setters for the other three fields
}
```

You can/should still have all of your `newBuilder` helpers, but I think this is a vastly simpler approach and allows you to match our builder pattern (can create empty, can create from existing object, etc), and allows you to late-bind the validation. Also it removes the `UpdateOptions` which, if we're not going to expose as is, there's no need building until needed internally.

temporal-sdk/src/main/java/io/temporal/client/WorkflowOptions.java
```diff
@@ -450,6 +467,8 @@ public WorkflowOptions validateBuildWithDefaults() {
 
   private final boolean disableEagerExecution;
 
+  private final StartWorkflowOperation startOperation;
```
Quinn-With-Two-Ns, line 470: You need to update the hash, equals and toString method as well

temporal-sdk/src/main/java/io/temporal/client/WorkflowStartOperationUpdate.java
```diff
@@ -0,0 +1,258 @@
...
+  }
+
+  @Override
+  public UpdateHandle<R> get(long timeout, TimeUnit unit)
+      throws InterruptedException, ExecutionException, TimeoutException {
+    return handle.get(timeout, unit);
+  }
+
+  @Override
+  public boolean equals(Object o) {
```
cretz, line 207: I am not sure it is safe to exclude the future in the equals/hash. In fact, I wonder if we should just not implement these and let them remain identity equality. I should be able to have a `Set<WorkflowStartOperationUpdate>` from separate start calls and not have two update operation instances be considered the same. IMO there is no value in explicit equals/hashCode for this class.

temporal-sdk/src/main/java/io/temporal/client/WorkflowStartOperationUpdate.java
Quinn-With-Two-Ns: Yeah the other methods on the `Future` interface will cause confusion and restrict our design in the future

temporal-sdk/src/main/java/io/temporal/client/WorkflowStartOperationUpdate.java
```diff
@@ -0,0 +1,240 @@
...
+
+package io.temporal.client;
+
+import io.temporal.api.common.v1.WorkflowExecution;
+import io.temporal.workflow.Functions;
+import java.lang.reflect.Type;
+import java.util.Arrays;
+import java.util.concurrent.*;
+
+/** TBD */
```
stephanos, line 29: I still owe those here. Will craft these once the overall design is acceptable.

temporal-sdk/src/main/java/io/temporal/client/WorkflowStartOperationUpdate.java
stephanos: I wasn't 100% sure if the new builder fulfills all the suggestions here; so I left this comment unresolved.

re `UpdateOptions`: I'm inclined to re-use it here; for example, because it allows users to obtain the updateId before executing the request via `.getOptions().getUpdateId()`.

temporal-sdk/src/test/java/io/temporal/workflow/updateTest/UpdateWithStartTest.java
```diff
@@ -0,0 +1,464 @@
...
+            .setWaitForStage(WorkflowUpdateStage.COMPLETED)
+            .build();
+
+    WorkflowUpdateHandle<String> handle1 =
+        WorkflowClient.startWithOperation(workflow::execute, updateOp);
+    assertEquals(options.getWorkflowId(), handle1.getExecution().getWorkflowId());
+    assertEquals("Hello Update", handle1.getResultAsync().get());
+
+    WorkflowUpdateHandle<String> updHandle = updateOp.getUpdateHandle().get();
+    assertEquals(updateOp.getResult(), updHandle.getResultAsync().get());
```
stephanos, line 74: I wonder if the `WorkflowUpdateHandle` should also have `getResult`()?

temporal-sdk/src/main/java/io/temporal/client/StartWorkflowAdditionalOperation.java
```diff
@@ -0,0 +1,36 @@
...
+ * limitations under the License.
+ */
+
+package io.temporal.client;
+
+import io.temporal.workflow.Functions;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+/** Base class for operations that can be added to a Start Workflow. */
+public abstract class StartWorkflowAdditionalOperation<R> {
```
cretz, line 27: If we are using the term "Additional" here can we also use that term in Go? I think it's reasonable for us to call this concept an "additional operation" to start every where when we document/discuss it.

temporal-sdk/src/main/java/io/temporal/client/WorkflowOptions.java
```diff
@@ -112,6 +110,8 @@ public static final class Builder {
 
     private Duration startDelay;
 
+    private StartWorkflowAdditionalOperation startOperation;
```
cretz, line 113:
```suggestion
    private StartWorkflowAdditionalOperation additionalOperation;
```

I think this makes more sense as "additional" instead of "start" considering everything on this this builder/options is implied to be for start, but no strong opinion here.

temporal-sdk/src/test/java/io/temporal/workflow/updateTest/UpdateWithStartTest.java
cretz: > I wonder if the WorkflowUpdateHandle should also have getResult()?

Would there be a meaningful difference between `getResult()` and `getResultAsync()` besides that the former just waits on the future of the latter? Also, @Quinn-With-Two-Ns are we sure we want to suffix the word `Async` to `getResult`? And if we do, should `WorkflowStartOperationUpdate.getUpdateHandle` be renamed to `WorkflowStartOperationUpdate.getUpdateHandleAsync` for consistency?

temporal-sdk/src/main/java/io/temporal/client/WorkflowStartOperationUpdate.java
```diff
@@ -0,0 +1,243 @@
...
+
+package io.temporal.client;
+
+import io.temporal.workflow.Functions;
+import java.lang.reflect.Type;
+import java.util.Arrays;
+import java.util.concurrent.*;
+
+/** TBD */
+public class WorkflowStartOperationUpdate<R>
```
cretz, line 29: IMO it doesn't make sense to have `Workflow` + `Start` + `Operation` + `Update` extend `Start` + `Workflow` + `Additional` + `Operation` with the adjectives inconsistent or missing.

Like `WorkflowOptions` are not `WorkflowStartOptions` because start is implied, `StartWorkflowAdditionalOperation` may be able to imply `Start`. Some possible options:

* `UpdateWorkflowAdditionalOperation` extends `WorkflowAdditionalOperation`
* `AdditionalUpdateWorkflowOperation` extends `AdditionalWorkflowOperation`
* `UpdateWithStartWorkflowAdditionalOperation` extends `StartWorkflowAdditionalOperation`
* `AdditionalUpdateWithStartWorkflowOperation` extends `AdditionalStartWorkflowOperation`

Or if we have decided to remove the term "Additional", then we should remove it from the base class too (though not as much of a fan, we should decide whether the adjective is generally valuable or not, not selectively), e.g. 

* `StartWorkflowOperationUpdate` extends `StartWorkflowOperation`
* `WorkflowUpdateStartOperation` extends `WorkflowStartOperation`

Or something like that. No strong opinion here.

cretz (commented):
LGTM, just minor things about naming

temporal-sdk/src/main/java/io/temporal/client/WorkflowOptions.java
stephanos: Good call 👍

temporal-serviceclient/src/main/java/io/temporal/serviceclient/StatusUtils.java
```diff
@@ -87,7 +87,7 @@ public static <T extends GeneratedMessageV3> StatusRuntimeException newException
    * This method does exactly what {@link Any#pack(Message)} does. But it doesn't go into reflection
    * to fetch the {@code descriptor}, which allows us to avoid a bunch of Graal reflection configs.
    */
-  private static <T extends GeneratedMessageV3> Any packAny(
+  public static <T extends GeneratedMessageV3> Any packAny(
```
Quinn-With-Two-Ns, line 90: Can we move this to internal if it needs to be public?

temporal-sdk/src/main/java/io/temporal/client/WorkflowClient.java
```diff
@@ -568,6 +568,302 @@ static <A1, A2, A3, A4, A5, A6, R> WorkflowExecution start(
...
   }
 
+  /**
+   * Executes zero argument workflow with void return type together with an additional operation.
+   *
+   * @param workflow The only supported value is method reference to a proxy created through {@link
+   *     #newWorkflowStub(Class, WorkflowOptions)}.
+   * @param additionalOperation to run at the beginning of the workflow (e.g. {@link
+   *     WorkflowStartOperationUpdate} for Update-With-Start)
+   * @return WorkflowExecution that contains WorkflowId and RunId of the started workflow.
```
Quinn-With-Two-Ns, line 578: The returns are wrong now no?

temporal-sdk/src/main/java/io/temporal/client/WorkflowStub.java
```diff
@@ -149,6 +149,9 @@ <R> WorkflowUpdateHandle<R> getUpdateHandle(
 
   WorkflowExecution start(Object... args);
 
+  <R> WorkflowExecution startWithOperation(
```
Quinn-With-Two-Ns, line 152: Can we document this? I know `signalWithStart ` isn't but that is more a deficiency of that API.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/StatusUtils.java
stephanos: Like so https://github.com/temporalio/sdk-java/pull/2199/commits/caf4fa1e18b703ddf3fc20b6b952425b80ad74cc ?

temporal-sdk/src/main/java/io/temporal/client/WorkflowClient.java
stephanos: 👍 let me fix that.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/StatusUtils.java
Quinn-With-Two-Ns: Yeah, this is just meant to be an internal helper method not something we want users of the SDK to use

temporal-sdk/src/main/java/io/temporal/client/WorkflowClient.java
stephanos: addressed here https://github.com/temporalio/sdk-java/pull/2199/commits/cb7565900ff4d33696fb36025723dd523a3e2927

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
```diff
@@ -912,6 +921,125 @@ public void pollWorkflowExecutionUpdate(
     }
   }
 
+  @Override
+  public void executeMultiOperation(
```
Quinn-With-Two-Ns, line 925: Do we plan to add explicit test coverage for the multi-ops API as part of this PR?

temporal-sdk/src/main/java/io/temporal/client/WorkflowStartOperationUpdate.java
cretz: I'll resolve it, this applied to a previous form of this PR.

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
stephanos: Right; it's only tested implicitly right now. To be honest, until now I didn't see that the test server actually has its own tests. I can add explicit tests there.

temporal-sdk/src/test/java/io/temporal/workflow/updateTest/UpdateWithStartTest.java
drewhoskins-temporal: Is there still a use case for getting a handle from this operation if `updateWithStart` is changed to return an update handle?

cretz: Yes, I think getting a handle always has value so you can pass it around, so you can see the update ID, and so that you can do other things on it once we add them (I hope we add "describe" soon, CLI already saw a need for it).

cretz: Correct, there is probably not value. It doesn't make too much sense to have two ways to get a handle.

cretz: Only reason I could think of is as a promise, e.g. to give you something to wait on before the start call even starts, e.g.

```java
public Optional<WorkflowStartOperationUpdate<String>> addUpdateIfNeeded(SomeParam someParam) {
  if (!someParam.isUpdateNeeded()) {
    return Optional.empty();
  }
  var update = WorkflowStartOperationUpdate.newBuilder(workflow::update, 1, "Hello Update")
        .setWaitForStage(WorkflowUpdateStage.COMPLETED)
        .build();
  this.asynchronouslyDealWithUpdateResult(update.getUpdateHandle());
  return Optional.of(update);
}
```

I could derive a few helpers, but suffice to say it is a use case to be able to await on a future once it's created, not once it's applied. Having said that we could probably get away with not having it.

temporal-sdk/src/main/java/io/temporal/client/UpdateWithStartWorkflowOperation.java
```diff
@@ -0,0 +1,407 @@
...
+  WorkflowUpdateHandle<R> invoke(Functions.Proc workflow) {
+    WorkflowInvocationHandler.initAsyncInvocation(
+        WorkflowInvocationHandler.InvocationType.UPDATE_WITH_START, this);
+    try {
+      request.apply();
+      workflow.apply();
+      stub.updateWithStart(this, this.updateArgs);
+      return this.handle.get();
+    } catch (ExecutionException | InterruptedException e) {
+      throw new RuntimeException(e);
```
stephanos, line 300: Is this the best way to convert this from a checked exception to an unchecked? (annotating all methods with the two unchecked exceptions seems undesirable to me)

temporal-sdk/src/main/java/io/temporal/client/WorkflowStartOperationUpdate.java
stephanos: Apologies, I overlooked this one yesterday. To make sure I understand this right: the idea is to provide the above method on the Update operation builder? And it would set the operation's update options to the options of `builder.build()`?

temporal-sdk/src/main/java/io/temporal/client/UpdateWithStartWorkflowOperation.java
Quinn-With-Two-Ns: `ExecutionException ` should be unwrapped, it is just a wrapper around the real exception from the future. I think for `InterruptedException ` we should just set the interrupted flag so 
` Thread.currentThread().interrupt();`

stephanos: Ah, thank you for the pointer! I found another place where we do this now and will copy that approach (just like you described).

Quinn-With-Two-Ns: The Java SDK does this in other places https://github.com/temporalio/sdk-java/blob/a173dbeb6e1b89f7ad184f2e3d9e21e185d51597/temporal-sdk/src/main/java/io/temporal/client/EagerPaginator.java#L99

Quinn-With-Two-Ns: Make sure your test coverage also includes `getResult()` when the update has failed

temporal-sdk/src/test/java/io/temporal/workflow/updateTest/UpdateWithStartTest.java
```diff
@@ -0,0 +1,484 @@
...
+    TestWorkflows.WorkflowWithUpdate workflow =
+        workflowClient.newWorkflowStub(TestWorkflows.WorkflowWithUpdate.class, options);
+
+    UpdateWithStartWorkflowOperation<String> updateOp =
+        UpdateWithStartWorkflowOperation.newBuilder(workflow::update, 1, "Hello Update")
+            .setWaitForStage(WorkflowUpdateStage.COMPLETED)
+            .build();
+
+    WorkflowUpdateHandle<String> handle1 =
+        WorkflowClient.updateWithStart(workflow::execute, updateOp);
```
cretz, line 69: Can you show how to get/wait on the workflow execution/result?

temporal-sdk/src/main/java/io/temporal/client/WorkflowClient.java
```diff
@@ -568,6 +568,288 @@ static <A1, A2, A3, A4, A5, A6, R> WorkflowExecution start(
...
 
+  /**
+   * Executes zero argument workflow with void return type together with an update workflow request.
+   *
+   * @param workflow The only supported value is method reference to a proxy created through {@link
+   *     #newWorkflowStub(Class, WorkflowOptions)}.
+   * @param updateOperation update workflow operation
+   * @return WorkflowUpdateHandle that can be used to get the result of the update
+   */
+  static <R> WorkflowUpdateHandle<R> updateWithStart(
```
cretz, line 579: Not the biggest fan of only this SDK only having an explicit, single-purpose update-with-start call, but I guess if it's limited to just Java and all other SDKs can do better, at least the blast radius around this decision is limited.

temporal-sdk/src/test/java/io/temporal/workflow/updateTest/UpdateWithStartTest.java
stephanos: Line 77 has
```
assertEquals("Hello Update complete", workflow.execute());
```
isn't that how you do that in Java? (ie call the WorkflowMethod again)

temporal-sdk/src/test/java/io/temporal/workflow/updateTest/UpdateWithStartTest.java
```diff
@@ -0,0 +1,484 @@
...
+
+  public static class TestUpdatedWorkflowImpl implements TestWorkflows.TestUpdatedWorkflow {
+
+    @Override
+    public String execute() {
+      return "done";
+    }
+
+    @Override
+    public void update(String arg) {}
```
Quinn-With-Two-Ns, line 482: Can we test calling this as part of an update with start successfully as well? I see we use it in a test, but I don't think we actually test it can work. The reason I want to test it is because it returns a `void` type and in and sometimes that can cause problems.

stephanos: 👍 good idea! done: https://github.com/temporalio/sdk-java/pull/2199/commits/7ed0ddd439bbb074d940f838eae711d178c0221b

temporal-sdk/src/main/java/io/temporal/client/WorkflowInvocationHandler.java
```diff
@@ -394,4 +399,65 @@ public <R> R getResult(Class<R> resultClass) {
       throw new IllegalStateException("No result is expected");
     }
   }
+
+  private static class UpdateWithStartInvocationHandler implements SpecificInvocationHandler {
+
+    private int step;
```
Quinn-With-Two-Ns, line 405: What is the purpose of this variable? I think it is trying to create a little state machine, but it isn't super obvious to me.

temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkflowClientCallsInterceptor.java
```diff
@@ -61,6 +62,8 @@ public interface WorkflowClientCallsInterceptor {
 
   WorkflowSignalWithStartOutput signalWithStart(WorkflowSignalWithStartInput input);
 
+  <R> WorkflowUpdateWithStartOutput<R> updateWithStart(WorkflowUpdateWithStartInput<R> input);
```
Quinn-With-Two-Ns, line 65: Can we add javadocs here?

Quinn-With-Two-Ns: We should probably also call out if you are overriding `start` you may want to override `updateWithStart` like we do for `signalWithStart `

temporal-sdk/src/main/java/io/temporal/client/WorkflowClient.java
```diff
@@ -568,6 +568,288 @@ static <A1, A2, A3, A4, A5, A6, R> WorkflowExecution start(
...
+  /**
+   * Executes one argument workflow with void return type together with an update workflow request.
+   *
+   * @param workflow The only supported value is method reference to a proxy created through {@link
+   *     #newWorkflowStub(Class, WorkflowOptions)}.
+   * @param arg1 first workflow function parameter
+   * @param updateOperation update workflow operation
+   * @return WorkflowUpdateHandle that can be used to get the result of the update
+   */
+  static <R, A1> WorkflowUpdateHandle<R> updateWithStart(
```
Quinn-With-Two-Ns, line 593: Also these should be tagged `Experimental` right? I see tags in other places for Update-With-Start but not here.

temporal-sdk/src/main/java/io/temporal/client/UpdateWithStartWorkflowOperation.java
```diff
@@ -0,0 +1,413 @@
...
+   * Returns a new builder for an {@link UpdateWithStartWorkflowOperation} for a five argument
+   * request.
+   *
+   * @param request method reference to a proxy created through {@link
+   *     WorkflowClient#newWorkflowStub(Class, WorkflowOptions)}.
+   * @param arg1 first request function parameter
+   * @param arg2 second request function parameter
+   * @param arg3 third request function parameter
+   * @param arg4 fourth request function parameter
+   * @param arg5 sixth request function parameter
```
Quinn-With-Two-Ns, line 118:
```suggestion
   * @param arg5 fifth request function parameter
```

temporal-sdk/src/main/java/io/temporal/client/UpdateWithStartWorkflowOperation.java
```diff
@@ -0,0 +1,413 @@
...
+   * Returns a new builder for an {@link UpdateWithStartWorkflowOperation} for a six argument
+   * request.
+   *
+   * @param request method reference to a proxy created through {@link
+   *     WorkflowClient#newWorkflowStub(Class, WorkflowOptions)}.
+   * @param arg1 first request function parameter
+   * @param arg2 second request function parameter
+   * @param arg3 third request function parameter
+   * @param arg4 fourth request function parameter
+   * @param arg5 sixth request function parameter
```
Quinn-With-Two-Ns, line 135:
```suggestion
   * @param arg5 fifth request function parameter
```

temporal-sdk/src/test/java/io/temporal/workflow/updateTest/UpdateWithStartTest.java
cretz: Does that start a different workflow? Update with start I would expect to be a single call that does both update and start at the same time and I would expect to be able to get the execution (or result) of the workflow alongside the handle of the update. Would not expect to make a separate workflow call potentially starting it again.

temporal-sdk/src/main/java/io/temporal/client/UpdateWithStartWorkflowOperation.java
```diff
@@ -0,0 +1,413 @@
...
+   * Returns a new builder for an {@link UpdateWithStartWorkflowOperation} for a five argument
+   * request.
+   *
+   * @param request method reference to a proxy created through {@link
+   *     WorkflowClient#newWorkflowStub(Class, WorkflowOptions)}.
+   * @param arg1 first request function parameter
+   * @param arg2 second request function parameter
+   * @param arg3 third request function parameter
+   * @param arg4 fourth request function parameter
+   * @param arg5 sixth request function parameter
```
Quinn-With-Two-Ns, line 230:
```suggestion
   * @param arg5 fifth request function parameter
```

temporal-sdk/src/main/java/io/temporal/client/UpdateWithStartWorkflowOperation.java
```diff
@@ -0,0 +1,413 @@
...
+   * Returns a new builder for an {@link UpdateWithStartWorkflowOperation} for a six argument
+   * request.
+   *
+   * @param request method reference to a proxy created through {@link
+   *     WorkflowClient#newWorkflowStub(Class, WorkflowOptions)}.
+   * @param arg1 first request function parameter
+   * @param arg2 second request function parameter
+   * @param arg3 third request function parameter
+   * @param arg4 fourth request function parameter
+   * @param arg5 sixth request function parameter
```
Quinn-With-Two-Ns, line 247:
```suggestion
   * @param arg5 fifth request function parameter
```

temporal-sdk/src/main/java/io/temporal/client/WorkflowInvocationHandler.java
stephanos: `invoke` is called twice: once for update and once for start. This keeps track of which one we're expecting. I can rename it to `invocation`?

temporal-sdk/src/main/java/io/temporal/client/UpdateWithStartWorkflowOperation.java
stephanos: I copied that from `BatchRequest` and didn't notice the issue. I'll fix it there, too 👍

stephanos: There were 16 places: https://github.com/temporalio/sdk-java/pull/2199/commits/3df72ba7e38e757d8f6d6209cdc8aee64ba9fc23

temporal-sdk/src/main/java/io/temporal/client/WorkflowClient.java
stephanos: 👍 addressed in https://github.com/temporalio/sdk-java/pull/2199/commits/5833afe1eafb1103d18c20573d3fc78d4c9d105f

temporal-sdk/src/main/java/io/temporal/client/UpdateWithStartWorkflowOperation.java
```diff
@@ -0,0 +1,413 @@
...
+
+  private UpdateOptions<R> options;
+
+  private Object[] updateArgs;
+
+  private final CompletableFuture<WorkflowUpdateHandle<R>> handle;
+
+  private final Functions.Proc request;
+
+  public UpdateWithStartWorkflowOperation(
```
Quinn-With-Two-Ns, line 283: Why does this need to be public? Generally if a class has a builder the constructor is private so users have to use the builder.

stephanos: 👍 good catch; made it private now

temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkflowClientCallsInterceptor.java
stephanos: @Quinn-With-Two-Ns regarding javadocs, anything in particular you'd expect? None of the other ones have real javadocs other than referring to what else you might want to override.

Quinn-With-Two-Ns: I would say something like

```
Intercepts update with start calls from {@link WorkflowStub#updateWithStart} and {@link WorkflowClient#updateWithStart}
```

How does that sounds

temporal-sdk/src/main/java/io/temporal/client/WorkflowInvocationHandler.java
Quinn-With-Two-Ns: Since each invocation is different can we declare an enum inside this class and use that to keep track of the state?

```java
if(state.equals(NOT_STARTED)) {
       UpdateMethod updateMethod = method.getAnnotation(UpdateMethod.class);
        if (updateMethod == null) {
              "Method '" + method.getName() + "' is not an UpdateMethod");
        }
        this.operation.prepareUpdate(
            untyped,
            methodMetadata.getName(),
            method.getReturnType(),
            method.getGenericReturnType(),
            args);
         state = UPDATE_RECIEVED
        return;
} else if (state.equals(UPDATE_RECIEVED)) {

        WorkflowMethod workflowMethod = method.getAnnotation(WorkflowMethod.class);
        if (workflowMethod == null) {
              "Method '" + method.getName() + "' is not a WorkflowMethod");
        }
        this.operation.prepareStart(untyped);
        state = WORKFLOW_RECIEVED
        return;
} else {
      throw new IllegalArgumentException("UpdateWithStartInvocationHandler called too many times");
}
```

What do you think?

temporal-sdk/src/main/java/io/temporal/client/WorkflowClient.java
drewhoskins-temporal: @cretz We should still consider this approach for other SDKs.  But we don't have to decide now.
Would be curious to discuss hear what you think are the concrete user downsides of this.  
If it's just a question of lowering method count, I'm not super concerned.

cretz: Specific to this method and not past discussions about multi-op, the primary downside is that we are expanding the number of ways to start a workflow. I think there is developer clarity value in having common, documented ways to start a workflow. For instance, I am glad we created a workflow ID conflict policy instead of being tempted to add a `getOrStart` call. Hopefully we can limit this to signal with start and update with start and there is not a `startPaused` or similar in the future.

Since we can't have a common start call, it just made more sense to at least have a common start-with-operation call.

temporal-sdk/src/main/java/io/temporal/client/WorkflowInvocationHandler.java
stephanos: That looks better; done! https://github.com/temporalio/sdk-java/pull/2199/commits/7626657a7d96780e6f77930e7f850cfdfe098bd1

temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkflowClientCallsInterceptor.java
stephanos: Done! https://github.com/temporalio/sdk-java/pull/2199/commits/e77581535609824811e76cb36227daa515531ce5

temporal-sdk/src/main/java/io/temporal/client/UpdateWithStartWorkflowOperation.java
stephanos: Added! https://github.com/temporalio/sdk-java/pull/2199/commits/c5d8b34d7c0b78a506e872f36816d2dff9e6c237

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
stephanos: Added explicit tests here: https://github.com/temporalio/sdk-java/pull/2199/commits/90a6e0ce3513a5d8a52c7f393a2d5d9aa48bdcd2

temporal-sdk/src/test/java/io/temporal/workflow/updateTest/UpdateWithStartTest.java
stephanos: > Would not expect to invoke the method I already passed to update with start.

Looking at BatchRequest, I see that the tests are doing `WorkflowStub.fromTyped(workflow).getResult(<type>)`. I've updated this test accordingly: https://github.com/temporalio/sdk-java/pull/2199/commits/db5956e539273cd7a7cc9535d4c560019c8547b9

I'll say that having to pass in the workflow result type again is not ideal. But right now I fail to see an obvious alternative without that problem. 

> And what if I just want the workflow execution?

That's done via `handle.getExecution()`; for example see
```
assertEquals(options.getWorkflowId(), handle1.getExecution().getWorkflowId());
```

cretz: I think getting started workflow information from the update could be a bit confusing for users, but I think we all now recognize Java is just going to be different from everyone else.

cretz (approved):
Approving as is, though I liked the previous way better (update handle came from update operation, workflow execution returned from start).

temporal-test-server/src/test/java/io/temporal/testserver/functional/MultiOperationTest.java
```diff
@@ -0,0 +1,446 @@
...
+import io.temporal.testserver.functional.common.TestWorkflows;
+import io.temporal.workflow.Workflow;
+import java.time.Duration;
+import java.util.UUID;
+import java.util.concurrent.ExecutionException;
+import java.util.function.Consumer;
+import org.junit.Rule;
+import org.junit.Test;
+
+public class MultiOperationTest {
```
Quinn-With-Two-Ns, line 51: Can we verify the multi op API timeout also works properly? I don't see any test coverage here and in the implementation I don't think the underlying update request is being called with a deadline? You can see the test coverage we have for regular update like `updateNotAcceptedTimeout`

temporal-sdk/src/main/java/io/temporal/internal/client/external/GenericWorkflowClientImpl.java
```diff
@@ -386,4 +386,15 @@ public GetWorkerTaskReachabilityResponse GetWorkerTaskReachability(
...
         grpcRetryerOptions);
   }
+
+  @Override
+  public ExecuteMultiOperationResponse executeMultiOperation(ExecuteMultiOperationRequest req) {
+    return grpcRetryer.retryWithResult(
+        () ->
+            service
+                .blockingStub()
+                .withOption(METRICS_TAGS_CALL_OPTIONS_KEY, metricsScope)
```
Quinn-With-Two-Ns, line 396: Should we be including more metric tags here? `start` includes the workflow type and task queue, update includes the update name. It might make sense to include them here as well.

temporal-sdk/src/main/java/io/temporal/internal/client/external/GenericWorkflowClientImpl.java
```diff
@@ -386,4 +386,15 @@ public GetWorkerTaskReachabilityResponse GetWorkerTaskReachability(
                 .getWorkerTaskReachability(req),
         grpcRetryerOptions);
   }
+
+  @Override
+  public ExecuteMultiOperationResponse executeMultiOperation(ExecuteMultiOperationRequest req) {
```
Quinn-With-Two-Ns, line 391: `executeMultiOperation` is a potentially long operation correct because we won't return until the update reaches the desired stage. If so we need to take a deadline here like the `update` call.

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
```diff
@@ -200,6 +170,151 @@ public WorkflowSignalWithStartOutput signalWithStart(WorkflowSignalWithStartInpu
...
+            .addOperations(
+                1,
+                ExecuteMultiOperationRequest.Operation.newBuilder()
+                    .setUpdateWorkflow(updateRequest)
+                    .build())
+            .build();
+
+    ExecuteMultiOperationResponse response;
+    try {
+      response = genericClient.executeMultiOperation(request);
```
Quinn-With-Two-Ns, line 224: Do we not need to call this in a loop like regular update to avoid loosing the update requests ?

temporal-sdk/src/test/java/io/temporal/workflow/updateTest/UpdateWithStartTest.java
```diff
@@ -0,0 +1,542 @@
...
+import io.temporal.workflow.shared.TestMultiArgWorkflowFunctions;
+import io.temporal.workflow.shared.TestWorkflows;
+import java.util.*;
+import java.util.concurrent.ExecutionException;
+import java.util.function.BiFunction;
+import org.junit.Rule;
+import org.junit.Test;
+
+public class UpdateWithStartTest {
+
```
Quinn-With-Two-Ns, line 44: Do we have any test here that tests the update was executed before the main workflow method? Maybe I just missed it.

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
```diff
@@ -200,6 +170,151 @@ public WorkflowSignalWithStartOutput signalWithStart(WorkflowSignalWithStartInpu
...
+        throw Status.fromCodeValue(updateStatus.getCode())
+            .withDescription("Invalid StartOperation: " + updateStatus.getMessage())
+            .asRuntimeException();
+      }
+
+      throw err; // no detailed failure was found
+    }
+
+    if (response.getResponsesCount() != request.getOperationsCount()) {
+      throw new Error(
```
Quinn-With-Two-Ns, line 261: We should not throw `Error` from the workflow client. The SDK never throws  `Error` from the client, generally `Error` should only be thrown if there is something critically wrong and unrecoverable within the application itself, not some downstream service. (In Workflow code we throw errors as well, but that is to avoid users catching internal exceptions and is more an exception then the rule)

temporal-sdk/src/main/java/io/temporal/client/UpdateWithStartWorkflowOperation.java
```diff
@@ -0,0 +1,415 @@
...
+        () -> {
+          request.apply();
+        });
+  }
+
+  /**
+   * Returns a new builder for an {@link UpdateWithStartWorkflowOperation} for a one argument
+   * request.
+   *
+   * @param request method reference to a proxy created through {@link
```
Quinn-With-Two-Ns, line 57: This method reference has to be a a method annotated as an Update correct? Can we clarify that in the docs here? I don't think it is immediately obvious looking at just this class what methods it can take

temporal-sdk/src/main/java/io/temporal/client/UpdateWithStartWorkflowOperation.java
```diff
@@ -0,0 +1,415 @@
...
+  /** Returns the result of the update request. */
+  public R getResult() throws ExecutionException, InterruptedException {
+    return this.getUpdateHandle().get().getResultAsync().get();
+  }
+
+  public Future<WorkflowUpdateHandle<R>> getUpdateHandle() {
+    return this.handle;
+  }
+
+  public void setUpdateHandle(WorkflowUpdateHandle<R> updateHandle) {
```
Quinn-With-Two-Ns, line 339: This should not be public

Quinn-With-Two-Ns: I see your calling it from the root interceptor and that is probably the root of the issue. I think this should be set at the top of the interceptor chain and _I think_ that will also fix the visibility here

temporal-sdk/src/main/java/io/temporal/client/UpdateWithStartWorkflowOperation.java
stephanos: good point! https://github.com/temporalio/sdk-java/pull/2199/commits/a05748d90f12edb44ec8451cf4cb81ecc31abd5a

temporal-sdk/src/main/java/io/temporal/client/UpdateWithStartWorkflowOperation.java
stephanos: Yeah, I had called that out before (in the first version of the PR); that's a good idea on how to solve that. https://github.com/temporalio/sdk-java/pull/2199/commits/c2e0b4237bfe1a87d2ef20bbc0c8b5401e18c096

temporal-sdk/src/main/java/io/temporal/internal/client/external/GenericWorkflowClientImpl.java
stephanos: Good idea! https://github.com/temporalio/sdk-java/pull/2199/commits/019690efdb6a005dbb5e89b4bd0bd22276e9ce4b

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
stephanos: Good callout. We do. I don't think I did that in Go (https://github.com/temporalio/sdk-go/blob/592ae12141a56d9c52a1cb40eb28fa06d5550254/internal/internal_workflow_client.go#L1718). I'll file a ticket for that.

temporal-sdk/src/test/java/io/temporal/workflow/updateTest/UpdateWithStartTest.java
stephanos: I'm not sure I follow; why would that be an Update-with-Start-specific test? Isn't that already covered by the Update-specific tests?

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
stephanos: Okay, addressed here: https://github.com/temporalio/sdk-java/pull/2199/commits/3652e31b2a667047f8a6ea4b2c20a432e8acc6ea - I couldn't see a good way to test that, though?

temporal-sdk/src/main/java/io/temporal/internal/client/external/GenericWorkflowClientImpl.java
stephanos: Update is throwing a `WorkflowUpdateTimeoutOrCancelledException` upon timeout; I'm not quite sure if I should re-use it here or not. What do you think?

temporal-test-server/src/test/java/io/temporal/testserver/functional/MultiOperationTest.java
stephanos: Okay, I'm writing that test right now and it seems that the retries I added are causing ALREADY_EXISTS errors. From what I can tell right now, it seems like the Test Server is not deduping start requests based on request ID. I'm might have to fix that first.

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
Quinn-With-Two-Ns: I think it is something that needs to be tested with a unit test mocking the grpc stub

temporal-sdk/src/test/java/io/temporal/workflow/updateTest/UpdateWithStartTest.java
Quinn-With-Two-Ns: Because in the past there was not a good way to guarantee the update was in the first workflow task

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
```diff
@@ -1054,6 +1063,171 @@ public void pollWorkflowExecutionUpdate(
...
+        throw multiOperationExecutionFailure(
+            MultiOperationExecutionFailure.OperationStatus.newBuilder()
+                .setCode(e.getStatus().getCode().value())
+                .setMessage(e.getMessage())
+                .build(),
+            null); // update aborted
+      }
+
+      // execute update
+      UpdateWorkflowExecutionResponse updateResult;
```
Quinn-With-Two-Ns, line 1168: How do you know this update is always sent as part of the first workflow task?

temporal-sdk/src/test/java/io/temporal/workflow/updateTest/UpdateWithStartTest.java
Quinn-With-Two-Ns: My main concern is how do you know the test server is sending the update in the first workflow task when using update-with-start?

stephanos: 👍 Okay, I see what you mean.

stephanos: Okay, so I think the `startAndSendUpdateTogetherWithNullUpdateResult` test covers this. The workflow is

```
public static class TestUpdatedWorkflowImpl implements TestWorkflows.TestUpdatedWorkflow {

    private String state;

    @Override
    public String execute() {
      return state;
    }

    @Override
    public void update(String arg) {
      this.state = arg;
    }
  }
```

And it asserts that the workflow result equals the update input. If the update _wasn't_ executed first, it would return `null` instead and the test fails.

temporal-test-server/src/test/java/io/temporal/testserver/functional/MultiOperationTest.java
stephanos: Okay, I'm having an issue here: when mimicking the `updateNotAcceptedTimeout` test for MultiOps, I see a `DEADLINE_EXCEEDED: deadline exceeded after 9.999713584s.` client-side error. I can also see in the debugger that the server attempted to respond, but the client timeout throws first.

I couldn't yet figure out how to bump the client-side timeout up. I tried `.withDeadlineAfter(30, TimeUnit.SECONDS)` but it has no effect, it still times out after ~10s.

temporal-sdk/src/test/java/io/temporal/workflow/updateTest/UpdateWithStartTest.java
```diff
@@ -288,6 +294,41 @@ public void startVariousProcs() throws ExecutionException, InterruptedException
...
+    final AtomicReference<WorkflowServiceException> exception = new AtomicReference<>();
+    ScheduledExecutorService scheduledExecutor = Executors.newScheduledThreadPool(1);
+    Context.current()
+        .withDeadlineAfter(500, TimeUnit.MILLISECONDS, scheduledExecutor)
+        .run(
+            () ->
+                exception.set(
+                    assertThrows(
+                        WorkflowServiceException.class,
```
stephanos, line 324: The equivalent Update test receives the `WorkflowUpdateTimeoutOrCancelledException` directly here. I'm not sure why here it's wrapped inside a `WorkflowServiceException`.

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
stephanos: Is there a good example for this? I couldn't really find one yet.

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
Quinn-With-Two-Ns: The test is good, but looking at the code it is not clear to me the test server implementation actually **guarantees** the update will be in the same workflow task? My specific concern is the  `startWorkflowExecutionImpl` and the `updateWorkflowExecutionImpl` are separate calls?

Quinn-With-Two-Ns: I checked out your branch and added a short 1s sleep between the call `startWorkflowExecutionImpl ` and `updateWorkflowExecutionImpl ` and confirmed the update will not be delivered in the first workflow task

stephanos: Okay, I've poked around the server a little more, in particular `TestWorkflowMutableStateImpl`, and can see how this happens now. I don't see an easy fix right now. The two MS updates need to happen together to guarantee they are also delivered together to the worker. But I don't see any kind of "transaction" API. I'll need to dig deeper.

temporal-test-server/src/test/java/io/temporal/testserver/functional/MultiOperationTest.java
stephanos: @Quinn-With-Two-Ns would you know what's going on here?

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
stephanos: Addressed here: https://github.com/temporalio/sdk-java/pull/2199/commits/2a39d5edb5b73ed5a9cb976fe5038105d7ea2f1c

temporal-test-server/src/test/java/io/temporal/testserver/functional/MultiOperationTest.java
stephanos: Addressed in https://github.com/temporalio/sdk-java/pull/2199/commits/df7a43d285bd050101d835e6f1a5eae56bee49e4

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
stephanos: Addressed in https://github.com/temporalio/sdk-java/pull/2199/commits/67bfbb4916ed87585dcaa23596128f5f49fa4eb5

Quinn-With-Two-Ns (approved):
Workflow client and test server changes LGTM

Issue #1106, opened 2022-03-31
Workflow metadata resolution code doesn't respect generics specified down the hierarchy


Spikhalskiy:
## Expected Behavior

```
@WorkflowInterface
public interface ReportingWorkflow extends TemporalWorkflow<ReportQueriesResponse> {
  @WorkflowMethod
  ReportQueriesResponse runReport(ReportQueries reportQueries);
}

public interface TemporalWorkflow<T> extends TemporalRunnable {
  @QueryMethod
  T getResults();
}
```

Temporal JavaSDK should be able to understand that `ReportingWorkflow#getResults` returns ReportQueriesResponse and use this class as a target in data conversion.

## Actual Behavior

Data Converters try to decode query results into `Object`.

## Notes

Worth checking the same problem for Activity Metadata resolution. 

## Specifications

  - Version: 1.8.x

mfateev:
I don't think this is a bug. Given that Java erases generics information at runtime, it is impossible to implement.

cretz:
> it is impossible to implement.

True, but there are utilities if we ever want to have user-specified generic values (e.g. maybe a default method that returns https://guava.dev/releases/21.0/api/docs/com/google/common/reflect/TypeToken.html for `T`)

But what I would recommend is to disallow annotations on generic methods. If Temporal needs to know concrete types, it should be an error not to be able to reflectively obtain them. So in the pasted sample, they'd need to provide an annotation on the concrete interface override instead.

Spikhalskiy:
> Given that Java erases generics information at runtime, it is impossible to implement.

It's not exactly true. Java erases generic information for dynamically created instances of gentrified classes. But Java do preserve information about generic typed for defined standalone classes. Basically, if you have `class A extends/implements B<C>` - type of C is preserved.
So, it's possible and it's a bug.

cretz:
You're going to have to have really complicated rules about how to derive `C` in complicated inheritance scenarios. And what if they never provide a type parameter? in an inherited interface? Or what if they do `? extends List<String>` or something. Or additional type constraints in multiple levels? Or if they have an impl that uses a more specific generic value than the interface?

I think this is a can of worms that you can only half support. I can give lots of confusing scenarios. To me the most straightforward answer is "you cannot define Temporal attributes on generic items". It's harmless to ask someone to have a concrete override for use by stubs and other items.

Spikhalskiy:
> I think this is a can of worms that you can only half support.
[...]

Issue #1107, opened 2022-03-31
Workflow metadata resolution code registers an overridden method from parent class as a second query


Spikhalskiy:
## Expected Behavior

Such a workflow interfaces hierarchy:

```
@WorkflowInterface
public interface ReportingWorkflow extends TemporalWorkflow<ReportQueriesResponse> {
  @WorkflowMethod
  ReportQueriesResponse runReport(ReportQueries reportQueries);

  @QueryMethod
  @Override
  ReportQueriesResponse getResults();
}

public interface TemporalWorkflow<T> extends TemporalRunnable {
  T getResults();
}
```

should be fine.

## Actual Behavior

Triggers `java.lang.IllegalArgumentException: Duplicated name of QUERY: "getResults" declared at "public abstract xxx.ReportingProto$ReportQueriesResponse xxx.ReportingWorkflow.getResults()" and "public default java.lang.Object xxx.ReportingWorkflow.getResults()"`

## Specifications

  - Version: 1.8.x

shabalin:
The same happened if you override with @WorkflowMethod
```
public interface TemporalWorkflow<T> {
    void execute(T params);
}
```
and then define the actual workflow interface:
```
@WorkflowInterface
public interface SearchTaskWorkflow extends TemporalWorkflow<SearchWorkflowParams> {
    @WorkflowMethod
    @Override
    void execute(SearchWorkflowParams params);
}
```
Workflow registration fails with the error:
```
java.lang.IllegalArgumentException: Duplicated @WorkflowMethod: public default void io.dev.analytics.temporal.SearchTaskWorkflow.execute(java.lang.Object) and public abstract void io.dev.analytics.temporal.SearchTaskWorkflow.execute(io.dev.analytics.service.search.async.SearchWorkflowParams)
	at io.temporal.common.metadata.POJOWorkflowInterfaceMetadata.getWorkflowInterfaceMethods(POJOWorkflowInterfaceMetadata.java:296) ~[temporal-sdk-1.16.0.jar:na]
	at io.temporal.common.metadata.POJOWorkflowInterfaceMetadata.<init>(POJOWorkflowInterfaceMetadata.java:205) ~[temporal-sdk-1.16.0.jar:na]
	at io.temporal.common.metadata.POJOWorkflowInterfaceMetadata.newInstanceInternal(POJOWorkflowInterfaceMetadata.java:184) ~[temporal-sdk-1.16.0.jar:na]
[...]

PR #2201, opened 2024-08-28
Support generics on workflow methods

rorueda:
## What was changed

Support for generics on workflow methods, both in parameters and return types. The aim is to support use cases like the following:

1. Use of a `TypeVariable` that is specified.
```java
public interface QueryBase<T> {
  @QueryMethod
  T query();
}
@WorkflowInterface
public interface Workflow extends QueryBase<WorkflowState> {}
```

2. Use of a `TypeVariable` that is bound.
```java
public interface QueryBase<T> {
  @QueryMethod
  T query();
}
@WorkflowInterface
public interface Workflow<T extends WorkflowState> extends QueryBase<T> {}
```

3. Use of a `TypeVariable` in a `ParameterizedType`.
```java
public interface QueryBase<T> {
  @QueryMethod
  Wrapper<T> query();
}
@WorkflowInterface
public interface Workflow extends QueryBase<WorkflowState> {}
```

4. Use of a `TypeVariable` in a `GenericArrayType`.
```java
public interface QueryBase<T> {
  @QueryMethod
  T[] query();
}
@WorkflowInterface
public interface Workflow extends QueryBase<WorkflowState> {}
```

5. Use of a `TypeVariable` in a `GenericArrayType`, in a `ParameterizedType`.
```java
public interface QueryBase<T> {
  @QueryMethod
  Wrapper<T[]> query();
}
@WorkflowInterface
public interface Workflow extends QueryBase<WorkflowState> {}
```

6. Use of a `TypeVariable` as an upper bound of a `WildcardType` in a `ParameterizedType`.
```java
public interface QueryBase<T> {
  @QueryMethod
  Wrapper<? extends T> query();
}
@WorkflowInterface
public interface Workflow extends QueryBase<WorkflowState> {}
```

## Why?

To be able to reuse a base interface in all my workflow interfaces. Together, the two issues mentioned below in the checklist make this currently impossible.

Reading the comments on the first issue, I get that this support might be controversial, but I couldn't live without it. It would be great to have this merged, so others can benefit and I don't need to keep a fork.

## Checklist

1. Closes 
https://github.com/temporalio/sdk-java/issues/1106
https://github.com/temporalio/sdk-java/issues/1107
2. How was this tested:

Unit tests.

3. Any docs updates needed?

Maybe it is good to point out what is supported at [workflow-parameters](https://docs.temporal.io/develop/java/core-application#workflow-parameters) and [workflow-return-values](https://docs.temporal.io/develop/java/core-application#workflow-return-values).


Quinn-With-Two-Ns:
Hey @rorueda Thank you for the contribution. First impression it looks good, but I need to take a closer look over a couple days to verify and make sure we aren't breaking any existing user. Can you think of any existing cases that may break from this change?

rorueda:
Great, @Quinn-With-Two-Ns. 

I certainly hope this doesn't break existing code, or at least break during startup, instead of resulting in wrong deserialization, because the changes made shouldn't result in already fully resolved types, with no type variables, being modified. 

I made some minor, I think, changes to try to encapsulate more stuff inside the `POJOWorkflowMethod`, but I would expect it to be covered by existing tests. 

I am not sure, but people who use native images might need to take additional steps to make this work. 

For people that will start using generics, the `PayloadConverter.fromData` will receive a different `valueType`, that is usually passed on to the serialization library, but a custom `PayloadConverter` might not do this for whatever reason or the serialization library might not handle the complex types well.

That is all I can think at the moment.

PS: I see that the features test is broken because of one of the changes and I'll push a fix.

rorueda:
So, there were 2 issues with the features test:
- I've changed the return type of a method in `POJOWorkflowMethodMetadata` that is used in the runner. Even though it is a public method, I assumed it's kind of internal;
- At least one method on the workflow interfaces there was declaring a type variable. It's just a helper method it seems, but as it's a candidate for turning into a workflow method, it is processed and the code was not handling this situation.

I'd like to keep the change to `POJOWorkflowMethodMetadata` because I think it is best long-term, but it would require the features to be compiled against the modified sdk. Does the CI modifies the `features/build.gradle` to use the sdk snaphost? If not, will probably have to keep the old method and add a new one.

Quinn-With-Two-Ns:
Tank you again for the contribution, after internal discussion we decided we do not want to support or encourage generic interfaces across our SDKs at this time. We recommend users create specific interfaces for each use case.

https://github.com/temporalio/sdk-python/issues/322

Issue #623, opened 2021-08-09
Upsert memo support


vkoby:
Memo can be utilized by users as an "additional" workflow result that is not explicitly populated in the workflow result object.
Workflow Executions should have an ability to modify/upsert the memo during the execution, not just at the start time.
There is a `ModifyWorkflowPropertiesCommandAttributes` command in place for this already.

aleskovets:
Sad that java sdk does not have it. Great feature to optimize number of requests with advanced visibility

PR #2202, opened 2024-08-30
Add support for upsert memo

Quinn-With-Two-Ns:
Add support for upsert memo to the Java SDK and test server.

closes https://github.com/temporalio/sdk-java/issues/623

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkflowOutboundCallsInterceptorBase.java
+++ b/temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkflowOutboundCallsInterceptorBase.java
@@ -158,7 +158,7 @@ public class WorkflowOutboundCallsInterceptorBase implements WorkflowOutboundCal
 
   @Override
   public void upsertMemo(Map<String, Object> memo) {
-    next.upsertSearchAttributes(memo);
+    next.upsertMemo(memo);
   }
 
   @Override
--- a/temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkflowOutboundCallsInterceptorBase.java
+++ b/temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkflowOutboundCallsInterceptorBase.java
@@ -157,6 +157,11 @@ public class WorkflowOutboundCallsInterceptorBase implements WorkflowOutboundCal
   }
 
   @Override
+  public void upsertMemo(Map<String, Object> memo) {
+    next.upsertSearchAttributes(memo);
+  }
+
+  @Override
   public Object newChildThread(Runnable runnable, boolean detached, String name) {
     return next.newChildThread(runnable, detached, name);
   }
--- a/temporal-sdk/src/main/java/io/temporal/workflow/Workflow.java
+++ b/temporal-sdk/src/main/java/io/temporal/workflow/Workflow.java
@@ -1189,8 +1189,9 @@ public final class Workflow {
   /**
    * Updates a Workflow Memos by applying {@code memoUpdates} to the existing Memos set attached to
    * the workflow. Memos are additional non-indexed information attributed to workflow and can
-   * returned by describing or listing a workflow. The type of value can be any object that are
-   * serializable by {@link io.temporal.common.converter.DataConverter}
+   * return by describing or listing a workflow. The type of value can be any object that are
+   * serializable by {@link io.temporal.common.converter.DataConverter}. To remove a memo set the
+   * value null.
    */
   public static void upsertMemo(Map<String, Object> memo) {
     WorkflowInternal.upsertMemo(memo);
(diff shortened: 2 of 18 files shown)


temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkflowOutboundCallsInterceptorBase.java
```diff
@@ -156,6 +156,11 @@ public void upsertTypedSearchAttributes(SearchAttributeUpdate<?>... searchAttrib
     next.upsertTypedSearchAttributes(searchAttributeUpdates);
   }
 
+  @Override
+  public void upsertMemo(Map<String, Object> memo) {
+    next.upsertSearchAttributes(memo);
```
cretz, line 161:
```suggestion
    next.upsertMemo(memo);
```

temporal-sdk/src/main/java/io/temporal/workflow/Workflow.java
```diff
@@ -1186,6 +1186,16 @@ public static void upsertTypedSearchAttributes(
     WorkflowInternal.upsertTypedSearchAttributes(searchAttributeUpdates);
   }
 
+  /**
+   * Updates a Workflow Memos by applying {@code memoUpdates} to the existing Memos set attached to
```
cretz, line 1190: Can you clarify how a memo value is removed? Is this done by setting a null object? So can memo contain a null value for a key? In .NET we accept [memo update objects](https://dotnet.temporal.io/api/Temporalio.Workflows.Workflow.html#Temporalio_Workflows_Workflow_UpsertMemo_Temporalio_Workflows_MemoUpdate___) to remove any ambiguity. I forget what Go does here.

Quinn-With-Two-Ns: Yes by setting a null value same as Go.

temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkflowOutboundCallsInterceptorBase.java
dandavison: Tests passed so looks like we must be missing some test coverage?

temporal-sdk/src/main/java/io/temporal/workflow/Workflow.java
cretz: Is there a difference between a null value with key present and an absent key? Specifically, if I do an `upsertMemo(Collections.singletonMap("key-that-exists-from-start", null))`, what will the raw memo be if I do describe workflow?

temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkflowOutboundCallsInterceptorBase.java
```diff
@@ -156,6 +156,11 @@ public void upsertTypedSearchAttributes(SearchAttributeUpdate<?>... searchAttrib
     next.upsertTypedSearchAttributes(searchAttributeUpdates);
   }
 
+  @Override
+  public void upsertMemo(Map<String, Object> memo) {
```
cretz, line 160: Does Java offer the ability for me to see the whole memo collection inside the workflow? For instance if I wanted to know all memo keys in a workflow, can I? Do we need to add that?

Quinn-With-Two-Ns: No and No, at least no one has asked for it and I see no reason this PR would need to add it.

temporal-sdk/src/main/java/io/temporal/workflow/Workflow.java
Quinn-With-Two-Ns: `key-that-exists-from-start` would be removed by the server just like Go

cretz: :+1: May be worth mentioning in docs that that is how to remove a key

temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkflowOutboundCallsInterceptorBase.java
cretz: Works for me. My concern is if/when we do we need to make sure to more explicitly make sure null means remove from a `Map` POV.

temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkflowOutboundCallsInterceptorBase.java
Quinn-With-Two-Ns: Yep added some test coverage for this, will need to think long term if there is some way I can test all the interceptors programatically

cretz (approved):

Issue #865, opened 2021-11-10
Add init method to workflow


Spikhalskiy:
We should add an init method to workflows that allow some code to be run before any query, signal, or workflow method is processed.
This method should have access to workflow method input parameters.
This method doesn't have the same properties and guarantees as a workflow method. It shouldn't omit commands and doesn't work with WFT, because it would make it indistinguishable from the workflow method. It should be just a regular Java method.

This should solve the following situation:
1. A signal method can arrive before the workflow main method execution (reset preserving the signals on the first workflow task)
2. A signal method may require some initialization to be done.
3. General advice is to perform the initialization in the constructor or init block, BUT some initialization may require access to workflow input parameters. In this situation, users have to perform non-trivial gymnastics saving the signals in the field and processing them in a delayed manner in the workflow code.

It will also make it extra clear that Signals may arrive before the workflow code in triggered and the initialization should be handled with caution in some cases.

An alternative is to make the constructor such an init method. It will play nicely because it will be clear that it's not a workflow code and doesn't have properties of the workflow code.

# Dynamic Workflow
With regular typed workflows, listeners (signals and queries) are registered by SDK before processing signals. Currently, it's impossible to mimic this behavior with Dynamic Workflows.
Right now the only way to register listeners in Dynamic Workflow is to call the register methods from the workflow code. 
But with signalToStart, signal is getting delivered before the execution of the workflow method, at the moment when Dynamic Workflow code didn't execute yet, hence no listener is present. 
This makes init() method a necessity for Dynamic Workflows, not a convenience.

Related to #864
Related to a discussion under a user request in a community slack: https://temporalio.slack.com/archives/CTT84KXK9/p1636375128042600

cretz:
Saw #2166 was marked as dupe.

I do want to make sure there is a clear difference between "run any code before start/signal/update" and "initialize state before start/signal/update". I think `@WorkflowInit` can be a constructor annotation that, when set, means the constructor accepts the exact params as the primary workflow method (enforced at registration). Users can use this to set `final` fields that rely on start parameters. But I don't think we'd want to encourage command-creating logic in the constructor (though I guess it's technically possible in Java where you don't have to mark it async). I think general initialization code that creates commands is a different topic.

Quinn-With-Two-Ns:
> I think general initialization code that creates commands is a different topic.

I agree, but that is not covered by the [feature request](https://github.com/temporalio/features/issues/400)

PR #2206, opened 2024-09-06
Add workflow init support (💥 - incompatible change)

Quinn-With-Two-Ns:
Add workflow init support to allow users to run code before signal/update handlers:
* If the workflow implementation defines a no-arg constructor that is used to maintain backwards compatibility
* If the workflow does not, the SDK checks if a constructor with the same args as the main workflow method exists
* 💥 Exceptions throw from constructors now behave the same as normal workflow code

closes https://github.com/temporalio/sdk-java/issues/865

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/POJOWorkflowImplementationFactory.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/POJOWorkflowImplementationFactory.java
@@ -29,6 +29,7 @@ import io.temporal.api.common.v1.WorkflowExecution;
 import io.temporal.api.common.v1.WorkflowType;
 import io.temporal.common.context.ContextPropagator;
 import io.temporal.common.converter.DataConverter;
+import io.temporal.common.converter.EncodedValues;
 import io.temporal.common.interceptors.Header;
 import io.temporal.common.interceptors.WorkerInterceptor;
 import io.temporal.common.interceptors.WorkflowInboundCallsInterceptor;
@@ -49,6 +50,7 @@ import io.temporal.worker.WorkflowImplementationOptions;
 import io.temporal.workflow.DynamicWorkflow;
 import io.temporal.workflow.Functions;
 import io.temporal.workflow.Functions.Func;
+import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.Collections;
@@ -89,7 +91,8 @@ public final class POJOWorkflowImplementationFactory implements ReplayWorkflowFa
   private final Map<Class<?>, Functions.Func<?>> workflowInstanceFactories =
       Collections.synchronizedMap(new HashMap<>());
   /** If present then it is called for any unknown workflow type. */
-  private Functions.Func<? extends DynamicWorkflow> dynamicWorkflowImplementationFactory;
+  private Functions.Func1<EncodedValues, ? extends DynamicWorkflow>
+      dynamicWorkflowImplementationFactory;
 
   private final Map<String, WorkflowImplementationOptions> implementationOptions =
       Collections.synchronizedMap(new HashMap<>());
@@ -135,7 +138,8 @@ public final class POJOWorkflowImplementationFactory implements ReplayWorkflowFa
             "DynamicWorkflow",
             "An implementation of DynamicWorkflow or its factory is already registered with the worker");
       }
-      dynamicWorkflowImplementationFactory = (Func<? extends DynamicWorkflow>) factory;
+      dynamicWorkflowImplementationFactory =
+          (unused) -> ((Func<? extends DynamicWorkflow>) factory).apply();
       return;
     }
     workflowInstanceFactories.put(clazz, factory);
@@ -175,23 +179,30 @@ public final class POJOWorkflowImplementationFactory implements ReplayWorkflowFa
   private <T> void registerWorkflowImplementationType(
       WorkflowImplementationOptions options, Class<T> workflowImplementationClass) {
     if (DynamicWorkflow.class.isAssignableFrom(workflowImplementationClass)) {
-      addWorkflowImplementationFactory(
-          options,
-          workflowImplementationClass,
-          () -> {
+      if (dynamicWorkflowImplementationFactory != null) {
+        throw new TypeAlreadyRegisteredException(
+            "DynamicWorkflow",
+            "An implementation of DynamicWorkflow or its factory is already registered with the worker");
+      }
+      dynamicWorkflowImplementationFactory =
+          (encodedValues) -> {
             try {
-              return workflowImplementationClass.getDeclaredConstructor().newInstance();
+              try {
+                return (DynamicWorkflow)
+                    workflowImplementationClass.getDeclaredConstructor().newInstance();
+              } catch (NoSuchMethodException e) {
+                return (DynamicWorkflow)
+                    workflowImplementationClass
+                        .getDeclaredConstructor(EncodedValues.class)
+                        .newInstance(encodedValues);
+              }
             } catch (NoSuchMethodException
                 | InstantiationException
                 | IllegalAccessException
                 | InvocationTargetException e) {
-              // Error to fail workflow task as this can be fixed by a new deployment.
-              throw new Error(
-                  "Failure instantiating workflow implementation class "
-                      + workflowImplementationClass.getName(),
-                  e);
+              throw wrap(e);
             }
-          });
+          };
       return;
     }
     POJOWorkflowImplMetadata workflowMetadata =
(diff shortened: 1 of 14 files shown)


Quinn-With-Two-Ns:
Note: I still need to add some coverage for signal-with-start

temporal-sdk/src/main/java/io/temporal/internal/sync/POJOWorkflowImplementationFactory.java
```diff
@@ -349,22 +363,30 @@ public WorkflowOutput execute(WorkflowInput input) {
...
+      protected void newInstance(Optional<Payloads> input) {
         Func<?> factory = workflowInstanceFactories.get(workflowImplementationClass);
         if (factory != null) {
           workflow = factory.apply();
         } else {
           try {
-            workflow = workflowImplementationClass.getDeclaredConstructor().newInstance();
+            try {
+              workflow = workflowImplementationClass.getDeclaredConstructor().newInstance();
+            } catch (NoSuchMethodException e) {
```
cretz, line 374: What if the user in their constructor makes a reflection call that has a no-such-method? I know it's not likely, but in general I would recommend a more explicit approach instead of exceptions for control flow here.

temporal-sdk/src/main/java/io/temporal/internal/sync/POJOWorkflowImplementationFactory.java
```diff
@@ -349,22 +363,30 @@ public WorkflowOutput execute(WorkflowInput input) {
...
-      protected void newInstance() {
+      protected void newInstance(Optional<Payloads> input) {
         Func<?> factory = workflowInstanceFactories.get(workflowImplementationClass);
         if (factory != null) {
           workflow = factory.apply();
         } else {
           try {
-            workflow = workflowImplementationClass.getDeclaredConstructor().newInstance();
+            try {
+              workflow = workflowImplementationClass.getDeclaredConstructor().newInstance();
```
cretz, line 373: Do we validate other things at worker registration time with regards to workflow structure and annotation use? Meaning something like no duplicate signals? Can we validate the constructor there and also know which one we're using? I think we should validate that either no constructor is present, a parameterless constructor is present, or a constructor matching the workflow run parameters is present. And if _both_ parameterless and parameter-based constructors are present, we should log a warning IMO since that we choose the parameterless may not be obvious in ambiguous situations.

Arguably we should warn if there is more than one public constructor no matter what. IMO there are not very many good use cases for a non-Temporal-invoked constructor on a workflow. It'd either be a user wanting to instantiate themselves somehow in a test (but we need to control the environment for testing workflow code) or a user having a constructor that may be invoked by a subclass (they should mark it protected).

temporal-sdk/src/main/java/io/temporal/internal/sync/POJOWorkflowImplementationFactory.java
Quinn-With-Two-Ns: That would not throw a `NoSuchMethodException` that would throw an `InvocationException` with cause `NoSuchMethodException`

cretz: When I said "makes a reflection call" I didn't mean necessarily "makes a reflection method invocation". They could call `getDeclaredConstructor` just like us (or their serialization library might or whatever). The general statement of not using reflection exceptions for control flow here I think still applies.

If you _must_, I think you should call `newInstance()` outside of this try/catch, but I mention in another comment, I think validating the constructor to use should be done at registration time.

Quinn-With-Two-Ns: Where are they calling `getDeclaredConstructor`?

temporal-sdk/src/main/java/io/temporal/internal/sync/POJOWorkflowImplementationFactory.java
Quinn-With-Two-Ns: Yeah I hesitate to add validation where it didn't exist before, but if it is just warning that is probably fine and useful. Libraries normally only require an annotation if their is some ambiguity, Jackson does not require `@JsonCreator` you can have annotationless constructors.

temporal-sdk/src/main/java/io/temporal/internal/sync/POJOWorkflowImplementationFactory.java
cretz: Oh, my mistake, I see what you're saying, yes this exception can only happen if the declared constructor does not exist. I still think we should pre-validate the constructor's presence/visibility/params/etc at registration time. Can close this thread in favor of the other one if you want.

temporal-sdk/src/main/java/io/temporal/internal/sync/POJOWorkflowImplementationFactory.java
cretz: Right, IMO the annotation should only be required if you are leaving our long-standing default behavior. You are basically opting in to a special type of constructor.

Quinn-With-Two-Ns: I think only reason you would require an annotation is if their is ambiguity. The only case I can think of where there is ambiguity is if both a no-arg and argument constructor exist. So if we want to add an annotation and require it in the ambiguous case I'm OK with that just seems like a lot of extra work for a niche case.

dandavison:
I think it might be desirable for users to be able to define a constructor that accepts workflow input, and a run method that does not. That would make it clear that the workflow operates on the input after the constructor has done whatever it does to it, rather than having two parallel routes for the input to get into the workflow, with the possibility that processing done in the constructor is ignored. (It would also be a familiar pattern similar to the Thread/Runnable interface).

However, I'm not suggesting we should block this PR to make that possible today, and I haven't thought about how the typing/overloads would work.

Quinn-With-Two-Ns:
> desirable for users to be able to define a constructor that accepts workflow input, and a run method that does not.

That is not possible without redesigning the core ways users interact with workflows in the Java. Not commenting on the idea itself but it is more of a Java SDK v2 discussion .

temporal-sdk/src/main/java/io/temporal/internal/sync/POJOWorkflowImplementationFactory.java
```diff
@@ -175,23 +179,30 @@ public <R> void addWorkflowImplementationFactory(
...
-              return workflowImplementationClass.getDeclaredConstructor().newInstance();
+              try {
+                return (DynamicWorkflow)
+                    workflowImplementationClass.getDeclaredConstructor().newInstance();
+              } catch (NoSuchMethodException e) {
+                return (DynamicWorkflow)
+                    workflowImplementationClass
+                        .getDeclaredConstructor(EncodedValues.class)
+                        .newInstance(encodedValues);
+              }
```
dandavison, line 198: In Python, I am worried that making the decision automatically by inspecting the constructor signatures is too dangerous. For example, I'm worried about the behavior "flipping" without warning as a result of someone changing method resolution in a scenario with workflow class inheritance, like the example below. I think an analogous scenario exists in Java, right?

<details>

<summary>python example</summary>

```python
class Mixin:
    def __init__(self, arg: str) -> None:
        # Some unused __init__ method that someone accidentally left on this class.
        self.important_field = arg
        pass


@workflow.defn
class Base:
    def __init__(self, arg: str = "default-value") -> None:
        # Suppose someone deletes this __init__ method because they think it's
        # unnecessary. They have now switched the workflow without warning into
        # "workflow init" mode, and `important_field` will be set equal to the
        # workflow input payload.
        self.important_field = arg

    @workflow.run
    async def run(self, arg: str) -> str:
        self.important_field = "some-value"
        ...
        return self.important_field


@workflow.defn
class Workflow(Base, Mixin):
    @workflow.run
    async def run(self, arg: str) -> str:
        return self.important_field
```

</details>

temporal-sdk/src/main/java/io/temporal/internal/sync/POJOWorkflowImplementationFactory.java
cretz: That a user has to think about whether they are ambiguous or not for the annotation makes it less worth it. I think it's simpler for a user to understand it's always required if they want the args.

temporal-sdk/src/main/java/io/temporal/internal/sync/POJOWorkflowImplementationFactory.java
Quinn-With-Two-Ns: I am not  fully sure I understand this argument and if it applies to Java, but if python is also going to require an annotation then I am fine with Java requiring it to be consistent with dotnet and python

Quinn-With-Two-Ns:
Closing for now as I rework a few points to use annotations

PR #2210, opened 2024-09-10
Fix Dynamic Update Handler return type

Quinn-With-Two-Ns:
Fix Dynamic Update Handler return type to be consistent with all other `Dynamic*Handler`. The workflow code also expected Dynamic Update Handler to return an Object so no workflow side code changed.


Quinn-With-Two-Ns:
Note: I also noticed `getUnfinishedPolicy` doesn't have good integration test coverage for dynamic handlers. I will open a seprate PR to cover that

cretz (approved):

tsurdilo (approved):

Issue #2216, opened 2024-09-15
MVP support for user meta data on workflows and events 


Quinn-With-Two-Ns:
**Is your feature request related to a problem? Please describe.**
Add support for user meta data on events 

* Fixed "summary" and "details" on workflow start (includes signal with start and child workflow and scheduled workflows)
* "summary" on timer start (can serve as "timer ID")

https://github.com/temporalio/features/issues/486

Issue #2217, opened 2024-09-15
workflow metadata query


Quinn-With-Two-Ns:
Add support for workflow metadata query as described [here](https://github.com/temporalio/features/issues/486) and implemented in the Go SDK [here](https://github.com/temporalio/sdk-go/issues/1565).

Quinn-With-Two-Ns:
Duplicate of https://github.com/temporalio/sdk-java/issues/335

PR #2218, opened 2024-09-15
Add support for user meta data 

Quinn-With-Two-Ns:
Add support for user metadata on workflows and timer events. Other events can be added in the future, just picked these events to match the Go SDK.

Note: This does NOT cover `__temporal_workflow_metadata`, that feature is tracked [here](https://github.com/temporalio/sdk-java/issues/2217) 

closes https://github.com/temporalio/sdk-java/issues/2216

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowInternal.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowInternal.java
@@ -89,6 +89,11 @@ public final class WorkflowInternal {
     return getWorkflowOutboundInterceptor().newTimer(duration);
   }
 
+  public static Promise<Void> newTimer(Duration duration, TimerOptions options) {
+    assertNotReadOnly("schedule timer");
+    return getWorkflowOutboundInterceptor().newTimer(duration, options);
+  }
+
   /**
    * @param capacity the maximum size of the queue
    * @return new instance of {@link WorkflowQueue}
--- a/temporal-sdk/src/main/java/io/temporal/workflow/ChildWorkflowOptions.java
+++ b/temporal-sdk/src/main/java/io/temporal/workflow/ChildWorkflowOptions.java
@@ -73,6 +73,8 @@ public final class ChildWorkflowOptions {
     private List<ContextPropagator> contextPropagators;
     private ChildWorkflowCancellationType cancellationType;
     private VersioningIntent versioningIntent;
+    private String staticSummary;
+    private String staticDetails;
 
     private Builder() {}
 
@@ -96,6 +98,8 @@ public final class ChildWorkflowOptions {
       this.contextPropagators = options.getContextPropagators();
       this.cancellationType = options.getCancellationType();
       this.versioningIntent = options.getVersioningIntent();
+      this.staticSummary = options.getStaticSummary();
+      this.staticDetails = options.getStaticDetails();
     }
 
     /**
@@ -303,6 +307,31 @@ public final class ChildWorkflowOptions {
       return this;
     }
 
+    /**
+     * Single-line fixed summary for this workflow execution that will appear in UI/CLI. This can be
+     * in single-line Temporal Markdown format.
+     *
+     * <p>Default is none/empty.
+     */
+    @Experimental
+    public Builder setStaticSummary(String staticSummary) {
+      this.staticSummary = staticSummary;
+      return this;
+    }
+
+    /**
+     * General fixed details for this workflow execution that will appear in UI/CLI. This can be in
+     * Temporal Markdown format and can span multiple lines. This is a fixed value on the workflow
+     * that cannot be updated.
+     *
+     * <p>Default is none/empty.
+     */
+    @Experimental
+    public Builder setStaticDetails(String staticDetails) {
+      this.staticDetails = staticDetails;
+      return this;
+    }
+
     public ChildWorkflowOptions build() {
       return new ChildWorkflowOptions(
           namespace,
@@ -320,7 +349,9 @@ public final class ChildWorkflowOptions {
           typedSearchAttributes,
           contextPropagators,
           cancellationType,
-          versioningIntent);
+          versioningIntent,
+          staticSummary,
+          staticDetails);
     }
 
     public ChildWorkflowOptions validateAndBuildWithDefaults() {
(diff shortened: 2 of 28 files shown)


temporal-sdk/src/main/java/io/temporal/workflow/ChildWorkflowOptions.java
```diff
@@ -303,6 +307,31 @@ public Builder setVersioningIntent(VersioningIntent versioningIntent) {
...
     }
 
+    /**
+     * Single-line fixed summary for this workflow execution that will appear in UI/CLI. This can be
+     * in single-line Temporal Markdown format.
+     *
+     * <p>Default is none/empty.
+     */
+    @Experimental
+    public Builder setStaticSummary(String staticSummary) {
```
cretz, line 317: This is not gonna work until https://github.com/temporalio/temporal/pull/6446 is in a released server version, unsure if you want to wait to expose until then. I do see the test is skipped currently for real servers.

temporal-sdk/src/main/java/io/temporal/workflow/TimerOptions.java
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    /**
+     * Single-line fixed summary for this timer that will appear in UI/CLI. This can be in
+     * single-line Temporal Markdown format.
+     *
+     * <p>Default is none/empty.
+     */
+    @Experimental
+    public Builder setStaticSummary(String staticSummary) {
```
cretz, line 68: We are not calling it "static summary" on a timer, just summary. From a user POV it is unrelated to workflow static summary and details (even if it does use the same data type behind the scenes).

temporal-sdk/src/main/java/io/temporal/workflow/TimerOptions.java
```diff
@@ -0,0 +1,136 @@
...
+
+    /**
+     * General fixed details for this timer that will appear in UI/CLI. This can be in Temporal
+     * Markdown format and can span multiple lines. This is a fixed value on the workflow that
+     * cannot be updated.
+     *
+     * <p>Default is none/empty.
+     */
+    @Experimental
+    public Builder setStaticDetails(String staticDetails) {
```
cretz, line 81: We had not considered "details" for timers to be a thing (even if the underlying data structure in history technically has it).

temporal-sdk/src/test/java/io/temporal/workflow/childWorkflowTests/ChildWorkflowMetadataTest.java
```diff
@@ -0,0 +1,160 @@
...
+import io.temporal.workflow.Workflow;
+import io.temporal.workflow.shared.TestWorkflows.*;
+import java.time.Duration;
+import java.util.List;
+import java.util.stream.Collectors;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+
+public class ChildWorkflowMetadataTest {
```
cretz, line 46: Any integration/e2e tests for regular (i.e. non-child) workflows with static summary/details? Or is that more difficult because it requires test server support?

temporal-sdk/src/main/java/io/temporal/workflow/ChildWorkflowOptions.java
Quinn-With-Two-Ns: Yes this should be in `1.25.0` and the tests are skipped for the test server, not the real server

temporal-sdk/src/test/java/io/temporal/workflow/childWorkflowTests/ChildWorkflowMetadataTest.java
Quinn-With-Two-Ns: This test also covers that, we test the parent workflow as well

temporal-sdk/src/main/java/io/temporal/workflow/TimerOptions.java
Quinn-With-Two-Ns: OK I can leave out for now

temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowInternal.java
```diff
@@ -89,6 +89,11 @@ public static Promise<Void> newTimer(Duration duration) {
     return getWorkflowOutboundInterceptor().newTimer(duration);
   }
 
+  public static Promise<Void> newTimer(Duration duration, TimerOptions options) {
```
Quinn-With-Two-Ns, line 92: @cretz one thing I debated on was if I should include `duration` in `TimerOptions`. We didn't in Go, but I didn't couldn't find a strong reason why?

temporal-sdk/src/main/java/io/temporal/workflow/ChildWorkflowOptions.java
cretz: Hrmm, child workflows may not have made the initial version (they weren't initially wired up). @gow - can you confirm whether or not the child workflow user metadata propagation is in a tagged server version?

temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowInternal.java
cretz: No strong reason. We just figured in Go (and other langs?) that having required duration in both overloaded forms makes sense, and these are just "additional" options beyond duration.

temporal-sdk/src/main/java/io/temporal/workflow/ChildWorkflowOptions.java
Quinn-With-Two-Ns: https://github.com/temporalio/temporal/commit/dd0c6ca01265298483408636da4d52da2b8b327a is part of [release/v1.25.x](https://github.com/temporalio/temporal/compare/release/v1.25.x) and a parent of https://github.com/temporalio/temporal/commit/be848c89192af6c5f9f01477ab2320b6a40d4b67

gow: Yes. It's included in 1.25 release - https://github.com/temporalio/temporal/compare/v1.25.0...main

temporal-sdk/src/main/java/io/temporal/workflow/TimerOptions.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+  }
+
+  private final String staticSummary;
+
+  private TimerOptions(String staticSummary) {
+    this.staticSummary = staticSummary;
+  }
+
+  public String getStaticSummary() {
```
cretz, line 82: When changing the other, didn't remove "Static" from this one

cretz (commented):
One minor thing, then LGTM

cretz (approved):

Issue #2215, opened 2024-09-14
Test Server does not treat internal errors as retryable from Nexus operations 


Quinn-With-Two-Ns:
## Expected Behavior
Test server treats internal errors from nexus operations as retryable

## Actual Behavior
Test server treats internal errors from nexus operations as non-retryable

## Steps to Reproduce the Problem
```
package io.temporal.workflow.nexus;

import static org.junit.Assume.assumeTrue;

import io.nexusrpc.handler.OperationHandler;
import io.nexusrpc.handler.OperationImpl;
import io.nexusrpc.handler.ServiceImpl;
import io.temporal.client.WorkflowFailedException;
import io.temporal.failure.NexusOperationFailure;
import io.temporal.failure.TimeoutFailure;
import io.temporal.testing.internal.SDKTestWorkflowRule;
import io.temporal.workflow.*;
import io.temporal.workflow.shared.TestNexusServices;
import io.temporal.workflow.shared.TestWorkflows;
import java.time.Duration;
import org.junit.Assert;
import org.junit.Before;
import org.junit.Rule;
import org.junit.Test;

public class SyncOperationTimeoutTest extends BaseNexusTest {
  @Rule
  public SDKTestWorkflowRule testWorkflowRule =
      SDKTestWorkflowRule.newBuilder()
          .setWorkflowTypes(TestNexus.class)
          .setNexusServiceImplementation(new TestNexusServiceImpl())
          .build();

  @Override
  protected SDKTestWorkflowRule getTestWorkflowRule() {
    return testWorkflowRule;
  }

  @Test
  public void typedOperationTimeout() {
    TestWorkflows.TestWorkflow1 workflowStub =
        testWorkflowRule.newWorkflowStubTimeoutOptions(TestWorkflows.TestWorkflow1.class);
    WorkflowFailedException exception =
        Assert.assertThrows(WorkflowFailedException.class, () -> workflowStub.execute(""));
    Assert.assertTrue(exception.getCause() instanceof NexusOperationFailure);
    NexusOperationFailure nexusFailure = (NexusOperationFailure) exception.getCause();
    Assert.assertTrue(nexusFailure.getCause() instanceof TimeoutFailure);
    TimeoutFailure timeoutFailure = (TimeoutFailure) nexusFailure.getCause();
    Assert.assertEquals("operation timed out", timeoutFailure.getOriginalMessage());
  }

  public static class TestNexus implements TestWorkflows.TestWorkflow1 {
    @Override
[...]

PR #2221, opened 2024-09-16
Test Server: Fix Nexus operation timeout during retry

pdoerner:
## What was changed

Changed test server logic to not immediately record a failure if the next attempt schedule time would exceed an operation's schedule to close timeout.

## Why?

Closes https://github.com/temporalio/sdk-java/issues/2215

### DIFF
--- a/temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
+++ b/temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
@@ -789,7 +789,12 @@ class StateMachines {
   private static State failNexusOperation(
       RequestContext ctx, NexusOperationData data, Failure failure, long notUsed) {
     RetryState retryState = attemptNexusOperationRetry(ctx, Optional.of(failure), data);
-    if (retryState == RetryState.RETRY_STATE_IN_PROGRESS) {
+    if (retryState == RetryState.RETRY_STATE_IN_PROGRESS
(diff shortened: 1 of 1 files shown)


temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
```diff
@@ -789,7 +789,12 @@ private static void timeoutNexusOperation(
   private static State failNexusOperation(
       RequestContext ctx, NexusOperationData data, Failure failure, long notUsed) {
     RetryState retryState = attemptNexusOperationRetry(ctx, Optional.of(failure), data);
-    if (retryState == RetryState.RETRY_STATE_IN_PROGRESS) {
+    if (retryState == RetryState.RETRY_STATE_IN_PROGRESS
```
Quinn-With-Two-Ns, line 792: Hm but in the test why are we even reaching this point since the error thrown should never fail the operation?

pdoerner: This function is also for handling task failures. I guess maybe @bergundy was right and I should separate out a `FAIL_ATTEMPT` from failing the operation as a whole. But this is how workflow and activity task failures/retries are handled and I though it would be better to stay consistent and refactor them all at the same time.

Quinn-With-Two-Ns (approved):

Issue #2208, opened 2024-09-09
Add `getMetricsScope` to `WorkflowOutboundCallsInterceptor`


Quinn-With-Two-Ns:
**Is your feature request related to a problem? Please describe.**
I want to add custom tags to all metrics emitted in my workflow

**Describe the solution you'd like**
Add `getMetricsScope` to `WorkflowOutboundCallsInterceptor`

PR #2224, opened 2024-09-18
Add getMetricsScope interceptor

Quinn-With-Two-Ns:
Add `getMetricsScope` interceptor in `WorkflowOutboundCallsInterceptor`

closes https://github.com/temporalio/sdk-java/issues/2208

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkflowOutboundCallsInterceptorBase.java
+++ b/temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkflowOutboundCallsInterceptorBase.java
@@ -20,6 +20,7 @@
 
 package io.temporal.common.interceptors;
 
+import com.uber.m3.tally.Scope;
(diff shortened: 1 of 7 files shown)


temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkflowOutboundCallsInterceptorBase.java
```diff
@@ -167,6 +168,11 @@ public void upsertMemo(Map<String, Object> memo) {
     next.upsertMemo(memo);
   }
 
+  @Override
+  public Scope getMetricsScope() {
+    return next.getMetricsScope();
```
cretz, line 173: Took a sec for me to realize `SyncWorkflowContext` already has `getMetricsScope`. May want to toss an `@Override` on it, but no biggie.

cretz (approved):

PR #2230, opened 2024-09-21
Add helper for Nexus to Kotlin Integration

Quinn-With-Two-Ns:
Add helper for Nexus to Kotlin Integration


cretz (approved):
I assume without checking that these are equivalent to little Kotlin helpers done for other options classes

Quinn-With-Two-Ns:
@bergundy not sure if you have any thoughts/want to review these type of Nexus PRs?

Issue #2235, opened 2024-09-23
Add Nexus Support to TestWorkflowExtension


Quinn-With-Two-Ns:
**Is your feature request related to a problem? Please describe.**
Add Nexus Support to TestWorkflowExtension

PR #2238, opened 2024-09-24
Add Nexus support to TestWorkflowExtension

Quinn-With-Two-Ns:
Add Nexus support to TestWorkflowExtension. This PR also adds support to automatically create a Nexus Endpoint in the `TestWorkflowExtension` to make testing easier.

closes: https://github.com/temporalio/sdk-java/issues/2235

### DIFF
--- a/temporal-testing/src/main/java/io/temporal/testing/TestWorkflowExtension.java
+++ b/temporal-testing/src/main/java/io/temporal/testing/TestWorkflowExtension.java
@@ -477,6 +477,7 @@ public class TestWorkflowExtension
      * When set to true the {@link TestWorkflowEnvironment} will not automatically create a Nexus
      * Endpoint. This is useful when you want to manually create a Nexus Endpoint for your test.
      */
+    @Experimental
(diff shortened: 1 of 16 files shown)


temporal-testing/src/main/java/io/temporal/testing/TestWorkflowExtension.java
```diff
@@ -212,13 +217,26 @@ public void beforeEach(ExtensionContext context) {
...
+          }
+          worker.registerWorkflowImplementationTypes(o, wft);
+        });
     worker.registerActivitiesImplementations(activityImplementations);
+    worker.registerNexusServiceImplementation(nexusServiceImplementations);
 
     if (!doNotStart) {
       testEnvironment.start();
     }
+    if (createNexusEndpoint) {
```
Quinn-With-Two-Ns, line 237: I may also want to avoid creating the endpoint if the worker is not started

bergundy (approved):

cretz (approved):
Assuming there's no real cost to creating the endpoint by default, makes sense to me

PR #2239, opened 2024-09-24
Add support for auto discovery of Nexus Services

Quinn-With-Two-Ns:
Add support for auto discovery of Nexus Services and add Nexus worker options

Note to reviewers: I would like to add a couple more tests, but need https://github.com/temporalio/sdk-java/pull/2238 merged to make it easy to setup the Nexus endpoint in the tests


temporal-spring-boot-autoconfigure/src/main/java/io/temporal/spring/boot/NexusServiceImpl.java
```diff
@@ -0,0 +1,48 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Enables the Nexus service bean to be discovered by the Workers auto-discovery. This annotation is
+ * not needed if only an explicit config is used.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.TYPE})
+public @interface NexusServiceImpl {
```
Quinn-With-Two-Ns, line 34: Wasn't sure on the name here. This matches what we did for activity where we use `ActivityImpl`

cretz (approved):

Issue #2132, opened 2024-07-02
Add option to configure the deadline of `ChannelManager#getServerCapabilities`


fextr:
## Expected Behavior

When I [setRpcTimeout](https://github.com/temporalio/sdk-java/blob/abd9f2d1371a0355c01e85bc03763c79b1846ace/temporal-serviceclient/src/main/java/io/temporal/serviceclient/ServiceStubsOptions.java#L774) in WorkflowServiceStubsOptions it overrides the deadline in [getServerCapabilitiesWithRetryOrThrow](https://github.com/temporalio/sdk-java/blob/abd9f2d1371a0355c01e85bc03763c79b1846ace/temporal-serviceclient/src/main/java/io/temporal/serviceclient/SystemInfoInterceptor.java#L95)

## Actual Behavior

I'm enforced to use deprecated [setHealthCheckAttemptTimeout](https://github.com/temporalio/sdk-java/blob/abd9f2d1371a0355c01e85bc03763c79b1846ace/temporal-serviceclient/src/main/java/io/temporal/serviceclient/ServiceStubsOptions.java#L696) because it is still used in the [ChannelManager](https://github.com/temporalio/sdk-java/blob/master/temporal-serviceclient/src/main/java/io/temporal/serviceclient/ChannelManager.java#L327)

## Steps to Reproduce the Problem

1. Set `rpcTimeout` in `WorkflowServiceStubsOptions`.
2. Observe that it does not override the deadline in `getServerCapabilitiesWithRetryOrThrow`
3. Note the usage of the deprecated `healthCheckAttemptTimeout` in `ChannelManager`.

## Specifications

  - Version: Java SDK 1.24.0

Quinn-With-Two-Ns:
`getServerCapabilities` is intentionally overriding the rpc timeout. If you need to configure a different deadline on  `getServerCapabilities` it would need to be a separate option.

Quinn-With-Two-Ns:
Yes, it should be called `getSystemInfoTimeout` though `serverCapabilities` is just an internal name to the Java SDK.

Quinn-With-Two-Ns:
Why do you need to configure the deadline on `getSystemInfo`?

PR #2240, opened 2024-09-25
Add option to configure the deadline of getSystemInfo

Quinn-With-Two-Ns:
Add option to configure the deadline of getSystemInfo like the Go SDK has

closes https://github.com/temporalio/sdk-java/issues/2132


cretz:
What was the issue reusing the health check timeout here?

Quinn-With-Two-Ns:
>What was the issue reusing the health check timeout here?

Yes the health check timeout is deprecated and it wasn't even using the timeout it was using the attempt timeout

cretz (approved):

Issue #2209, opened 2024-09-09
Nexus handler - well known Temporal error translation


bergundy:
Replicate the behavior implemented in this Go SDK PR: https://github.com/temporalio/sdk-go/pull/1626

PR #2242, opened 2024-09-25
Nexus: Well know error translation

Quinn-With-Two-Ns:
Translate some well know errors like what we did with the Go SDK. This does not cover all possible SDK exception, just some we except users to throw in this initial release.

closes https://github.com/temporalio/sdk-java/issues/2209

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/client/WorkflowException.java
+++ b/temporal-sdk/src/main/java/io/temporal/client/WorkflowException.java
@@ -59,7 +59,6 @@ public abstract class WorkflowException extends TemporalException {
         + execution.getWorkflowId()
         + "', runId='"
         + execution.getRunId()
-        + (workflowType == null ? "" : "', workflowType='" + workflowType + '\'')
-        + '}';
+        + (workflowType == null ? "" : "', workflowType='" + workflowType + '\'');
   }
 }
--- a/temporal-sdk/src/main/java/io/temporal/internal/nexus/NexusTaskHandlerImpl.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/nexus/NexusTaskHandlerImpl.java
@@ -30,7 +30,9 @@ import io.nexusrpc.handler.*;
 import io.temporal.api.common.v1.Payload;
 import io.temporal.api.nexus.v1.*;
 import io.temporal.client.WorkflowClient;
+import io.temporal.client.WorkflowException;
 import io.temporal.common.converter.DataConverter;
+import io.temporal.failure.ApplicationFailure;
 import io.temporal.internal.common.NexusUtil;
 import io.temporal.internal.worker.NexusTask;
 import io.temporal.internal.worker.NexusTaskHandler;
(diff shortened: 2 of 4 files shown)


temporal-sdk/src/test/java/io/temporal/workflow/nexus/OperationFailureConversionTest.java
```diff
@@ -0,0 +1,143 @@
...
+
+  @Rule
+  public SDKTestWorkflowRule testWorkflowRule =
+      SDKTestWorkflowRule.newBuilder()
+          .setWorkflowTypes(TestNexus.class)
+          .setNexusServiceImplementation(new TestNexusServiceImpl())
+          .build();
+
+  @Before
+  public void checkRealServer() {
```
Quinn-With-Two-Ns, line 54: @pdoerner The test server does not appear to match the real server's behaviour here

temporal-sdk/src/main/java/io/temporal/client/WorkflowException.java
```diff
@@ -59,7 +59,6 @@ private static String getMessage(WorkflowExecution execution, String workflowTyp
         + execution.getWorkflowId()
         + "', runId='"
         + execution.getRunId()
-        + (workflowType == null ? "" : "', workflowType='" + workflowType + '\'')
-        + '}';
+        + (workflowType == null ? "" : "', workflowType='" + workflowType + '\'');
```
Quinn-With-Two-Ns, line 62: Not sure what this random `}` was for

temporal-sdk/src/test/java/io/temporal/workflow/nexus/OperationFailureConversionTest.java
Quinn-With-Two-Ns: Thanks PJ!

temporal-sdk/src/main/java/io/temporal/internal/nexus/NexusTaskHandlerImpl.java
```diff
@@ -177,12 +179,34 @@ private CancelOperationResponse handleCancelledOperation(
...
+  private void convertKnownFailures(Throwable failure) {
+    if (failure instanceof ApplicationFailure) {
+      if (((ApplicationFailure) failure).isNonRetryable()) {
+        throw new OperationHandlerException(
+            OperationHandlerException.ErrorType.BAD_REQUEST, failure.getMessage());
+      }
+      throw new OperationHandlerException(
+          OperationHandlerException.ErrorType.INTERNAL, failure.getMessage());
+    }
+    if (failure instanceof WorkflowException) {
```
bergundy, line 200: Can you confirm that this roughly maps to `WorkflowExecutionError` or `QueryRejectedError` in Go?

This is fine as an initial first pass at this but I think eventually we may want a different approach. We'll see if we get any feedback on the current approach.

bergundy (approved):

temporal-sdk/src/main/java/io/temporal/internal/nexus/NexusTaskHandlerImpl.java
Quinn-With-Two-Ns: `WorkflowQueryException` and `WorkflowFailedException` extend `WorkflowException`

Quinn-With-Two-Ns: So they are both treated as `BAD_REQUEST`

PR #2244, opened 2024-09-28
Nexus Interceptors

Quinn-With-Two-Ns:
Add support for Nexus worker interceptors and tracing support

Note: Marking as draft as I don't think this is the approach we should take. Since the Nexus SDK controls deserialization the UX of this interceptor is very different from all other interceptors.


temporal-opentracing/src/main/java/io/temporal/opentracing/internal/ContextAccessor.java
```diff
@@ -72,4 +74,20 @@ public SpanContext readSpanContextFromHeader(Header header, Tracer tracer) {
...
+  public Span writeSpanContextToHeader(
+      Supplier<Span> spanSupplier, Map<String, String> header, Tracer tracer) {
+    Span span = spanSupplier.get();
+    writeSpanContextToHeader(span.context(), header, tracer);
+    return span;
+  }
+
+  public void writeSpanContextToHeader(
+      SpanContext spanContext, Map<String, String> header, Tracer tracer) {
+    tracer.inject(spanContext, Format.Builtin.HTTP_HEADERS, new TextMapAdapter(header));
```
Quinn-With-Two-Ns, line 87: We can't use the same serialization strategy we use for other operations because the Headers are only a String to String map.

PR #2257, opened 2024-10-07
[#1727] Use default RetryOptions for LocalActivities when not set

### LINKED ISSUE #1727: inconsistency on MaximumAttempts attribute between local activities and "normal" activities

ghaskins:
Setting setMaximumAttempts=0 (or not setting any value, default = 0) for local activities should make the activity retry forever.

This is currently broken because the logic treats both no policy set and a policy with all default values (including MaxiumAttempts = 0) as abort conditional.  What should happen (at least according the doc and to align with normal activities) is that no policy set should be processed according to the default values of the retry policy, which includes unlimited retries when the MaxiumAttempts = 0.

Fixes #1727

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java
@@ -127,6 +127,14 @@ final class LocalActivityWorker implements Startable, Shutdownable {
     slotQueue.submitAttempt(reservationDat, isRetry, task);
   }
 
+  private RetryOptions getRetryOptions(PollActivityTaskQueueResponseOrBuilder activityTask) {
+    if (isRetryPolicyNotSet(activityTask)) {
+      return RetryOptions.getDefaultInstance();
+    } else {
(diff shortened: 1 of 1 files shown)


Quinn-With-Two-Ns:
Hey @ghaskins I know we discussed this a bit offline, but if we do want to change this we should add some tests (like the one in the reproduction you provided) that this change fixes.

ghaskins:
@Quinn-With-Two-Ns ok, I can take a stab at that.  Normally I would have already submitted them but I was having trouble with the UT framework on my mac (most tests seem to be failing even with unmodified code).  I'll see if I can sort that out.

PR #2258, opened 2024-10-08
Test server support for bidi links

pdoerner:
## What was changed

Added test server support for bidirectional links on Nexus operations
Cleaned up some fields in the DescribeWorkflowExecution response related to callbacks and Nexus operation cancellation.

### DIFF
--- a/temporal-test-server/src/main/java/io/temporal/internal/testservice/LinkConverter.java
+++ b/temporal-test-server/src/main/java/io/temporal/internal/testservice/LinkConverter.java
@@ -1,0 +1,69 @@
+package io.temporal.internal.testservice;
+
+import io.temporal.api.common.v1.Link;
+import io.temporal.api.enums.v1.EventType;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.StringTokenizer;
+
+public class LinkConverter {
+
+  private static final String linkPathFormat = "temporal:///namespaces/%s/workflows/%s/%s/history";
+  private static final String linkQueryFormat =
+      "?eventID=%d&eventType=%s&referenceType=EventReference";
+
+  public static io.temporal.api.nexus.v1.Link WorkflowEventToNexusLink(Link.WorkflowEvent we) {
+    String url =
+        String.format(linkPathFormat, we.getNamespace(), we.getWorkflowId(), we.getRunId());
+    if (we.hasEventRef()) {
+      url +=
+          String.format(
+              linkQueryFormat, we.getEventRef().getEventId(), we.getEventRef().getEventType());
+    }
+    return io.temporal.api.nexus.v1.Link.newBuilder()
+        .setUrl(url)
+        .setType(we.getDescriptorForType().getFullName())
+        .build();
+  }
+
+  public static Link NexusLinkToWorkflowEvent(io.temporal.api.nexus.v1.Link nexusLink) {
+    Link.Builder link = Link.newBuilder();
+    try {
+      URI uri = new URI(nexusLink.getUrl());
+
+      StringTokenizer st = new StringTokenizer(uri.getPath(), "/");
+      st.nextToken(); // /namespaces/
+      String namespace = st.nextToken();
+      st.nextToken(); // /workflows/
+      String workflowID = st.nextToken();
+      String runID = st.nextToken();
+
+      Link.WorkflowEvent.Builder we =
+          Link.WorkflowEvent.newBuilder()
+              .setNamespace(namespace)
+              .setWorkflowId(workflowID)
+              .setRunId(runID);
+
+      if (uri.getQuery() != null) {
+        Link.WorkflowEvent.EventReference.Builder eventRef =
+            Link.WorkflowEvent.EventReference.newBuilder();
+        st = new StringTokenizer(uri.getQuery(), "&");
+        while (st.hasMoreTokens()) {
+          String[] param = st.nextToken().split("=");
+          switch (param[0]) {
+            case "eventID":
+              eventRef.setEventId(Long.parseLong(param[1]));
+              continue;
+            case "eventType":
+              eventRef.setEventType(EventType.valueOf(param[1]));
+          }
+        }
+        we.setEventRef(eventRef);
+      }
+    } catch (URISyntaxException e) {
+      // Swallow un-parsable links since they are not critical to processing
+      return null;
+    }
+    return link.build();
+  }
+}
--- a/temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
+++ b/temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
@@ -20,6 +20,7 @@
 
 package io.temporal.internal.testservice;
 
+import static io.temporal.internal.testservice.LinkConverter.*;
 import static io.temporal.internal.testservice.StateMachines.Action.CANCEL;
 import static io.temporal.internal.testservice.StateMachines.Action.COMPLETE;
 import static io.temporal.internal.testservice.StateMachines.Action.CONTINUE_AS_NEW;
@@ -60,6 +61,7 @@ import io.temporal.api.failure.v1.NexusOperationFailureInfo;
 import io.temporal.api.failure.v1.TimeoutFailureInfo;
 import io.temporal.api.history.v1.*;
 import io.temporal.api.nexus.v1.*;
+import io.temporal.api.nexus.v1.Link;
 import io.temporal.api.protocol.v1.Message;
 import io.temporal.api.query.v1.WorkflowQueryResult;
 import io.temporal.api.taskqueue.v1.StickyExecutionAttributes;
@@ -347,6 +349,7 @@ class StateMachines {
     RetryPolicy retryPolicy = defaultNexusRetryPolicy();
 
     long scheduledEventId = NO_EVENT_ID;
+    Timestamp cancelRequestedTime;
 
     TestServiceRetryState retryState;
     long lastAttemptCompleteTime;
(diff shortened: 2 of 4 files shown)


temporal-test-server/src/main/java/io/temporal/internal/testservice/LinkConverter.java
```diff
@@ -0,0 +1,69 @@
...
+              eventRef.setEventId(Long.parseLong(param[1]));
+              continue;
+            case "eventType":
+              eventRef.setEventType(EventType.valueOf(param[1]));
+          }
+        }
+        we.setEventRef(eventRef);
+      }
+    } catch (URISyntaxException e) {
+      // Swallow un-parsable links since they are not critical to processing
```
Quinn-With-Two-Ns, line 64: hm does the real server skip un-parsable links?

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -2974,6 +2974,13 @@ private DescribeWorkflowExecutionResponse describeWorkflowExecutionInsideLock()
                 .setParentNamespaceId(p.getExecutionId().getNamespace())
                 .setParentExecution(p.getExecutionId().getExecution()));
 
+    if (this.startRequest.getCompletionCallbacksCount() > 0) {}
```
Quinn-With-Two-Ns, line 2977: Is this a typo or is github rendering this wrong?

temporal-test-server/src/main/java/io/temporal/internal/testservice/LinkConverter.java
pdoerner: yes it does

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
pdoerner: typo

temporal-test-server/src/main/java/io/temporal/internal/testservice/LinkConverter.java
bergundy: It does but it logs when this happens.

temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
```diff
@@ -1238,12 +1267,14 @@ private static void startWorkflow(
...
-    HistoryEvent event =
+    HistoryEvent.Builder event =
         HistoryEvent.newBuilder()
             .setEventType(EventType.EVENT_TYPE_WORKFLOW_EXECUTION_STARTED)
-            .setWorkflowExecutionStartedEventAttributes(a)
-            .build();
-    ctx.addEvent(event);
+            .setWorkflowExecutionStartedEventAttributes(a);
+    if (request.getLinksCount() > 0) {
+      event.addAllLinks(request.getLinksList());
```
bergundy, line 1275: We should do the same validation for these links as we do in the real server.

temporal-test-server/src/main/java/io/temporal/internal/testservice/LinkConverter.java
pdoerner: Yeah I didn't see a great way to inject a logger here but I could do some refactoring to include it if it is important.

temporal-test-server/src/main/java/io/temporal/internal/testservice/LinkConverter.java
```diff
@@ -0,0 +1,89 @@
...
+      String workflowID = st.nextToken();
+      String runID = st.nextToken();
+
+      Link.WorkflowEvent.Builder we =
+          Link.WorkflowEvent.newBuilder()
+              .setNamespace(namespace)
+              .setWorkflowId(workflowID)
+              .setRunId(runID);
+
+      if (uri.getQuery() != null) {
```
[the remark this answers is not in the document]
pdoerner: There is not and I didn't want to add a dependency for it.

temporal-test-server/src/main/java/io/temporal/internal/testservice/LinkConverter.java
Quinn-With-Two-Ns: To get a logger you can use `LoggerFactory.getLogger`

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -3143,6 +3159,30 @@ private static PendingNexusOperationState convertNexusOperationState(
     }
   }
 
+  private static NexusOperationCancellationState convertNexusOperationCancellationState(
+      State state, NexusOperationData data) {
+    // Terminal states have already been filtered out, so only handle pending states.
+    if (data.getAttempt() > 1) {
```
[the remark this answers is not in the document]
pdoerner: If attempt is greater than 1 that means we are retrying which I thought meant we always return backing off

temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
pdoerner: I don't think we are doing any validation on these in the real server, did you have something in mind?
https://github.com/temporalio/temporal/blob/main/service/history/historybuilder/history_builder.go#L190

bergundy: I'll track.

temporal-test-server/src/main/java/io/temporal/internal/testservice/LinkConverter.java
pdoerner: Oh right. Been too long since I wrote much Java.

temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
bergundy: I forgot we don't use the Nexus link here, it's a Temporal Link.

bergundy: Still requires validation though.

bergundy: https://github.com/temporalio/temporal/pull/6625

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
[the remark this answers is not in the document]
bergundy: No backing off is for between attempts. If a nexus task is currently scheduled, we should report that correctly.

temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
pdoerner: Added the same for the test server.

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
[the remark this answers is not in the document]
pdoerner: Ah okay, I've updated to track that. It's hard to write functional tests for that logic though because the test server allows time skipping while waiting for retries.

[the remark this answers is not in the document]
bergundy: WDYM updated to track that? I think it's fine if the tests gloss over this detail to some degree. E.g. you can accept either scheduled or backing off.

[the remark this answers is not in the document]
pdoerner: I added a new field to track whether the operation is waiting on retry. So it should be accurate now.

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -250,6 +250,25 @@ private StartWorkflowExecutionRequest.Builder validateStartWorkflowExecutionRequ
...
+      if (request.getLinksCount() > 10) {
+        throw Status.INVALID_ARGUMENT
+            .withDescription(
+                String.format(
+                    "cannot attach more than %d links per request, got %d",
+                    10, request.getLinksCount()))
+            .asRuntimeException();
+      }
+      for (Link l : request.getLinksList()) {
+        if (l.getSerializedSize() > 4000) {
```
bergundy, line 263: Note that I improved the validation here: https://github.com/temporalio/temporal/pull/6629/files#diff-c9a59445e489bd70346630f2773ad69e1723b2298fadbae50ff5c1a424917e1cR4613.
Might as well copy that logic while we're at it.

bergundy (approved):
LGTM. Had one small comment about improving the link validation logic that I would like addressed before merging but I'd also be okay doing a quick followup.

temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
```diff
@@ -846,7 +872,10 @@ private static RetryState attemptNexusOperationRetry(
       ctx.onCommit(
           (historySize) -> {
             data.retryState = nextAttempt;
-            data.nextAttemptScheduleTime = ctx.currentTime().getSeconds();
+            data.isBackingOff = true;
```
bergundy, line 875: You should reset this to false when the backoff timer fires.

pdoerner: I added it to where the timer actually fires here:
https://github.com/temporalio/sdk-java/pull/2258/files#diff-83c1231c017389012549bc7181908362062be65ffabc65ae405e7ad3cdedaa31R2284

bergundy: I just saw it now. Thanks.

bergundy (approved):

Quinn-With-Two-Ns (approved):

PR #2260, opened 2024-10-08
Add some nexus option validation

Quinn-With-Two-Ns:
Add some nexus option validation based on what we did in Go 

https://github.com/temporalio/sdk-go/pull/1661

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflowContext.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflowContext.java
@@ -733,9 +733,9 @@ final class SyncWorkflowContext implements WorkflowContext, WorkflowOutboundCall
   public <R> ExecuteNexusOperationOutput<R> executeNexusOperation(
       ExecuteNexusOperationInput<R> input) {
     Preconditions.checkArgument(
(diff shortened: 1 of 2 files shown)


temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflowContext.java
```diff
@@ -732,6 +732,11 @@ public <R> ChildWorkflowOutput<R> executeChildWorkflow(ChildWorkflowInput<R> inp
   @Override
   public <R> ExecuteNexusOperationOutput<R> executeNexusOperation(
       ExecuteNexusOperationInput<R> input) {
+    Preconditions.checkArgument(
+        input.getEndpoint() != null && !input.getEndpoint().equals(""), "endpoint must be set");
```
cretz, line 736:
```suggestion
        input.getEndpoint() != null && !input.getEndpoint().isEmpty(), "endpoint must be set");
```

Can use `isEmpty` on these

cretz (approved):

temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflowContext.java
Quinn-With-Two-Ns: done

PR #2261, opened 2024-10-08
Change when nexus_task_execution_failed is emitted

Quinn-With-Two-Ns:
Align with https://github.com/temporalio/sdk-go/pull/1664


bergundy (approved):

cretz (approved):

cretz:
Made comment at https://github.com/temporalio/sdk-go/pull/1664#discussion_r1793585032 I didn't think about here. I am concerned that we are confusing operation failure and task failure.

Issue #2203, opened 2024-08-31
Async activity inputs potential memory leak


hansen-jake:
When starting multiple async activities within a workflow, references to the activity inputs are never null’d out so the inputs are never garbage collected, leading to a memory leak. This has caused our workers to run out of memory.

## Expected Behavior

The expectation/assumption is that as activities within a workflow complete, references to their inputs are null’d out, making them eligible for garbage collection.

## Actual Behavior

After running the below code for a while, a heap dump shows that a strong reference to each activity input is held by the class `WorkflowOutboundCallsInterceptor$ActivityInput`, even after each activity completes, causing a memory leak.

## Steps to Reproduce the Problem

See below code for reproduction.

## Specifications

  - Version: Java SDK 1.25.1

```java
// NOT A CONTRIBUTION

public class Main {
    public static void main(String[] args) {
        WorkflowServiceStubs service = WorkflowServiceStubs.newLocalServiceStubs();
        WorkflowClient client = WorkflowClient.newInstance(service);
        WorkerFactory factory = WorkerFactory.newInstance(client);
        Worker worker = factory.newWorker("task_queue");
        worker.registerWorkflowImplementationTypes(EventLoopWorkflowImpl.class);
        worker.registerActivitiesImplementations(new ProcessingActivityImpl(client));
        factory.start();

        EventLoopWorkflow workflow = client.newWorkflowStub(
                EventLoopWorkflow.class,
                WorkflowOptions.newBuilder()
                        .setWorkflowId(UUID.randomUUID().toString())
                        .setTaskQueue("task_queue")
                        .build());

        workflow.startEventLoop();
    }

    @WorkflowInterface
    public interface EventLoopWorkflow {

        @WorkflowMethod
        void startEventLoop();

        @SignalMethod
        void addBatch(List<String> batch);
    }

    public static class EventLoopWorkflowImpl implements EventLoopWorkflow {

        private final ConcurrentLinkedQueue<Supplier<?>> eventQueue = new ConcurrentLinkedQueue<>();

        private final ProcessingActivity activity = Workflow.newActivityStub(
                ProcessingActivity.class,
                ActivityOptions.newBuilder()
                        .setStartToCloseTimeout(Duration.ofHours(24))
                        .build());

        @Override
        public void startEventLoop() {
            var getBatches = Async.procedure(activity::getBatches);

            do {
                Workflow.await(() -> getBatches.isCompleted() || !eventQueue.isEmpty());

                Supplier<?> event;
[...]

PR #2262, opened 2024-10-09
Fix memory leak of command inputs

Quinn-With-Two-Ns:
Fix memory leak of command inputs. Before this change the Java SDK was holding onto the input of various operations (activities, local activities, child workflow) after the operation had already started for two reasons: 
* The state machine was holding a strong reference to the command even after it had been sent to the server
* The output handle needed some of the type informations from the input, but that kept a reference to the input parameters as well

This could cause an OOM error if a lot of async operations were started in a specific workflow.

There is another memory leak in the SDK where the JVM cannot garbage collect complete state machines because they are being held by the cancellation callback in the CancellationScope. This PR does not address this issue because it is significantly more complicated, given the above fixes the impact is much less since only the relatively small amount of data for the statemachine is retained

closes https://github.com/temporalio/sdk-java/issues/2203

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/statemachines/EntityStateMachineInitialCommand.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/statemachines/EntityStateMachineInitialCommand.java
@@ -20,22 +20,17 @@
 
 package io.temporal.internal.statemachines;
 
+import static io.temporal.internal.common.WorkflowExecutionUtils.isCommandEvent;
+
 import io.temporal.api.command.v1.Command;
 import io.temporal.api.enums.v1.CommandType;
+import io.temporal.api.history.v1.HistoryEvent;
 import io.temporal.workflow.Functions;
-import java.lang.ref.WeakReference;
 import javax.annotation.Nullable;
 
 class EntityStateMachineInitialCommand<State, ExplicitEvent, Data>
     extends EntityStateMachineBase<State, ExplicitEvent, Data> {
-
-  // Holds a reference to the command that was created by this state machine.
-  // Used to cancel the command if needed. We use a weak reference to avoid holding
-  // the command in memory longer than needed. This is safe because if no other state machine holds
-  // a reference
-  // to the command then cancelling it wouldn't have any effect.
-  private WeakReference<CancellableCommand> commandRef;
-
+  private CancellableCommand command;
   private long initialCommandEventId;
 
   public EntityStateMachineInitialCommand(
@@ -57,15 +52,22 @@ class EntityStateMachineInitialCommand<State, ExplicitEvent, Data>
     if (command.getCommandType() == CommandType.COMMAND_TYPE_UNSPECIFIED) {
       throw new IllegalArgumentException("unspecified command type");
     }
-    CancellableCommand cancellableCommand = new CancellableCommand(command, this);
-    this.commandRef = new WeakReference<>(cancellableCommand);
-    commandSink.apply(cancellableCommand);
+    this.command = new CancellableCommand(command, this);
+    commandSink.apply(this.command);
   }
 
   protected final void cancelCommand() {
-    if (commandRef.get() != null) {
-      commandRef.get().cancel();
+    if (command != null) {
+      command.cancel();
     }
+  }
+
+  public WorkflowStateMachines.HandleEventStatus handleEvent(
+      HistoryEvent event, boolean hasNextEvent) {
+    if (isCommandEvent(event)) {
+      command = null;
+    }
+    return super.handleEvent(event, hasNextEvent);
   }
 
   protected long getInitialCommandEventId() {
--- a/temporal-sdk/src/main/java/io/temporal/internal/statemachines/EntityStateMachineInitialCommand.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/statemachines/EntityStateMachineInitialCommand.java
@@ -23,12 +23,18 @@ package io.temporal.internal.statemachines;
 import io.temporal.api.command.v1.Command;
 import io.temporal.api.enums.v1.CommandType;
 import io.temporal.workflow.Functions;
+import java.lang.ref.WeakReference;
 import javax.annotation.Nullable;
 
 class EntityStateMachineInitialCommand<State, ExplicitEvent, Data>
     extends EntityStateMachineBase<State, ExplicitEvent, Data> {
 
-  private CancellableCommand command;
+  // Holds a reference to the command that was created by this state machine.
+  // Used to cancel the command if needed. We use a weak reference to avoid holding
+  // the command in memory longer than needed. This is safe because if no other state machine holds
+  // a reference
+  // to the command then cancelling it wouldn't have any effect.
+  private WeakReference<CancellableCommand> commandRef;
 
   private long initialCommandEventId;
 
@@ -51,12 +57,15 @@ class EntityStateMachineInitialCommand<State, ExplicitEvent, Data>
     if (command.getCommandType() == CommandType.COMMAND_TYPE_UNSPECIFIED) {
       throw new IllegalArgumentException("unspecified command type");
     }
-    this.command = new CancellableCommand(command, this);
-    commandSink.apply(this.command);
+    CancellableCommand cancellableCommand = new CancellableCommand(command, this);
+    this.commandRef = new WeakReference<>(cancellableCommand);
+    commandSink.apply(cancellableCommand);
   }
 
   protected final void cancelCommand() {
-    command.cancel();
+    if (commandRef.get() != null) {
+      commandRef.get().cancel();
+    }
   }
 
   protected long getInitialCommandEventId() {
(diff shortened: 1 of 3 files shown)


temporal-sdk/src/main/java/io/temporal/internal/statemachines/EntityStateMachineInitialCommand.java
```diff
@@ -23,12 +23,18 @@
...
 class EntityStateMachineInitialCommand<State, ExplicitEvent, Data>
     extends EntityStateMachineBase<State, ExplicitEvent, Data> {
 
-  private CancellableCommand command;
+  // Holds a reference to the command that was created by this state machine.
+  // Used to cancel the command if needed. We use a weak reference to avoid holding
+  // the command in memory longer than needed. This is safe because if no other state machine holds
+  // a reference
+  // to the command then cancelling it wouldn't have any effect.
+  private WeakReference<CancellableCommand> commandRef;
```
cretz, line 37: Hrmm. Can you clarify "This is safe because if no other state machine holds a reference to the command then cancelling it wouldn't have any effect" a bit? For example, I am reading `TimerStateMachine` and not seeing where another state machine would hold this command.

In general, why is the reference to this overall `EntityStateMachineInitialCommand` being held on to longer than its state is needed? And if the lifetime of `EntityStateMachineInitialCommand` is longer than its lifetime of being cancellable, is there a more explicit way to mark this as no longer cancellable instead of relying on Java references?

Quinn-With-Two-Ns: >Can you clarify "This is safe because if no other state machine holds a reference to the command then cancelling it wouldn't have any effect"

If nothing else is holding a reference to the command then nothing can observe that it was cancelled.

>Why is the reference to this overall EntityStateMachineInitialCommand 

The state machine needs to be held as long as whatever operation is running 

> is there a more explicit way to mark this as no longer cancellable instead of relying on Java references

No, there is nothing generic

cretz: > If nothing else is holding a reference to the command then nothing can observe that it was cancelled

So does this turn cancel into a no-op? Meaning is there a possibility that nothing holds a reference but we still need to issue the cancel command for determinism correctness? Can you demonstrate a case from a workflow author POV where the `TimerStateMachine` is still held but it cannot be canceled? Does this only apply for detached cancellation scope because otherwise all other cancellation scopes are parented to the root and must issue cancels when the root is canceled? (I am not completely familiar with cancellation scope hierarchy in Java SDK)

> The state machine needs to be held as long as whatever operation is running

I guess I am a bit confused where a state machine is running but cannot be cancelled.

I wonder if there's a better way to split off cancellation from the command. It seems that the issue is that the ability to cancel is holding more information than it needs. Can we make cancel only hold the information it needs? Then that even helps people that do still need to cancel things I assume.

Quinn-With-Two-Ns: Cancelling was a no-op before and after this change. This change just means we don't hold only the command ,which may potentially be large, in memory

>I guess I am a bit confused where a state machine is running but cannot be cancelled.

This is the command being canceled not the state machine 

This change has not effect on the correctness of the SDK, it is purely about releasing memory when it is not needed.

cretz: Can you demonstrate a case from a workflow author POV where the `TimerStateMachine` is still held but this `commandRef` reference may be gone? And if `commandRef` is gone, does that mean it cannot be canceled?

> This change has not effect on the correctness of the SDK, it is purely about releasing memory when it is not needed.

Right, I am just a bit concerned relying on Java reference counts now for cancellation just because some input/output state is being held along with it and we want to save memory (instead of separating cancellation from the other memory it holds).

I may just need to see the situation where this weak reference can be removed.

Quinn-With-Two-Ns: If you want to see any example the linked issue has an example, but most async invocation will benefit from this optimization

>The question becomes why does the command have to be all or none? Can cancellation part/ability of a command be separated from the large-memory part of the command?

In principle yes, you would have to introduce some third object the `CancellableCommand` observes which, I think, is just more confusing then the current approach.

>Right, I am just a bit concerned relying on Java reference counts now for cancellation j

We are not relying on Java reference counts now for cancellation, we are relying on Java reference count for memory management

cretz: > If you want to see any example the linked issue has an example, but most async invocation will benefit from this optimization

I didn't see the timer there, but I may be able to set aside some time and fashion one with `Async` `procedure`/`function`. I am not finding docs on it, so to save me from some lookup, can you confirm whether `CancellationScope` automatically propagates to `Async` calls or if they have a detached cancellation scope from the root?

Quinn-With-Two-Ns: Sorry I thought you just wanted an example of a workflow improved by this change, timer is probably the least useful state machine for this issue since its' command doesn't consume much memory. `CancellationScope` 
 does automatically propagates to Async calls thought that I am not sure the relevance here. This change is purely a memory optimization, not behaviour is changed, all state machines tests that cover all state machine transitions continue to pass

cretz: I think my current concern is if I create a cancellation scope, start a timer (that does not finish) in a async procedure, then cancel that cancellation scope for the timer, will the weak reference for this timer state machine be gone and affect whether the command is sent? Meaning is it possible before this PR that the timer cancel was sent but after this PR it won't be because the command is no longer referenced?

This is why I was wanting to see a situation where the weak reference disappears but cancel still succeeds in creating a command after the weak reference was gone. But I may be off base and whether the cancel command is sent for that no-longer-referenced timer may be unrelated to whether this weak reference still exists.

Quinn-With-Two-Ns: If the command is not sent then something else is still referencing the command by definition. Timer cancelation is also covered in the [state machine tests ](https://github.com/temporalio/sdk-java/blob/d1dc2e186ae4c23097de2bb52c3cd6e0ebeb8676/temporal-sdk/src/test/java/io/temporal/internal/statemachines/TimerStateMachineTest.java#L47) we have a workflow test like that for activity as well https://github.com/temporalio/sdk-java/blob/d1dc2e186ae4c23097de2bb52c3cd6e0ebeb8676/temporal-sdk/src/test/java/io/temporal/workflow/activityTests/cancellation/CancellingScheduledActivityTest.java#L87 and it passes

cretz: > we have a workflow test like that for activity as well 

Hrmm, not sure that test covers whether a cancel command is sent (or even if the activity gets scheduled or maybe the signal is sent in the same task). I may write a test just to check. So question, in that test, is the weak reference for the activity cancellable command present by the time this line is reached: https://github.com/temporalio/sdk-java/blob/d1dc2e186ae4c23097de2bb52c3cd6e0ebeb8676/temporal-sdk/src/test/java/io/temporal/workflow/activityTests/cancellation/CancellingScheduledActivityTest.java#L95

I'll try to set some time aside.

Quinn-With-Two-Ns: I think you are confused , cancelling a command does not send a different command.  Again this PR does not change any behaviour it is and we have state machine tests for all of these transitions 

>I'll try to set some time aside to check. If you need to make haste on this, I can mark approved in the meantime though.

Yes please

cretz: > cancelling a command does not send a different command

I was thinking `COMMAND_TYPE_CANCEL_TIMER` command may come through `cancelCommand` somehow. But I suppose not. I have not followed the logic for `WorkflowStateMachine.cancellableCommands` enough to see when the data is no longer referenced.

> Again this PR does not change any behaviour it is and we have state machine tests for all of these transitions

I am not sure we have tests that confirm the behavior across versions for replay safety. My concern was more about version change rather than whether the current code works within itself.

> Yes please

Ok, marking approved.

cretz (approved):
Approving apprehensively just because I am unable to devote enough time to obtaining higher confidence in using weak references in workflow state machines.

temporal-sdk/src/main/java/io/temporal/internal/statemachines/EntityStateMachineInitialCommand.java
Quinn-With-Two-Ns: >I am not sure we have tests that confirm the behavior across versions for replay safety. My concern was more about version change rather than whether the current code works within itself.

So for your peace of mind would a replay test with a  workflow that cancels a timer in the same WFT it is scheduled ?

cretz: Yes, specifically a test that does a replay of older-version workflow history and runs this `cancelCommand` call now with the weak reference unset would give peace of mind (you may have to invoke `System.gc()` manually). But by no means is my peace of mind required :-) If you're confident, you don't have to spend extra time on this. Most of this is my lack of time to devote to understanding the different code paths based on GC status.

PR #2266, opened 2024-10-11
Add Nexus operation link support

Quinn-With-Two-Ns:
Add Nexus operation link support

Note: Will note build yet as I need to do make a new release of the Java Nexus SDK


cretz (approved):

temporal-sdk/src/main/java/io/temporal/internal/nexus/NexusTaskHandlerImpl.java
```diff
@@ -208,15 +214,27 @@ private void convertKnownFailures(Throwable failure) {
...
-      throws InvalidProtocolBufferException {
+      OperationContext.Builder ctx, StartOperationRequest task) {
     ctx.setService(task.getService()).setOperation(task.getOperation());
 
     OperationStartDetails.Builder operationStartDetails =
         OperationStartDetails.newBuilder()
             .setCallbackUrl(task.getCallback())
             .setRequestId(task.getRequestId());
     task.getCallbackHeaderMap().forEach(operationStartDetails::putCallbackHeader);
+    task.getLinksList()
```
Quinn-With-Two-Ns, line 225: @bergundy is it fair to say their is no way to set this right now?

bergundy: Are you referring to setting link from the caller workflow? If so, then yes, the server will set this automatically without an ability to override.

temporal-sdk/src/main/java/io/temporal/nexus/WorkflowRunOperation.java
```diff
@@ -49,7 +56,32 @@ public OperationStartResult<R> start(
...
+      OperationStartResult.Builder<R> result =
+          OperationStartResult.<R>newBuilder().setAsyncOperationId(workflowExec.getWorkflowId());
+      if (nexusLink != null) {
+        result.addLink(nexusProtoLinkToLink(nexusLink));
+      }
+      return result.build();
+    } catch (URISyntaxException e) {
+      // Not expected as the link is constructed by the SDK.
+      throw new OperationHandlerException(
+          OperationHandlerException.ErrorType.BAD_REQUEST, "failed to construct result URL", e);
```
bergundy, line 83: Wouldn't this be an internal error since it originated in the handler code?

temporal-sdk/src/test/java/io/temporal/workflow/nexus/WorkflowOperationLinkingTest.java
```diff
@@ -0,0 +1,139 @@
...
+              .build();
+      TestNexusServices.TestNexusService1 serviceStub =
+          Workflow.newNexusServiceStub(TestNexusServices.TestNexusService1.class, serviceOptions);
+      // Start an asynchronous operation backed by a workflow
+      NexusOperationHandle<String> asyncOpHandle =
+          Workflow.startNexusOperation(serviceStub::operation, input);
+      NexusOperationExecution asyncExec = asyncOpHandle.getExecution().get();
+      // Signal the operation to unblock, this makes sure the operation doesn't complete before the
+      // operation
+      // started event is written to history
```
bergundy, line 96: You can add a TODO to remove this once the server properly handles this race.

bergundy (commented):
I don't see a link being added to the StartWorkflowExecution request (which will add link the WorkflowExecutionStarted event to the caller's operation), you've only attached the back-link.

temporal-sdk/src/test/java/io/temporal/workflow/nexus/WorkflowOperationLinkingTest.java
Quinn-With-Two-Ns: Is there a github issue for this?

temporal-sdk/src/main/java/io/temporal/nexus/WorkflowRunOperation.java
Quinn-With-Two-Ns: Yeah I think you are right.

Quinn-With-Two-Ns:
>I don't see a link being added to the StartWorkflowExecution request 

A good catch fixed that

temporal-sdk/src/test/java/io/temporal/workflow/nexus/WorkflowOperationLinkingTest.java
bergundy: Unfortunately no, we're doing most of this tracking with an internal tool. In any case we'll need to wait for the next server release.
@pdoerner I realized you'll need to fix this race in the test server too when you get to it but let's start with the real server.

temporal-sdk/src/main/java/io/temporal/client/WorkflowOptions.java
```diff
@@ -444,6 +449,17 @@ public Builder setCompletionCallbacks(List<Callback> completionCallbacks) {
       return this;
     }
 
+    /**
+     * Links to be associated with the workflow.
+     *
+     * <p>WARNING: Not intended for User Code.
+     */
+    @Experimental
+    public Builder setLinks(List<Link> links) {
```
bergundy, line 458: Works for me as long as it's experimental. @cretz do you have a concern with adding this public API?

bergundy (approved):

temporal-sdk/src/main/java/io/temporal/client/WorkflowOptions.java
Quinn-With-Two-Ns: Yeah ideally I wouldn't expose it, but there is not good way to hide this given the package layout of the Java SDK

PR #2278, opened 2024-10-17
Add Nexus Worker interceptor

Quinn-With-Two-Ns:
Add Nexus Worker interceptor based on the Nexus Java SDK interceptors proposed here https://github.com/nexus-rpc/sdk-java/pull/14

### DIFF
--- a/temporal-opentracing/src/main/java/io/temporal/opentracing/SpanOperationType.java
+++ b/temporal-opentracing/src/main/java/io/temporal/opentracing/SpanOperationType.java
@@ -34,7 +34,10 @@ public enum SpanOperationType {
   UPDATE_WORKFLOW("UpdateWorkflow"),
   HANDLE_QUERY("HandleQuery"),
   HANDLE_SIGNAL("HandleSignal"),
-  HANDLE_UPDATE("HandleUpdate");
+  HANDLE_UPDATE("HandleUpdate"),
+  EXECUTE_NEXUS_OPERATION("ExecuteNexusOperation"),
+  START_NEXUS_OPERATION("StartNexusOperation"),
+  CANCEL_NEXUS_OPERATION("CancelNexusOperation");
 
   private final String defaultPrefix;
 
--- a/temporal-opentracing/src/main/java/io/temporal/opentracing/SpanOperationType.java
+++ b/temporal-opentracing/src/main/java/io/temporal/opentracing/SpanOperationType.java
@@ -35,9 +35,9 @@ public enum SpanOperationType {
   HANDLE_QUERY("HandleQuery"),
   HANDLE_SIGNAL("HandleSignal"),
   HANDLE_UPDATE("HandleUpdate"),
-  EXECUTE_NEXUS_OPERATION("ExecuteNexusOperation"),
   START_NEXUS_OPERATION("StartNexusOperation"),
-  CANCEL_NEXUS_OPERATION("CancelNexusOperation");
+  RUN_START_NEXUS_OPERATION("RunStartNexusOperationHandler"),
+  RUN_CANCEL_NEXUS_OPERATION("RunCancelNexusOperationHandler");
 
   private final String defaultPrefix;
 
--- a/temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowOutboundCallsInterceptor.java
+++ b/temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowOutboundCallsInterceptor.java
@@ -30,6 +30,8 @@ import io.temporal.opentracing.OpenTracingOptions;
 import io.temporal.workflow.Workflow;
 import io.temporal.workflow.WorkflowInfo;
 import io.temporal.workflow.unsafe.WorkflowUnsafe;
+import java.util.HashMap;
+import java.util.Map;
 
 public class OpenTracingWorkflowOutboundCallsInterceptor
     extends WorkflowOutboundCallsInterceptorBase {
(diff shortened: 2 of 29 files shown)


Quinn-With-Two-Ns:
Note: CI will fail until a new Nexus Java SDK is released

temporal-opentracing/src/main/java/io/temporal/opentracing/SpanOperationType.java
```diff
@@ -34,7 +34,10 @@ public enum SpanOperationType {
   UPDATE_WORKFLOW("UpdateWorkflow"),
   HANDLE_QUERY("HandleQuery"),
   HANDLE_SIGNAL("HandleSignal"),
-  HANDLE_UPDATE("HandleUpdate");
+  HANDLE_UPDATE("HandleUpdate"),
+  EXECUTE_NEXUS_OPERATION("ExecuteNexusOperation"),
+  START_NEXUS_OPERATION("StartNexusOperation"),
+  CANCEL_NEXUS_OPERATION("CancelNexusOperation");
```
cretz, line 40: To match other span names, I wonder if this should be

```suggestion
  START_NEXUS_OPERATION("StartNexusOperation"),
  RUN_NEXUS_OPERATION("RunNexusOperation"),
  CANCEL_NEXUS_OPERATION("CancelNexusOperation");
```

It seems we use start for outbound things and run for inbound things. I would usually not consider this a blocker, but we copy these names in the other SDKs too.

cretz (commented):
Looks great, only one minor span naming thing worth discussing

temporal-opentracing/src/main/java/io/temporal/opentracing/SpanOperationType.java
Quinn-With-Two-Ns: Yes I was inconsistent here since the handler side is called `startOperation`

Quinn-With-Two-Ns: in the future we will have types like `GetResult` or `GetInfo` so I think the types here should reflect the method being called

Quinn-With-Two-Ns: But unlike activities the Nexus operation is not being "Run" in this span, it is simply being started. if we wanted to keep the  `Run` prefix I would be OK with `StartNexusOperation ` -> `RunStartNexusOperation `

cretz: I think the same thing happens to child workflows though, especially with sticky disabled, and especially in SDKs that do not recreate this span during replay. The operation is being run, it just may say it's not completing yet. I think we would have to use at least `RunStartNexusOperation` since `StartNexusOperation` is for the caller side.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/SpanFactory.java
```diff
@@ -59,6 +59,18 @@ public Tracer.SpanBuilder createWorkflowStartSpan(
     return createSpan(context, tracer, null, References.FOLLOWS_FROM);
   }
 
+  public Tracer.SpanBuilder createStartNexusOperationSpan(
+      Tracer tracer, String serviceName, String operationName, String workflowId, String runId) {
+    SpanCreationContext context =
+        SpanCreationContext.newBuilder()
+            .setSpanOperationType(SpanOperationType.START_NEXUS_OPERATION)
+            .setActionName(serviceName + "." + operationName)
```
bergundy, line 67: I wonder if it'd be better to use a slash for the delimiter. That's the standard for gRPC services and likely will be the same for Nexus.

temporal-opentracing/src/test/java/io/temporal/opentracing/NexusOperationTest.java
```diff
@@ -0,0 +1,198 @@
+/*
+ * Copyright (C) 2022 Temporal Technologies, Inc. All Rights Reserved.
```
bergundy, line 2: Don't you want to update the year?

temporal-sdk/src/main/java/io/temporal/common/interceptors/NexusOperationInboundCallsInterceptor.java
```diff
@@ -0,0 +1,117 @@
...
+
+  /**
+   * Intercepts a call to start a Nexus operation.
+   *
+   * @param input input to the operation start.
+   * @return result of the operation start.
+   * @throws OperationUnsuccessfulException if the operation start failed.
+   */
+  StartOperationOutput startOperation(StartOperationInput input)
+      throws OperationUnsuccessfulException;
```
bergundy, line 108: I wonder if this shouldn't follow the operation handler interface but I understand that what you have here is more consistent with Temporal interceptors.

bergundy (approved):

cretz (approved):

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowOutboundCallsInterceptor.java
```diff
@@ -100,6 +100,25 @@ public <R> ChildWorkflowOutput<R> executeChildWorkflow(ChildWorkflowInput<R> inp
...
+      Span nexusOperationExecuteSpan =
+          contextAccessor.writeSpanContextToHeader(
+              () -> createStartNexusOperationSpanBuilder(input).start(),
+              input.getHeaders(),
+              tracer);
+      try (Scope ignored = tracer.scopeManager().activate(nexusOperationExecuteSpan)) {
+        return super.executeNexusOperation(input);
+      } finally {
+        nexusOperationExecuteSpan.finish();
+      }
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: After the request to schedule it has been sent, we do the same thing for activities and child workflows in the Java SDK open tracing interceptor

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/SpanFactory.java
cretz: I think this may be an opportunity for Nexus spec to define what a single-string qualified operation name looks like. I think I like slash because that represents how it looks in a URL. I also think we should consider disallowing slashes (and maybe other non-url-safe characters) in operation names, but that's not that important.

Quinn-With-Two-Ns: Going with `/` for now, can resist since we are still not GA and this is the open tracing interceptor

Issue #2285, opened 2024-10-22
Workflow update with start does not pass Workflow arguments 


Quinn-With-Two-Ns:
## Expected Behavior
`WorkflowClient.updateWithStart` passes workflow arguments 

## Actual Behavior
`WorkflowClient.updateWithStart` does not pass workflow arguments 

## Steps to Reproduce the Problem
```java
  @Test
  public void startVariousFuncs() throws ExecutionException, InterruptedException {
    WorkflowClient workflowClient = testWorkflowRule.getWorkflowClient();

    BiFunction<Functions.Func1<Integer, String>, Integer, UpdateWithStartWorkflowOperation<String>>
        newUpdateOp =
            (request, input) ->
                UpdateWithStartWorkflowOperation.newBuilder(request, input)
                    .setWaitForStage(WorkflowUpdateStage.COMPLETED)
                    .build();

    // no arg
    TestMultiArgWorkflowFunctions.TestNoArgsWorkflowFunc stubF =
        workflowClient.newWorkflowStub(
            TestMultiArgWorkflowFunctions.TestNoArgsWorkflowFunc.class, createOptions());
    UpdateWithStartWorkflowOperation<String> updateOp0 = newUpdateOp.apply(stubF::update, 0);
    WorkflowUpdateHandle<String> handle0 = WorkflowClient.updateWithStart(stubF::func, updateOp0);

    // 1 arg
    TestMultiArgWorkflowFunctions.Test1ArgWorkflowFunc stubF1 =
        workflowClient.newWorkflowStub(
            TestMultiArgWorkflowFunctions.Test1ArgWorkflowFunc.class, createOptions());
    UpdateWithStartWorkflowOperation<String> updateOp1 = newUpdateOp.apply(stubF1::update, 1);
    WorkflowUpdateHandle<String> handle1 =
        WorkflowClient.updateWithStart(stubF1::func1, "1", updateOp1);

    // 2 args
    TestMultiArgWorkflowFunctions.Test2ArgWorkflowFunc stubF2 =
        workflowClient.newWorkflowStub(
            TestMultiArgWorkflowFunctions.Test2ArgWorkflowFunc.class, createOptions());
    UpdateWithStartWorkflowOperation<String> updateOp2 = newUpdateOp.apply(stubF2::update, 2);
    WorkflowUpdateHandle<String> handle2 =
        WorkflowClient.updateWithStart(stubF2::func2, "1", 2, updateOp2);

    // 3 args
[...]

PR #2286, opened 2024-10-22
Fix UpdateWithStart workflow args

stephanos:
## What was changed

Pass workflow args - instead of update args - to workflow start.

## Why?

Bug.

## Checklist

1. Closes https://github.com/temporalio/sdk-java/issues/2285

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/client/UpdateWithStartWorkflowOperation.java
+++ b/temporal-sdk/src/main/java/io/temporal/client/UpdateWithStartWorkflowOperation.java
@@ -278,6 +278,8 @@ public final class UpdateWithStartWorkflowOperation<R> {
 
   private Object[] updateArgs;
 
+  private Object[] workflowArgs;
+
   private final CompletableFuture<WorkflowUpdateHandle<R>> handle;
 
   private final Functions.Proc request;
@@ -296,7 +298,7 @@ public final class UpdateWithStartWorkflowOperation<R> {
     try {
       request.apply();
       workflow.apply();
-      stub.updateWithStart(this, this.updateArgs);
+      stub.updateWithStart(this, this.workflowArgs);
       return this.handle.get();
     } catch (InterruptedException e) {
       Thread.currentThread().interrupt();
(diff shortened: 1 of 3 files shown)


temporal-sdk/src/main/java/io/temporal/client/WorkflowInvocationHandler.java
```diff
@@ -490,7 +490,7 @@ public void invoke(
           throw new IllegalArgumentException(
               "Method '" + method.getName() + "' is not a WorkflowMethod");
         }
-        this.operation.prepareStart(untyped);
+        this.operation.prepareStart(untyped, args);
         state = State.UPDATE_RECEIVED;
```
Quinn-With-Two-Ns, line 494: Do these enum names make sense? isn't the state now `START_RECEIVED`?

temporal-sdk/src/main/java/io/temporal/client/UpdateWithStartWorkflowOperation.java
```diff
@@ -278,6 +278,8 @@ public static <R> Builder<R> newBuilder(String updateName, Class<R> resultClass,
 
   private Object[] updateArgs;
 
+  private Object[] workflowArgs;
```
Quinn-With-Two-Ns, line 281: Can we add a comment here explaining where this is set and when? I think the current code isn't super obvious how this gets set since the AsyncInvocation is hard to follow.

temporal-sdk/src/main/java/io/temporal/client/WorkflowInvocationHandler.java
stephanos: Yup, I noticed this was wrong, too. It's fixed in https://github.com/temporalio/sdk-java/pull/2288/files#diff-62667c49d1f873a13d0894606bc2163f4db202d4f351c83bc1cf8c141e9342a8R481

temporal-sdk/src/main/java/io/temporal/client/UpdateWithStartWorkflowOperation.java
stephanos: I agree, it took me a bit to re-figure it out myself.

Quinn-With-Two-Ns: Yeah I think long term we should move this logic out of `UpdateWithStartWorkflowOperation.java` into its own class internal class that just deals with the the async invocation, but for now this is fine

Quinn-With-Two-Ns (approved):

PR #2288, opened 2024-10-22
Fix UpdateWithStart untyped operation

stephanos:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


temporal-sdk/src/main/java/io/temporal/client/UpdateWithStartWorkflowOperation.java
```diff
@@ -280,22 +280,24 @@ public static <R> Builder<R> newBuilder(String updateName, Class<R> resultClass,
 
   private final CompletableFuture<WorkflowUpdateHandle<R>> handle;
 
-  private final Functions.Proc request;
+  private final Functions.Proc updateRequest;
```
Quinn-With-Two-Ns, line 283: nit: Would be good to annotate this fields with @Nullable

Quinn-With-Two-Ns (approved):

PR #2294, opened 2024-10-24
Update gradle 7.6.1 -> 8.10.2

Quinn-With-Two-Ns:
Update gradle 7.6.1 -> 8.10.2


cretz:
Makes sense, but curious what the impetus was to upgrade

Quinn-With-Two-Ns:
We will need to upgrade to support Java 21 for virtual thread support

https://docs.gradle.org/current/userguide/compatibility.html

cretz:
This is just my curiosity now (approving anyways), but is that chart basically saying "you have to have this Java version for running this Gradle version" or "you have to have this Gradle version for running/compiling with this Java version"? It's a bit surprising older Gradle that just delegates to `javac` would care what version of `javac` is in use or the source compatibility given to it, but I guess there are Java version-specific things Gradle needs to know about.

cretz (approved):

Quinn-With-Two-Ns:
>This is just my curiosity now (approving anyways), but is that chart basically saying "you have to have this Java version for running this Gradle version" or "you have to have this Gradle version for running/compiling with this Java version"?

These charts show the latest java version you can use , not the min. The first column shows what you can build your source with , the second shows what you can run gradle with

Issue #2299, opened 2024-10-28
Springboot workflow implementation with non default constructors fail in 1.26.0


Quinn-With-Two-Ns:
## Expected Behavior 
Workflow implementation with non default constructors succeeds as Spring can autowires the parameters 

## Actual Behavior
Workflow implementation with non default constructors fail
```
java.lang.IllegalArgumentException: No default constructor or constructor annotated with @WorkflowInit found: 
```

## Steps to Reproduce the Problem

Have a workflow implementation with a constructor with parameters that can be auto wired and no default constructor 

## Specifications

  - Version: 1.26.0

PR #2300, opened 2024-10-28
Fix workflow implementation in springboot failing if no default const…

Quinn-With-Two-Ns:
Fix workflow implementation in springboot failing if no default constructor is present. Unfortunately the validation we added [in](https://github.com/temporalio/sdk-java/commit/21d15aef0294f0845be9af179bbbb82d6d277650) to require a default public constructor is not a valid assertion if a user is using our springboot integration because workflow implementation objects are constructed with the `BeanFactory`.

closes https://github.com/temporalio/sdk-java/issues/2299


cretz (approved):
Note for .NET we disallow DI for workflows since it is usually a smell that you are using something whose lifetime or options you don't control inside a workflow where you need to be very much in control of the lifetimes and options of things you use for determinism reasons.

Granted if people have not already been using this for worker-specific configuration beans and such, maybe it's not as much of an issue here.

Quinn-With-Two-Ns:
Yeah I wish we didn't allow it, but there are reasonable uses cases and we already have so I can't change it now

Issue #335, opened 2021-02-06
Support workflow metadata query


mfateev:
**Is your feature request related to a problem? Please describe.**
Currently, it is not possible to get a list of queries and signals a workflow is listening on. So UI has to emit an unsupported query and parse the list of supported ones from the message. 

**Describe the solution you'd like**
Add a query that would return metadata about workflow. Lists of supported queries and signals should be included into the result.

**Describe alternatives you've considered**
Use worker query (not implemented yet) instead. But as the set of queries and signals that workflow accepts is dynamic a normal query would make more sense.

**Additional context**
[An error is logged by the service ](https://community.temporal.io/t/io-temporal-internal-sync-syncworkflowcontext-query-uncategorized-exception-often-seen-in-server-log/1454)every time UI needs to find out the list of supported queries.

The query should be supported by all SDKs.

cretz:
See https://github.com/temporalio/features/issues/51

PR #2301, opened 2024-10-30
Add workflow metadata query

Quinn-With-Two-Ns:
Add workflow metadata query

https://github.com/temporalio/sdk-java/issues/335


cretz (approved):
Looks great

PR #2313, opened 2024-11-08
Increase timeout of canRunWithResourceBasedTuner

Quinn-With-Two-Ns:
Increase timeout of canRunWithResourceBasedTuner to try to fix flakes like seen here https://github.com/temporalio/sdk-java/actions/runs/11802877846/job/32883424690


cretz (approved):
Hrmm, so that test before failed after 10s but the same test in this CI run succeeded after only 3s. My guess is that whatever caused the 10s hang will cause a 5m hang too, but no harm in trying I suppose.

PR #2319, opened 2024-11-14
(Breaking Change) Generate update ID at call time if not set

Quinn-With-Two-Ns:
Current the SDK generates an Update ID, if not is set, when `UpdateOptionsBuilder.build` is called. This PR changes the behaviour so the Update ID is generated when the update request is made. This is done to:
* Match the behaviour of `WorkflowID` in `WorkflowOptions`
* Match other SDK's behaviour
* Prevent the confusing case where reusing an option without an ID set causes one one update request when the customer tried to make two, see the test `updateWorkflowReuseOptions`.


cretz (approved):
Makes sense to delay this until inside of root interceptor

Quinn-With-Two-Ns:
@cretz I set the ID before the interceptor so the interceptor has access to the ID, this is important for things like tracing interceptors . This is also consistent with the rest of the Java SDK.

cretz:
Hrmm, ok. Some other SDKs do differently, but I agree it's more important to be intra-SDK consistent here than inter-SDK consistent.

Issue #2265, opened 2024-10-10
Add high-level workflow describe


cretz:
**Describe the solution you'd like**

* Add high-level workflow describe call
  * Probably `describe` on the workflow stub
* Return a `WorkflowExecutionDescription` that contains at least the things list does and access to the raw describe response for now
  * In other SDKs, this would be an extension of the `WorkflowExecutionMetadata` class since it has all of that and more
* Make sure it is interceptable

See TypeScript, Python, and .NET SDKs if looking for design inspiration for this call, or reference https://github.com/temporalio/sdk-go/issues/1667.

PR #2326, opened 2024-11-20
Add high-level workflow describe

Quinn-With-Two-Ns:
closes https://github.com/temporalio/sdk-java/issues/2265

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/client/WorkflowExecutionDescription.java
+++ b/temporal-sdk/src/main/java/io/temporal/client/WorkflowExecutionDescription.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright (C) 2022 Temporal Technologies, Inc. All Rights Reserved.
+ *
+ * Copyright (C) 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Modifications copyright (C) 2017 Uber Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this material except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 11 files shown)


temporal-sdk/src/main/java/io/temporal/client/WorkflowExecutionDescription.java
```diff
@@ -0,0 +1,41 @@
...
+
+import io.temporal.api.workflowservice.v1.DescribeWorkflowExecutionResponse;
+import io.temporal.common.converter.DataConverter;
+import javax.annotation.Nonnull;
+
+/** Contains information about a workflow execution. */
+public class WorkflowExecutionDescription extends WorkflowExecutionMetadata {
+  private final @Nonnull DescribeWorkflowExecutionResponse response;
+
+  public WorkflowExecutionDescription(
```
cretz, line 31: Can we make this constructor package private like its super class? I assume this is public because it is interceptable? Not sure the describe call needs to be interceptable if the list call isn't. Or if we want to make the list call interceptable, we can and then expose the constructor of `WorkflowExecutionMetadata`.

Quinn-With-Two-Ns: Yes it is public because it is intercept-able, It was a requirement on the issue. The list call is not intercept-able right now but it makes sense to open an issue to add that as well and expose the constructor of WorkflowExecutionMetadata, but I would prefer to do that in a separate issue.

Quinn-With-Two-Ns: Opened an issue to track making list intercept-able https://github.com/temporalio/sdk-java/issues/2328

PR #2332, opened 2024-11-27
[WIP] Fix CI after GHA's drop of node16 actions

### LINKED ISSUE #2331: [Bug] CI build will soon get broken due to GHA dropping support for node16-based actions
### Describe the bug

GitHub Action’s just started enforcing the long-announced EOL on Node16-based actions. Basically, the latest release of GHA’s Runner agent ignores the fact that an action is requesting node16, and instead just calls node20 anyway. 

Until now, we were relying on an older GLIBC 2.27-based runner to ensure that the native artifacts of the Test Server would run on some older Linux distros that some of our customers were still using.

However, given that Node20 is incompatible with GLIBC 2.27 or earlier, this should no longer work. We may get some extra time to resolve this issue, as it is believed that the buildjet runner image we currently use may not be immediately updated to the newer GHA runner software, but it is expected that our release workflows may start failing in the very near future.

mjameswh:
## What was changed

- Modified our GHA workflows so that when building the native executable of the Test Server on Linux targets, only the actual compilation step runs inside an older GLIBC 2.17-based Docker container; all other operations are performed directly on the runner, which can now be any recent Ubuntu.  

  This was required because GHA no longer allows actions based on Node16 (i.e. actions are forced to use Node20 instead), but Node20 requires at least GLIBC 2.28. It is therefore not possible to have a single execution environment that can both execute Node20-based actions, yet produces native binaries that can then execute on the GLIBC 2.17-based environments that some of our users have.

  Fixes #2331.

### DIFF
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -128,7 +128,7 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - runner: buildjet-2vcpu-ubuntu-1804
+          - runner: ubuntu-latest
             os_family: linux
             arch: amd64
           - runner: macos-13
@@ -138,16 +138,9 @@ jobs:
             os_family: windows
             arch: amd64
     runs-on: ${{ matrix.runner }}
-    env:
-      # This is required to allow continuing usage of Node 16 for actions,
-      # as Node 20 won't run on the builder we use for linux builds
-      # (Node 20 require glibc 2.28+, but ubuntu-1804 has glibc 2.27).
-      # https://github.blog/changelog/2024-05-17-updated-dates-for-actions-runner-using-node20-instead-of-node16-by-default/
-      ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: true
     steps:
       - name: Checkout repo
-        # FIXME: v4+ requires Node 20
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
           ref: ${{ env.INPUT_REF }}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/prepare-release.yml
```diff
@@ -156,25 +149,20 @@ jobs:
...
-          java-version: "11"
-          distribution: "temurin"
-
-      - name: Set up Gradle
-        # FIXME: v3+ requires Node 20
-        uses: gradle/gradle-build-action@v2
-      
       - name: Build native test server
-        run: ./gradlew :temporal-test-server:build
+        run: |
```
Quinn-With-Two-Ns, line 153: Is there a reason you didn't run this task in a `container`?

.github/workflows/prepare-release.yml
```diff
@@ -156,25 +149,26 @@ jobs:
...
-        uses: gradle/gradle-build-action@v2
-      
-      - name: Build native test server
-        run: ./gradlew :temporal-test-server:build
+      - name: Build native test server (Docker)
+        if: matrix.os_family == 'Linux'
+        run: |
+          docker run \
+            --rm -v "$(pwd):/workspace" -w /workspace \
+            ghcr.io/graalvm/graalvm-community:21-ol7 \
```
Quinn-With-Two-Ns, line 162: Also the graal compiler should come from our plugin not the environment so why did you need to use a graal docker image here?

.github/workflows/prepare-release.yml
Quinn-With-Two-Ns: Wouldn't this step still make sense to run in a container?

mjameswh: The `container` parameter only exists at the job-level. Steps run with whatever runner/container was set by the job.

A possible alternative is to use a docker image as an action, but there are a few limitations to that approach. But trying it that now to see if that would work.

.github/workflows/prepare-release.yml
mjameswh: Replaced by `openjdk:21-jdk-oraclelinux7`.

PR #2337, opened 2024-11-29
New Update-with-Start API

stephanos:
## What was changed

Changed the Update-with-Start API to use a `WithStartWorkflowOperation` that expresses the workflow start.


temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
```diff
@@ -1153,7 +1153,16 @@ public void executeMultiOperation(
...
+            null); // update aborted
+      }
+
+      if (startRequest.hasWorkflowStartDelay()) {
+        throw multiOperationExecutionFailure(
+            MultiOperationExecutionFailure.OperationStatus.newBuilder()
+                .setCode(Status.INVALID_ARGUMENT.getCode().value())
+                .setMessage("INVALID_ARGUMENT: WorkflowStartDelay is not supported.")
+                .build(),
+            null); // update aborted
```
stephanos, line 1165: Moved this up to the other Start Operation validation.

temporal-sdk/src/main/java/io/temporal/client/WorkflowStubImpl.java
```diff
@@ -131,45 +131,71 @@ public WorkflowExecution start(Object... args) {
...
+          "Required parameter WorkflowIdConflictPolicy in WorkflowOptions is missing in WorkflowStub");
     }
-    checkExecutionIsNotStarted();
+    if (!startOp.markInvoked()) {
+      throw new IllegalStateException("WithStartWorkflowOperation was already executed");
+    }
+    if (startOp.getStub() == null) {
+      throw new IllegalStateException(
+          "WithStartWorkflowOperation was created with different WorkflowStub");
+    }
```
stephanos, line 152: So this is worth discussing: the Start Operation's `getResult` needs a class for the Workflow's result type to work. When using the typed stub to initialize it, it is inferred. When using the untyped, it is provided by the user explicitly. However, when initialize it via the typed stub _and_ using the untyped stub to run startUpdateWithStart, the type inference step was skipped and any call to `startOp.getResult()` fails. I don't see a way around this other than removing the `getResult` shortcut entirely. Then this check can be safely removed.

temporal-sdk/src/main/java/io/temporal/client/WorkflowInvocationHandler.java
```diff
@@ -91,17 +83,18 @@ static <T> void initAsyncInvocation(InvocationType type, T value) {
...
+      NexusStartWorkflowRequest request = (NexusStartWorkflowRequest) values[0];
       invocationContext.set(new StartNexusOperationInvocationHandler(request));
     } else if (type == InvocationType.UPDATE) {
-      UpdateOptions<?> updateOptions = (UpdateOptions) value;
+      UpdateOptions<?> updateOptions = (UpdateOptions<?>) values[0];
       invocationContext.set(new UpdateInvocationHandler(updateOptions));
     } else if (type == InvocationType.UPDATE_WITH_START) {
-      UpdateWithStartWorkflowOperation operation = (UpdateWithStartWorkflowOperation) value;
-      invocationContext.set(new UpdateWithStartInvocationHandler(operation));
+      UpdateOptions<?> options = (UpdateOptions<?>) values[0];
```
Quinn-With-Two-Ns, line 95: I would make a module private type to pass these two values over making this whole function take a vararg

temporal-sdk/src/main/java/io/temporal/client/WorkflowStubImpl.java
Quinn-With-Two-Ns: I think it is worth considering if the untyped stub `startUpdateWithStart` API should even take a `WithStartWorkflowOperation` ?

Quinn-With-Two-Ns: I guess `startUpdateWithStart ` is supposed to take a `vargs` , and you can't have two varags

stephanos: Yup, @dandavison and I had that conversation and we were leaning towards using the `WithStartWorkflowOperation` for the untyped `startUpdateWithStart` API for these reasons:

(1) a signature like `startUpdateWithStart(UpdateOptions<R> options, Object[] updateArgs, Object[] startArgs)` is potentially dangerous because users might mix up the order of the args (counterpoint: the existing `signalWithStart` has the same problem)
(2) consistent use of the `startOp` parameter across typed and untyped APIs
(3) allows convenience function `getResult()`

(1) is the main reason, (2) and (3) are minor reasons.

Having said that, neither of us - so far - felt very strongly about it.

temporal-sdk/src/main/java/io/temporal/client/WorkflowStubImpl.java
```diff
@@ -131,45 +131,71 @@ public WorkflowExecution start(Object... args) {
...
     if (options == null) {
-      throw new IllegalStateException("Required parameter WorkflowOptions is missing");
+      throw new IllegalStateException(
+          "Required parameter WorkflowOptions is missing in WorkflowStub");
     }
-    if (!updateOperation.markInvoked()) {
-      throw new IllegalStateException("UpdateWithStartWorkflowOperation was already executed");
+    if (options.getWorkflowIdConflictPolicy() == null) {
+      throw new IllegalStateException(
+          "Required parameter WorkflowIdConflictPolicy in WorkflowOptions is missing in WorkflowStub");
```
stephanos, line 144: Important change: we're making the `WorkflowIdConflictPolicy` a required field for UwS across all SDKs.

temporal-sdk/src/main/java/io/temporal/client/WorkflowStubImpl.java
```diff
@@ -131,45 +131,71 @@ public WorkflowExecution start(Object... args) {
...
           new WorkflowClientCallsInterceptor.WorkflowUpdateWithStartInput<>(
-              startInput, updateOperation);
+              startInput, updateInput);
+
       WorkflowClientCallsInterceptor.WorkflowUpdateWithStartOutput<R> output =
           workflowClientInvoker.updateWithStart(input);
+
+      // gather outputs
       workflowExecution = output.getWorkflowStartOutput().getWorkflowExecution();
       populateExecutionAfterStart(workflowExecution);
```
stephanos, line 176: In other SDKs where we'll have a Future to obtain the workflow handle from the start operation, that future is immediately fulfilled once the run id is known. This can happen before the update wait stage has been reached.

I wonder if we want to do sth similar here. We could assign the workflow execution to the stub as soon as the run id is known. But it's not clear to me that it would do much since calling other functions on the stub are not waiting for the execution to be available.

I think we shouldn't do anything right now and re-consider based on feedback.

dandavison:
Summary of some parts of user-facing API in this PR:

- `WithStartWorkflowOperation`:
  - constructors
    - `WithStartWorkflowOperation(startMethod, args...)`
    - `WithStartWorkflowOperation(WorkflowStub, Class<? extends R> resultClass, Object...)`
  - `R getResult`

- static `WorkflowClient` methods
  - `WorkflowUpdateHandle<R> startUpdateWithStart(updateMethod, updateArgs..., UpdateOptions<R>, WithStartWorkflowOperation)`
  - `R executeUpdateWithStart(updateMethod, updateArgs..., UpdateOptions<R>, WithStartWorkflowOperation)`
  
- `WorkflowStub` methods
  - `WorkflowUpdateHandle<R> startUpdateWithStart(UpdateOptions<R>, Object[], WithStartWorkflowOperation)`
  - `R executeUpdateWithStart(UpdateOptions<R>, Object[], WithStartWorkflowOperation)`

- Interceptor
  - `WorkflowUpdateWithStartInput<R> => WorkflowUpdateWithStartOutput<R>`, both of which contain a start and update component


So the untyped stub methods obtain the update name from the `UpdateOptions`, and the `WorkflowClient` methods take an updateMethod, but also take `UpdateOptions<R>`. Can you confirm this is all as intended?

stephanos:
@dandavison yup, that's all correct!

> So the untyped stub methods obtain the update name from the UpdateOptions, and the WorkflowClient methods take an updateMethod, but also take UpdateOptions<R>. Can you confirm this is all as intended?

That's true. I believe it's the same for the existing `startUpdate` APIs in `WorkflowClient` (plus the `WithStartWorkflowOperation` at the end).

Now I wonder how it picks the final update name: the method's annotation or the update options' `updateName`? 👀

temporal-sdk/src/main/java/io/temporal/client/WorkflowStubImpl.java
Quinn-With-Two-Ns: Yeah +1 to this wording

temporal-sdk/src/main/java/io/temporal/client/WorkflowClient.java
```diff
@@ -858,299 +858,668 @@ static <R, A1, A2, A3, A4, A5, A6> WorkflowUpdateHandle<R> startUpdate(
...
+   *     {@link #newWorkflowStub(Class, WorkflowOptions)}.
+   * @param arg1 first workflow argument
+   * @param withStartOperation start workflow operation
+   * @return update result
+   */
+  @Experimental
+  static <A1, R> R executeUpdateWithStart(
+      Func1<A1, R> updateMethod,
+      A1 arg1,
+      @Nonnull UpdateOptions<R> options,
```
Quinn-With-Two-Ns, line 1394: It is a bit odd to use `UpdateOptions` here since almost none of the `options` here make any sense to set... The only option is makes sense to set is the update ID.

Quinn-With-Two-Ns: I guess when we have durable on admitted support we will need to have a `WaitStage` parameter for that...

stephanos: That's absolutely true. I'd support simplifying this `executeUpdateWithStart(updateMethod, args ..., startOp)`. We can always add a new overload when we need it.

Quinn-With-Two-Ns: I really want to avoid adding overloads to these methods if we can avoid it because the number of methods will explode with all the combinations for different numbers of parameters. I would support adding a `TypedUpdateOptions` that only takes `updateID` and `WaitStage` and is used for all the update APIs on `WorkflowClient`

Quinn-With-Two-Ns:
A general comment can we verify exceptions are consistent with `update`? While working on another PR https://github.com/temporalio/sdk-java/pull/2339/files#diff-deb992241240a3c5bff6838f447cddb38ed7978751405c2efe1daf563a00259bR481 I noticed that the`updateWithStart` API was returning a `WorkflowServiceException` instead of a `WorkflowUpdateException`. My PR fixed it for this test, but something we should double check I think.

temporal-sdk/src/main/java/io/temporal/client/UpdateOptions.java
```diff
@@ -154,6 +154,13 @@ public void validate() {
     }
   }
 
+  void validateWaitForCompleted() {
+    if (waitForStage != null && waitForStage != WorkflowUpdateStage.COMPLETED) {
+      throw new IllegalArgumentException(
+              "waitForStage must be unspecified or " + WorkflowUpdateStage.COMPLETED);
+    }
+  }
```
stephanos, line 162: So I added this because I think users would rather ensure that the UwS call waits for "completed" when they call `executeUpdateWithStart`, than have it be "accepted" (accidentally) and potentially impacting their experience (since an additional polling step might be needed).

This also seems better than silently setting it, but I don't feel strongly about this.

stephanos: Whatever we decide, the behavior for the "execute update" API(s) should do the same.

stephanos:
> I noticed that theupdateWithStart API was returning a WorkflowServiceException instead of a WorkflowUpdateException.

@Quinn-With-Two-Ns My understanding is that the method invocations of `startUpdate`, `startUpdateWithStart` and `executeUpdateWithStart` on `WorkflowStubImpl` throw a `WorkflowServiceException` if anything goes wrong with the RPC call. But then the `getResult` or `getResultAsync().get()` calls throw `WorkflowUpdateException` if there was an issue obtaining the result?

Quinn-With-Two-Ns:
@stephanos 

>WorkflowStubImpl throw a WorkflowServiceException if anything goes wrong with the RPC call (e.g. it was invalid).

Yes, but if the update is rejected or failed that is not an RPC issue and should not be surfaced as such. It should return an `WorkflowUpdateException`

stephanos:
@Quinn-With-Two-Ns From what I can tell, `WorkflowUpdateException` is _only_ thrown when the update request or the poll update request receive a failure outcome.

UwS throws does the same, and for [returns a `WorkflowServiceException` for anything else](https://github.com/temporalio/sdk-java/pull/2337/files#diff-64e98c2533a94b01fbda951826a90567fb17dd80b70af7f6f44c5c6dd1759c69R185).

This includes when there's (a) server validation/rate limiting error (for start or update) or (b) the update never becomes durable and times out (client-side). Which seems to match the vanilla update behavior.

Quinn-With-Two-Ns (approved):

temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkflowClientCallsInterceptor.java
```diff
@@ -233,21 +234,20 @@ public WorkflowStartOutput getWorkflowStartOutput() {
 
   final class WorkflowUpdateWithStartInput<R> {
     private final WorkflowStartInput workflowStartInput;
-    private final UpdateWithStartWorkflowOperation<R> updateOperation;
+    private final StartUpdateInput<R> workflowUpdateInput;
```
Quinn-With-Two-Ns, line 237: @cretz pointed out in the Go SDK PR interceptors shouldn't reuse input/output from other interceptors, which I agree with , but the Java SDK has done that historically. I don't think we need to block the PR on this , but we should resist this before we make UwS stable in the Java SDK.

temporal-sdk/src/main/java/io/temporal/client/WorkflowStub.java
```diff
@@ -157,16 +158,35 @@ <R> WorkflowUpdateHandle<R> getUpdateHandle(
...
+      UpdateOptions<R> updateOptions, Object[] updateArgs, Object[] starArgs);
+
+  /**
+   * Synchronously update a workflow execution by invoking its update handler, and start the
+   * workflow according to the option's {@link WorkflowIdConflictPolicy}. It returns the update
+   * result.
+   *
+   * @param updateOptions options that will be used to configure and start a new update request
+   * @param updateArgs update method arguments
+   * @param starArgs workflow start arguments
```
dandavison, line 183: typo, also in parameter names in code.

dandavison (approved):
LGTM

PR #2338, opened 2024-11-29
Fix CI after GHA's drop of node16 actions

Quinn-With-Two-Ns:
Build the test server in an ubuntu 18 container

### DIFF
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -24,6 +24,11 @@ on:
         description: "Publish Java Artifacts"
         required: true
         default: "true"
+      do_create_release:
+        type: boolean
+        description: "Create Github draft release"
+        required: true
+        default: "true"
 env:
   INPUT_REF: ${{ github.event.inputs.ref }}
   INPUT_TAG: ${{ github.event.inputs.tag }}
@@ -31,6 +36,7 @@ env:
 jobs:
   create_draft_release:
     name: Create Github draft release
+    if: github.event.inputs.do_create_release == 'true'
     runs-on: ubuntu-latest
     steps:
       - name: Audit gh version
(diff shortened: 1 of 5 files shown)


.github/workflows/prepare-release.yml
```diff
@@ -24,13 +24,19 @@ on:
...
+        required: true
+        default: "true"
 env:
   INPUT_REF: ${{ github.event.inputs.ref }}
   INPUT_TAG: ${{ github.event.inputs.tag }}
 
 jobs:
   create_draft_release:
     name: Create Github draft release
+    if: github.event.inputs.do_create_release == 'true'
```
Quinn-With-Two-Ns, line 39: This is just for testing and will be removed

.github/workflows/prepare-release.yml
```diff
@@ -92,7 +98,7 @@ jobs:
           distribution: "temurin"
 
       - name: Set up Gradle
-        uses: gradle/actions/setup-gradle@v3
+        uses: gradle/actions/setup-gradle@v4
```
mjameswh, line 101: There are other places we use `gradle/actions/setup-gradle@v3`. Should update those too.

.github/workflows/prepare-release.yml
```diff
@@ -138,43 +144,45 @@ jobs:
...
+        run: |
+          ./gradlew :temporal-test-server:build
 
+      - name: Build native test server (Docker)
+        if: matrix.os_family == 'Linux'
+        run: |
+          docker run \
+            --rm -w /github/workspace -v "$(pwd):/github/workspace" \
+            $(docker build -q ./docker/native-image) \
+            sh -c "./gradlew :temporal-test-server:build"
```
mjameswh, line 181: Now that you moved installation stuff to a docker image, you probably no longer need the `sh -c` and quotes.

.github/workflows/prepare-release.yml
Quinn-With-Two-Ns: Done

.github/workflows/prepare-release.yml
Quinn-With-Two-Ns: Yep, fixed

docker/native-image/dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM ubuntu:18.04
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: Added a comment

mjameswh (approved):
Thanks for completing that.

PR #2342, opened 2024-12-04
Make headers in ExecuteNexusOperationInput case insensitive

Quinn-With-Two-Ns:
Make headers in ExecuteNexusOperationInput case insensitive, this will align with other Nexus headers that are all case insensitive.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkflowOutboundCallsInterceptor.java
+++ b/temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkflowOutboundCallsInterceptor.java
@@ -33,6 +33,7 @@ import java.time.Duration;
 import java.util.*;
 import java.util.function.BiPredicate;
 import java.util.function.Supplier;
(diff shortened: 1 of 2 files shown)


temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkflowOutboundCallsInterceptor.java
```diff
@@ -285,7 +286,14 @@ public ExecuteNexusOperationInput(
...
       this.options = options;
-      this.headers = headers;
+      this.headers =
+          headers.entrySet().stream()
+              .collect(
+                  Collectors.toMap(
+                      (k) -> k.getKey().toLowerCase(),
+                      Map.Entry::getValue,
+                      (a, b) -> a,
+                      () -> new TreeMap<>(String.CASE_INSENSITIVE_ORDER)));
```
cretz, line 296: A nice bonus of having a deterministically sorted map too

cretz (approved):

PR #2343, opened 2024-12-05
fix: Nexus operation headers should be modifiable

robzienert:
## What was changed
Changes Nexus start/execute operation headers from an immutable empty map to an empty HashMap.

## Why?
Changing headers as part of a `WorkerInterceptor` is not currently possible.

## Checklist

1. How was this tested:
Not manually tested.


robzienert:
Closing, since it's likely desired behavior to just create a new instance of `input` for next interceptors.

PR #2348, opened 2024-12-12
Test server support for Nexus operation complete before start

pdoerner:
## What was changed

Added support to the test server for completing Nexus operations before the start response is received.

### DIFF
--- a/temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
+++ b/temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
@@ -21,6 +21,7 @@
 package io.temporal.internal.testservice;
 
 import static io.temporal.api.enums.v1.UpdateWorkflowExecutionLifecycleStage.*;
+import static io.temporal.internal.common.LinkConverter.workflowEventToNexusLink;
 import static io.temporal.internal.testservice.CronUtils.getBackoffInterval;
 import static io.temporal.internal.testservice.StateMachines.*;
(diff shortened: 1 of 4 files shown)


temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -2218,6 +2234,32 @@ public void completeNexusOperation(NexusOperationRef ref, Payload result) {
...
+  public void completeAsyncNexusOperation(
+      NexusOperationRef ref,
+      Payload result,
+      String operationID,
+      io.temporal.api.nexus.v1.Link startLink) {
+    update(
+        ctx -> {
+          StateMachine<NexusOperationData> operation =
+              getPendingNexusOperation(ref.getScheduledEventId());
+          if (operation.getState() == State.INITIATED) {
```
bergundy, line 2247: To save me from reading, does this cover when the activity is backing off and scheduled?

bergundy (commented):
LGTM but I'm not super familiar with this codebase. Will defer to @Quinn-With-Two-Ns for approval.

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
pdoerner: Yes, it goes back to INITIATED when backing off.

Quinn-With-Two-Ns (approved):

Quinn-With-Two-Ns:
@pdoerner is this good to merge?

PR #2357, opened 2024-12-27
Lower refreshNexusEndpointsMinWait

Quinn-With-Two-Ns:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/docker/github/dynamicconfig/development.yaml
+++ b/docker/github/dynamicconfig/development.yaml
@@ -28,3 +28,5 @@ component.callbacks.allowedAddresses:
   - value:
       - Pattern: "localhost:7243"
         AllowInsecure: true
+system.refreshNexusEndpointsMinWait:
+  - value: 1ms


temporal-sdk/src/main/java/io/temporal/failure/DefaultFailureConverter.java
```diff
@@ -189,7 +189,14 @@ private TemporalFailure failureToExceptionImpl(Failure failure, DataConverter da
         }
       case FAILUREINFO_NOT_SET:
       default:
-        throw new IllegalArgumentException("Failure info not set");
+        // All unknown types are considered to be retryable ApplicationError.
```
Quinn-With-Two-Ns, line 192: I checked all other SDKs and the Java SDK was the only SDK that was throwing an exception here instead of converting to an application failure

temporal-sdk/src/test/java/io/temporal/workflow/nexus/SyncOperationCancelledTest.java
```diff
@@ -58,17 +58,6 @@ public void syncOperationImmediatelyCancelled() {
         "operation canceled before it was started", canceledFailure.getOriginalMessage());
   }
 
-  @Test
-  public void syncOperationCancelled() {
```
Quinn-With-Two-Ns, line 62: Removed this test since the Server dropped support for cancelling failing sync operations

temporal-sdk/src/test/java/io/temporal/workflow/nexus/TerminateWorkflowAsyncOperationTest.java
```diff
@@ -54,10 +55,19 @@ public void terminateAsyncOperation() {
         Assert.assertThrows(WorkflowFailedException.class, () -> workflowStub.execute(""));
     Assert.assertTrue(exception.getCause() instanceof NexusOperationFailure);
     NexusOperationFailure nexusFailure = (NexusOperationFailure) exception.getCause();
-    Assert.assertTrue(nexusFailure.getCause() instanceof ApplicationFailure);
-    Assert.assertEquals(
-        "operation terminated",
-        ((ApplicationFailure) nexusFailure.getCause()).getOriginalMessage());
+    // TODO: Test server needs to be fixed to return the correct type
```
cretz, line 58: If this TODO is to be done separately, can we open an issue and reference it here?

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -86,6 +87,17 @@
 import org.slf4j.LoggerFactory;
 
 class TestWorkflowMutableStateImpl implements TestWorkflowMutableState {
+  static final Failure FAILED_UPDATE_ON_WF_COMPLETION =
```
cretz, line 90: Do we have a test confirming that this at least looks mostly like the real server's version? Unsure if Java has tests against actual server and test server with the same code, but I just want to make sure if/when this changes server side we can catch it here.

cretz (approved):
Two minor non-blocking comments

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
Quinn-With-Two-Ns: Yeah I can do that

PR #2360, opened 2025-01-07
doc: Clarify `UnableToAcquireLockException` (#2359)

### LINKED ISSUE #2359: Clarify `UnableToAcquireLockException`
[This Slack thread](https://temporalio.slack.com/archives/C05JRT1GKEE/p1735812460229109?thread_ts=1735802152.698209&cid=C05JRT1GKEE): a user had a question about this Exception, and they asked Kapa, and Kapa gave a wrong answer (which Tiho corrected 🙏 ). I suspect Kapa could have given better direction if this exception message contained the info that Tiho corrected with.

I ported Tiho's response over to the code and submitted a PR

I recognize that the message I'm putting here is less general than the one that was there previously. If any of it is incorrect in more general scenarios, please correct me 👍 👍 

**Additional context**
None

GSmithApps:
Closes #2359 (which is to help clarify [this Slack thread](https://temporalio.slack.com/archives/C05JRT1GKEE/p1735812460229109?thread_ts=1735802152.698209&cid=C05JRT1GKEE))

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
@@ -355,8 +355,12 @@ final class WorkflowWorker implements SuspendableWorker {
 
           if (!locked) {
             throw new UnableToAcquireLockException(
-                "Workflow lock for the run id hasn't been released by one of previous execution attempts, "
-                    + "consider increasing workflow task timeout.");
+                "Workflow lock for the run id hasn't been released by one of previous execution attempts.  "
+                    + "This likely means the workflow task timed out, and the Service attempted to "
+                    + "retry the task on the same Worker while the timed out Task had not finished yet.  "
+                    + "We suggest investigating why the Workflow Task timed out. This could mean "
+                    + "looking at Worker resource utilization at the time in which this exception was thrown or "
+                    + "increasing workflow task timeout.");
           }
         }
 


temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
```diff
@@ -355,8 +355,12 @@ public void handle(WorkflowTask task) throws Exception {
 
           if (!locked) {
             throw new UnableToAcquireLockException(
-                "Workflow lock for the run id hasn't been released by one of previous execution attempts, "
-                    + "consider increasing workflow task timeout.");
+                "Workflow lock for the run id hasn't been released by one of previous execution attempts.  "
+                    + "This likely means the workflow task timed out, and the Service attempted to "
```
Quinn-With-Two-Ns, line 359: This exception can be thrown any time the workflow lock cannot be acquire because another task is holding the lock. It can happen because of workflow task timeout, but that is misleading to suggest it is likely or to recommend investigating Workflow Task timed out without doing some triage. This error can have multiple causes including something happening on the Temporal Server side

```suggestion
"Workflow lock for the run id hasn't been released by one of previous execution attempts. This means another workflow task is holding the lock. If seen repeatedly users should investigate the failing workflow executions for more details."
```

GSmithApps:
Hi Quinn, thanks so much for looking at this PR!

I'm not sure if you saw the [Slack thread](https://app.slack.com/client/TNWA8QCGZ/C05JRT1GKEE) in the comment from me above, but I was trying to get ahead of a question similar to that in the future, because judging by Tiho's comment, Kapa went down the wrong path. I figured if we gave it a little more direction in the exception description, we might be able to avoid that. 

Do you think there's any clarifying or expanding to do, or should we leave it as-is?

Also, if this warrants more discussion or anything, I'm happy to chat over Slack 🙂

PR #2363, opened 2025-01-07
Add Graal reflection config for UwS

Quinn-With-Two-Ns:
Add Graal reflection config for UwS. Generated by running UwS tests with tracing agent


cretz (approved):
Can you share how you generated this? Were the contributing notes enough or did you have to do some manual things? No need to explain in depth or anything (especially if we hope to improve in the future), just curious.

Quinn-With-Two-Ns:
>Can you share how you generated this?

I just build the test server and ran with the [tracing agent](https://www.graalvm.org/latest/reference-manual/native-image/metadata/AutomaticMetadataCollection/). When I get us onto the official plugin it should be a lot more straightforward

PR #2366, opened 2025-01-13
[#1962] Wrap returned promises in OpenTracingOutboundCallsInterceptor

### LINKED ISSUE #1962: Tracing context does not propagate into .thenCompose
## Expected Behavior

Temporal workflows and activities should generate related spans when the Temporal client is properly configured with OpenTracing interceptors.

## Actual Behavior

Some spans end up severed from the parent.  This seems to be isolated to spans that are generated with a promise chain using .thenCompose.  Other spans seem to work as expected.

  1. Enable trace interceptors
  1. Write a workflow that has at least two activities that are composed serially via .executeAsync -> .thenCompose
  1. Observe that the workflow and first activity end up in one span, and the second activity ends up orphaned within its own span.

## Specifications

  - Version: Temporal Java SDK 1.22.3

ghaskins:
This change allows the OpenTracing interceptor to transfer any active trace/span from
the point of invocation to the point of promise callback execution, thus preserving
the span within the callback.

Without this, spans become disjoint across Promise.thenCompose, due
to the nature of the deferring the execution of the .thenCompose to a callback thread.

Fixes #1962

## Checklist

1. Closes #1962

2. How was this tested:
A unit test was developed first to confirm the issue.  The solution was iterated over until the unit test passed satisfactorily.  The unit-test is included in the MR for inclusion with the code-base

3. Any docs updates needed?
No.

### DIFF
--- a/temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowOutboundCallsInterceptor.java
+++ b/temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowOutboundCallsInterceptor.java
@@ -27,15 +27,90 @@ import io.opentracing.Tracer;
 import io.temporal.common.interceptors.WorkflowOutboundCallsInterceptor;
 import io.temporal.common.interceptors.WorkflowOutboundCallsInterceptorBase;
 import io.temporal.opentracing.OpenTracingOptions;
+import io.temporal.workflow.Functions;
+import io.temporal.workflow.Promise;
 import io.temporal.workflow.Workflow;
 import io.temporal.workflow.WorkflowInfo;
 import io.temporal.workflow.unsafe.WorkflowUnsafe;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 
 public class OpenTracingWorkflowOutboundCallsInterceptor
     extends WorkflowOutboundCallsInterceptorBase {
   private final SpanFactory spanFactory;
   private final Tracer tracer;
   private final ContextAccessor contextAccessor;
+
+  private class PromiseWrapper<R> implements Promise<R> {
+    private final Span capturedSpan;
+    private final Promise<R> delegate;
+
+    PromiseWrapper(Span capturedSpan, Promise<R> delegate) {
+      this.capturedSpan = capturedSpan;
+      this.delegate = delegate;
+    }
+
+    private <O> O wrap(Functions.Func<O> fn) {
+      Span activeSpan = tracer.scopeManager().activeSpan();
+      if (activeSpan == null && capturedSpan != null) {
+        try (Scope ignored = tracer.scopeManager().activate(capturedSpan)) {
+          return fn.apply();
+        }
+      } else {
+        return fn.apply();
+      }
+    }
+
+    @Override
+    public boolean isCompleted() {
+      return delegate.isCompleted();
+    }
+
+    @Override
+    public R get() {
+      return delegate.get();
+    }
+
+    @Override
+    public R cancellableGet() {
+      return delegate.cancellableGet();
+    }
+
+    @Override
+    public R get(long timeout, TimeUnit unit) throws TimeoutException {
+      return delegate.get(timeout, unit);
+    }
+
+    @Override
+    public R cancellableGet(long timeout, TimeUnit unit) throws TimeoutException {
+      return delegate.cancellableGet(timeout, unit);
+    }
+
+    @Override
+    public RuntimeException getFailure() {
+      return delegate.getFailure();
+    }
+
+    @Override
+    public <U> Promise<U> thenApply(Functions.Func1<? super R, ? extends U> fn) {
+      return delegate.thenApply((r) -> wrap(() -> fn.apply(r)));
+    }
+
+    @Override
+    public <U> Promise<U> handle(Functions.Func2<? super R, RuntimeException, ? extends U> fn) {
+      return delegate.handle((r, e) -> wrap(() -> fn.apply(r, e)));
+    }
+
+    @Override
+    public <U> Promise<U> thenCompose(Functions.Func1<? super R, ? extends Promise<U>> fn) {
+      return delegate.thenCompose((r) -> wrap(() -> fn.apply(r)));
+    }
+
+    @Override
+    public Promise<R> exceptionally(Functions.Func1<Throwable, ? extends R> fn) {
+      return delegate.exceptionally((t) -> wrap(() -> fn.apply(t)));
+    }
+  }
 
   public OpenTracingWorkflowOutboundCallsInterceptor(
       WorkflowOutboundCallsInterceptor next,
@@ -51,13 +126,16 @@ public class OpenTracingWorkflowOutboundCallsInterceptor
   @Override
   public <R> ActivityOutput<R> executeActivity(ActivityInput<R> input) {
     if (!WorkflowUnsafe.isReplaying()) {
+      Span capturedSpan = tracer.scopeManager().activeSpan();
       Span activityStartSpan =
           contextAccessor.writeSpanContextToHeader(
               () -> createActivityStartSpanBuilder(input.getActivityName()).start(),
               input.getHeader(),
               tracer);
       try (Scope ignored = tracer.scopeManager().activate(activityStartSpan)) {
-        return super.executeActivity(input);
+        ActivityOutput<R> output = super.executeActivity(input);
+        return new ActivityOutput<>(
+            output.getActivityId(), new PromiseWrapper<>(capturedSpan, output.getResult()));
       } finally {
         activityStartSpan.finish();
       }
@@ -69,13 +147,15 @@ public class OpenTracingWorkflowOutboundCallsInterceptor
   @Override
   public <R> LocalActivityOutput<R> executeLocalActivity(LocalActivityInput<R> input) {
     if (!WorkflowUnsafe.isReplaying()) {
+      Span capturedSpan = tracer.scopeManager().activeSpan();
       Span activityStartSpan =
           contextAccessor.writeSpanContextToHeader(
               () -> createActivityStartSpanBuilder(input.getActivityName()).start(),
               input.getHeader(),
               tracer);
       try (Scope ignored = tracer.scopeManager().activate(activityStartSpan)) {
-        return super.executeLocalActivity(input);
+        LocalActivityOutput<R> output = super.executeLocalActivity(input);
+        return new LocalActivityOutput<>(new PromiseWrapper<>(capturedSpan, output.getResult()));
       } finally {
         activityStartSpan.finish();
       }
@@ -87,11 +167,15 @@ public class OpenTracingWorkflowOutboundCallsInterceptor
   @Override
   public <R> ChildWorkflowOutput<R> executeChildWorkflow(ChildWorkflowInput<R> input) {
     if (!WorkflowUnsafe.isReplaying()) {
+      Span capturedSpan = tracer.scopeManager().activeSpan();
       Span childWorkflowStartSpan =
           contextAccessor.writeSpanContextToHeader(
               () -> createChildWorkflowStartSpanBuilder(input).start(), input.getHeader(), tracer);
       try (Scope ignored = tracer.scopeManager().activate(childWorkflowStartSpan)) {
-        return super.executeChildWorkflow(input);
+        ChildWorkflowOutput<R> output = super.executeChildWorkflow(input);
+        return new ChildWorkflowOutput<>(
+            new PromiseWrapper<>(capturedSpan, output.getResult()),
+            new PromiseWrapper<>(capturedSpan, output.getWorkflowExecution()));
       } finally {
         childWorkflowStartSpan.finish();
       }
@@ -104,13 +188,17 @@ public class OpenTracingWorkflowOutboundCallsInterceptor
   public <R> ExecuteNexusOperationOutput<R> executeNexusOperation(
       ExecuteNexusOperationInput<R> input) {
     if (!WorkflowUnsafe.isReplaying()) {
+      Span capturedSpan = tracer.scopeManager().activeSpan();
       Span nexusOperationExecuteSpan =
           contextAccessor.writeSpanContextToHeader(
               () -> createStartNexusOperationSpanBuilder(input).start(),
               input.getHeaders(),
               tracer);
       try (Scope ignored = tracer.scopeManager().activate(nexusOperationExecuteSpan)) {
-        return super.executeNexusOperation(input);
+        ExecuteNexusOperationOutput<R> output = super.executeNexusOperation(input);
+        return new ExecuteNexusOperationOutput<>(
+            new PromiseWrapper<>(capturedSpan, output.getResult()),
+            new PromiseWrapper<>(capturedSpan, output.getOperationExecution()));
       } finally {
         nexusOperationExecuteSpan.finish();
       }
@@ -122,6 +210,7 @@ public class OpenTracingWorkflowOutboundCallsInterceptor
   @Override
   public SignalExternalOutput signalExternalWorkflow(SignalExternalInput input) {
     if (!WorkflowUnsafe.isReplaying()) {
+      Span capturedSpan = tracer.scopeManager().activeSpan();
       WorkflowInfo workflowInfo = Workflow.getInfo();
       Span childWorkflowStartSpan =
           contextAccessor.writeSpanContextToHeader(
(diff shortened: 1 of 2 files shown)


cretz:
Hrmm, this is a very specific interceptor-oriented solution to a general problem. I wonder if a `WorkflowAsyncLocal` is best which behaves like thread local, but also propagates across promises.

ghaskins:
@cretz, if I understand what you are saying, I think we'd still need to use interception at some level unless OTEL can be made to utilize the WorkflowAsyncLocal instead of ThreadLocal because something would need to initialize the state in the callback context, which currently runs outside of the RootThread complex.  I'm not saying OTEL can't use WorkflowAsyncLocal; I just don't know

temporal-sdk/src/main/java/io/temporal/internal/sync/CompletablePromiseImpl.java
```diff
@@ -275,9 +280,8 @@ private <U> Promise<U> then(Functions.Proc1<CompletablePromise<U>> proc) {
    * @return true if there were any handlers invoked
    */
   private boolean invokeHandlers() {
-    for (Functions.Proc handler : handlers) {
-      handler.apply();
-    }
+    executor.execute(() -> handlers.forEach(Functions.Proc::apply));
```
mfateev, line 283: I believe this change is not backward compatible and can break determinism.

temporal-opentracing/src/test/java/io/temporal/opentracing/CallbackContextTest.java
```diff
@@ -0,0 +1,134 @@
...
+            ActivityOptions.newBuilder()
+                .setStartToCloseTimeout(Duration.ofMinutes(1))
+                .validateAndBuildWithDefaults());
+
+    @Override
+    public String workflow(String input) {
+      return Async.function(activity::activity)
+          .thenCompose(
+              (r) -> {
+                Span activeSpan = mockTracer.activeSpan();
```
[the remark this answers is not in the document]
ghaskins: That is a good point.  I will check this.

temporal-sdk/src/main/java/io/temporal/internal/sync/CompletablePromiseImpl.java
ghaskins: @mfateev are you referring to the forEach or the executor deferment?

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowOutboundCallsInterceptor.java
```diff
@@ -30,13 +30,34 @@
...
+    private final Span activeSpan;
+
+    public SpanTransferringExecutor(Executor passthrough) {
+      this.passthrough = passthrough;
+      activeSpan = tracer.scopeManager().activeSpan();
+    }
+
+    public void execute(Runnable r) {
+      if (activeSpan != null) {
+        try (Scope ignored = tracer.scopeManager().activate(activeSpan)) {
```
Quinn-With-Two-Ns, line 52: For consistency with OTEL and backwards compatibility, If we have an active span here we should not set the `activeSpan`

ghaskins: This might be a poor-nomenclature thing than a problem, unless I am misunderstanding you.

In the above, "activeSpan" isnt the current threads activeSpan, but that which was captured during callback/promise creation.  Perhaps I should rename it to something else, like "parentSpan"?  Or are you referring to a different problem?

ghaskins: I changed this to "capturedSpan" to be more explicit.  Let me know if you still have concerns here.

Quinn-With-Two-Ns: Yes so i am saying if the thread already has an active span, don't set a new one

temporal-sdk/src/main/java/io/temporal/internal/sync/CompletablePromiseImpl.java
ghaskins: I added documentation to clarify the synchronous requirement and changed to a plain for() loop.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowOutboundCallsInterceptor.java
ghaskins: Ah, I get you now.  I will add that check

temporal-sdk/src/main/java/io/temporal/internal/sync/CompletablePromiseImpl.java
mfateev: My point is that switching from direct call to an executor that might change the order of execution of these calls can lead to non determinism. Such changes should be protected by versioning/patching.

temporal-opentracing/src/test/java/io/temporal/opentracing/CallbackContextTest.java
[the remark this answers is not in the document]
ghaskins: You were correct: I changed the test to this

```
@Override
    public String workflow() {
      String workflowSpanId = mockTracer.activeSpan().context().toSpanId();
      return Async.function(activity::activity)
          .thenCompose(
              (r) -> {
                Span activeSpan = mockTracer.activeSpan();

                if (activeSpan != null) {
                  String promiseSpanId = activeSpan.context().toSpanId();
                  if (promiseSpanId.equals(workflowSpanId)) {
                    return Workflow.newPromise(activeSpan.context().toTraceId());
                  } else {
                    return Workflow.newPromise("bad-span");
                  }
                } else {
                  return Workflow.newPromise("not-found");
                }
              })
          .get();
    }
    ```
    
    and it fails with "bad-span"

 Ill take a look at how to account for this.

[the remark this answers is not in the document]
Comment (Quinn-With-Two-Ns) [reply temporal-opentracing/src/test/java/io/temporal/opentracing/CallbackContextTest.java:109]: This was part of why In the slack thread I suggested creating a promise wrapper since to deal with this you need to capture the span before the outbound interceptor

File: temporal-sdk/src/main/java/io/temporal/internal/sync/CompletablePromiseImpl.java (hunk 1 above)
Comment (ghaskins) [reply temporal-sdk/src/main/java/io/temporal/internal/sync/CompletablePromiseImpl.java]: Ah, I see.  FWIW: I was careful to design the executor chain in a way that should respect the original order of interceptors, and I retained the original order of the for loop as well.  I suspect your concern would be if someone added an interceptor that creates an NDE situation, but I think that if that were the case, they should follow patching procedures in the intercept code.  I don't think we could solve that at this level.  Thoughts?

File: temporal-opentracing/src/test/java/io/temporal/opentracing/CallbackContextTest.java (hunk 2 above)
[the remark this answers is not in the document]
Comment (ghaskins) [reply temporal-opentracing/src/test/java/io/temporal/opentracing/CallbackContextTest.java:110]: Wrapper implemented...nice suggestion!

File: temporal-sdk/src/main/java/io/temporal/internal/sync/CompletablePromiseImpl.java (hunk 1 above)
Comment (ghaskins) [reply temporal-sdk/src/main/java/io/temporal/internal/sync/CompletablePromiseImpl.java]: @mfateev I looked, and I don't see any apparent examples of SDK code using patching internally.  Since I don't believe the PR code changes the ordering, the only NDE potential that I see would be if an interceptor introduced a determinism issue in its handler.  However, that would also concern any existing interception points, so it's unclear if this proposed one should be treated differently.  

This is a blocking issue for me, so I am highly motivated to see it get merged.  So please let me know what you'd like to see done here, and I will do my best to make it happen.

File: temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowOutboundCallsInterceptor.java [hunk 4]
```diff
@@ -27,16 +27,141 @@
...
+    @Override
+    public RuntimeException getFailure() {
+      return delegate.getFailure();
+    }
+
+    @Override
+    public <U> Promise<U> thenApply(Functions.Func1<? super R, ? extends U> fn) {
+      return delegate.thenApply(
+          (r) -> {
+            if (capturedSpan != null) {
```
Quinn-With-Two-Ns, line 112: We should copy the logic we had for `SpanTransferringExecutor` where we ignored the `capturedSpan` if one is already active

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowOutboundCallsInterceptor.java
```diff
@@ -27,16 +27,141 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 
 public class OpenTracingWorkflowOutboundCallsInterceptor
     extends WorkflowOutboundCallsInterceptorBase {
   private final SpanFactory spanFactory;
   private final Tracer tracer;
   private final ContextAccessor contextAccessor;
 
+  private class SpanTransferringExecutor implements Executor {
```
Quinn-With-Two-Ns, line 45: We can remove this executor and the interceptor for it now since it is not needed

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowOutboundCallsInterceptor.java
```diff
@@ -87,11 +217,14 @@ public <R> LocalActivityOutput<R> executeLocalActivity(LocalActivityInput<R> inp
...
     if (!WorkflowUnsafe.isReplaying()) {
+      Span capturedSpan = tracer.scopeManager().activeSpan();
       Span childWorkflowStartSpan =
           contextAccessor.writeSpanContextToHeader(
               () -> createChildWorkflowStartSpanBuilder(input).start(), input.getHeader(), tracer);
       try (Scope ignored = tracer.scopeManager().activate(childWorkflowStartSpan)) {
-        return super.executeChildWorkflow(input);
+        ChildWorkflowOutput<R> output = super.executeChildWorkflow(input);
+        return new ChildWorkflowOutput<>(
+            new PromiseWrapper<>(capturedSpan, output.getResult()), output.getWorkflowExecution());
```
Quinn-With-Two-Ns, line 227: `output.getWorkflowExecution()` is also a Promise that needs to be wrapped

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowOutboundCallsInterceptor.java
```diff
@@ -104,13 +237,16 @@ public <R> ChildWorkflowOutput<R> executeChildWorkflow(ChildWorkflowInput<R> inp
...
       Span nexusOperationExecuteSpan =
           contextAccessor.writeSpanContextToHeader(
               () -> createStartNexusOperationSpanBuilder(input).start(),
               input.getHeaders(),
               tracer);
       try (Scope ignored = tracer.scopeManager().activate(nexusOperationExecuteSpan)) {
-        return super.executeNexusOperation(input);
+        ExecuteNexusOperationOutput<R> output = super.executeNexusOperation(input);
+        return new ExecuteNexusOperationOutput<>(
+            new PromiseWrapper<>(capturedSpan, output.getResult()), output.getOperationExecution());
```
Quinn-With-Two-Ns, line 249: `output.getWorkflowExecution()` is a promise that needs to be wrapped

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowOutboundCallsInterceptor.java
ghaskins: I intentionally left that out because it was doing wrong by inheriting the context created by the SpanTransferringExecutor inside the activity span.  If we remove the executor, as you said in the comment below, this will be less of an issue. Still, it got me thinking about the correct behavior here, and it felt like the capturedSpan is always the proper choice.  I am not married to this, though.  Thoughts?

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowOutboundCallsInterceptor.java
ghaskins: Yeah, I was debating that too.  My thought was leaving it would mean any Promise could benefit from the wrapping. Still, my primary concern is supporting propagation to Activity/LocalActivity, so I am happy either way and can remove this.  This decision can always be revisited in a later PR if generalizing propagation to any Promise becomes a desired feature.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowOutboundCallsInterceptor.java
ghaskins: Good catch.  Will fix.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowOutboundCallsInterceptor.java
ghaskins: Will fix

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowOutboundCallsInterceptor.java
Quinn-With-Two-Ns: So, in insolation, I think it is debatable if the `capturedSpan` or the span of the caller of `thenApply` should be used. That being said, the current behaviour and the behaviour of OTEL + Completable futures is more closely mimicked by ignoring the `capturedSpan` if one is already active.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowOutboundCallsInterceptor.java
Quinn-With-Two-Ns: We can never remove an extension point after we add it. So if it is not needed to solve this problem we shouldn't add it. We can always add it later if the need arises.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowOutboundCallsInterceptor.java
ghaskins: ok, I will re-add the activeSpan checks.

ghaskins: done

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowOutboundCallsInterceptor.java
ghaskins: removed

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowOutboundCallsInterceptor.java
ghaskins: fixed

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowOutboundCallsInterceptor.java
ghaskins: fixed

Quinn-With-Two-Ns:
@ghaskins Looks like part of CI is failing because your branch is missing the version tags, can you make sure you fetched tags when fetched? I will try to prioritize adding a fallback in our CI but for now the quickest would be just to push make sure tags are present.

temporal-opentracing/src/main/java/io/temporal/opentracing/internal/OpenTracingWorkflowOutboundCallsInterceptor.java
```diff
@@ -51,13 +126,16 @@ public OpenTracingWorkflowOutboundCallsInterceptor(
   @Override
   public <R> ActivityOutput<R> executeActivity(ActivityInput<R> input) {
     if (!WorkflowUnsafe.isReplaying()) {
+      Span capturedSpan = tracer.scopeManager().activeSpan();
```
Quinn-With-Two-Ns, line 129: There are a lot of futures in the interceptor that are not wrapped, I will need to go and wrap them, that is one big drawback with this approach, but given the other constraints I can't think of a better approach.

ghaskins:
@Quinn-With-Two-Ns pushed tags

Quinn-With-Two-Ns (approved):

Issue #2362, opened 2025-01-07
Provide client from activity context


cretz:
**Describe the solution you'd like**

Allow users to access Temporal client from within an activity. This is likely a `getWorkflowClient() WorkflowClient` instance method on `ActivityExecutionContext`. See https://github.com/temporalio/features/issues/203.

PR #2369, opened 2025-01-15
Add method to get workflow client from an activity

Quinn-With-Two-Ns:
Add method to get workflow client from an activity.

closes https://github.com/temporalio/sdk-java/issues/2362

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/activity/ActivityExecutionContext.java
+++ b/temporal-sdk/src/main/java/io/temporal/activity/ActivityExecutionContext.java
@@ -22,6 +22,7 @@ package io.temporal.activity;
 
 import com.uber.m3.tally.Scope;
 import io.temporal.client.ActivityCompletionException;
+import io.temporal.client.WorkflowClient;
 import io.temporal.serviceclient.WorkflowServiceStubsOptions;
 import io.temporal.worker.WorkerOptions;
 import java.lang.reflect.Type;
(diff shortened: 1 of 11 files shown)


temporal-sdk/src/main/java/io/temporal/activity/ActivityExecutionContext.java
```diff
@@ -140,4 +141,10 @@ public interface ActivityExecutionContext {
    * WorkflowServiceStubsOptions.Builder#setMetricsScope(Scope)} when a worker starts up.
    */
   Scope getMetricsScope();
+
+  /**
+   * Get a {@link WorkflowClient} that can be used to start interact with the Temporal service from
+   * an activity.
+   */
+  WorkflowClient getWorkflowClient();
```
cretz, line 149: I assume we don't expect users to have their own implementations of `ActivityExecutionContext`? I know we have a `TestActivityEnvironment`, just wanted to check. If you do make another commit in this PR, may be worth a warning at the top of this interface saying that methods may be added without default implementations and that users should not implement this interface.

cretz (approved):

temporal-sdk/src/main/java/io/temporal/activity/ActivityExecutionContext.java
Quinn-With-Two-Ns: We have `ActivityExecutionContextBase` if users want to provide their own wrapper implementation but not have to implement every method.

Issue #2349, opened 2024-12-17
Failure to allocate thread can lock up local activities


Sushisource:
## Expected Behavior
Local activities should be kept in the queue if a thread can't be allocated to run them

## Actual Behavior
Silent lock up

## Steps to Reproduce the Problem
1. When a local activity task was taken off the LocalActivitySlotSupplierQueue in #processQueue [here](https://github.com/temporalio/sdk-java/blob/53a8af7ec44b262fdd00efa6358893f5e2e7a13c/temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivitySlotSupplierQueue.java#L79-L102), it was passed to the afterReservedCallback [here](https://github.com/temporalio/sdk-java/blob/53a8af7ec44b262fdd00efa6358893f5e2e7a13c/temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivitySlotSupplierQueue.java#L97)
2. This callback is set up [here](https://github.com/temporalio/sdk-java/blob/53a8af7ec44b262fdd00efa6358893f5e2e7a13c/temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivityWorker.java#L94-L96) in LocalActivityWorker, where it forwards each task to the PollTaskExecutor#process method [defined here](https://github.com/temporalio/sdk-java/blob/53a8af7ec44b262fdd00efa6358893f5e2e7a13c/temporal-sdk/src/main/java/io/temporal/internal/worker/PollTaskExecutor.java#L98-L118)
3. That method calls ThreadPoolExecutor#execute [here](https://github.com/temporalio/sdk-java/blob/53a8af7ec44b262fdd00efa6358893f5e2e7a13c/temporal-sdk/src/main/java/io/temporal/internal/worker/PollTaskExecutor.java#L101), which [may spin up a new thread](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ThreadPoolExecutor.html#execute-java.lang.Runnable-)
4. The creation up of that thread failed (see [0]), which thankfully our JVM deployment seems to always log (since I don't believe this comes from the Temporal SDK)
We suspect that this is due to the process running out of memory used for allocating thread stacks
[...]

PR #2371, opened 2025-01-16
Avoid lock up on unexpected ExecutorService errors while executing Local Activities

Sushisource:
## What was changed
Avoid putting the worker into a useless state if submitting a local activity for execution fails (ex: OOM / other thread starting problems). 

## Why?
Bad times when you don't know why your worker is stuck

## Checklist

1. Closes https://github.com/temporalio/sdk-java/issues/2349

2. How was this tested:
Added an integration test with an executor service that fails intentionally

3. Any docs updates needed?

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivitySlotSupplierQueue.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivitySlotSupplierQueue.java
@@ -109,7 +109,8 @@ class LocalActivitySlotSupplierQueue implements Shutdownable {
         if (request != null) {
           LocalActivityExecutionContext executionContext = request.task.getExecutionContext();
           executionContext.callback(
-              LocalActivityResult.processingFailed(executionContext.getActivityId(), 1, e));
+              LocalActivityResult.processingFailed(
+                  executionContext.getActivityId(), request.task.getAttemptTask().getAttempt(), e));
         }
       }
     }
--- a/temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivitySlotSupplierQueue.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivitySlotSupplierQueue.java
@@ -102,7 +102,7 @@ class LocalActivitySlotSupplierQueue implements Shutdownable {
       } catch (Throwable e) {
         // Fail the workflow task if something went wrong executing the local activity (at the
         // executor level, otherwise, the LA handler itself should be handling errors)
-        log.error("Unexpected error executing local activity", e);
+        log.error("Unexpected error submitting local activity task to worker", e);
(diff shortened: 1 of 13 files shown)


temporal-sdk/src/main/java/io/temporal/internal/worker/PollTaskExecutor.java
```diff
@@ -74,18 +75,7 @@ public interface TaskHandler<TT> {
...
-              // for LinkedBlockingQueue we have to set slots to workerTaskSlots to avoid situation
-              // when the queue grows, but the amount of threads is not, because the queue is not
-              // (and
-              // never) full
-              synchronousQueue ? 0 : workerTaskSlots,
-              workerTaskSlots,
-              10,
-              TimeUnit.SECONDS,
-              synchronousQueue ? new SynchronousQueue<>() : new LinkedBlockingQueue<>());
+          new ThreadPoolExecutor(0, threadPoolMax, 10, TimeUnit.SECONDS, new SynchronousQueue<>());
```
Quinn-With-Two-Ns, line 78: So I am fine with this change, but it will mean that any submission to the `threadPoolTaskExecutor` will not be blocking if the `threadPoolMax` is full. Did you confirm the submitter can handle this and we have appropriate test coverage?

temporal-sdk/src/main/java/io/temporal/internal/worker/LocalActivitySlotSupplierQueue.java
```diff
@@ -95,9 +97,22 @@ private void processQueue() {
         }
         request.task.getExecutionContext().setPermit(slotPermit);
         afterReservedCallback.apply(request.task);
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
+      } catch (Throwable e) {
+        // Fail the workflow task if something went wrong executing the local activity (at the
+        // executor level, otherwise, the LA handler itself should be handling errors)
+        log.error("Unexpected error executing local activity", e);
```
Quinn-With-Two-Ns, line 105: Can we make this error message more clear? Without reading the comment it sounds like the error could be from the users activity.
```suggestion
        log.error("Unexpected error submitting local activity task to worker", e);
```
Does this seem accurate?

Sushisource: Yeah, I thought for a bit about what to write here. I had "scheduling" but that felt unclear too. Your wording works for me.

temporal-sdk/src/main/java/io/temporal/internal/worker/PollTaskExecutor.java
Sushisource: Made sure this works via a small change to `testParallelLocalActivities`

Quinn-With-Two-Ns (approved):

PR #2376, opened 2025-01-17
Make sure GetVersion never yields

Quinn-With-Two-Ns:
Make sure GetVersion never yields

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowContext.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowContext.java
@@ -284,7 +284,7 @@ public interface ReplayWorkflowContext extends ReplayAware {
    * @param callback used to return version
    * @return True if the identifier is not present in history
    */
-  boolean getVersion(
+  Integer getVersion(
       String changeId,
       int minSupported,
       int maxSupported,
--- a/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowContext.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowContext.java
@@ -418,6 +418,11 @@ public interface ReplayWorkflowContext extends ReplayAware {
   boolean tryUseSdkFlag(SdkFlag flag);
 
   /**
+   * @return true if this flag is currently set.
+   */
+  boolean checkSdkFlag(SdkFlag flag);
+
+  /**
    * @return The Build ID of the worker which executed the current Workflow Task. May be empty the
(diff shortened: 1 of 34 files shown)


temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowContext.java
```diff
@@ -284,7 +284,7 @@ void mutableSideEffect(
    * @param callback used to return version
    * @return True if the identifier is not present in history
    */
-  boolean getVersion(
+  Integer getVersion(
```
Sushisource, line 287: Docstring needs update

temporal-sdk/src/main/java/io/temporal/internal/statemachines/VersionStateMachine.java
```diff
@@ -379,7 +379,7 @@ private VersionStateMachine(
    * @param callback used to return version
    * @return True if the identifier is not present in history
    */
-  public boolean getVersion(
+  public Integer getVersion(
```
Sushisource, line 382: Docstring needs update here too

temporal-sdk/src/test/java/io/temporal/workflow/versionTests/GetVersionOutOfOrderFailTest.java
```diff
@@ -75,6 +75,9 @@ public void testGetVersionOutOfOrderFail() {
     assertEquals(
         NonDeterministicException.class.getName(),
         ((ApplicationFailure) e.getCause().getCause().getCause()).getType());
+    assertEquals(
+        "[TMPRL1100] getVersion call before the existing version marker event. The most probable cause is retroactive addition of a getVersion call with an existing 'changeId'",
```
Sushisource, line 79:
```suggestion
        "[TMPRL1100] getVersion called before the existing version marker event. The most probable cause is retroactive addition of a getVersion call with an existing 'changeId'",
```

Sushisource (approved):
Makes sense :+1: 

I'm assuming this means you decided against doing the whole two-phase thing?

cretz (approved):

Quinn-With-Two-Ns:
>I'm assuming this means you decided against doing the whole two-phase thing?

@Sushisource not sure what you mean by "the whole two-phase thing"?

Sushisource:
> > I'm assuming this means you decided against doing the whole two-phase thing?
> 
> @Sushisource not sure what you mean by "the whole two-phase thing"?

Doing the change in two parts so rolling back one version would work. But, maybe I'm actually getting this mixed up with a different change.

Issue #1811, opened 2023-07-11
SpringBoot - add @Primary to temporalWorkflowClient WorkflowClient Bean


tsurdilo:
Springboot autoconfig defines the primary WorkflowClient bean thats generated from config as:

      @Bean(name = "temporalWorkflowClient")
      public WorkflowClient client(ClientTemplate clientTemplate) {
        return clientTemplate.getWorkflowClient();
      }

For client app use case sometimes we need to create additional WorkflowClient beans (for example one for a different namespace thats part of the use case). Its not possible to do this via config so
to do this currently once a custom WorkflowClient bean is created, for example:

       @Bean(name = "customWorkflowClient")
       public WorkflowClient secondClient() {
         WorkflowServiceStubs service = WorkflowServiceStubs.newLocalServiceStubs();
         return WorkflowClient.newInstance(
        service, WorkflowClientOptions.newBuilder().setNamespace("xyz").build());
      }

There are two issues with this:
1. you cannot autowire the default WorkflowServiceStubs here because of cyclic dependency on the default template. user has to create a new WorkflowServiceStubs which seems unnecessary. 
2. Once you define this you no longer can resolve autowiring default WorkflowClient generated from config unless you use its name Qualifier, for example you have to do:

        @Autowired
        @Qualifier("temporalWorkflowClient") // gotta do this even if you don't autowire the custom one
        WorkflowClient client;
      
        @Autowired
        @Qualifier("customWorkflowClient")
        WorkflowClient customClient;

I think it would help if we annotate the default WorkflowClient bean with @Primary in order to allow autowiring the default
one without having to use Qualifier directly everywhere in your code. 
Another idea could be to allow config of multiple clients from config directly, but not sure this is best approach maybe.

Wdyt?

cretz:
> For client app use case sometimes we need to create additional WorkflowClient beans

May need this for worker app too not just client app

This is similar to #1799 but more generic. There was internal discussion with myself and @Quinn-With-Two-Ns about how to possibly support multiple clients on multiple workers.

> annotate the default WorkflowClient bean with @primary in order to allow autowiring the default
one without having to use Qualifier directly everywhere in your code.

But what if you want to use that `secondClient` in a different worker? I think multiple clients and multiple workers in Spring needs to be solved in general. What is the general Spring solution for named things being required by other named things?

robzienert:
FWIW, this would work so `@Qualifier` is unnecessary; perhaps already known:

```
@Autowired
[...]

PR #2378, opened 2025-01-19
feat: multi namespace support

### LINKED ISSUE #1799: [Springboot] add support for multiple namespaces
Add support in springboot integration to create clients and workers with different options including different namespaces

Create a springboot application per namespace, doesn't help with cross namespace calls or different client options for different workers in the same namespace

**Additional context**

Yogurt-lei:
## What was changed

The previous namespace is called RootNamespace. 
A new configuration item named "**temporal.namespaces**" is added, and the configurations under it are all Non-rootNamespaces. 
There are different workers and clients will register in different namespaces, and Activities and Workflows corresponding to a specific namespace can be discovered from different namespaces.

## Why?

The current implementation only supports a single namespace. 
In this PR, multi-namespaces support has been added. 

## Checklist

1. Closes 
#1799
#1811

2. How was this tested:

i. add new configuration
```yaml
spring:
  temporal:
    connection:
      target: localhost
    start-workers: false
    namespaces:
      - namespace: assign
        alias: assign
        workers-auto-discovery:
          packages: com.component.temporal.assign
        workers:
          - task-queue: global
      - namespace: unassign
        alias: unassign
        workers-auto-discovery:
          packages: com.component.temporal.unassign
        workers:
          - task-queue: global
```

you can also define per-namespace specific bean.
```java
// Note: TemporalOptionsCustomizer type bean must start with namespace/alias you defined. 
// And end with function class  name you want to customizer and concat Customizer as the bean name. 
// If we found none TemporalOptionsCustomizer defined. will give you a tips when you run the application.
  @Bean
    TemporalOptionsCustomizer<WorkflowServiceStubsOptions.Builder> assignWorkflowServiceStubsCustomizer() {
        return builder -> builder.setKeepAliveTime(Duration.ofHours(1));
    }

// also data converter is support
    @Bean
    DataConverter assignDataConverter() {
        return DataConverter.getDefaultInstance();
    }

```

ii. Wherever you want, we can inject the per-non-namespace bean via byName mode.

```java
   // temporalWorkflowClient  this is the rootNamespace bean. 
    @Resource
    WorkflowClient workflowClient;

   // NOTE: bean name is corresponding namespace alias + ClassSimpleName
    @Resource(name = "assignWorkflowClient")
    private WorkflowClient assignWorkflowClient;

    @Resource(name = "unassignWorkflowClient")
    private WorkflowClient unassignWorkflowClient;
```
iii. Next, you can use beans from different namespaces anywhere.

4. Any docs updates needed?


temporal-spring-boot-autoconfigure/src/main/java/io/temporal/spring/boot/autoconfigure/properties/NonRootNamespaceProperties.java
```diff
@@ -0,0 +1,95 @@
...
+import javax.annotation.Nullable;
+import org.springframework.boot.context.properties.ConstructorBinding;
+
+public class NonRootNamespaceProperties extends NamespaceProperties {
+
+  public static final String NAMESPACE_NON_ROOT = "non-root-" + UUID.randomUUID();
+  private final @Nonnull String alias;
+
+  /** Whether to auto-register the namespace. The default is false. */
+  private final @Nullable Boolean namespaceAutoRegister;
```
Quinn-With-Two-Ns, line 37: A worker should not be registering it's own namespace or even have permission to in a correct setup so we should remove this feature

temporal-spring-boot-autoconfigure/src/main/java/io/temporal/spring/boot/autoconfigure/properties/NonRootNamespaceProperties.java
```diff
@@ -0,0 +1,95 @@
...
+
+  /** The state of the history archival. The default is false means disable. */
+  private final @Nullable Boolean historyArchivalState;
+
+  /** The state of the visibility archival. The default is false means disable. */
+  private final @Nullable Boolean visibilityArchivalState;
+
+  @ConstructorBinding
+  public NonRootNamespaceProperties(
+      @Nonnull String alias,
```
Quinn-With-Two-Ns, line 50: Why is an alias required?

temporal-spring-boot-autoconfigure/src/main/java/io/temporal/spring/boot/autoconfigure/properties/NonRootNamespaceProperties.java
```diff
@@ -0,0 +1,95 @@
...
+import java.time.Duration;
+import java.util.List;
+import java.util.UUID;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import org.springframework.boot.context.properties.ConstructorBinding;
+
+public class NonRootNamespaceProperties extends NamespaceProperties {
+
+  public static final String NAMESPACE_NON_ROOT = "non-root-" + UUID.randomUUID();
```
Quinn-With-Two-Ns, line 33: Why do we generate a random namespace name?

temporal-spring-boot-autoconfigure/src/main/java/io/temporal/spring/boot/autoconfigure/properties/NonRootNamespaceProperties.java
```diff
@@ -0,0 +1,95 @@
...
+  public NonRootNamespaceProperties(
+      @Nonnull String alias,
+      @Nullable String namespace,
+      @Nullable Boolean namespaceAutoRegister,
+      @Nullable Boolean historyArchivalState,
+      @Nullable Boolean visibilityArchivalState,
+      @Nullable Duration retentionPeriod,
+      @Nullable WorkersAutoDiscoveryProperties workersAutoDiscovery,
+      @Nullable List<WorkerProperties> workers,
+      @Nullable WorkflowCacheProperties workflowCache) {
```
Quinn-With-Two-Ns, line 58: We also need to support:
* Different connection options for different namespaces
* Different data converter for different namespaces

Quinn-With-Two-Ns (commented):
Hey, thanks for opening this PR. Multi-namepace support is something we are interesting in adding. I left some comments on some other per namespace configurations we need.

temporal-spring-boot-autoconfigure/src/main/java/io/temporal/spring/boot/autoconfigure/properties/NonRootNamespaceProperties.java
Yogurt-lei: Do you mean application should not register their namespace, all namespace should register via **tctl**? Maybe its the best practice i guess.

temporal-spring-boot-autoconfigure/src/main/java/io/temporal/spring/boot/autoconfigure/properties/NonRootNamespaceProperties.java
Yogurt-lei: Currently use the alias as per-namespace bean prefix. maybe sometimes namespace it's not simple and easy to use. or i can change it to bean-prefix to easy understand.

temporal-spring-boot-autoconfigure/src/main/java/io/temporal/spring/boot/autoconfigure/properties/NonRootNamespaceProperties.java
Yogurt-lei: Maybe need let it non-null. what's your opinion?

temporal-spring-boot-autoconfigure/src/main/java/io/temporal/spring/boot/autoconfigure/properties/NonRootNamespaceProperties.java
Yogurt-lei: Different namespace can connect to different temporal server? 
Currently temporal cluster can support huge request i guess. so i'm not sure it is a necessary feature.

Let me explain more. In my project. why we need this feature is we have a ring-test run every 15 mins. and it will trigger workflow because we only support one namespace now, and with one retention date, 
As time goes by, huge test data has been stored. these test data take up huge disk space, putting great pressure on database. also real-biz data is mixed with test data , it's not good experience


Anyway, if real want Different connection options for different namespaces, i can handle it.

temporal-spring-boot-autoconfigure/src/main/java/io/temporal/spring/boot/autoconfigure/properties/NonRootNamespaceProperties.java
Quinn-With-Two-Ns: Applications should not register their namespace or even have permission to register a namespace.

temporal-spring-boot-autoconfigure/src/main/java/io/temporal/spring/boot/autoconfigure/properties/NonRootNamespaceProperties.java
Quinn-With-Two-Ns: Yeah so I understand why a user may want an alias, what isn't clear is why it would be required?

temporal-spring-boot-autoconfigure/src/main/java/io/temporal/spring/boot/autoconfigure/properties/NonRootNamespaceProperties.java
Quinn-With-Two-Ns: Yeah the namespace should be required, I can't think of a use case where you are connecting to a namespace but don't know what it is called?

temporal-spring-boot-autoconfigure/src/main/java/io/temporal/spring/boot/autoconfigure/properties/NonRootNamespaceProperties.java
Quinn-With-Two-Ns: Different namespaces can be in seprate Temporal clusters. Even if the namespaces are in the same cluster they can have different requirements like different connection setting (like TLS certs. or API keys)

temporal-spring-boot-autoconfigure/src/main/java/io/temporal/spring/boot/autoconfigure/properties/NonRootNamespaceProperties.java
Yogurt-lei: Ok. will set alias as nullable, if user set value, use alias as bean prefix, if not, use namespace

Yogurt-lei:
@Quinn-With-Two-Ns  Support customizer and support different connection by namespace.
@cretz Hope you still remeber this feature. 
Let me know if we need more support for this feature.

temporal-sdk/src/main/java/io/temporal/common/converter/DataConverter.java
```diff
@@ -70,6 +70,10 @@
  */
 public interface DataConverter {
 
+  default boolean useInRootNamespace() {
```
Quinn-With-Two-Ns, line 73: The Dataconverter interface should not have any reference to springboot

Quinn-With-Two-Ns: The data converter should follow the same principle as all the other per NS beans

temporal-spring-boot-autoconfigure/src/main/java/io/temporal/spring/boot/autoconfigure/RootNamespaceAutoConfiguration.java
```diff
@@ -80,21 +86,20 @@ public NamespaceTemplate rootNamespaceTemplate(
       @Autowired(required = false) @Nullable Tracer otTracer,
       @Qualifier("temporalTestWorkflowEnvironmentAdapter") @Autowired(required = false) @Nullable
           TestWorkflowEnvironmentAdapter testWorkflowEnvironment,
-      @Autowired(required = false) @Nullable
+      @Qualifier("temporalWorkerFactoryCustomizer") @Autowired(required = false) @Nullable
```
Quinn-With-Two-Ns, line 89: Wouldn't adding this `@Qualifier` here filter break existing users since their `TemporalOptionsCustomizer` bean may not be named this?

Yogurt-lei: Resolved, currently lookup specify bean from autowired map, if it is not non root namespace bean, found the first one like before.

Quinn-With-Two-Ns:
Thank you, I will take a final review tomorrow (PST)

temporal-spring-boot-autoconfigure/src/test/java/io/temporal/spring/boot/autoconfigure/OptionalWorkerOptionsTest.java
```diff
@@ -96,7 +97,7 @@ public TemporalOptionsCustomizer<WorkerFactoryOptions.Builder> workerFactoryCust
 
     @Bean
     @SuppressWarnings("unchecked")
-    public TemporalOptionsCustomizer<WorkerOptions.Builder> workerCustomizer() {
+    public TemporalOptionsCustomizer<WorkerOptions.Builder> temporalWorkerCustomizer() {
```
Quinn-With-Two-Ns, line 100: These renames are no longer necessary correct?

Yogurt-lei: yes, correct. let me revert this.

Yogurt-lei: one question to ask, how to resolve this features integration test issue? can you give me some guide?

Quinn-With-Two-Ns: Yes sorry, please make sure you pushed the git `tags` to your branch. Our CI currently uses that to do some tests across versions.

Quinn-With-Two-Ns: let me know when you have and I will re-run it

Yogurt-lei: got it, thank you , i have pushed 1.28.1 .

Yogurt-lei:
@Quinn-With-Two-Ns  sorry, made mistask, forget add tag prefix **v**

temporal-spring-boot-autoconfigure/src/test/java/io/temporal/spring/boot/autoconfigure/OptionalWorkerOptionsTest.java
Quinn-With-Two-Ns: To clarify you just need to pull the tags from `master` and add them to your branch, don't need to add a new tag

Yogurt-lei: ok, let me retry this. thanks your time

Quinn-With-Two-Ns (approved):

Issue #2361, opened 2025-01-07
Access to workflow/activity instance from context


cretz:
**Describe the solution you'd like**

Provide access to the Temporal created/owned workflow and activity instance from their respective contexts. This means adding something like `ActivityExecutionContext.getInstance()` instance method and `Workflow.getInstance()` static method. See https://github.com/temporalio/features/issues/572.

PR #2384, opened 2025-01-22
Access to workflow/activity instance from context

Quinn-With-Two-Ns:
Provide access to workflow/activity instance from their context. There is some subtly around when the workflow instance is created, I tried to document this in the user visible method.

Note: I did not do anything for Nexus here because the `OperationHandler` is already passed to the interceptor.

closes: https://github.com/temporalio/sdk-java/issues/2361

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/activity/ActivityExecutionContext.java
+++ b/temporal-sdk/src/main/java/io/temporal/activity/ActivityExecutionContext.java
@@ -147,4 +147,7 @@ public interface ActivityExecutionContext {
    * an activity.
    */
   WorkflowClient getWorkflowClient();
+
+  /** Get the currently running activity instance. */
+  Object getInstance();
 }
--- a/temporal-sdk/src/main/java/io/temporal/workflow/Workflow.java
+++ b/temporal-sdk/src/main/java/io/temporal/workflow/Workflow.java
@@ -29,6 +29,7 @@ import io.temporal.common.RetryOptions;
 import io.temporal.common.SearchAttributeUpdate;
 import io.temporal.common.SearchAttributes;
 import io.temporal.common.converter.DataConverter;
+import io.temporal.common.interceptors.WorkflowOutboundCallsInterceptor;
 import io.temporal.failure.ActivityFailure;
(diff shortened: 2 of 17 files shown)


temporal-sdk/src/main/java/io/temporal/activity/ActivityExecutionContext.java
```diff
@@ -147,4 +147,7 @@ public interface ActivityExecutionContext {
    * an activity.
    */
   WorkflowClient getWorkflowClient();
+
+  /** Get the currently running activity instance. */
+  Object getInstance();
```
cretz, line 152: General question, is this available in the main execute-activity inbound interceptor? Meaning there's no case where it is null (and Java has no concept of accessing the execution context from inside activity class constructors since users invoke those I assume).

temporal-sdk/src/main/java/io/temporal/workflow/Workflow.java
```diff
@@ -1385,6 +1386,20 @@ public static String getCurrentDetails() {
     return WorkflowInternal.getCurrentDetails();
   }
 
+  /**
+   * Get the currently running workflow instance.
+   *
+   * @apiNote The instance is only available after it has been initialized. This function will
+   *     return null if called before the workflow has been initialized. For example, this could
+   *     happen if the function is called from a {@link WorkflowInit} constructor or {@link
+   *     io.temporal.common.interceptors.WorkflowInboundCallsInterceptor#init(WorkflowOutboundCallsInterceptor)}.
```
cretz, line 1395: So long as available in `execute` of this same interceptor, I think we're good

temporal-sdk/src/main/java/io/temporal/activity/ActivityExecutionContext.java
Quinn-With-Two-Ns: Correct

temporal-sdk/src/main/java/io/temporal/workflow/Workflow.java
Quinn-With-Two-Ns: Yeah in the test I show this is not null in `execute`

PR #2390, opened 2025-01-27
Enable speculative workflow tasks

Quinn-With-Two-Ns:
Enable speculative workflow tasks


temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowRunTaskHandler.java
```diff
@@ -219,7 +219,7 @@ public QueryResult handleDirectQueryWorkflowTask(
 
   @Override
   public void resetStartedEvenId(Long eventId) {
```
Sushisource, line 221: This one is still "even id"

Sushisource (approved):

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowRunTaskHandler.java
Quinn-With-Two-Ns: 💀💀💀

cretz (approved):

Issue #2018, opened 2024-03-26
Kotlin `ActivityStubExt.execute*` extension functions wrap arguments in list


rgrochowicz:
## Expected Behavior

When I use the `ActivityStubExt.execute` extension function, I expect the `args` parameter to be given as-is to the underlying `ActivityStub.execute` method (keeping the varargs intact).

## Actual Behavior

The varargs `args` parameters to `ActivityStubExt.execute` is wrapped in a list rather than provided as a varargs parameter to `ActivityStub.execute`.

## Steps to Reproduce the Problem

  1. Create a new ActivityStub via `Workflow.newUntypedActivityStub`
  1. Import the Kotlin extension method via `import io.temporal.workflow.execute`
  1. Call `activityStub.execute<Void>("activityName", object { val foo = 1 })`
  1. Observe that the arguments were wrapped in a list rather than given directly to `ActivityStub.execute`

More complete example:

```kotlin
import io.temporal.workflow.execute
import io.temporal.activity.ActivityOptions

// ... within workflow ...

val activityStub = Workflow.newUntypedActivityStub(ActivityOptions {})
activityStub.execute("activityName", object { val foo = 1 })

```

The `ActivityTaskScheduled` event has an input of:

```json
[
  [
    {
      "foo": 1
    }
  ]
]
```

Rather than:

```json
[
  {
    "foo": 1
  }
]
```

The incorrect input is wrapped in an extra list.

To fix, I suspect that the call to `ActivityStub.execute` with varargs in [ActivityStubExt.kt](https://github.com/temporalio/sdk-java/blob/master/temporal-kotlin/src/main/kotlin/io/temporal/workflow/ActivityStubExt.kt#L36) needs to use the spread operator (`*`).

## Specifications

  - Version: Java SDK v1.23.0
  - Platform: macOS / Linux

Quinn-With-Two-Ns:
Yes I think we are missing a spread operator, thank you for reporting this. I will follow up with my team and consider if there are any backwards compatibility issues with fixing this.

Quinn-With-Two-Ns:
https://github.com/temporalio/sdk-java/pull/2395/files

PR #2395, opened 2025-01-30
Fix kotlin child workflow execute varargs

adamlehenbauer:
## What was changed

The kotlin `execute` extension functions on child workflows aren't working for me. My args are wrapped in an array, causing my child workflow to fail when deserializing it's workflow arguments.

I'm no expert on kotlin/java varargs, but it seems to me that the varargs need to be spread so that the java code sees varargs, not a one element array of varargs.

I've added test cases that fail in the way that I'm seeing if the spread operator is dropped, and ensure correct functioning going forward.

Related issues: https://github.com/temporalio/sdk-java/issues/2018

## Why?

I'm using untyped child workflows myself, and passing my arguments through is leading to workflow failures.

## Checklist

1. Closes n/a

2. How was this tested:

* The unit tests provided exercise this fix and demonstrate the bug without the fix
* I've tested in a real application with a copy/pasted `execute` function that includes the spread, and this works as expected

3. Any docs updates needed?

n/a


temporal-kotlin/src/main/kotlin/io/temporal/workflow/ChildWorkflowStubExt.kt
```diff
@@ -28,13 +28,13 @@ import kotlin.reflect.typeOf
  */
 @OptIn(ExperimentalStdlibApi::class)
 inline fun <reified T> ChildWorkflowStub.execute(vararg args: Any?): T {
-  return execute(T::class.java, typeOf<T>().javaType, args)
+  return execute(T::class.java, typeOf<T>().javaType, *args)
```
adamlehenbauer, line 31: If not spread here, the `args` are considered a single argument to the java `execute` function. 

So if passing in a single string arg like "my-arg", the child workflow will receive `[ "my-arg" ]` as it's first argument, rather than `"my-arg"`

Quinn-With-Two-Ns:
Hey, thanks for the contribution. I think Activities also have the same [issue](https://github.com/temporalio/sdk-java/issues/2018), are you able to resolve that as part of this PR as well?

adamlehenbauer:
> Hey, thanks for the contribution. I think Activities also have the same [issue](https://github.com/temporalio/sdk-java/issues/2018), are you able to resolve that as part of this PR as well?

Updated to include the activity executes

Quinn-With-Two-Ns (approved):

adamlehenbauer:
@Quinn-With-Two-Ns anything I need to do here?

Quinn-With-Two-Ns:
Looks like CI is failing, can you please make sure tags are pushed to your branch?  don't think I can push them for you. Then let me know and I will re-run CI. Thanks!

adamlehenbauer:
@Quinn-With-Two-Ns which tags? I haven't created any new ones. Pushing all failed for me (` ! [remote rejected]   v0.29.0 -> v0.29.0 (refusing to allow an OAuth App to create or update workflow `.github/workflows/gradle-wrapper-validation.yml` without `workflow` scope)`.

I've synced the fork now, not sure if that helps

Issue #2058, opened 2024-05-09
Set Temporal-Namespace header on every namespace-specific gRPC request


cretz:
**Describe the solution you'd like**

See https://github.com/temporalio/features/issues/475. It needs to be an interceptor surely, but unsure how the implementation will do this generically. One way is via reflection to call `getNamespace() string` assuming it's not deemed too expensive (probably not), but this is left up to the implementer. All that matters is that new RPCs that come in with namespaces automatically have the header set.

cretz:
Note, once this is released, any samples/docs that reference explicit namespace setting we can probably alter

PR #2400, opened 2025-02-03
Inject namespace header

Quinn-With-Two-Ns:
Inject Namespace in grpc headers. Unfortunately I couldn't use gRPC interceptors as originally proposed because java's gRPC interceptors cannot modify the header based of the content of the message https://github.com/grpc/grpc-java/issues/4684#issuecomment-410315965

closes https://github.com/temporalio/sdk-java/issues/2058

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/client/NamespaceInjectWorkflowServiceStubs.java
+++ b/temporal-sdk/src/main/java/io/temporal/client/NamespaceInjectWorkflowServiceStubs.java
@@ -34,16 +34,15 @@ import java.util.concurrent.TimeUnit;
 import java.util.function.Supplier;
 import javax.annotation.Nullable;
 
+/** Inject the namespace into the gRPC header */
 class NamespaceInjectWorkflowServiceStubs implements WorkflowServiceStubs {
   private static Metadata.Key<String> TEMPORAL_NAMESPACE_HEADER_KEY =
       Metadata.Key.of("temporal-namespace", Metadata.ASCII_STRING_MARSHALLER);
   private final Metadata metadata;
-  private WorkflowServiceStubs next;
-  private String namespace;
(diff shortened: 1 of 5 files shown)


temporal-sdk/src/main/java/io/temporal/client/NamespaceInjectWorkflowServiceStubs.java
```diff
@@ -0,0 +1,114 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+import javax.annotation.Nullable;
+
+class NamespaceInjectWorkflowServiceStubs implements WorkflowServiceStubs {
+  private static Metadata.Key<String> TEMPORAL_NAMESPACE_HEADER_KEY =
+      Metadata.Key.of("temporal-namespace", Metadata.ASCII_STRING_MARSHALLER);
+  private final Metadata metadata;
+  private WorkflowServiceStubs next;
+  private String namespace;
```
cretz, line 42: Pedantic but these can be final too

cretz (approved):
LGTM. As a separate issue, we should consider a test with cloud data plane and API key to confirm this works there, but I may open some separate issues on that.

nzuguem:
Since release v1.28.0, I've had a problem (`io.grpc.StatusRuntimeException: PERMISSION_DENIED: Request unauthorized`) with authentication via API Key to Temporal Cloud. After investigation (via `git bisect`) I can see that it's this PR that's causing the problem.

I may create a separate issue

nzuguem:
Thanks for your answer. I'm going to create a ticket for Temporal Cloud.

I'd like to remind you that this worked normally with SDK v1.27.1, but since v1.28.0 with the addition of interceptors in this PR, it no longer works. I've even tried removing the interceptors I added manually, but it still doesn't work.

PR #2403, opened 2025-02-07
Remove Nexus sync client handler

Quinn-With-Two-Ns:
Remove Nexus sync client handler, users can now get the client with `Nexus.getOperationContext().getWorkflowClient()`


temporal-sdk/src/main/java/io/temporal/nexus/WorkflowClientOperationHandlers.java
```diff
@@ -52,9 +36,8 @@ public static <T, R> OperationHandler<T, R> sync(
   public static <T, R> OperationHandler<T, R> fromWorkflowMethod(
       WorkflowMethodFactory<T, R> startMethod) {
     return new RunWorkflowOperation<>(
```
bergundy, line 38: Doesn't have to be in this PR but I would to call this `WorkflowRunOperation` across all SDKs.

temporal-sdk/src/main/java/io/temporal/nexus/NexusOperationContext.java
```diff
@@ -39,4 +40,10 @@ public interface NexusOperationContext {
    * WorkflowServiceStubsOptions.Builder#setMetricsScope(Scope)} when a worker starts up.
    */
   Scope getMetricsScope();
+
+  /**
+   * Get a {@link WorkflowClient} that can be used to start interact with the Temporal service from
+   * a Nexus handler.
+   */
+  WorkflowClient getWorkflowClient();
```
bergundy, line 48: Maybe add a method to get the nexus context too?

bergundy (approved):
No blocking comments for this PR but we should do another round on the API to see if we can get more consistency across the languages.

temporal-sdk/src/main/java/io/temporal/nexus/NexusOperationContext.java
Quinn-With-Two-Ns: If a use case comes up yes, at least internally.

Sushisource (approved):

PR #2405, opened 2025-02-07
Rename WorkflowClientOperationHandlers->WorkflowRunOperation

Quinn-With-Two-Ns:
Rename `WorkflowClientOperationHandlers`->`WorkflowRunOperation`. Previously we thought we would have operation wrappers for client operations like signal. Now it looks like we are going to go with a helper approach so a generic `WorkflowClientOperationHandlers` isn't very useful. It also brings the Java API closer to Go


bergundy (approved):
Thanks!

cretz (approved):

PR #2411, opened 2025-02-10
Java test server support for attaching links to signal and signal with start requests

justinp-tt:
## What was changed

Java test server support for attaching links to signal and signal with start requests

## Why?

We need links support

## Checklist

1. Closes 
SDK-3344

2. How was this tested:

Unit tests
3. Any docs updates needed?


temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -38,43 +31,129 @@
 import io.grpc.Deadline;
 import io.grpc.Status;
 import io.grpc.StatusRuntimeException;
-import io.temporal.api.command.v1.*;
-import io.temporal.api.common.v1.*;
-import io.temporal.api.enums.v1.*;
+import io.temporal.api.command.v1.CancelTimerCommandAttributes;
```
bergundy, line 34: Seems like you IDE has different settings from those generally used in this codebase, you should probably revert to using `*` imports.

temporal-test-server/src/test/java/io/temporal/testserver/functional/SignalLinksTest.java
```diff
@@ -0,0 +1,166 @@
...
+                        .blockingStub()
+                        .signalWithStartWorkflowExecution(signalWithStartRequest);
+
+        WorkflowExecution execution =
+                WorkflowExecution.newBuilder()
+                        .setWorkflowId(workflowId)
+                        .setRunId(response.getRunId())
+                        .build();
+
+        verifySignalLink(execution, testLink);
```
bergundy, line 123: Please verify that there's also a link on the WorkflowExecutionStarted event with signal-with-start.

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
justinp-tt: I'm surprised we allow wildcard imports

justinp-tt: Unless there is a strong aversion, I'd rather not use wildcard imports

Quinn-With-Two-Ns: Yes please do not change the conventions in this PR

justinp-tt: What are the conventions? After N imports, use wildcards. After M static imports, use wildcards. What are N and M?

bergundy (approved):
Approving based on the test. Please fix the import changes.

Will defer to @Quinn-With-Two-Ns for the final approval.

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
Quinn-With-Two-Ns: 5 imports and 3 static

Quinn-With-Two-Ns (approved):

PR #2415, opened 2025-02-13
Add OnConflictOptions Support

justinp-tt:
## What was changed

Add OnConflictOptions support in the test server

## Why?

Test server needs OnConflictOptions support

## Checklist

1. Closes 
SDK-3343
2. How was this tested:

3. Any docs updates needed?


temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
```diff
@@ -169,27 +158,26 @@ static final class WorkflowData {
...
-          + ", firstExecutionRunId='"
-          + firstExecutionRunId
-          + '\''
-          + ", originalExecutionRunId='"
-          + originalExecutionRunId
-          + '\''
-          + ", continuedExecutionRunId="
-          + continuedExecutionRunId
-          + '}';
+      return toStringHelper(this)
```
Quinn-With-Two-Ns, line 171: Please leave this as is, it is not consistent with the rest of SDK codebase

temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
```diff
@@ -2098,7 +2086,7 @@ private static RetryState attemptActivityRetry(
       }
       if (info.get().hasNextRetryDelay()) {
         nextRetryDelay =
-            Optional.ofNullable(ProtobufTimeUtils.toJavaDuration(info.get().getNextRetryDelay()));
+            ofNullable(ProtobufTimeUtils.toJavaDuration(info.get().getNextRetryDelay()));
```
Quinn-With-Two-Ns, line 2089:
```suggestion
            Optional.ofNullable(ProtobufTimeUtils.toJavaDuration(info.get().getNextRetryDelay()));
```

justinp-tt: Are you anti-static imports? :)

temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
```diff
@@ -148,6 +134,8 @@ static final class WorkflowData {
 
     Functions.Proc runTimerCancellationHandle;
 
+    private final Map<String, StartWorkflowExecutionResponse> requestIdToResponse = new HashMap<>();
```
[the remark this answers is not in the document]
bergundy: Also, did we not have request ID deduping before this feature?
You should make sure that any request ID is deduped regardless of `OnConflictOptions` to match the real server's behavior.

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -613,6 +614,61 @@ public void completeWorkflowTask(
...
+      StateMachines.WorkflowData data = workflow.getData();
+      OnConflictOptions options = request.getOnConflictOptions();
+
+      if (options.getAttachRequestId()) {
+        Optional<StartWorkflowExecutionResponse> cachedResponse =
+            data.getResponseForRequestId(request.getRequestId());
+        if (cachedResponse.isPresent()) {
+          return cachedResponse.get();
+        }
+      }
```
bergundy, line 631: This condition should be checked outside of this function, there should be one location where any request ID is deduped regardless of onConflictOptions.

bergundy (commented):
Overall LGTM.

Could you please also add tests?

temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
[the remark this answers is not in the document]
justinp-tt: I am not aware of any prior request ID deduping, as this is my first foray into the Java SDK. @Quinn-With-Two-Ns -- do you know the status on this?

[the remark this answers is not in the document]
justinp-tt: Switched Map to Set

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
justinp-tt: That's fair. It looks like the workflow data is not exposed outside of TestWorkflowMutableStateImpl. We could expose another interface method to encapsulate the dedupe logic. Let's see if @Quinn-With-Two-Ns has any preference

temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
```diff
@@ -20,33 +20,18 @@
...
-import static io.temporal.internal.testservice.StateMachines.State.FAILED;
-import static io.temporal.internal.testservice.StateMachines.State.INITIATED;
-import static io.temporal.internal.testservice.StateMachines.State.NONE;
-import static io.temporal.internal.testservice.StateMachines.State.STARTED;
-import static io.temporal.internal.testservice.StateMachines.State.TERMINATED;
-import static io.temporal.internal.testservice.StateMachines.State.TIMED_OUT;
+import static io.temporal.internal.common.LinkConverter.nexusLinkToWorkflowEvent;
+import static io.temporal.internal.common.LinkConverter.workflowEventToNexusLink;
+import static io.temporal.internal.testservice.StateMachines.Action.*;
+import static io.temporal.internal.testservice.StateMachines.State.*;
```
rodrigozhou, line 26: This looks like it will have conflicting symbols eventually.

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
```diff
@@ -261,40 +263,41 @@ StartWorkflowExecutionResponse startWorkflowExecutionImpl(
...
+                                    .build())
+                            .build())
+                    .build());
+          } else if (conflictPolicy
+              == WorkflowIdConflictPolicy.WORKFLOW_ID_CONFLICT_POLICY_USE_EXISTING) {
+            if (startRequest.hasOnConflictOptions()) {
+              return existing.applyOnConflictOptions(startRequest);
+            }
+            return StartWorkflowExecutionResponse.newBuilder()
+                .setRunId(existing.getExecutionId().getExecution().getRunId())
```
rodrigozhou, line 292: Why did you remove `.setStarted(false)`? Shouldn't it be kept?

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -613,6 +614,59 @@ public void completeWorkflowTask(
...
+  public StartWorkflowExecutionResponse applyOnConflictOptions(
+      @Nonnull StartWorkflowExecutionRequest request) {
+    lock.lock();
+    try {
+      StateMachines.WorkflowData data = workflow.getData();
+      OnConflictOptions options = request.getOnConflictOptions();
+
+      StartWorkflowExecutionResponse response =
+          StartWorkflowExecutionResponse.newBuilder()
+              .setRunId(getExecutionId().getExecution().getRunId())
```
rodrigozhou, line 627: Add `.setStarted(false)`?

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -613,6 +614,59 @@ public void completeWorkflowTask(
...
+            HistoryEvent.newBuilder()
+                .setEventType(EventType.EVENT_TYPE_WORKFLOW_EXECUTION_OPTIONS_UPDATED)
+                .setWorkflowExecutionOptionsUpdatedEventAttributes(attrs)
+                .addAllLinks(request.getLinksList())
+                .build();
+
+        RequestContext ctx = new RequestContext(clock, this, nextEventId);
+        ctx.addEvent(event);
+        store.save(ctx);
+      }
```
rodrigozhou, line 658: We're always adding the history event if `OnConflictOptions` is not null, not only when `AttachLinks` is true.

rodrigozhou (changes requested):

temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
Quinn-With-Two-Ns: I am just pro consistency with the rest of the codebase

temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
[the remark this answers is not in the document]
Quinn-With-Two-Ns: The test server doesn't have request ID deduping. I agree in principle it should to match the real server, but the test server is not used over a unreliable connection so this doesn't seem incredibly important. I think we can just add whatever request ID deduping we need for this issue and open a separate issue to add request ID deduping in general.

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
Quinn-With-Two-Ns: Yeah I think we should have a method that handles deduping and as part of that applies the conflict options?

temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
justinp-tt: Just following codebase conventions. When conflict arises with use, it's easy to resolve the import issue.

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
justinp-tt: It will default to `false`

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
justinp-tt: It will default to `false`

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
justinp-tt: Thanks! Resolved!

temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
justinp-tt: I think the static import increases code readability, but I don't have a strong preference.

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -613,6 +614,58 @@ public void completeWorkflowTask(
...
+      StartWorkflowExecutionResponse response =
+          StartWorkflowExecutionResponse.newBuilder()
+              .setRunId(getExecutionId().getExecution().getRunId())
+              .build();
+      if (options.getAttachRequestId()) {
+        boolean hasRequestId = data.hasRequestId(request.getRequestId());
+        if (hasRequestId) {
+          return response;
+        }
+      }
```
rodrigozhou, line 634: nit

```suggestion
      if (options.getAttachRequestId() && data.hasRequestId(request.getRequestId())) {
        return response;
      }
```

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -613,6 +614,58 @@ public void completeWorkflowTask(
...
+        eventBuilder.addAllLinks(request.getLinksList());
+      }
+
+      RequestContext ctx = new RequestContext(clock, this, nextEventId);
+      ctx.addEvent(eventBuilder.build());
+      store.save(ctx);
+
+      if (options.getAttachRequestId()) {
+        data.addRequestId(request.getRequestId());
+      }
```
rodrigozhou, line 661: Why not do this together in L639?

rodrigozhou (commented):
We changed the server validation of `StartWorkflowExecutionRequest.OnConflictOptions`. Now, if `AttachCompletionCallbacks` is true, then `AttachRequestId` has to be true as well. Not sure if we need to make any changes here.

temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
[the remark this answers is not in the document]
justinp-tt: Tracking this under the other open conversation below. Thx

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
```diff
@@ -257,44 +259,51 @@ StartWorkflowExecutionResponse startWorkflowExecutionImpl(
...
-              .build();
-        } else if (status == WorkflowExecutionStatus.WORKFLOW_EXECUTION_STATUS_RUNNING
-            || reusePolicy == WorkflowIdReusePolicy.WORKFLOW_ID_REUSE_POLICY_REJECT_DUPLICATE) {
-          return throwDuplicatedWorkflow(startRequest, existing);
-        } else if (reusePolicy
-                == WorkflowIdReusePolicy.WORKFLOW_ID_REUSE_POLICY_ALLOW_DUPLICATE_FAILED_ONLY
-            && (status == WorkflowExecutionStatus.WORKFLOW_EXECUTION_STATUS_COMPLETED
-                || status == WorkflowExecutionStatus.WORKFLOW_EXECUTION_STATUS_CONTINUED_AS_NEW)) {
+        if (status == WORKFLOW_EXECUTION_STATUS_RUNNING) {
+          if (reusePolicy == WORKFLOW_ID_REUSE_POLICY_TERMINATE_IF_RUNNING
```
bergundy, line 272: There's some edge cases that we handled in the server (or at least planned to) for reuse policy terminate if running and the different conflict policies.

Someone can address that in a follow up PR though.

CC @rodrigozhou

Quinn-With-Two-Ns:
As part of this PR please remove the checks for the real server on these tests since this PR should enable them to pass on the test server:

- https://github.com/temporalio/sdk-java/blob/master/temporal-sdk/src/test/java/io/temporal/workflow/nexus/WorkflowHandleUseExistingOnConflictTest.java#L53
- https://github.com/temporalio/sdk-java/blob/master/temporal-sdk/src/test/java/io/temporal/workflow/nexus/WorkflowHandleUseExistingOnConflictCancelTest.java#L55
- https://github.com/temporalio/sdk-java/blob/master/temporal-sdk/src/test/java/io/temporal/workflow/nexus/WorkflowHandleFailOnConflictTest.java#L56

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
```diff
@@ -257,44 +259,51 @@ StartWorkflowExecutionResponse startWorkflowExecutionImpl(
...
+                            .build())
+                    .build());
+          } else if (conflictPolicy
+              == WorkflowIdConflictPolicy.WORKFLOW_ID_CONFLICT_POLICY_USE_EXISTING) {
+            if (startRequest.hasOnConflictOptions()) {
+              existing.applyOnConflictOptions(startRequest);
+            }
+            return StartWorkflowExecutionResponse.newBuilder()
+                .setRunId(existing.getExecutionId().getExecution().getRunId())
+                .setStarted(true)
```
bergundy, line 298: This should be false, the workflow wasn't started in this request.

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
```diff
@@ -1856,6 +1865,30 @@ public WorkflowServiceStubs newClientStub() {
...
+        String existingRequestId = state.getStartRequest().getRequestId();
+        if (requestId.equals(existingRequestId)) {
+          if (state.getWorkflowExecutionStatus()
+              == WorkflowExecutionStatus.WORKFLOW_EXECUTION_STATUS_RUNNING) {
+            if (startRequest.hasOnConflictOptions()) {
+              state.applyOnConflictOptions(startRequest);
+            }
+            return StartWorkflowExecutionResponse.newBuilder()
+                .setRunId(state.getExecutionId().getExecution().getRunId())
+                .setStarted(true)
```
bergundy, line 1882: This should be false if the workflow is already running.

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -613,6 +614,43 @@ public void completeWorkflowTask(
         request.hasStickyAttributes() ? request.getStickyAttributes() : null);
   }
 
+  @Override
+  public void applyOnConflictOptions(@Nonnull StartWorkflowExecutionRequest request) {
```
bergundy, line 618: Note that we added validation that attachCompletionCallbacks cannot be true if attachRequestId is false in the server, you should validate that in the test server and reject such requests with an invalid argument error.

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -613,6 +614,43 @@ public void completeWorkflowTask(
...
+      OnConflictOptions options = request.getOnConflictOptions();
+      WorkflowExecutionOptionsUpdatedEventAttributes.Builder attrs =
+          WorkflowExecutionOptionsUpdatedEventAttributes.newBuilder();
+      if (options.getAttachRequestId()) {
+        attrs.setAttachedRequestId(request.getRequestId());
+      }
+
+      if (options.getAttachCompletionCallbacks()
+          && !request.getCompletionCallbacksList().isEmpty()) {
+        attrs.addAllAttachedCompletionCallbacks(request.getCompletionCallbacksList());
```
Quinn-With-Two-Ns, line 630: Do we resolve all these callbacks we are attaching? I can't find it anywhere in this PR

justinp-tt: applyOnConflictOptions calls `ctx.commitChanges(store);` The save implementation makes a call to `fireCallbacks`. Is that where we would expect callbacks to be invoked?

bergundy: I was assuming that we would trigger any callback attached to the workflow, that should be validated when tests are added.

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
bergundy: @rodrigozhou have you reviewed this code and verified the different edge cases?
Can wait for a follow up PR.

rodrigozhou: I'll do in a follow up PR, this one is already quite big.

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
rodrigozhou: It's already resolving. I just had to change to get the callbacks from this new attribute I added to `TestWorkflowMutableStateImpl` instead of reading from the `startRequest`.

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -1718,8 +1748,16 @@ private void processWorkflowCompletionCallbacks(RequestContext ctx) {
...
+      try {
+        service.completeNexusOperation(
+            ref, ctx.getExecution().getWorkflowId(), startLink, completionEvent.get());
+      } catch (StatusRuntimeException e) {
+        // Callback destination not found should not block processing the callbacks nor
+        // completing the workflow.
+        if (e.getStatus().getCode() != Status.Code.NOT_FOUND) {
+          throw e;
+        }
+      }
```
rodrigozhou, line 1760: @pdoerner This is what I did to address callback "not found".

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
```diff
@@ -1465,6 +1488,7 @@ public String continueAsNew(
       String identity,
       ExecutionId continuedExecutionId,
       String firstExecutionRunId,
+      TestWorkflowMutableState previousExecutionState,
```
rodrigozhou, line 1491: Need the previous mutable state so we can copy all callbacks to the continued as new workflow (L1511-1514 below).

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
pdoerner: I think if we do get NOT_FOUND (or really any error) we should log an error and/or store the failure somewhere so that it can be returned as part of the DescribeWorkflow response which has fields for callback's last failure, attempt time, etc

pdoerner: But we can also address that in a followup, doesn't necessarily have to be in this PR

Quinn-With-Two-Ns (approved):

Issue #2418, opened 2025-02-18
workflow_task_execution_failed metric sometime doesn't have `failure_reason` tag


Quinn-With-Two-Ns:
## Expected Behavior

`workflow_task_execution_failed` metric should always have `failure_reason` tag

## Actual Behavior

`workflow_task_execution_failed` metric sometimes doesn't have `failure_reason` tag

cretz:
What is the situation where this exception happens in that PR? Am just curious because I can't see where it reports a task failure in that catch clause.

PR #2419, opened 2025-02-18
Make sure workflow_task_execution_failed always has a failure_reason

Quinn-With-Two-Ns:
closes https://github.com/temporalio/sdk-java/issues/2418


cretz (approved):
This was a bit confusing to follow with `handle`, but since it's just adding a tag to existing metric, LGTM. I am now curious the situation this code is called in because I am afraid it actually won't return a task failure to the server.

Quinn-With-Two-Ns:
> I am now curious the situation this code is called in because I am afraid it actually won't return a task failure to the server.

Yes that is correct, we don't report a task to the server if we are retrying a task multiple times to control the task retry behaviour

PR #2422, opened 2025-02-20
Add sdk name/version to task completions when unset or changed

Sushisource:
## What was changed
Write name and version to task completions

## Why?
Parity 

## Checklist

1. Closes https://github.com/temporalio/sdk-java/issues/1838

2. How was this tested:
Added unit tests

3. Any docs updates needed?

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
@@ -251,12 +251,17 @@ public final class ReplayWorkflowTaskHandler implements WorkflowTaskHandler {
     String writeSdkName = result.getWriteSdkName();
     String writeSdkVersion = result.getWriteSdkVersion();
     if (!sdkFlags.isEmpty() || writeSdkName != null || writeSdkVersion != null) {
-      completedRequest.setSdkMetadata(
-          WorkflowTaskCompletedMetadata.newBuilder()
-              .addAllLangUsedFlags(sdkFlags)
-              .setSdkName(writeSdkName)
-              .setSdkVersion(writeSdkVersion)
-              .build());
+      WorkflowTaskCompletedMetadata.Builder md = WorkflowTaskCompletedMetadata.newBuilder();
+      if (!sdkFlags.isEmpty()) {
+        md.addAllLangUsedFlags(sdkFlags);
+      }
+      if (writeSdkName != null) {
+        md.setSdkName(writeSdkName);
+      }
+      if (writeSdkVersion != null) {
+        md.setSdkVersion(writeSdkVersion);
+      }
+      completedRequest.setSdkMetadata(md.build());
     }
     return new Result(
         workflowType,
--- a/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
@@ -247,12 +247,16 @@ public final class ReplayWorkflowTaskHandler implements WorkflowTaskHandler {
       }
       completedRequest.setStickyAttributes(attributes);
     }
-    if (!result.getSdkFlags().isEmpty()) {
-      completedRequest =
-          completedRequest.setSdkMetadata(
-              WorkflowTaskCompletedMetadata.newBuilder()
-                  .addAllLangUsedFlags(result.getSdkFlags())
-                  .build());
+    List<Integer> sdkFlags = result.getSdkFlags();
+    String writeSdkName = result.getWriteSdkName();
+    String writeSdkVersion = result.getWriteSdkVersion();
+    if (!sdkFlags.isEmpty() || writeSdkName != null || writeSdkVersion != null) {
+      completedRequest.setSdkMetadata(
+          WorkflowTaskCompletedMetadata.newBuilder()
+              .addAllLangUsedFlags(sdkFlags)
+              .setSdkName(writeSdkName)
+              .setSdkVersion(writeSdkVersion)
+              .build());
     }
     return new Result(
         workflowType,
--- a/temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
@@ -43,11 +43,13 @@ import io.temporal.internal.history.LocalActivityMarkerUtils;
 import io.temporal.internal.history.VersionMarkerUtils;
 import io.temporal.internal.sync.WorkflowThread;
 import io.temporal.internal.worker.LocalActivityResult;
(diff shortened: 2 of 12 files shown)


temporal-sdk/src/test/java/io/temporal/internal/replay/ReplayWorkflowRunTaskHandlerTaskHandlerTests.java
```diff
@@ -208,6 +210,143 @@ public void ifStickyExecutionAttributesAreSetThenWorkflowsAreCached() throws Thr
...
+    WorkflowTaskHandler.Result result = taskHandler.handleWorkflowTask(workflowTask);
+
+    assertTrue(result.isCompletionCommand());
+    assertEquals(Version.SDK_NAME, result.getTaskCompleted().getSdkMetadata().getSdkName());
+    assertEquals(
+        Version.LIBRARY_VERSION, result.getTaskCompleted().getSdkMetadata().getSdkVersion());
+  }
+
+  @Test
+  public void setsSdkNameAndVersionIfDifferentFromHistory() throws Throwable {
```
Sushisource, line 242: This test and the next I can't get to work because the setup in this suite isn't meant to work with more than one completion. I tried adjusting the mocks but it just led down dead end of the mock really not being suitable.

@Quinn-With-Two-Ns I can't find any tests that will allow me to really do what I want, which is just pass in some history and get a WFT response out that I can inspect, at this level. There are a bunch that operate at the workflow state machines level, but that's one level too deep. Any pointers would be welcome

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
```diff
@@ -247,13 +248,17 @@ private Result createCompletedWFTRequest(
...
-                  .build());
+      metadata.addAllLangUsedFlags(result.getSdkFlags());
     }
+    if (!Strings.isNullOrEmpty(result.getWriteSdkName())) {
+      metadata.setSdkName(result.getWriteSdkName());
+    }
+    if (!Strings.isNullOrEmpty(result.getWriteSdkVersion())) {
+      metadata.setSdkVersion(result.getWriteSdkVersion());
+    }
+    completedRequest.setSdkMetadata(metadata.build());
```
cretz, line 261: Will this set an empty workflow task completed metadata even if there are no SDK flags and no SDK name/version to set? If so I think this means it appears in every history JSON dump and technically takes a byte. I wonder if this whole chunk of code should be surrounded by `if (!result.getSdkFlags().isEmpty() || !Strings.isNullOrEmpty(result.getWriteSdkName()) || !Strings.isNullOrEmpty(result.getWriteSdkVersion())) {`.

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
```diff
@@ -384,6 +387,12 @@ private void handleSingleEventLookahead(HistoryEvent event) {
           }
           flags.setSdkFlag(sdkFlag);
         }
+        if (!Strings.isNullOrEmpty(completedEvent.getSdkMetadata().getSdkName())) {
```
cretz, line 390: Pardon my lack of Java proto knowledge, can you confirm (like some other lang proto impls) that a getter for a message returns an empty form of the message even if the underlying value is null?

temporal-sdk/src/test/java/io/temporal/internal/replay/ReplayWorkflowRunTaskHandlerTaskHandlerTests.java
Quinn-With-Two-Ns: The tests in `io/temporal/internal/statemachines` take in a history and you can inspect the result. All the tests there are focused on a particular state machine, but no reason you can't add tests that focus on the workflow task state machine.

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
Sushisource: Yes it does

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
Sushisource: Yeah, it will do that, I'll change.

temporal-sdk/src/test/java/io/temporal/internal/replay/ReplayWorkflowRunTaskHandlerTaskHandlerTests.java
Sushisource: Problem is this logic isn't in the workflow task state machine, it's in `ReplayWorkflowRunTaskHandler`, but, whatever. For the sake of making this easier to test with the existing stuff I'll just push it down and require one extra getter and just make sure it's set for completions with an integ test

Sushisource: Ok, this is all fixed. The one remaining test at that level ensures the name/ver returned by state machines is actually attached to the response

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
```diff
@@ -247,12 +247,16 @@ private Result createCompletedWFTRequest(
...
-                  .build());
+    List<Integer> sdkFlags = result.getSdkFlags();
+    String writeSdkName = result.getWriteSdkName();
+    String writeSdkVersion = result.getWriteSdkVersion();
+    if (!sdkFlags.isEmpty() || writeSdkName != null || writeSdkVersion != null) {
+      completedRequest.setSdkMetadata(
+          WorkflowTaskCompletedMetadata.newBuilder()
+              .addAllLangUsedFlags(sdkFlags)
+              .setSdkName(writeSdkName)
+              .setSdkVersion(writeSdkVersion)
```
cretz, line 258: I am not sure proto builder setters are null safe

Sushisource: God they aren't. That's absolutely silly.

Quinn-With-Two-Ns:
Can we add an integration test as well that tests this end to end , ideally the test workflow should contain multiple workflow tasks and evict the workflow from cache during the test

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
cretz: :+1: May be showing a test coverage hole too. Alongside Quinn's suggestion of integration test, can we make sure there is some kind of test that would have failed if this code had not changed?

Quinn-With-Two-Ns (approved):

PR #2424, opened 2025-02-20
Add support for attaching to a running workflow

Quinn-With-Two-Ns:
Add support for attaching to a running workflow.

See similar Go PR: https://github.com/temporalio/sdk-go/pull/1797

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/client/OnConflictOptions.java
+++ b/temporal-sdk/src/main/java/io/temporal/client/OnConflictOptions.java
@@ -20,6 +20,7 @@
 
 package io.temporal.client;
 
+import com.google.common.base.Preconditions;
 import io.temporal.common.Experimental;
 import java.util.Objects;
 
@@ -116,7 +117,10 @@ public class OnConflictOptions {
       return this;
     }
 
-    /** Attaches the completion callbacks to the running workflow. */
+    /**
+     * Attaches the completion callbacks to the running workflow. If true, AttachRequestId must be
+     * true.
+     */
     public Builder setAttachCompletionCallbacks(boolean attachCompletionCallbacks) {
       this.attachCompletionCallbacks = attachCompletionCallbacks;
       return this;
@@ -129,6 +133,10 @@ public class OnConflictOptions {
     }
 
     public OnConflictOptions build() {
+      if (attachCompletionCallbacks) {
+        Preconditions.checkState(
+            attachRequestId, "AttachRequestId must be true if AttachCompletionCallbacks is true");
+      }
       return new OnConflictOptions(attachRequestId, attachCompletionCallbacks, attachLinks);
     }
   }
--- a/temporal-sdk/src/main/java/io/temporal/client/OnConflictOptions.java
+++ b/temporal-sdk/src/main/java/io/temporal/client/OnConflictOptions.java
@@ -1,0 +1,135 @@
+/*
+ * Copyright (C) 2022 Temporal Technologies, Inc. All Rights Reserved.
+ *
+ * Copyright (C) 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Modifications copyright (C) 2017 Uber Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this material except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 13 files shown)


Quinn-With-Two-Ns:
Marking as draft while I add tests, will need https://github.com/temporalio/sdk-java/pull/2415 merged and probably a server release to merge

temporal-sdk/src/main/java/io/temporal/client/WorkflowOptions.java
```diff
@@ -460,6 +463,20 @@ public Builder setLinks(List<Link> links) {
       return this;
     }
 
+    /**
+     * Set workflow ID conflict options used in conjunction with conflict policy
+     * WORKFLOW_ID_CONFLICT_POLICY_USE_EXISTING. If onConflictOptions is set and a workflow is
+     * already running, the options specifies the actions to be taken on the running workflow. If
+     * not set or use together with any other WorkflowIDConflictPolicy, this parameter is ignored.
+     *
+     * <p>WARNING: Not intended for User Code.
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: Can't be private because it needs to be accessible by the `Internal` package. Passing them through a thread local is an option, but puts a requirement on the calling thread. I think we should revisit before we GA

temporal-sdk/src/main/java/io/temporal/internal/statemachines/NexusOperationStateMachine.java
```diff
@@ -186,46 +186,47 @@ private void cancelNexusOperationCommand() {
...
-        String operationId =
-            currentEvent.getNexusOperationStartedEventAttributes().getOperationId();
-        startedCallback.apply(
-            Optional.of(operationToken.isEmpty() ? operationId : operationToken), null);
-      }
-    }
+    async = true;
+    String operationToken =
+        currentEvent.getNexusOperationStartedEventAttributes().getOperationToken();
+    String operationId = currentEvent.getNexusOperationStartedEventAttributes().getOperationId();
```
bergundy, line 192: Wanna put a TODO here to remove this before GA?

temporal-sdk/src/main/java/io/temporal/internal/statemachines/NexusOperationStateMachine.java
```diff
@@ -186,46 +186,47 @@ private void cancelNexusOperationCommand() {
...
+        Optional.of(operationToken.isEmpty() ? operationId : operationToken), null);
   }
 
   private void notifyCompleted() {
-    notifyStarted();
     NexusOperationCompletedEventAttributes attributes =
         currentEvent.getNexusOperationCompletedEventAttributes();
+    if (!async) {
+      startedCallback.apply(Optional.empty(), null);
+    }
```
bergundy, line 202: I would just keep all of this logic in `notifyStarted()` but this is fine.

temporal-sdk/src/main/java/io/temporal/client/OnConflictOptions.java
```diff
@@ -0,0 +1,143 @@
...
+import com.google.common.base.Preconditions;
+import io.temporal.common.Experimental;
+import java.util.Objects;
+
+/**
+ * OnConflictOptions specifies the actions to be taken when using the {@link
+ * io.temporal.api.enums.v1.WorkflowIdConflictPolicy#WORKFLOW_ID_CONFLICT_POLICY_USE_EXISTING}
+ */
+@Experimental
+public class OnConflictOptions {
```
cretz, line 32: Hrmm, it was my understanding that we were going to hive this from users, though I know that is difficult in Java. If instead we're ok with showing to users but marking experimental, I think we should do the same in Go.

cretz (approved):
Nothing blocking

temporal-sdk/src/main/java/io/temporal/client/OnConflictOptions.java
Quinn-With-Two-Ns: We can't avoid exposing it given the structure of the Java SDK, just because the Java SDK has this unfortunate limitation doesn't mean other SDKs should copy it

cretz: So the question becomes should other SDKs follow Go's lead or Java's lead here. Sounds like we are saying Java will be the outlier?

Quinn-With-Two-Ns: Yes other SDK should not copy it

Issue #1456, opened 2022-09-25
Pollers should use just one thread to perform async gRPC requests


Spikhalskiy:
**Is your feature request related to a problem? Please describe.**
Currently, we allocate a lot of physical threads to perform blocking gRPC long poll requests. This doesn't make sense as it creates a lot of semi-active threads with a lot of context switching that just does a blocking io inside.

**Describe the solution you'd like**
We should have two threads for Pollers. One for Workflow Task Pollers, one for Activity Task Pollers. They should handle all corresponding long poll requests in an async manner.

PR #2433, opened 2025-02-28
:boom: [Breaking] Asyncify slot suppliers

Sushisource:
## What was changed
Made the `reserveSlot` method on `SlotSupplier` interface async.

Note that, as it stands, this can actually slightly _increase_ the number of threads used because the slot suppliers have been changed in a way that they will not block the caller (which is the whole point) but in order to do that, the resource based one at least needs a couple threads that it didn't before. 

For the fixed size supplier, I think I've managed to come up with something that should always be non-blocking.

## Why?
This is to support https://github.com/temporalio/sdk-java/issues/1456 where the pollers themselves will be made async, and thus will be able to take advantage of async slot reservation as well.

## Checklist

1. Closes 

2. How was this tested:
Existing tests

3. Any docs updates needed?

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/worker/ActivityPollTask.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/worker/ActivityPollTask.java
@@ -98,12 +98,18 @@ final class ActivityPollTask implements Poller.PollTask<ActivityTask> {
     PollActivityTaskQueueResponse response;
     SlotPermit permit;
     boolean isSuccessful = false;
-    CompletableFuture<SlotPermit> future =
-        slotSupplier.reserveSlot(
-            new SlotReservationData(
-                pollRequest.getTaskQueue().getName(),
-                pollRequest.getIdentity(),
-                pollRequest.getWorkerVersionCapabilities().getBuildId()));
+    CompletableFuture<SlotPermit> future;
+    try {
+      future =
+          slotSupplier.reserveSlot(
+              new SlotReservationData(
+                  pollRequest.getTaskQueue().getName(),
+                  pollRequest.getIdentity(),
+                  pollRequest.getWorkerVersionCapabilities().getBuildId()));
+    } catch (Exception e) {
+      log.warn("Error while trying to reserve a slot for an activity", e.getCause());
+      return null;
+    }
     try {
       permit = future.get();
     } catch (InterruptedException e) {
--- a/temporal-sdk/src/main/java/io/temporal/internal/worker/ActivityPollTask.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/worker/ActivityPollTask.java
@@ -35,8 +35,8 @@ import io.temporal.serviceclient.WorkflowServiceStubs;
 import io.temporal.worker.MetricsType;
 import io.temporal.worker.tuning.*;
 import java.util.Objects;
-import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
 import java.util.function.Supplier;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
@@ -98,7 +98,7 @@ final class ActivityPollTask implements Poller.PollTask<ActivityTask> {
     PollActivityTaskQueueResponse response;
     SlotPermit permit;
     boolean isSuccessful = false;
-    CompletableFuture<SlotPermit> future;
+    Future<SlotPermit> future;
     try {
       future =
           slotSupplier.reserveSlot(
--- a/temporal-sdk/src/main/java/io/temporal/internal/worker/ActivityPollTask.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/worker/ActivityPollTask.java
@@ -35,8 +35,6 @@ import io.temporal.serviceclient.WorkflowServiceStubs;
 import io.temporal.worker.MetricsType;
 import io.temporal.worker.tuning.*;
 import java.util.Objects;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.Future;
 import java.util.function.Supplier;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
@@ -97,8 +95,8 @@ final class ActivityPollTask implements Poller.PollTask<ActivityTask> {
     }
     PollActivityTaskQueueResponse response;
     SlotPermit permit;
+    SlotSupplierFuture future;
     boolean isSuccessful = false;
-    Future<SlotPermit> future;
     try {
       future =
           slotSupplier.reserveSlot(
@@ -110,16 +108,8 @@ final class ActivityPollTask implements Poller.PollTask<ActivityTask> {
       log.warn("Error while trying to reserve a slot for an activity", e.getCause());
       return null;
     }
-    try {
-      permit = future.get();
-    } catch (InterruptedException e) {
-      future.cancel(true);
-      Thread.currentThread().interrupt();
-      return null;
-    } catch (ExecutionException e) {
-      log.warn("Error while trying to reserve a slot for an activity", e.getCause());
-      return null;
-    }
+    permit = Poller.getSlotPermitAndHandleInterrupts(future, slotSupplier);
+    if (permit == null) return null;
 
     try {
       response =
--- a/temporal-sdk/src/main/java/io/temporal/internal/worker/ActivityPollTask.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/worker/ActivityPollTask.java
@@ -35,6 +35,8 @@ import io.temporal.serviceclient.WorkflowServiceStubs;
 import io.temporal.worker.MetricsType;
 import io.temporal.worker.tuning.*;
 import java.util.Objects;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
 import java.util.function.Supplier;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
@@ -96,18 +98,19 @@ final class ActivityPollTask implements Poller.PollTask<ActivityTask> {
     PollActivityTaskQueueResponse response;
     SlotPermit permit;
     boolean isSuccessful = false;
-
+    CompletableFuture<SlotPermit> future =
+        slotSupplier.reserveSlot(
+            new SlotReservationData(
+                pollRequest.getTaskQueue().getName(),
+                pollRequest.getIdentity(),
+                pollRequest.getWorkerVersionCapabilities().getBuildId()));
     try {
-      permit =
-          slotSupplier.reserveSlot(
-              new SlotReservationData(
-                  pollRequest.getTaskQueue().getName(),
-                  pollRequest.getIdentity(),
-                  pollRequest.getWorkerVersionCapabilities().getBuildId()));
+      permit = future.get();
     } catch (InterruptedException e) {
+      future.cancel(true);
       Thread.currentThread().interrupt();
       return null;
-    } catch (Exception e) {
+    } catch (ExecutionException e) {
       log.warn("Error while trying to reserve a slot for an activity", e.getCause());
       return null;
     }
--- a/temporal-sdk/src/main/java/io/temporal/internal/worker/TrackingSlotSupplier.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/worker/TrackingSlotSupplier.java
@@ -26,6 +26,7 @@ import io.temporal.worker.tuning.*;
 import java.util.Collections;
 import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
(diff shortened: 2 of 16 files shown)


temporal-sdk/src/main/java/io/temporal/worker/tuning/SlotSupplier.java
```diff
@@ -36,16 +37,19 @@
...
   /**
-   * This function is called before polling for new tasks. Your implementation should block until a
-   * slot is available then return a permit to use that slot.
+   * This function is called before polling for new tasks. Your implementation should return a
+   * Promise that is completed with a {@link SlotPermit} when one becomes available.
+   *
+   * <p>These futures may be cancelled if the worker is shutting down or otherwise abandons the
+   * reservation. This can cause an {@link InterruptedException} to be thrown, in the thread running
+   * your implementation. You may want to catch it to perform any necessary cleanup, and then you
+   * should rethrow the exception.
```
cretz, line 46: > and then you should rethrow the exception

They can't rethrow without wrapping because the checked exception was removed from this interface. We should add back `throws InterruptedException`, though could consider `throws Exception` as a way of saying we have code to handle any exception.

temporal-sdk/src/main/java/io/temporal/worker/tuning/ResourceBasedSlotSupplier.java
```diff
@@ -32,6 +33,18 @@ public class ResourceBasedSlotSupplier<SI extends SlotInfo> implements SlotSuppl
   private final ResourceBasedController resourceController;
   private final ResourceBasedSlotOptions options;
   private Instant lastSlotIssuedAt = Instant.EPOCH;
+  // For slot reservations that are waiting to re-check resource usage
+  private static final ScheduledExecutorService scheduler =
```
cretz, line 37: I think you should consider accepting this in the constructor. Can have a default, but we should allow people to control thread creation if they want.

temporal-sdk/src/main/java/io/temporal/worker/tuning/ResourceBasedSlotSupplier.java
```diff
@@ -139,29 +152,43 @@ private ResourceBasedSlotSupplier(
...
+    return permitFuture.thenCompose(
+        ignored -> {
+          Optional<SlotPermit> permit = tryReserveSlot(ctx);
+          // If we couldn't get a slot this time, delay for a short period and try again
+          return permit
+              .map(CompletableFuture::completedFuture)
+              .orElseGet(
+                  () ->
+                      CompletableFuture.supplyAsync(() -> null, delayedExecutor(10))
+                          .thenCompose(ig -> scheduleSlotAcquisition(ctx)));
```
cretz, line 190: ChatGPT is telling me that cancellation does not propagate across `orElseGet` or `thenCompose` operators, but that doesn't seem right and I can't find this in the docs at https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletionStage.html. May need a test somehow confirming cancel _after_ this `orElseGet` does actually prevent the delayed call from being invoked. If it doesn't work, I can make some suggestions I think.

Sushisource: I think this works fine because the returned future is `permitFuture` which is `thenCompose`d with the future from here. Added a test that confirms.

temporal-sdk/src/main/java/io/temporal/worker/tuning/SlotSupplier.java
Sushisource: Added Exception

temporal-sdk/src/main/java/io/temporal/worker/tuning/ResourceBasedSlotSupplier.java
cretz: My concern is that ChatGPT says that a cancel of the outer future does not propagate the cancel to the `thenCompose`d future. Not sure I believe it because I haven't tested, but it does concern me. I would have to test Java behavior of cancel when the `thenCompose` has already run to create a new future. I may set aside some time to do this.

But I fear reading https://stackoverflow.com/questions/25417881/canceling-a-completablefuture-chain and poking around, there may need to be some other mechanism to make sure the outer completable future cancel propagates to cancelling the delayed executor.

temporal-sdk/src/main/java/io/temporal/worker/tuning/SlotSupplier.java
```diff
@@ -36,16 +37,19 @@
...
    * @param ctx The context for slot reservation.
-   * @return A permit to use the slot which may be populated with your own data.
-   * @throws InterruptedException The worker may choose to interrupt the thread in order to cancel
-   *     the reservation, or during shutdown. You may perform cleanup, and then should rethrow the
-   *     exception.
+   * @return A future that will be completed with a permit to use the slot when one becomes
+   *     available. Never return null, or complete the future with null.
    */
-  SlotPermit reserveSlot(SlotReserveContext<SI> ctx) throws InterruptedException;
+  CompletableFuture<SlotPermit> reserveSlot(SlotReserveContext<SI> ctx) throws Exception;
```
cretz, line 52:
```suggestion
  Future<SlotPermit> reserveSlot(SlotReserveContext<SI> ctx) throws Exception;
```

I wonder if this is more flexible. That some implementations happen to use `CompletableFuture` doesn't affect us right?

Sushisource: I feel like it's less flexible. You can't compose raw Futures or really do much of anything with them besides `get` them

temporal-sdk/src/main/java/io/temporal/worker/tuning/SlotSupplier.java
```diff
@@ -36,16 +37,19 @@
 @Experimental
 public interface SlotSupplier<SI extends SlotInfo> {
   /**
-   * This function is called before polling for new tasks. Your implementation should block until a
-   * slot is available then return a permit to use that slot.
+   * This function is called before polling for new tasks. Your implementation should return a
+   * Promise that is completed with a {@link SlotPermit} when one becomes available.
+   *
+   * <p>These futures may be cancelled if the worker is shutting down or otherwise abandons the
```
Quinn-With-Two-Ns, line 43: >These futures may be cancelled if the worker is shutting down or otherwise abandons the

You are aware `CompletableFuture` cancellation does not propagate upstream correct?

temporal-sdk/src/main/java/io/temporal/worker/tuning/ResourceBasedSlotSupplier.java
Quinn-With-Two-Ns: https://stackoverflow.com/questions/23320407/how-to-cancel-java-8-completable-future

temporal-sdk/src/main/java/io/temporal/internal/worker/ActivityPollTask.java
```diff
@@ -96,18 +98,19 @@ public ActivityTask poll() {
     PollActivityTaskQueueResponse response;
     SlotPermit permit;
     boolean isSuccessful = false;
-
+    CompletableFuture<SlotPermit> future =
```
Quinn-With-Two-Ns, line 101: `reserveSlot` is a user defined function so I would except any use case to be under a `try` block so exceptions don't leek out and crash the worker or cause undesirable behaviour

temporal-sdk/src/main/java/io/temporal/worker/tuning/ResourceBasedSlotSupplier.java
cretz: :+1: So I do not know of what a good way in Java to tell the implementer of something async that it is cancelled? I am guessing we use thread interruption for synchronous activity code today and disallow async/future-returning activity code? Any ideas here?

Quinn-With-Two-Ns: So the standard library would use `Future` for this like here https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ScheduledFuture.html. The other option is to return a special interface like `SlotSupplierFuture` interface that implements the  `CompletableFuture ` interface and another different "cancel" method that actually does what you want

Sushisource: I've dealt with this by returning future and using an anonymous class in the one spot (two counting test code) I actually need another stage. It works fine since the "callback" (incrementing the counter in the tracking slot supplier) only ever needs to be called if `get` is actually invoked, and it always is anyway.

Sushisource: I think the existing docstring is reasonably clear on that front. I could add a warning about using CompletableFuture specifically, and if they do that they maybe want to be aware of how that can go wrong (which wasn't obvious to me at least so maybe that's a good idea) but I suppose otherwise it's like any other Future returning Java API

cretz: > I suppose otherwise it's like any other Future returning Java API

Yup but still a bit confusing because most don't manually implement cancel. I propose we alter the javadoc to change from:

> These futures may be cancelled if the worker is shutting down or otherwise abandons the reservation. This can cause an {@link InterruptedException} to be thrown, in the thread running your implementation. You may want to catch it to perform any necessary cleanup, and then you should rethrow the exception.

To:

> These futures may get `cancel` invoked on them if the worker is shutting down or otherwise abandons the reservation. Users need to make sure they handle `cancel` on the resulting future. By default in Java, futures created by users or composed from other futures will not properly propagate `cancel` so users need to make sure this is handled properly.

temporal-sdk/src/main/java/io/temporal/internal/worker/TrackingSlotSupplier.java
```diff
@@ -48,10 +49,52 @@ public TrackingSlotSupplier(SlotSupplier<SI> inner, Scope metricsScope) {
     publishSlotsMetric();
   }
 
-  public SlotPermit reserveSlot(SlotReservationData dat) throws InterruptedException {
-    SlotPermit p = inner.reserveSlot(createCtx(dat));
-    issuedSlots.incrementAndGet();
-    return p;
+  public Future<SlotPermit> reserveSlot(SlotReservationData dat) {
```
Quinn-With-Two-Ns, line 52:
```suggestion
  public Future<SlotPermit> reserveSlot(SlotReservationData data) {
```

temporal-sdk/src/main/java/io/temporal/internal/worker/TrackingSlotSupplier.java
```diff
@@ -48,10 +49,52 @@ public TrackingSlotSupplier(SlotSupplier<SI> inner, Scope metricsScope) {
...
+      @Override
+      public SlotPermit get(long timeout, TimeUnit unit)
+          throws InterruptedException, ExecutionException, TimeoutException {
+        SlotPermit permit = originalFuture.get(timeout, unit);
+        return executeCallbackIfNeeded(permit);
+      }
+    };
   }
 
   public Optional<SlotPermit> tryReserveSlot(SlotReservationData dat) {
```
Quinn-With-Two-Ns, line 100:
```suggestion
  public Optional<SlotPermit> tryReserveSlot(SlotReservationData data) {
```

Sushisource:
Looks like there can be some reserve/release mismatches. Investigating that.

Sushisource:
Actually I've just realized this ends up being a huge mess because anything like `thenApply` you call on this type returns a `CompleteableFuture` instead of a `SlotSupplierFuture`, and there's, AFAICT, no good way to fix that without manually overriding every single method which seems awful.

Quinn-With-Two-Ns:
>thenApply you call on this type returns a CompleteableFuture instead of a SlotSupplierFuture

Why is that an issue?

Sushisource:
> > thenApply you call on this type returns a CompleteableFuture instead of a SlotSupplierFuture
> 
> Why is that an issue?

I suppose it's not really

temporal-sdk/src/main/java/io/temporal/worker/tuning/SlotSupplier.java
```diff
@@ -36,16 +36,19 @@
 @Experimental
 public interface SlotSupplier<SI extends SlotInfo> {
   /**
-   * This function is called before polling for new tasks. Your implementation should block until a
-   * slot is available then return a permit to use that slot.
+   * This function is called before polling for new tasks. Your implementation should return a
+   * Promise that is completed with a {@link SlotPermit} when one becomes available.
```
Quinn-With-Two-Ns, line 40:
```suggestion
   * Future that is completed with a {@link SlotPermit} when one becomes available.
```

Quinn-With-Two-Ns (approved):

Issue #2432, opened 2025-02-27
Expose static summary/details on workflow execution description


cretz:
**Describe the solution you'd like**

As part of user metadata, we should have `getStaticSummary()` and `getStaticDetails()` on the result of `describe` which is `WorkflowExecutionDescription`. Note these are not in the base class `WorkflowExecutionMetadata` which also comes back from `listExecutions` because it's not in visibility.

Issue #2430, opened 2025-02-25
Expose root workflow execution


cretz:
**Describe the solution you'd like**

See https://github.com/temporalio/features/issues/605

Issue #2219, opened 2024-09-15
Test Server: Add support for user metadata


Quinn-With-Two-Ns:
Add support for https://github.com/temporalio/api/commit/de9e0055a5067ee3e3a6105a207b7c78e8ea8668 on workflow start and commands.

PR #2441, opened 2025-03-11
Add support for metadata to test server

Quinn-With-Two-Ns:
This PR does a few things:
* Add support for metadata to test server
* Add summary and detail to describe
* expose root execution

closes: 
* https://github.com/temporalio/sdk-java/issues/2219
* https://github.com/temporalio/sdk-java/issues/2432
* https://github.com/temporalio/sdk-java/issues/2430


temporal-sdk/src/main/java/io/temporal/workflow/WorkflowInfo.java
```diff
@@ -136,6 +136,18 @@ public interface WorkflowInfo {
    */
   Optional<String> getParentRunId();
 
+  /**
+   * @return Workflow ID of the root Workflow
+   */
+  @Nullable
+  String getRootWorkflowId();
```
Quinn-With-Two-Ns, line 143: This is a bit awkward since this can only be null on old servers that didn't have this value.

cretz: Makes sense, may be worth documenting that here in the javadoc. I do appreciate that this is not using `Optional`.

temporal-sdk/src/main/java/io/temporal/client/WorkflowExecutionDescription.java
```diff
@@ -21,19 +21,60 @@
...
       @Nonnull DescribeWorkflowExecutionResponse response, @Nonnull DataConverter dataConverter) {
     super(response.getWorkflowExecutionInfo(), dataConverter);
+    this.dataConverter = dataConverter;
     this.response = response;
   }
 
+  /** Get the fixed summary for this workflow execution. */
+  @Experimental
+  @Nullable
+  public String getStaticSummary() {
```
cretz, line 45: Might consider memoizing this (e.g. backed by an `AtomicReference`). If not, at least worth clarifying in the javadoc each invocation does data conversion.

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
cretz, line 1: I wonder if it's worth a more generic implementation of command-to-event user metadata propagation like server instead of the only known current use cases. This would make it more future proof. I understand it'd change the code a bit since there is no generic place where an event is created from a command nor is there a common place to affect an event after it is created. Up to you.

cretz (approved):
Minor things, nothing blocking

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
Quinn-With-Two-Ns: Yeah possibly, child workflow would still need some special logic, but yeah this should be generic

PR #2444, opened 2025-03-13
Add Summary to Nexus Operations

Quinn-With-Two-Ns:
Add Summary to Nexus Operations

closes https://github.com/temporalio/sdk-java/issues/2442


cretz (approved):
I think we should consider updating https://github.com/temporalio/api/blob/9156239c8272003b6f228951602ca2b130a55876/temporal/api/history/v1/message.proto#L965-L973 set of "well known uses" to include activity and nexus operation events. We could discuss removing it, but I think it's nice to UI that they know what's expected.

PR #2446, opened 2025-03-14
Fix workflow ID reuse policy and conflict policy handling

rodrigozhou:
## What was changed

Fix logic handling workflow ID reuse policy and conflict policy.
Refactor the code, make it cleaner.

## Why?

Reflect recent changes in Temporal Server in https://github.com/temporalio/temporal/pull/7097 and https://github.com/temporalio/temporal/pull/7099.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


Quinn-With-Two-Ns (approved):
Is there any tests for the test server we should add to cover this?

Quinn-With-Two-Ns (approved):

Issue #1407, opened 2022-09-06
ARM64 build for Test Server


dominik-pakosz:
## Expected Behavior
Test server running

## Actual Behavior
Exception: `qemu-x86_64: Could not open '/lib64/ld-linux-x86-64.so.2': No such file or directory` thrown

## Steps to Reproduce the Problem

  1. download latest release binary
  2. run ./temporal-test-server
  3. no profit - seeing error

## Specifications

  - Version: latest - [v1.16.0](https://github.com/temporalio/sdk-java/releases/tag/v1.16.0)
  - Platform: MacOS ARM M1 Pro

cv65kr:
@Spikhalskiy thanks for explanation. I think is quite important topic, a lot companies started exchanging old intel macbooks with new one with arm. I am using Temporal and for me is a big problem because now I need to use CI which tooks ages and don't have a control.

Proper testing is basic of good application.

bergundy:
@cv65kr which download did you use?
Try this one: https://github.com/temporalio/sdk-java/releases/download/v1.16.0/temporal-test-server_1.16.0_macOS_amd64.tar.gz.
It's for Mac OS and should work with no additional steps, IIRC, with Rosetta.
Which SDK are you using? The TypeScript and Python SDKs both automatically install the right executable for you.

cv65kr:
The newest one for Temporal test server. PHP SDK - https://github.com/temporalio/sdk-php/blob/master/testing/src/Downloader.php#L29 looks like is not working out of box

bergundy:
Hmm.. the code you sent LGTM at first glance, can you check which file it downloads?

Spikhalskiy:
@cv65kr Could you download, unpack and try to run the specific artifact @bergundy pointed?
https://github.com/temporalio/sdk-java/releases/download/v1.16.0/temporal-test-server_1.16.0_macOS_amd64.tar.gz
This one is expected to work on M1. If it doesn't, let us know.

dominik-pakosz:
@Spikhalskiy I tried this artifact.
Result:
```
❯ docker-compose exec container-apache /bin/sh
# ./temporal-test-server
/bin/sh: 2: ./temporal-test-server: Exec format error
# exit
```

```
❯ docker-compose exec container-apache /bin/bash
root@a479b2380cc8:/src# ./temporal-test-server
bash: ./temporal-test-server: cannot execute binary file: Exec format error
root@a479b2380cc8:/src# exit
exit
```

[...]

Issue #2402, opened 2025-02-05
Support musl-based build of test server


cretz:
**Describe the solution you'd like**

See https://github.com/temporalio/features/issues/594.

Upon Java SDK release, support a musl-based binary in the assets with an obvious name that can be used by `temporal.download`.

PR #2448, opened 2025-03-17
ARM64 build for Test Server

Quinn-With-Two-Ns:
Create an `ARM64` build for time skipping test server. Move us off old palantir graal plugin to the official graal gradle [plugin](https://graalvm.github.io/native-build-tools/latest/gradle-plugin.html)

closes https://github.com/temporalio/sdk-java/issues/1407


.github/workflows/prepare-release.yml
```diff
@@ -164,24 +170,22 @@ jobs:
...
       # path ends in a wildcard because on windows the file ends in '.exe'
-      # path excludes *.txt because native-image also writes a build manifest txt file
       - name: Upload executable to workflow
         uses: actions/upload-artifact@v4
         with:
           name: ${{ matrix.os_family }}_${{ matrix.arch }}
           path: |
-            temporal-test-server/build/graal/temporal-test-server*
-            !temporal-test-server/build/graal/*.txt
+            temporal-test-server/build/native/nativeCompile/temporal-test-server*
```
cretz, line 188: Can I get confirmation that we believe the release artifact files will have the same name, just with arm ones now? So `temporal-test-server_<version>_linux_arm64.tar.gz` and `temporal-test-server_<version>_maxOS_arm64.tar.gz` will now be present?

temporal-test-server/build.gradle
```diff
@@ -99,28 +92,43 @@ idea {
...
+        binaries {
+            main {
+                mainClass = application.getMainClass().get()
+                sharedLibrary = false
+                // Signal handling so that ^C actually stops the process
+                buildArgs.add("--install-exit-handlers")
+                // If we're on linux, static link everything but libc. Otherwise link
+                // everything dynamically (note the '-' rather than '+' in front of
+                // StaticExecutable)
+                buildArgs.add(isLinux() ? "-H:+StaticExecutableWithDynamicLibC": "-H:-StaticExecutable")
```
cretz, line 123: I suspect this may change a tad when we do #2402, specifically that I guess we'll add `--libc=musl` instead of dynamic libc.

temporal-test-server/build.gradle
```diff
@@ -99,28 +92,43 @@ idea {
...
+                buildArgs.add("-H:+UnlockExperimentalVMOptions")
+                buildArgs.add("-O4")
+
+                runtimeArgs.add("7233")
```
cretz, line 127: Can you help me understand this a bit? Does it mean that this is the default arg if not provided? https://graalvm.github.io/native-build-tools/latest/gradle-plugin.html is not very clear to me.

.github/workflows/prepare-release.yml
Quinn-With-Two-Ns: Yep

temporal-test-server/build.gradle
Quinn-With-Two-Ns: It sets the arg for `./gradlew -Pagent -PnativeBuild :temporal-test-server:nativeRun` since there isn't another way I could find

PR #2453, opened 2025-03-19
Priorities for Workflows/Activities

Sushisource:
## What was changed
Added priority annotations to starting workflows & activities

## Why?
New feature!

## Checklist

1. Closes https://github.com/temporalio/sdk-java/issues/2399

2. How was this tested:
> [!NOTE]
> No tests yet. Ideally will write integ tests if we make a CLI RC 

3. Any docs updates needed?
Will need docs for Priority generally

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/activity/ActivityInfo.java
+++ b/temporal-sdk/src/main/java/io/temporal/activity/ActivityInfo.java
@@ -142,7 +142,13 @@ public interface ActivityInfo {
   /** Used to determine if the Activity Execution is a local Activity. */
   boolean isLocal();
 
-  /** Return the priority of the activity task. */
+  /**
+   * Return the priority of the activity task.
+   *
+   * @apiNote If unset or on an older server version, this method will return {@link
+   *     Priority#getDefaultInstance()}.
+   */
   @Experimental
+  @Nonnull
   Priority getPriority();
 }
--- a/temporal-sdk/src/main/java/io/temporal/activity/ActivityInfo.java
+++ b/temporal-sdk/src/main/java/io/temporal/activity/ActivityInfo.java
@@ -21,6 +21,8 @@
 package io.temporal.activity;
 
 import io.temporal.api.common.v1.Payloads;
+import io.temporal.common.Experimental;
+import io.temporal.common.Priority;
 import java.time.Duration;
 import java.util.Optional;
 import javax.annotation.Nonnull;
@@ -139,4 +141,8 @@ public interface ActivityInfo {
 
   /** Used to determine if the Activity Execution is a local Activity. */
   boolean isLocal();
+
+  /** Return the priority of the activity task. */
+  @Experimental
+  Priority getPriority();
 }
--- a/temporal-sdk/src/main/java/io/temporal/common/Priority.java
+++ b/temporal-sdk/src/main/java/io/temporal/common/Priority.java
@@ -1,0 +1,79 @@
+/*
+ * Copyright (C) 2022 Temporal Technologies, Inc. All Rights Reserved.
+ *
+ * Copyright (C) 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Modifications copyright (C) 2017 Uber Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this material except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.temporal.common;
+
+import java.util.Objects;
+
+/**
+ * Priority contains metadata that controls the relative ordering of task processing when tasks are
+ * backed up in a queue. The affected queues depend on the server version.
+ *
+ * <p>Priority is attached to workflows and activities. By default, activities and child workflows
+ * inherit Priority from the workflow that created them, but may override fields when an activity is
+ * started or modified.
+ *
+ * <p>For all fields, the field not present or equal to zero/empty string means to inherit the value
+ * from the calling workflow, or if there is no calling workflow, then use the default value.
+ */
+@Experimental
+public class Priority {
+  private int priorityKey;
+
+  public Priority(int priorityKey) {
+    this.priorityKey = priorityKey;
+  }
+
+  /**
+   * See {@link #setPriorityKey(int)}
+   *
+   * @return The priority key
+   */
+  public int getPriorityKey() {
+    return priorityKey;
+  }
+
+  /**
+   * A priority key is a positive integer from 1 to n, where smaller integers correspond to higher
+   * priorities (tasks run sooner). In general, tasks in a queue should be processed in close to
+   * priority order, although small deviations are possible.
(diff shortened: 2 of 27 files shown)


temporal-sdk/src/main/java/io/temporal/common/Priority.java
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * <p>Priority is attached to workflows and activities. By default, activities and child workflows
+ * inherit Priority from the workflow that created them, but may override fields when an activity is
+ * started or modified.
+ *
+ * <p>For all fields, the field not present or equal to zero/empty string means to inherit the value
+ * from the calling workflow, or if there is no calling workflow, then use the default value.
+ */
+@Experimental
+public class Priority {
```
Quinn-With-Two-Ns, line 37: Not a fan of one parameter builders, but since we know this field will expend we should make this a builder.

Sushisource:
> Do we need test server support? and if so what level of support?

I don't know, that's a question for the broader team I think. I would kind of hope not, since priority management seems like something you ought not to be dependent on for functional testing.

temporal-sdk/src/main/java/io/temporal/common/Priority.java
```diff
@@ -0,0 +1,79 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.temporal.common;
+
+import java.util.Objects;
+
+/**
+ * Priority contains metadata that controls the relative ordering of task processing when tasks are
```
dnr, line 26: I updated the text in the api repo slightly, not terribly important but you might want to grab the latest version just for consistency

temporal-sdk/src/main/java/io/temporal/common/Priority.java
cretz: Agreed, this object should be immutable. This likely includes making the constructor private, but we can have a static helper that creates it from just the integer.

cretz, line 37: Let's toss a basic `toString()` on here too

temporal-sdk/src/main/java/io/temporal/internal/common/PriorityUtils.java
```diff
@@ -0,0 +1,33 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.temporal.internal.common;
+
+import io.temporal.api.common.v1.Priority;
+
+public class PriorityUtils {
```
cretz, line 25: Pedantic - add a private constructor here if it's just static utilities

temporal-sdk/src/main/java/io/temporal/activity/ActivityInfo.java
```diff
@@ -139,4 +141,8 @@ public interface ActivityInfo {
 
   /** Used to determine if the Activity Execution is a local Activity. */
   boolean isLocal();
+
+  /** Return the priority of the activity task. */
+  @Experimental
+  Priority getPriority();
```
cretz, line 147: What situations is this nullable? Old servers only or always unless explicitly set? (just curious, though if confusing we should consider documenting the cases where it may be)

Also a bit confused on this class how a user can know whether a return is nullable or not. But given the confusion, best to be safe and make it clear in docs. I would consider `Optional<Priority>` if it can be nullable even on newer servers, but meh. Note this same concern does apply as much to user-set values like the activity options.

cretz (commented):
LGTM and is nice and simple. I understand the server implementation is fairly new, but would be nice to see what an end-to-end test may look like so that other langs can add it too. I think a simple test is:

With worker off, start 10 workflows with priority being maybe workflow 5 first, then 7, then 3, then the rest. Then start worker with only a single workflow task slot (so cache disabled) and confirm that 5, 7, and 3 are first in that order. It's a simple test that will give SDKs peace of mind in server presence/impl.

Also, we should discuss whether we should open an issue for test server. Obviously if the logic is complicated we will just not implement and not take priority into account, but if it's as simple as changing the `TaskQueue`s `LinkedList` to be a `SortedSet`, we should consider it.

temporal-sdk/src/main/java/io/temporal/activity/ActivityInfo.java
Quinn-With-Two-Ns: If it can be nullable we should bias towards tagging with `@Nullable` vs optional, it is the preferred approach in Java

dnr:
> would be nice to see what an end-to-end test may look like so that other langs can add it too

See https://github.com/temporalio/temporal/blob/24cc4dd82/tests/priority_fairness_test.go for an sdk-based test. That uses activities but it would work with workflows too. The catch is that order is only approximate due to partitions, so use a larger number and check that ordering is within a threshold. Or force a single partition.

temporal-sdk/src/main/java/io/temporal/activity/ActivityInfo.java
Sushisource: It can be nullable insofar as you might be using an old server. So, yeah I'll attach the annotation. I believe new servers will always attach the (default) value.

Quinn-With-Two-Ns: Would be worth documenting that and verifying the test server does match the real server here

temporal-sdk/src/main/java/io/temporal/common/Priority.java
```diff
@@ -0,0 +1,114 @@
...
+ * <p>Priority is attached to workflows and activities. By default, activities and child workflows
+ * inherit Priority from the workflow that created them, but may override fields when an activity is
+ * started or modified.
+ *
+ * <p>For all fields, the field not present or equal to zero/empty string means to inherit the value
+ * from the calling workflow, or if there is no calling workflow, then use the default value.
+ */
+@Experimental
+public class Priority {
+  public static Priority.Builder newBuilder() {
```
cretz, line 38: Wonder if we should have a `create` static helper that accepts the int and returns a `Priority` instance to make it easy for users

cretz (approved):
Nothing blocking. Only marking approved now because I see the priority test which is great in that it fails currently when running against real server so I can have confidence this won't be merged until it is in a real server (i.e. a user can actually use it).

temporal-sdk/src/main/java/io/temporal/common/Priority.java
Quinn-With-Two-Ns: Converted this to a builder to be consistent with the rest of the SDK

temporal-sdk/src/main/java/io/temporal/activity/ActivityInfo.java
Quinn-With-Two-Ns: So server leaves this as `null` if unset which is the equivalent of default. In Go, if the server left it as null, we return the default object so I did the same here

temporal-sdk/src/main/java/io/temporal/common/Priority.java
Quinn-With-Two-Ns: We don't have helpers for any other builders, I don't think priority is special

temporal-sdk/src/main/java/io/temporal/activity/ActivityInfo.java
Sushisource: I think we need to not do that as it will lie. The default values for priorty are not 0, and therefore we might lie and seem like they've been _set to zero_ for a server that actually just doesn't understand priority

Quinn-With-Two-Ns: Zero means default, so zero is not a lie

PR #2457, opened 2025-03-22
Add Workflow Info on activity cancellation and failure logs

kevinawoo:
## What was changed
Add the WorkflowId on activity cancelled and failure logs to make it easier to find the effected workflow.

## Why?
Users can quickly identify which workflow an activity failure belongs to.

This also is inline with [Go's version](https://github.com/temporalio/sdk-go/blob/master/internal/internal_task_handlers.go#L2286-L2288) and other SDKs.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/activity/ActivityTaskExecutors.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/activity/ActivityTaskExecutors.java
@@ -117,15 +117,23 @@ final class ActivityTaskExecutors {
         boolean local = info.isLocal();
         if (ex instanceof ActivityCanceledException) {
           log.info(
-              "{} canceled. ActivityId={}, activityType={}, attempt={}",
+              "{} canceled. Namespace={}, TaskQueue={}, WorkflowId={}, RunId={}, ActivityId={}, activityType={}, attempt={}",
               local ? "Local activity" : "Activity",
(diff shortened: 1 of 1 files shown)


temporal-sdk/src/main/java/io/temporal/internal/activity/ActivityTaskExecutors.java
```diff
@@ -117,15 +117,23 @@ public ActivityTaskHandler.Result execute(ActivityInfoInternal info, Scope metri
         boolean local = info.isLocal();
         if (ex instanceof ActivityCanceledException) {
           log.info(
-              "{} canceled. ActivityId={}, activityType={}, attempt={}",
+              "{} canceled. Namespace={}, TaskQueue={}, WorkflowId={}, RunId={}, ActivityId={}, activityType={}, attempt={}",
```
Quinn-With-Two-Ns, line 120: All of this Data is already on the MDC context so it can be included in any log from the activity woker

Quinn-With-Two-Ns: https://github.com/temporalio/sdk-java/blob/6c4c1835f261e3a6539292809976fdaea9d3fbaa/temporal-sdk/src/main/java/io/temporal/internal/worker/ActivityWorker.java#L245

Issue #2459, opened 2025-03-26
Test server can overwrite last heartbeat details when not on fail request


cretz:
## Expected Behavior

Test server should not overwrite last heartbeat details on activity failure unless it is actually set

## Actual Behavior

As of #2354, the last heartbeat details are always overridden regardless of whether set which, while that's ok for Java, is not for Core-based SDKs (at least until https://github.com/temporalio/sdk-core/issues/293 is done)

PR #2460, opened 2025-03-26
Ensure heartbeat details aren't cleared

Quinn-With-Two-Ns:
closes https://github.com/temporalio/sdk-java/issues/2459

### DIFF
--- a/temporal-test-server/src/test/java/io/temporal/testserver/functional/activity/ActivityHeartbeat.java
+++ b/temporal-test-server/src/test/java/io/temporal/testserver/functional/activity/ActivityHeartbeat.java
@@ -1,0 +1,109 @@
+/*
+ * Copyright (C) 2022 Temporal Technologies, Inc. All Rights Reserved.
+ *
+ * Copyright (C) 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Modifications copyright (C) 2017 Uber Technologies, Inc.
+ *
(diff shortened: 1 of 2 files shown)


temporal-test-server/src/test/java/io/temporal/testserver/functional/activity/ActivityHeartbeat.java
```diff
@@ -0,0 +1,109 @@
...
+    // clear the heartbeat details.
+    String result =
+        testWorkflowRule.newWorkflowStub(TestWorkflows.WorkflowReturnsString.class).execute();
+    assertEquals("", result);
+    assertEquals(2, activityHeartbeats.size());
+    assertFalse(activityHeartbeats.poll().isPresent());
+    assertEquals(
+        "heartbeat details",
+        DefaultDataConverter.STANDARD_INSTANCE.fromPayloads(
+            0, activityHeartbeats.poll(), String.class, String.class));
```
cretz, line 68: I assume this is the assertion that would have failed without the change because it'd still have 2 in the queue, but even the second would not be present due to overwrite (just kinda proving I read it and understand it heh)

cretz (approved):

temporal-test-server/src/test/java/io/temporal/testserver/functional/activity/ActivityHeartbeat.java
Quinn-With-Two-Ns: Yes this fails without the test server change

Issue #1692, opened 2023-03-08
Implement workflow_start_delay in test server


bergundy:
See API PR here: https://github.com/temporalio/api/pull/264

PR #2462, opened 2025-03-31
Add support for start delay to the time skipping test server

Quinn-With-Two-Ns:
Add support for start delay to the time skipping test server.

closes https://github.com/temporalio/sdk-java/issues/1692

### DIFF
--- a/temporal-sdk/src/test/java/io/temporal/client/functional/StartDelayTest.java
+++ b/temporal-sdk/src/test/java/io/temporal/client/functional/StartDelayTest.java
@@ -51,7 +51,7 @@ public class StartDelayTest {
   public void startWithDelay() {
     WorkflowOptions workflowOptions =
         SDKTestOptions.newWorkflowOptionsWithTimeouts(testWorkflowRule.getTaskQueue()).toBuilder()
-            .setStartDelay(Duration.ofSeconds(5))
+            .setStartDelay(Duration.ofSeconds(1))
             .build();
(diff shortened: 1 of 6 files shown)


temporal-sdk/src/test/java/io/temporal/client/functional/StartDelayTest.java
```diff
@@ -0,0 +1,93 @@
...
+            .build();
+    TestNoArgsWorkflowFunc stubF =
+        testWorkflowRule
+            .getWorkflowClient()
+            .newWorkflowStub(TestNoArgsWorkflowFunc.class, workflowOptions);
+    long start = testWorkflowRule.getTestEnvironment().currentTimeMillis();
+    stubF.func();
+    long end = testWorkflowRule.getTestEnvironment().currentTimeMillis();
+    // Assert that the workflow took at least 5 seconds to start
+    assertEquals(5000, end - start, 1000);
```
cretz, line 64: Can you confirm that if time skipping is not enabled/uinlocked, this waits the actual system amount of time?

cretz (approved):

temporal-sdk/src/test/java/io/temporal/client/functional/StartDelayTest.java
Quinn-With-Two-Ns: Yeah, that was what this test was supposed to do, but I shouldn't have used the test environments time

PR #2463, opened 2025-03-31
Worker Versioning Annotations & Options

Sushisource:
## What was changed
Added worker options for the latest iteration of the worker versioning feature, as well as annotations for workflow methods to specify their behavior.

## Why?
New feature!

## Checklist

1. Closes https://github.com/temporalio/sdk-java/issues/2458
2. How was this tested:
Added tests

3. Any docs updates needed?


temporal-sdk/src/main/java/io/temporal/internal/worker/WorkerVersioningOptions.java
```diff
@@ -0,0 +1,86 @@
...
+        || (workerDeploymentOptions != null && workerDeploymentOptions.isUsingVersioning());
+  }
+
+  @Nullable
+  public WorkerDeploymentOptions getWorkerDeploymentOptions() {
+    return workerDeploymentOptions;
+  }
+
+  public static io.temporal.api.deployment.v1.WorkerDeploymentOptions deploymentOptionsToProto(
+      @Nonnull WorkerDeploymentOptions options) {
```
Sushisource, line 64: It felt a little silly to have yet-another proto utils class for these, but can move if desired

temporal-sdk/src/main/java/io/temporal/worker/WorkerOptions.java
```diff
@@ -385,24 +387,25 @@ public Builder setDisableEagerExecution(boolean disableEagerExecution) {
 
     /**
      * Opts the worker in to the Build-ID-based versioning feature. This ensures that the worker
-     * will only receive tasks which it is compatible with. For more information see: TODO: Doc link
+     * will only receive tasks which it is compatible with.
      *
      * <p>Defaults to false
      */
     @Experimental
+    @Deprecated // TODO REVIEW: Are we actually deprecating these yet?
```
Sushisource, line 395: Not entirely sure if we are deprecating these or not

Quinn-With-Two-Ns:
Please make sure you also test Spring boot

temporal-testing/src/main/java/io/temporal/testing/TestWorkflowRule.java
```diff
@@ -401,6 +403,17 @@ public Builder setMetricsScope(Scope metricsScope) {
       return this;
     }
 
+    /**
+     * Enables worker versioning, and uses a unique deployment name based on the test.
+     *
+     * @param buildId Build Id to use for the rule's built-in worker.
+     * @return {@code this}
+     */
+    public Builder setEnableWorkerDeployment(String buildId) {
```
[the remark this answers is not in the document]
Sushisource: The existing tests I have now show how you can have more than one and why it's useful

Sushisource:
> Please make sure you also test Spring boot

Do you mean configuring the worker options via config files?

temporal-testing/src/main/java/io/temporal/testing/TestWorkflowRule.java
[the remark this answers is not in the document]
Quinn-With-Two-Ns: They are useful for us testing the implementation, but are they useful for a user?

temporal-sdk/src/test/java/io/temporal/worker/WorkerVersioningTest.java
```diff
@@ -0,0 +1,452 @@
...
+      WorkerDeploymentVersion v) {
+    DescribeWorkerDeploymentRequest req =
+        DescribeWorkerDeploymentRequest.newBuilder()
+            .setNamespace(testWorkflowRule.getTestEnvironment().getNamespace())
+            .setDeploymentName(v.getDeploymentName())
+            .build();
+    return Eventually.assertEventually(
+        Duration.ofSeconds(15),
+        () -> {
+          DescribeWorkerDeploymentResponse resp =
```
Quinn-With-Two-Ns, line 397: Don't we need a high level deployment client? Go has that.

Sushisource: No, we're not including that as part of the initial functionality

temporal-testing/src/main/java/io/temporal/testing/TestWorkflowRule.java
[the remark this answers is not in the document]
Sushisource: Maybe not. The underlying problem is that if you want to re-use something that uniquely identifies this worker (like the postfix to the task queue) and use it in worker options, you can't because you don't have access to it until after the worker is built.

So, maybe I can just instead expose a `setTaskQueuePostfix` or something, and that's generally useful.

temporal-sdk/src/main/java/io/temporal/common/WorkerDeploymentVersion.java
```diff
@@ -0,0 +1,86 @@
...
+    this.deploymentName = deploymentName;
+    this.buildId = buildId;
+  }
+
+  /**
+   * @return The canonical string representation of this worker deployment version.
+   */
+  public String toCanonicalString() {
+    return deploymentName + "." + buildId;
+  }
```
cretz, line 44: For this specific class, I think just having this be the `toString` instead of the generated one below would be fine

temporal-sdk/src/main/java/io/temporal/common/WorkerDeploymentVersion.java
cretz, line 1: Is there an issue tracking the client-side changes? Specifically, version info for list, describe, and scheduling, and the new versioning calls on the client. (asking at top of this file because this stuff is easier in a thread than PR comment and there's no better place)

temporal-sdk/src/main/java/io/temporal/worker/WorkerOptions.java
cretz: I think we can, though I think we should commit to deprecating all past versioning things I think

temporal-sdk/src/main/java/io/temporal/common/WorkerDeploymentVersion.java
Quinn-With-Two-Ns: I prefer the current approach since the canonical format is different then the standard `toString` implementation

temporal-sdk/src/main/java/io/temporal/workflow/WorkflowVersioningBehavior.java
```diff
@@ -0,0 +1,42 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Indicates the versioning behavior of this workflow. May only be applied to workflow
+ * implementations, not interfaces.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.METHOD)
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: Being on the workflow method is correct. Workflow implementations can contain multiple workflow methods

temporal-sdk/src/main/java/io/temporal/common/WorkerDeploymentVersion.java
cretz: Meh, not sure there is such thing as a "standard" `toString` implementation, especially for a basic dot-delimited tuple like this. Same as there isn't a standard builder expectation for such a simple class. Having said that, I don't mind if we keep them separate.

temporal-sdk/src/main/java/io/temporal/workflow/WorkflowVersioningBehavior.java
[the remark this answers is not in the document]
Quinn-With-Two-Ns: The Method is the workflow, a class can contain multiple unrelated workflows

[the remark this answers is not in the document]
cretz: The method is only part of the workflow, there is lots of code that makes up a workflow. We can only hope that when they use versioning and incompatibly change a signal handler, they remember to update the versioning enum on all primary workflow methods and not just one.

temporal-sdk/src/main/java/io/temporal/worker/WorkerOptions.java
```diff
@@ -494,6 +497,16 @@ public Builder setIdentity(String identity) {
       return this;
     }
 
+    /**
+     * Set deployment options for the worker. Exclusive with {@link #setUseBuildIdForVersioning} and
```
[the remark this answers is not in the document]
Sushisource: No, thanks.

temporal-sdk/src/main/java/io/temporal/common/WorkerDeploymentVersion.java
Sushisource: Yeah, I looked into this when I wrote it and it seems the consensus is that toString is more for debugging that canonical representations

temporal-sdk/src/main/java/io/temporal/common/VersioningBehavior.java
```diff
@@ -0,0 +1,41 @@
...
+@Experimental
+public enum VersioningBehavior {
+  /**
+   * An unspecified versioning behavior. By default, workers opting into worker versioning will be
+   * required to specify a behavior. See {@link
+   * io.temporal.worker.WorkerOptions.Builder#setDeploymentOptions(WorkerDeploymentOptions)}.
+   */
+  VERSIONING_BEHAVIOR_UNSPECIFIED,
+  /** The workflow will be pinned to the current Build ID unless manually moved. */
+  VERSIONING_BEHAVIOR_PINNED,
```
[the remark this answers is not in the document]
Sushisource: Yeah. I only did it because some of the other ones were lol. I guess this is more necessary in Go

temporal-sdk/src/main/java/io/temporal/common/VersioningBehavior.java
```diff
@@ -0,0 +1,41 @@
...
+
+/** Specifies when a workflow might move from a worker of one Build Id to another. */
+@Experimental
+public enum VersioningBehavior {
+  /**
+   * An unspecified versioning behavior. By default, workers opting into worker versioning will be
+   * required to specify a behavior. See {@link
+   * io.temporal.worker.WorkerOptions.Builder#setDeploymentOptions(WorkerDeploymentOptions)}.
+   */
+  VERSIONING_BEHAVIOR_UNSPECIFIED,
```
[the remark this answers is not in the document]
Sushisource: Yeah, that's the case. (with the minor exception of the one workflow method metadata thing where it makes sense that it should be able to represent both, because it needs to distinguish un-annotated).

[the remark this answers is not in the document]
cretz: :+1: So long as we never give this value back to the user as null and always give it back as unspecified, works for me

temporal-sdk/src/main/java/io/temporal/common/WorkerDeploymentVersion.java
Sushisource: https://github.com/temporalio/sdk-java/issues/2469

cretz (approved):

temporal-spring-boot-autoconfigure/src/test/java/io/temporal/spring/boot/autoconfigure/workerversioning/TestWorkflowImpl.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+package io.temporal.spring.boot.autoconfigure.workerversioning;
+
+import io.temporal.common.VersioningBehavior;
+import io.temporal.spring.boot.WorkflowImpl;
+import io.temporal.workflow.WorkflowVersioningBehavior;
+import org.springframework.context.ConfigurableApplicationContext;
+
+@WorkflowImpl(workers = "mainWorker")
```
Quinn-With-Two-Ns, line 28: Damn Springboot taking this great annotation name for itself hehe

temporal-sdk/src/main/java/io/temporal/common/metadata/POJOWorkflowImplMetadata.java
```diff
@@ -238,4 +243,35 @@ public List<POJOWorkflowMethodMetadata> getUpdateValidatorMethods() {
...
+
+  /**
+   * @return The {@link VersioningBehavior} for the workflow method on the implementation class. If
+   *     the method is annotated with {@link WorkflowVersioningBehavior}.
+   * @throws RuntimeException if the method is not found on the implementation class or is not a
+   *     workflow method.
+   */
+  @Experimental
+  @Nullable
+  public static VersioningBehavior getVersioningBehaviorForMethod(
```
Quinn-With-Two-Ns, line 255: We may want to make this a bit more generic if we plan to add other annotations to workflow impl. methods. Since this is experimental I am fine leaving it as is.

Quinn-With-Two-Ns (approved):

PR #2464, opened 2025-04-01
Issue 587

Quinn-With-Two-Ns:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


temporal-sdk/src/main/java/io/temporal/internal/history/VersionMarkerUtils.java
```diff
@@ -75,17 +93,46 @@ public static Integer getVersion(MarkerRecordedEventAttributes markerAttributes)
...
+      String changeId, Integer version, Boolean upsertVersionSA) {
     Preconditions.checkNotNull(version, "version");
     Map<String, Payloads> details = new HashMap<>();
     details.put(
         MARKER_CHANGE_ID_KEY, DefaultDataConverter.STANDARD_INSTANCE.toPayloads(changeId).get());
     details.put(
         MARKER_VERSION_KEY, DefaultDataConverter.STANDARD_INSTANCE.toPayloads(version).get());
+    details.put(
+        UPSERT_VERSION_SA_KEY,
+        DefaultDataConverter.STANDARD_INSTANCE.toPayloads(upsertVersionSA).get());
```
cretz, line 113: Can you help me understand the changing of the marker value here. Do other SDKs do this and/or are we using it as a flag of sorts?

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
```diff
@@ -508,6 +534,14 @@ private void handleCommandEvent(HistoryEvent event) {
     if (handleLocalActivityMarker(event)) {
       return;
     }
+    if (shouldSkipUpsertVersionSA) {
+      if (handleNonMatchingUpsertSearchAttribute(event)) {
+        shouldSkipUpsertVersionSA = false;
```
cretz, line 539: Is this bool only valid for this one event? Does setting it back to false here make a difference, or is it just to be safe?

temporal-sdk/src/main/java/io/temporal/internal/history/VersionMarkerUtils.java
Quinn-With-Two-Ns: Yes Go does the same thing, it tells the SDK if it should expect a search attribute after the marker.

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
Quinn-With-Two-Ns: Yes it is only valid for one event

temporal-sdk/src/main/java/io/temporal/worker/WorkflowImplementationOptions.java
```diff
@@ -174,6 +178,26 @@ public Builder setDefaultNexusServiceOptions(NexusServiceOptions defaultNexusSer
...
+     * the SDK to automatically add the <b>TemporalChangeVersion</b> search attributes to the
+     * workflow when getVersion is called. This search attribute is a keyword list of all the
+     * getVersion calls made in the workflow. The format of each entry is "<Change ID>-<Version>".
+     * This allows for easy discovery of what versions are being used in your namespace.
+     *
+     * <p>Note: This change is backwards compatible, so it is safe to enable or disable this options
+     * with running workflows. However, if this options is enabled, it is not safe to rollback to a
+     * previous version of the SDK that does not support this option.
+     *
+     * <p>The default value is false. This may change in future releases.
```
cretz, line 192: Wonder if we need an issue to track enabling this by default after a release or two

cretz (approved):
I admit I didn't do a deep dive, but the test coverage does provide some comfort here couple with the fact that it's not even enabled by default today. May be worth getting review from others if needing multiple opinions/eyes.

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowContextImpl.java
```diff
@@ -356,6 +361,18 @@ public long currentTimeMillis() {
...
+     * Temporal Change Version is a reserved field and should ideally not be set by the user.
+     * It is set by the SDK when getVersion is called. We know that users have been setting
+     * this field in the past, and we want to avoid breaking their workflows.
+     * */
+    if (searchAttributes.containsIndexedFields(TEMPORAL_CHANGE_VERSION.getName())) {
+      // When we enabled upserting of the search attribute by default, we should consider raising a
+      // warning here.
+      log.debug(
+          "{} is a reserved field. This can be set automatically by the SDK by calling `setEnableUpsertVersionSearchAttributes` on you `WorkflowImplementationOptions`",
+          TEMPORAL_CHANGE_VERSION.getName());
```
Sushisource, line 374:
```suggestion
          "{} is a reserved field. This can be set automatically by the SDK by calling `setEnableUpsertVersionSearchAttributes` on your `WorkflowImplementationOptions`",
          TEMPORAL_CHANGE_VERSION.getName());
```

Sushisource (approved):
I believe this does indeed match Core

PR #2476, opened 2025-04-15
Add support for activity pause

Quinn-With-Two-Ns:
closes https://github.com/temporalio/sdk-java/issues/2465

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/client/ActivityPausedException.java
+++ b/temporal-sdk/src/main/java/io/temporal/client/ActivityPausedException.java
@@ -1,3 +1,23 @@
+/*
+ * Copyright (C) 2022 Temporal Technologies, Inc. All Rights Reserved.
+ *
+ * Copyright (C) 2012-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Modifications copyright (C) 2017 Uber Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this material except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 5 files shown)


temporal-sdk/src/main/java/io/temporal/client/ActivityPausedException.java
```diff
@@ -0,0 +1,38 @@
...
+package io.temporal.client;
+
+import io.temporal.activity.ActivityInfo;
+
+/***
+ * Indicates that the activity was paused by the user.
+ *
+ * <p>Catching this exception directly is discouraged and catching the parent class {@link ActivityCompletionException} is recommended instead.<br>
+ */
+public final class ActivityPausedException extends ActivityCompletionException {
```
cretz, line 30: Hrmm, so I see this is in the `client` package. I am guessing some other extensions of this activity completion exception are in `client` package instead of the `activity` one because they can be the cause of the outer activity failure when caught inside a workflow, correct? My concern with this exception is that it is only for inside activities and never for inside workflows.

Not that it's wrong or anything, but I wonder if it's worth noting in javadoc that even if, say, max attempts were reached as a result of this being bubbled out of the activity, the workflow caller will never see this exception.

cretz (approved):
Nothing blockling

temporal-sdk/src/main/java/io/temporal/client/ActivityPausedException.java
Quinn-With-Two-Ns: All other `ActivityCompletionException` are in `client` , I agree it is not the ideal package but it is consistent with the rest of the SDK.

> the workflow caller will never see this exception.

That is just the basic failure conversion rules for activities https://docs.temporal.io/references/failures#errors-in-activities I don't think we should document it for every SDK exception.

PR #2485, opened 2025-04-22
Add application err category

### LINKED ISSUE #2475: Apply application failure logging and metrics behaviour according to `ApplicationErrorCategory`
See corresponding features issue:
https://github.com/temporalio/features/issues/614

THardy98:
## What was changed
Add support for application error category.

## Why?
SDK work for benign exceptions

1. Closes #2475

2. How was this tested:
Integration tests

3. Any docs updates needed?
Likely


THardy98:
Don't like the API changes to ApplicationFailure, but figured I'd get something up first for discussion

temporal-sdk/src/main/java/io/temporal/failure/ApplicationErrorCategory.java
cretz, line 1: It seems in other parts of the SDK, e.g. parent close policy, we just expose the raw enum from proto. Should we do the same here? It has "unrecognized", automatically gets new values, etc.

temporal-sdk/src/main/java/io/temporal/failure/ApplicationErrorCategory.java
```diff
@@ -0,0 +1,62 @@
...
+      return UNSPECIFIED;
+    }
+    switch (protoCategory) {
+      case APPLICATION_ERROR_CATEGORY_BENIGN:
+        return BENIGN;
+      case APPLICATION_ERROR_CATEGORY_UNSPECIFIED:
+      case UNRECOGNIZED:
+      default:
+        // Fallback unrecognized or unspecified proto values as UNSPECIFIED
+        return UNSPECIFIED;
```
cretz, line 47: I think the concept of unrecognized is ideal and different from unspecified (then again though, arguably we should be using raw API proto enums and not a new enumerate here).

temporal-sdk/src/main/java/io/temporal/failure/ApplicationFailure.java
```diff
@@ -218,4 +274,43 @@ private static String getMessage(String message, String type, boolean nonRetryab
         + ", nonRetryable="
         + nonRetryable;
   }
+
+  public static boolean isBenignApplicationFailure(@Nullable Throwable t) {
```
cretz, line 278: Easier to justify than the Go equivalent I suppose, but I still think this helper may not be needed and may be more confusing than it's worth

temporal-sdk/src/main/java/io/temporal/failure/ApplicationFailure.java
```diff
@@ -218,4 +274,43 @@ private static String getMessage(String message, String type, boolean nonRetryab
...
+      if (failure.hasApplicationFailureInfo()
+          && failure.getApplicationFailureInfo().getCategory()
+              == io.temporal.api.enums.v1.ApplicationErrorCategory
+                  .APPLICATION_ERROR_CATEGORY_BENIGN) {
+        return true;
+      }
+    }
+
+    // Handle ActivityFailure, which wraps the actual ApplicationFailure
+    if (t instanceof io.temporal.failure.ActivityFailure) {
```
cretz, line 301: I don't think we want to expose this kind of logic to users. A user checking whether an in-workflow-thrown application failure is benign may not want to have it match when an activity raises it. I think if a user wants to check a category for whatever reason, they can.

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowExecutor.java
```diff
@@ -153,7 +154,9 @@ private void completeWorkflow(@Nullable WorkflowExecutionException failure) {
       metricsScope.counter(MetricsType.WORKFLOW_CANCELED_COUNTER).inc(1);
     } else if (failure != null) {
       workflowStateMachines.failWorkflow(failure.getFailure());
-      metricsScope.counter(MetricsType.WORKFLOW_FAILED_COUNTER).inc(1);
+      if (!ApplicationFailure.isBenignApplicationFailure(failure)) {
```
cretz, line 157: I do not think you want to swallow activities that fail because of benign exceptions. They may reach a max attempts or something and throw this out. Just because an activity has a benign exception it doesn't want to be in telemetry doesn't mean a workflow doesn't want it in telemetry.

THardy98: Is this not the same pattern in:
https://github.com/temporalio/sdk-go/pull/1925
https://github.com/temporalio/sdk-core/pull/905
unless I'm mistaken.

What about if a user throws a benign exception in workflow code, not in an activity?

cretz: Sorry, I was unclear here. This is basically me restating https://github.com/temporalio/sdk-java/pull/2485/files#r2054385164.

> What about if a user throws a benign exception in workflow code, not in an activity?

That is fine to skip metrics, but we don't want to skip metrics if the exception _in the workflow code_ is an activity failure with a benign cause. There is a difference between not recording telemetry on benign exceptions and not recording telemetry on an exception that has a benign cause.

Here are the scenarios:

* Activity throws benign exception - do not record telemetry
* Workflow throws benign exception - do not record telemetry
* Activity or child fails with benign exception from workflow POV (thrown from activity and it is out of max attempts or something) - still record telemetry

Basically, change `isBenignApplicationFailure` to stop checking causes IMO. IMO throwing a benign exception is benign, but receiving a benign exception wrapped in another exception is not benign.

> Is this not the same pattern in:
> https://github.com/temporalio/sdk-go/pull/1925
> https://github.com/temporalio/sdk-core/pull/905
> Unless I'm mistaken.

Not exactly because Go stops at the first application error, not every application error no matter the depth. Also, this isn't recursively checking causes no matter how deep. Still, looking back on Go, we should only apply it to the current error and not check cause or `errors.As`. Didn't check Go, but I think the same applies there too.

If you think about how users use benign exceptions (as control flow), they aren't supposed to wrap them.

temporal-sdk/src/main/java/io/temporal/failure/ApplicationErrorCategory.java
THardy98: Didn't notice that, have changed to use raw proto

temporal-sdk/src/main/java/io/temporal/failure/ApplicationErrorCategory.java
THardy98: using raw API proto enums now

temporal-sdk/src/main/java/io/temporal/failure/ApplicationFailure.java
THardy98: Moved to an internal failure utils file, it's convenient to have imo

temporal-sdk/src/main/java/io/temporal/failure/ApplicationFailure.java
THardy98: Agree, moved to an internal util file

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowExecutor.java
THardy98: Won't they inherently be wrapped as other failure types though? Could be wrong here, but for example, an ApplicationFailure thrown from an activity, would it not always (or at least in the general case) be wrapped as an ActivityFailure?

THardy98: I agree in that it doesn't seem relevant/useful to check for benign causes at an arbitrary depth, but I though a depth of 0 (i.e. immediate) or 1 would be sufficient.

In Java SDK, how would one receive an ApplicationFailure that is **not** wrapped?

cretz: > Won't they inherently be wrapped as other failure types though?

No. They are only benign where they are thrown (impl side) and they are usually only wrapped where they are caught (caller side).

> Could be wrong here, but for example, an ApplicationFailure thrown from an activity, would it not always (or at least in the general case) be wrapped as an ActivityFailure?

Not on the activity side, only on the workflow caller side. But if an activity fails because of a benign error, even though it is benign on the activity side, if that activity failure bubbles out of a workflow, that is _not_ benign.

> In Java SDK, how would one receive an ApplicationFailure that is not wrapped?

What do you mean by "one"? If you mean how does our SDK internals receive an unwrapped thrown error, it is where you have changes in `ActivityTaskExecutors.java`. If you mean how does a workflow caller receive it unwrapped, they don't, that's the point.

An error is only benign where it is thrown impl side, not where it is caught caller side.

THardy98: Thanks for the clarification, that all sounds reasonable (and saved me some digging :))
I've removed the nested failure checks, only check immediate failures.

temporal-sdk/src/main/java/io/temporal/internal/common/FailureUtils.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+package io.temporal.internal.common;
+
+import io.temporal.api.enums.v1.ApplicationErrorCategory;
+import io.temporal.api.failure.v1.Failure;
+import io.temporal.failure.ApplicationFailure;
+import javax.annotation.Nullable;
+
+public class FailureUtils {
```
cretz, line 28: Pedantic, but would add a private constructor here

cretz (approved):
Looks great! Will want to wait for @Quinn-With-Two-Ns's approval here.

temporal-sdk/src/main/java/io/temporal/failure/ApplicationFailure.java
```diff
@@ -51,13 +52,15 @@
  *   <li>nonRetryable is set to false
  *   <li>details are set to null
  *   <li>stack trace is copied from the original exception
+ *   <li>stack category is set to ApplicationErrorCategory.APPLICATION_ERROR_CATEGORY_UNSPECIFIED
```
Quinn-With-Two-Ns, line 55:
```suggestion
 *   <li>category is set to ApplicationErrorCategory.APPLICATION_ERROR_CATEGORY_UNSPECIFIED
```

temporal-sdk/src/main/java/io/temporal/failure/ApplicationFailure.java
```diff
@@ -21,6 +21,7 @@
 package io.temporal.failure;
 
 import com.google.common.base.Strings;
+import io.temporal.api.enums.v1.ApplicationErrorCategory;
```
Quinn-With-Two-Ns, line 24: We shouldn't use the proto enum in the user facing API, we should declare our own enum in the Java SDK and convert.

Quinn-With-Two-Ns: sorry I didn't notice before you were still using the proto enum

THardy98: I formerly had this but: https://github.com/temporalio/sdk-java/pull/2485#discussion_r2054380000

It looked like we use raw protos, when do we use each pattern?

Quinn-With-Two-Ns: All new features should NOT use raw proto, only old features use raw protos

temp.txt
```diff
@@ -0,0 +1,483 @@
+diff --git a/temporal-sdk/src/main/java/io/temporal/failure/ApplicationErrorCategory.java b/temporal-sdk/src/main/java/io/temporal/failure/ApplicationErrorCategory.java
```
Quinn-With-Two-Ns, line 1: Please remove

temporal-sdk/src/main/java/io/temporal/failure/ApplicationErrorCategory.java
```diff
@@ -0,0 +1,34 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.temporal.failure;
+
+/**
+ * Mirrors the proto definition for ApplicationErrorCategory. Used to categorize application
```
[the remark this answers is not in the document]
THardy98: done

temp.txt
THardy98: whoops - rm'd

Quinn-With-Two-Ns (approved):

PR #2487, opened 2025-04-23
Handle async completion in TestActivityEnvironment

Quinn-With-Two-Ns:
Handle async completion in TestActivityEnvironment


temporal-sdk/src/test/java/io/temporal/internal/testing/ActivityTestingTest.java
```diff
@@ -111,6 +109,21 @@ public void testFailure() {
     }
   }
 
+  private static class AsyncActivityImpl implements TestActivity {
+    @Override
+    public String activity1(String input) {
+      Activity.getExecutionContext().doNotCompleteOnReturn();
```
cretz, line 115: When running in a non-test-environment, does this throw? If not, not sure we should in test environment either. If it does, can we throw the same thing in the test environment?

Quinn-With-Two-Ns: `doNotCompleteOnReturn ` behaves the same in the test environment and non-test-environment

cretz (approved):

PR #2490, opened 2025-04-23
Add native build to CI

Quinn-With-Two-Ns:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


.github/workflows/ci.yml
```diff
@@ -188,3 +188,9 @@ jobs:
     
     - name: Run copyright and code format checks
       run: ./gradlew --no-daemon checkLicenseMain checkLicenses spotlessCheck
+  
+  build_native_images:
```
cretz, line 192: Is it by intention that we are building native images for every CI run now or is this leftover from testing? If it is by intention, I would usually be concerned with unnecessarily using storage for all of these images, though I see we only retain for one day, but I wonder if it's even worth that vs just a workflow call option to not upload the artifacts.

.github/workflows/build-native-image.yml
cretz, line 1: Is it just me, or are the test server artifacts on the CI build much larger than they once were? For instance, Windows x64 on this PR's CI run is 62.4MB (exe inside about the same size), but the one at https://github.com/temporalio/sdk-java/releases is 18.4MB (exe inside about the same size). I wonder if it's due to Java upgrade or if we're inadvertently not trimming something anymore or what.

.github/workflows/ci.yml
Quinn-With-Two-Ns: Yes, I do want to build as part of CI so we don't discover issue when trying to cut a release. I would be OK only doing the builds post merge or not including artifacts for CI

.github/workflows/build-native-image.yml
Quinn-With-Two-Ns: hm good callout, not sure let me check

.github/workflows/ci.yml
cretz: Works for me, especially if it's concurrent and doesn't take longer than the existing jobs (doesn't seem to though that poor macos-arm one in last CI run took a really long time). I do think we should skip uploading the artifact though, but not a big deal.

.github/workflows/build-native-image.yml
Quinn-With-Two-Ns: OK I think I addressed the size by adjusting the optimization level and reverting the extra reflection stuff I was including by mistake

Quinn-With-Two-Ns: Once you double check the size I will remove the artifacts from the CI build

cretz: It's still slightly bigger. For instance, Linux amd64 in this CI is 22.5MB vs 18.8MB in the last release. I'm ok with that if you are.

Switching topics (sorry, just easy to keep the same thread), I kinda figured the musl-based binary would be bigger than the libc-based one since it should statically link musl. Can you confirm that, unlike glibc build, we statically link musl? (meaning it could technically run on any Linux not just musl-based distros)

Quinn-With-Two-Ns: musl is statically linked

cretz: And I assume you're ok with the ~12% size-after-extracted increase? Checked extracted sizes, for Linux amd64 on release it's 60.3MB but here on CI it's 67.8MB. I don't think I have a big problem with this, though is a bit strange I think to increase sizes for every binary just to add a new musl binary (though I assume/understand it's mostly due to our switching of how we build w/ graal).

PR #2492, opened 2025-04-28
Add RawValue support

Quinn-With-Two-Ns:
Add RawValue support

closes https://github.com/temporalio/sdk-java/issues/2425

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/common/converter/RawValue.java
+++ b/temporal-sdk/src/main/java/io/temporal/common/converter/RawValue.java
@@ -39,4 +39,22 @@ public final class RawValue {
   public Payload getPayload() {
     return payload;
   }
+
+  @Override
+  public boolean equals(Object o) {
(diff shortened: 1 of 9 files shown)


temporal-sdk/src/main/java/io/temporal/common/converter/RawValue.java
```diff
@@ -0,0 +1,42 @@
...
+import io.temporal.api.common.v1.Payload;
+import java.util.Objects;
+
+/**
+ * RawValue is a representation of an unconverted, raw payload.
+ *
+ * <p>This type can be used as a parameter or return type in workflows and activities to pass
+ * through a raw payload. Encoding/decoding of the payload is still done by the system.
+ */
+public final class RawValue {
```
cretz, line 32: May be worth the good 'ol hash code, equals, and toString

cretz (approved):

PR #2495, opened 2025-04-30
Add ApplicationFailure.Builder

Quinn-With-Two-Ns:
Add `ApplicationFailure.Builder` to handle the growing number of options on `ApplicationFailure`

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/failure/ApplicationFailure.java
+++ b/temporal-sdk/src/main/java/io/temporal/failure/ApplicationFailure.java
@@ -234,7 +234,7 @@ public final class ApplicationFailure extends TemporalFailure {
     this.nextRetryDelay = nextRetryDelay;
   }
 
-  public ApplicationErrorCategory getApplicationErrorCategory() {
+  public ApplicationErrorCategory getCategory() {
     return category;
   }
 
--- a/temporal-sdk/src/main/java/io/temporal/failure/ApplicationFailure.java
+++ b/temporal-sdk/src/main/java/io/temporal/failure/ApplicationFailure.java
@@ -51,7 +51,7 @@ import javax.annotation.Nullable;
  *   <li>nonRetryable is set to false
  *   <li>details are set to null
  *   <li>stack trace is copied from the original exception
- *   <li>category is set to ApplicationErrorCategory.APPLICATION_ERROR_CATEGORY_UNSPECIFIED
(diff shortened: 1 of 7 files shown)


temporal-sdk/src/main/java/io/temporal/failure/ApplicationFailure.java
```diff
@@ -274,4 +246,122 @@ private static String getMessage(String message, String type, boolean nonRetryab
...
+      this.nextRetryDelay = nextRetryDelay;
+      return this;
+    }
+
+    /**
+     * Set the optional category of the failure.
+     *
+     * <p>Default is {@link ApplicationErrorCategory#UNSPECIFIED}.
+     */
+    public Builder setCategory(ApplicationErrorCategory category) {
```
cretz, line 351: I think we should change the getter on `ApplicationFailure` to `getCategory`

cretz (approved):
Nothing blocking

temporal-sdk/src/main/java/io/temporal/failure/ApplicationFailure.java
Quinn-With-Two-Ns: I would also be fine changing the setter instead, any preference?

temporal-sdk/src/main/java/io/temporal/failure/ApplicationFailure.java
```diff
@@ -178,45 +188,7 @@ public static ApplicationFailure newNonRetryableFailureWithCause(
...
-   * @param message optional error message
-   * @param type error type
-   * @param category the category of the application failure.
-   * @param cause failure cause. Each element of the cause chain will be converted to
-   *     ApplicationFailure for network transmission across network if it doesn't extend {@link
-   *     TemporalFailure}
-   * @param details optional details about the failure. They are serialized using the same approach
-   *     as arguments and results.
-   */
-  public static ApplicationFailure newFailureWithCategory(
```
Sushisource, line 197: I thought this would be a breaking change but this isn't actually released, for any other readers.

temporal-sdk/src/main/java/io/temporal/failure/ApplicationFailure.java
Sushisource: +1 to the shorter version

Sushisource (approved):

Issue #2494, opened 2025-04-29
Test server not showing user metadata for workflow start in history


cretz:
## Expected Behavior

Workflow start with summary/details on test server should show this in user metadata of `EVENT_TYPE_WORKFLOW_EXECUTION_STARTED` event but it is not.

## Actual Behavior

User metadata not visible. Testing on latest release of Java SDK which should have #2441. May need to also check that child workflow propagates the metadata too.

First noticed in https://github.com/temporalio/sdk-python/issues/851

PR #2496, opened 2025-04-30
Make sure user metadata is set on WorkflowExecutionStartedEvent

Quinn-With-Two-Ns:
Make sure user metadata is set on WorkflowExecutionStartedEvent

closes https://github.com/temporalio/sdk-java/issues/2494


Sushisource (approved):
Absolute title gore :rofl:

Issue #1634, opened 2023-01-31
Add support for interceptor beans to Spring Boot


Spikhalskiy:
**Is your feature request related to a problem? Please describe.**

Users should be able to register interceptor beans in the Spring Boot context and these interceptors should be used for Workers and Client configuration.

PR #2500, opened 2025-05-01
Add interceptor support to springboot integration

Quinn-With-Two-Ns:
Add interceptor support to Springboot integration. With this PR any interceptor bean will be automatically picked up and registered with the worker/client. Similar to Quarkus support for [interceptors](https://docs.quarkiverse.io/quarkus-temporal/dev/index.html#_interceptors) in Temporal.

Note: Currently not supported in multi namespace.

closes: https://github.com/temporalio/sdk-java/issues/1634

### DIFF
--- a/temporal-spring-boot-autoconfigure/src/main/java/io/temporal/spring/boot/autoconfigure/template/WorkerFactoryOptionsTemplate.java
+++ b/temporal-spring-boot-autoconfigure/src/main/java/io/temporal/spring/boot/autoconfigure/template/WorkerFactoryOptionsTemplate.java
@@ -1,25 +1,32 @@
 package io.temporal.spring.boot.autoconfigure.template;
 
 import io.opentracing.Tracer;
+import io.temporal.common.interceptors.WorkerInterceptor;
 import io.temporal.opentracing.OpenTracingOptions;
 import io.temporal.opentracing.OpenTracingWorkerInterceptor;
 import io.temporal.spring.boot.TemporalOptionsCustomizer;
 import io.temporal.spring.boot.autoconfigure.properties.NamespaceProperties;
 import io.temporal.worker.WorkerFactoryOptions;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
 import java.util.Optional;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
 public class WorkerFactoryOptionsTemplate {
   private final @Nonnull NamespaceProperties namespaceProperties;
+  private final @Nullable List<WorkerInterceptor> workerInterceptors;
   private final @Nullable Tracer tracer;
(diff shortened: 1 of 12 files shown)


cretz (approved):

temporal-spring-boot-autoconfigure/src/main/java/io/temporal/spring/boot/autoconfigure/template/WorkerFactoryOptionsTemplate.java
```diff
@@ -58,12 +65,15 @@ public WorkerFactoryOptions createWorkerFactoryOptions() {
...
     }
 
+    List<WorkerInterceptor> interceptors =
+        new ArrayList<>(workerInterceptors != null ? workerInterceptors : Collections.EMPTY_LIST);
     if (tracer != null) {
       OpenTracingWorkerInterceptor openTracingClientInterceptor =
           new OpenTracingWorkerInterceptor(
               OpenTracingOptions.newBuilder().setTracer(tracer).build());
-      options.setWorkerInterceptors(openTracingClientInterceptor);
+      interceptors.add(openTracingClientInterceptor);
```
Quinn-With-Two-Ns, line 74: @cretz @Sushisource Any thoughts on the default order here? I am putting the tracing interceptor at the end of the chain. unless we want to trace the interceptors themselves maybe?

cretz: Interesting. In other SDKs we have them create the tracing interceptor instead of just the tracer, so it is their responsibility to choose order.

Hard to say whether at the end (is that innermost?) is a good default. Some users swallow or mutate errors, and the span may be better off reflecting the results of other people's interceptors.

Say I didn't want it at the end of the chain as a user, how hard would it be for me to provide a null `tracer` to this constructor here and put it at the beginning of the `workerInterceptors` list? Just curious, because both defaults have tradeoffs, so just wondering if there was a workaround.

Quinn-With-Two-Ns: Yeah end would be innermost, to work around a user would provide a customizer to change the interceptors

Quinn-With-Two-Ns: I can check what Quarkus did

cretz: :+1: If there's a workaround (e.g. don't inject tracer, make the interceptor in the list yourself), that'd be ideal. I do think the tracing intercepter being the outermost makes sense to have the span cover what other interceptors are doing (from a time or exception POV).

Issue #1984, opened 2024-02-05
[Temporal Spring-boot Starter] customizing workflow options by workflow type


sva02:
**[Temporal Spring-boot Starter] cannot customize workflow options by workflow type**
The temporal spring boot starter doesn't provide the capability of customizing workflow implementations by workflow type.
Although there is `WorkflowImplementationOptions.Builder` option to use in `TemporalOptionsCustomizer`, it doesn't provide the flexibility to apply these options by workflow type.

When using the java-sdk, however, this is possible to do so when registering workflow implementations by:
`registerWorkflowImplementationTypes(WorkflowImplementationOptions options, Class<?>... workflowImplementationClasses)`

**Support customizing workflow implementation options by workflow type**
Add the capability to customize workflow options by workflow type when using the spring boot starter, as we might not want to apply these options to all workflow types.

** Context / Use Case **
For scheduled workflows, we want to fail the workflow on `NonDeterministicException.class`.
For the rest of the workflows, this behavior is not desired.

Thank you.

PR #2503, opened 2025-05-02
Support WorkflowImplementationOptionsCustomizer

Quinn-With-Two-Ns:
Support WorkflowImplementationOptionsCustomizer so users can customize workflow implementation options per type in Springboot

closes https://github.com/temporalio/sdk-java/issues/1984

### DIFF
--- a/temporal-spring-boot-autoconfigure/src/main/java/io/temporal/spring/boot/WorkflowImplementationOptionsCustomizer.java
+++ b/temporal-spring-boot-autoconfigure/src/main/java/io/temporal/spring/boot/WorkflowImplementationOptionsCustomizer.java
@@ -1,0 +1,47 @@
+package io.temporal.spring.boot;
+
+import io.temporal.common.metadata.POJOWorkflowMethodMetadata;
+import io.temporal.worker.Worker;
+import io.temporal.worker.WorkflowImplementationOptions;
+import javax.annotation.Nonnull;
+
+/**
+ * Bean of this class can be added to Spring context to get a fine control over {@link
+ * WorkflowImplementationOptions} objects that are created by Temporal Spring Boot Autoconfigure
+ * module.
+ *
+ * <p>Only one bean of this or {@code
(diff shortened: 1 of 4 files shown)


temporal-spring-boot-autoconfigure/src/main/java/io/temporal/spring/boot/WorkflowImplementationOptionsCustomizer.java
cretz, line 1: I think if we had thought about this holistically with dynamic workflow versioning, we'd have one solution to customize workflow implementation options at runtime. As of now there is one way for spring, one way for dynamic workflows for only one option, and no way for non-spring, non-dynamic.

cretz (approved):

temporal-spring-boot-autoconfigure/src/main/java/io/temporal/spring/boot/WorkflowImplementationOptionsCustomizer.java
Quinn-With-Two-Ns: This is not related to runtime customization

This is for customizing workflow implementation options per workflow type at registration time, the Java SDK already supports this, this is just aligning spring boot with the regular Java SDK

Quinn-With-Two-Ns: Spring is not getting any special functionality

cretz: I think there could be an overlap to "customize options by workflow type" whether that workflow type is at registration time or runtime (though runtime obviously gives you more info than workflow type, so not that much overlap).

Quinn-With-Two-Ns: Yeah the point is this PR adds registration time customization for Spring which we already have for the Java SDK

Issue #1449, opened 2022-09-21
Spring Boot - add support for dynamic worklows


tsurdilo:
@WorkflowImpl(taskQueues = "MyTaskQueueName")
      public class MyWorkflow implements DynamicWorkflow { ... }

currently raises:

     Class doesn't implement any non empty public interface annotated with @WorkflowInterface:

If possible please add support for DynamicWorkflow

PR #2506, opened 2025-05-02
Add support for dynamics workflows to Springboot

Quinn-With-Two-Ns:
Add support for dynamics workflows to Springboot

closes https://github.com/temporalio/sdk-java/issues/1449

### DIFF
--- a/temporal-spring-boot-autoconfigure/src/main/java/io/temporal/spring/boot/autoconfigure/template/WorkersTemplate.java
+++ b/temporal-spring-boot-autoconfigure/src/main/java/io/temporal/spring/boot/autoconfigure/template/WorkersTemplate.java
@@ -1,14 +1,18 @@
 package io.temporal.spring.boot.autoconfigure.template;
 
+import static io.temporal.serviceclient.CheckedExceptionWrapper.wrap;
+
 import com.google.common.base.Preconditions;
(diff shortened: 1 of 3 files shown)


temporal-spring-boot-autoconfigure/src/main/java/io/temporal/spring/boot/autoconfigure/template/WorkersTemplate.java
```diff
@@ -512,6 +512,48 @@ private void configureWorkflowImplementationAutoDiscovery(
 
   @SuppressWarnings("unchecked")
   private <T> void configureWorkflowImplementation(Worker worker, Class<?> clazz) {
+    // Handle dynamic workflows
+    if (DynamicWorkflow.class.isAssignableFrom(clazz)) {
```
cretz, line 516: I figure lots of this comes from existing code? I'm guessing there's not a reasonable way to reuse (i.e. maybe we don't want to reference `internal` packages of another JAR)?

Quinn-With-Two-Ns: No we don't want to reference internal, it will break Java modules.

Issue #2508, opened 2025-05-06
The version of grpc needs to be updated


fatekingsama:
Some classes have been removed in the new version of grpc, and your grpc version is too old, causing startup errors

Caused by: java.lang.NoClassDefFoundError: io/grpc/internal/AbstractManagedChannelImplBuilder
	at java.base/java.lang.ClassLoader.defineClass1(Native Method)
	at java.base/java.lang.ClassLoader.defineClass(ClassLoader.java:1027)
	at java.base/java.security.SecureClassLoader.defineClass(SecureClassLoader.java:150)
	at java.base/jdk.internal.loader.BuiltinClassLoader.defineClass(BuiltinClassLoader.java:862)
	at java.base/jdk.internal.loader.BuiltinClassLoader.findClassOnClassPathOrNull(BuiltinClassLoader.java:760)
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClassOrNull(BuiltinClassLoader.java:681)
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:639)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:188)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:526)
	at io.temporal.serviceclient.ChannelManager.prepareChannel(ChannelManager.java:206)
	at io.temporal.serviceclient.ChannelManager.<init>(ChannelManager.java:118)
	at io.temporal.serviceclient.ChannelManager.<init>(ChannelManager.java:98)
	at io.temporal.serviceclient.WorkflowServiceStubsImpl.<init>(WorkflowServiceStubsImpl.java:84)
	at io.temporal.serviceclient.WorkflowServiceStubs.newServiceStubs(WorkflowServiceStubs.java:66)

fatekingsama:
I’ve submitted a pull request to resolve the problem, but there are a few uncertainties in this update that I’d appreciate your input on.

Quinn-With-Two-Ns:
The Java SDK intentionally builds with an older version of gRPC Java to maintain compatibility with the widest range of versions for our users. You are free and encouraged to use a higher version in your end application. The specific error on this issue doesn't really make sense to me since this is a very [old issue](https://github.com/grpc/grpc-java/issues/7552) that shouldn't be an issue for any library that compiled with any version <`1.36.0`

fatekingsama:
[...]

PR #2509, opened 2025-05-06
Upgrade gRPC dependencies to 1.68.2 (#2508)

fatekingsama:
## What was changed
Upgrade gRPC dependencies to 1.68.2 (#2508)

## Why?
The framework we are using is a newer version of grpc, which is not compatible with the current grpc version of the SDK, so we need to upgrade it. This should be a seamless upgrade for users

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -28,7 +28,7 @@ allprojects {
 
 ext {
     // Platforms
-    grpcVersion = '1.58.1' // [1.38.0,) Needed for io.grpc.protobuf.services.HealthStatusManager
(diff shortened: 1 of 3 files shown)


fatekingsama:
I commented out the line exclude '**/google/**/*' in the build.gradle file of :temporal-serviceclient and documented the reason in the code.
However, I noticed that this line is related to another issue, so I'm not sure whether this change might introduce potential side effects.
Please let me know if it causes any problems.

Quinn-With-Two-Ns:
@fatekingsama We intentionally do nut bump grpc to maintain compatibility with the widest range of gRPC versions, you are free to include a greater version in your build

Issue #2473, opened 2025-04-07
Cancellation scopes and async procedures are canceled in non-deterministic order when workflow is canceled


sajkol:
## Description of problem
I have a workflow that uses cancellation scopes and a timer to handle expiration, which can be reduced to something like this (using Java SDK):
```
var scope = Workflow.newCancellationScope(() -> {
    activity.doActivity();
});

Async.procedure(() -> {
    Workflow.sleep(Duration.ofMinutes(5));
    scope.cancel();
});

scope.run();
```
However, when this workflow is canceled during activity execution, it seems that the order in which the pending activity and the timer are canceled is not deterministic, which sometimes causes `NonDeterministicException` when replaying.

## Steps to Reproduce the Problem
Here's the issue reproduced in a unit test. Test is repeated because the result is random - sometimes the history generated in the replay happens to match the one resulting from initial execution.

```
package com.example;

import io.temporal.activity.ActivityInterface;
import io.temporal.activity.ActivityOptions;
import io.temporal.client.WorkflowClient;
import io.temporal.client.WorkflowOptions;
import io.temporal.client.WorkflowStub;
import io.temporal.testing.TestWorkflowEnvironment;
import io.temporal.testing.WorkflowReplayer;
import io.temporal.worker.Worker;
import io.temporal.workflow.Async;
import io.temporal.workflow.Workflow;
import io.temporal.workflow.WorkflowInterface;
import io.temporal.workflow.WorkflowMethod;
import java.time.Duration;
import org.junit.jupiter.api.BeforeEach;
import org.junit.jupiter.api.RepeatedTest;

@Slf4j
public class WorkflowCanceledTimerTest {

    @WorkflowInterface
    public interface TestWorkflow {
        @WorkflowMethod
        void start();
    }

    @ActivityInterface
    public interface TestActivity {
        void doActivity();
    }

    public static class TestActivityImpl implements TestActivity {
        @Override
        public void doActivity() {
            try {
                Thread.sleep(5000);
            } catch (InterruptedException e) {
                throw new RuntimeException(e);
            }
        }
    }

    public static class TestWorkflowImpl implements TestWorkflow {

        TestActivity activity = Workflow.newActivityStub(TestActivity.class,
                ActivityOptions.newBuilder().setScheduleToCloseTimeout(Duration.ofSeconds(60)).build());

        @Override
        public void start() {
            var scope = Workflow.newCancellationScope(() -> {
                activity.doActivity();
            });

[...]

PR #2512, opened 2025-05-07
Make CancellationScopeImpl more deterministic

Quinn-With-Two-Ns:
Make `CancellationScopeImpl` more deterministic, our use of `HashSet` is not deterministic since we iterate over the map and the order matters her. 

closes https://github.com/temporalio/sdk-java/issues/2473

### DIFF
--- a/temporal-sdk/src/test/java/io/temporal/workflow/cancellationTests/WorkflowCancellationScopeDeterminism.java
+++ b/temporal-sdk/src/test/java/io/temporal/workflow/cancellationTests/WorkflowCancellationScopeDeterminism.java
@@ -1,0 +1,100 @@
+package io.temporal.workflow.cancellationTests;
+
+import io.temporal.activity.ActivityInterface;
+import io.temporal.activity.ActivityOptions;
+import io.temporal.client.WorkflowClient;
+import io.temporal.client.WorkflowStub;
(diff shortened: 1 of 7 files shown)


temporal-sdk/src/test/java/io/temporal/workflow/cancellationTests/WorkflowCancellationScopeDeterminism.java
cretz, line 1: Is there a test that confirms the HashSet is a problem? I figure it should assert task failure (can be done in loop like other test below) and then when you enable the flag, can be altered to assert no task failure.

cretz (approved):

temporal-sdk/src/test/java/io/temporal/workflow/cancellationTests/WorkflowCancellationScopeDeterminism.java
cretz: Ah I see now it's running a workflow, grabbing history, and replaying it 100 times. I assume that's not too expensive of a test time wise.

Quinn-With-Two-Ns: No takes about 5s

PR #2513, opened 2025-05-07
Set links in Nexus callback

rodrigozhou:
## What was changed

Support for `RequestIdReference` link and setting the links in the Nexus callback

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


Quinn-With-Two-Ns:
Do we need any test server changes here?

rodrigozhou:
I don't think so. This is purely an SDK change.

rodrigozhou:
The server doesn't do anything with the links in the callback. It also doesn't do anything with links in the request besides deduping. I think deduping is the only thing the server does.

Quinn-With-Two-Ns (approved):

PR #2515, opened 2025-05-12
RequestIdInfo and links changes in test server

rodrigozhou:
## What was changed

Changes to the test server:
- Add `requestIdInfos` to the mutable state and return in the `DescribeWorkflowExecution` API.
- Return link in the `StartWorkflowExecution` API.
- Refactor start workflow dedup logic to lookup `requestIdInfos`.
- Dedup links in start workflow request.

## Checklist

1. Closes 

2. How was this tested:

Created/modified tests.

3. Any docs updates needed?


temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -3502,26 +3526,26 @@ private void addExecutionSignaledByExternalEvent(
...
+        HistoryEvent.newBuilder()
+            .setWorkerMayIgnore(true)
+            .setEventType(EVENT_TYPE_WORKFLOW_EXECUTION_OPTIONS_UPDATED);
+
     WorkflowExecutionOptionsUpdatedEventAttributes.Builder attrs =
         WorkflowExecutionOptionsUpdatedEventAttributes.newBuilder();
     if (requestId != null) {
       attrs.setAttachedRequestId(requestId);
-      this.attachedRequestIds.add(requestId);
+      this.attachRequestId(requestId, event.getEventType(), BUFFERED_EVENT_ID);
```
bergundy, line 3538: Is this necessarily buffered?

rodrigozhou: Unsure, but it doesn't really matter. In the test server, when you add the event to the context, it gets an event id, but it doesn't seem to necessarily the final event id (which is weird, but make sense since it might have the workflow task completed added before flushing the context).
I can assume that it is buffered here, and if the context flushes the events, it will assign the final event id at that point, and this will be updated.

Issue #1350, opened 2022-08-08
Wire #reason parameter in workflow cancellation request


Spikhalskiy:
Server added RequestCancelWorkflowExecutionRequest#reason field for cancellation requests. 
JavaSDK should allow this field to be set the same way it's allowed for termination requests.

Issue #2414, opened 2025-02-11
Expose Workflow cancel cause/reason


Sushisource:
WorkflowExecutionCancelRequestedEventAttributes has a cause field that should be exposed when workflows receive cancels, and we should additionally expose this as an input when users are issuing cancels from clients or commands

madhav2302:
@Sushisource does this reason also exposed to a running activity when it heartbeats and gets `ActivityCanceledException` execption?

Quinn-With-Two-Ns:
@madhav2302 No, this information is not returned by the Server on heartbeat. The only information the Server [returns](https://github.com/temporalio/api/blob/9156239c8272003b6f228951602ca2b130a55876/temporal/api/workflowservice/v1/request_response.proto#L515) is if to cancel the activity.

antmendoza:
Hi @madhav2302 , we have already submitted a feature request for this to our product team.

PR #2519, opened 2025-05-13
Wire reason parameter in workflow cancellation request

Quinn-With-Two-Ns:
Wire reason parameter in workflow cancellation request. Not this does not cover wiring the cancellation reason on the workflow side. There is a different issue to track that work.

closes https://github.com/temporalio/sdk-java/issues/1350

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/client/WorkflowStub.java
+++ b/temporal-sdk/src/main/java/io/temporal/client/WorkflowStub.java
@@ -370,6 +370,21 @@ public interface WorkflowStub {
   void cancel();
 
   /**
+   * Request cancellation of a workflow execution with a reason.
+   *
+   * <p>Cancellation cancels {@link io.temporal.workflow.CancellationScope} that wraps the main
+   * workflow method. Note that workflow can take long time to get canceled or even completely
+   * ignore the cancellation request.
+   *
+   * @param reason optional reason for the cancellation request
+   * @throws WorkflowNotFoundException if the workflow execution doesn't exist or is already
+   *     completed
+   * @throws WorkflowServiceException for all other failures including networking and service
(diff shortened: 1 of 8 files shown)


temporal-sdk/src/main/java/io/temporal/client/WorkflowStub.java
cretz, line 1: How does this relate to #2414 (as part of https://github.com/temporalio/features/issues/596)? Specifically, is this a partial implementation of that? And are other SDKs expected to follow suit with a "reason" from client side?

cretz, line 1: Does this apply to child workflows?

Quinn-With-Two-Ns: Yes as I said in the PR description this does not cover the workflow side

cretz: :+1: Just wanted to confirm it is part of that project since it wasn't linked

Quinn-With-Two-Ns: `reason` on the client side is something we already do for terminate. I think how we expose this on the workflow side needs a bit of SDK team consensus so I left it out for now.

cretz: :+1: I see `CancellationScope.cancel` accepts a reason, does that propagate to `RequestCancelExternalWorkflowExecutionCommandAttributes.reason` proto? Similarly, I think we may want to have an overload for `ExternalWorkflowStub.cancel` that accepts `reason`. Both of those can be separate issues though, not blocking this one.

Issue #2160, opened 2024-07-30
Child workflow cancellation can trigger SDK event loop without a WFT Started


Quinn-With-Two-Ns:
## Expected Behavior

Child workflow cancellation would not trigger the event loop without a WFT Started

## Actual Behavior

Child workflow cancellation can trigger SDK event loop without a WFT Started

## Steps to Reproduce the Problem

see: https://github.com/temporalio/sdk-java/issues/2155

https://github.com/temporalio/sdk-java/blob/eabd51fbdc62a7435c8f4d8dc1ac24da26f6022d/temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java#L884

Fixing this would be a non deterministic change so we will need to version this change.

PR #2520, opened 2025-05-13
Implement Nexus operation cancellation types

pdoerner:
## What was changed
Implemented cancellation types for Nexus operations which specify what action to take when the caller context is cancelled.
* `ABANDON` -> Do not request cancellation of the operation.
* `TRY_CANCEL` -> Initiate a cancellation request and immediately report cancellation to the caller.
* `WAIT_REQUESTED` -> Request cancellation of the operation and wait for confirmation that the request was received.
* `WAIT_COMPLETED` -> Wait for operation completion. Default.
These options are set on NexusOperationOptions which are passed in when starting an operation.

## Why?
To give users more control over what happens to their handler workflows when their caller workflows are cancelled.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/statemachines/CancelNexusOperationStateMachine.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/statemachines/CancelNexusOperationStateMachine.java
@@ -4,6 +4,8 @@ import io.temporal.api.command.v1.Command;
 import io.temporal.api.command.v1.RequestCancelNexusOperationCommandAttributes;
 import io.temporal.api.enums.v1.CommandType;
 import io.temporal.api.enums.v1.EventType;
+import io.temporal.api.failure.v1.CanceledFailureInfo;
+import io.temporal.api.failure.v1.Failure;
 import io.temporal.workflow.Functions;
 
 /** CancelNexusOperationStateMachine manges a request to cancel a nexus operation. */
@@ -15,23 +17,29 @@ final class CancelNexusOperationStateMachine
 
   private final RequestCancelNexusOperationCommandAttributes requestCancelNexusAttributes;
 
+  private final Functions.Proc2<Void, Failure> completionCallback;
+
   /**
    * @param attributes attributes to use to cancel a nexus operation
    * @param commandSink sink to send commands
    */
   public static void newInstance(
       RequestCancelNexusOperationCommandAttributes attributes,
+      Functions.Proc2<Void, Failure> completionCallback,
       Functions.Proc1<CancellableCommand> commandSink,
       Functions.Proc1<StateMachine> stateMachineSink) {
-    new CancelNexusOperationStateMachine(attributes, commandSink, stateMachineSink);
+    new CancelNexusOperationStateMachine(
+        attributes, completionCallback, commandSink, stateMachineSink);
   }
 
   private CancelNexusOperationStateMachine(
       RequestCancelNexusOperationCommandAttributes attributes,
+      Functions.Proc2<Void, Failure> completionCallback,
       Functions.Proc1<CancellableCommand> commandSink,
       Functions.Proc1<StateMachine> stateMachineSink) {
     super(STATE_MACHINE_DEFINITION, commandSink, stateMachineSink);
     this.requestCancelNexusAttributes = attributes;
+    this.completionCallback = completionCallback;
     explicitEvent(ExplicitEvent.SCHEDULE);
   }
 
@@ -42,14 +50,19 @@ final class CancelNexusOperationStateMachine
   enum State {
     CREATED,
     REQUEST_CANCEL_NEXUS_OPERATION_COMMAND_CREATED,
+    REQUEST_CANCEL_NEXUS_OPERATION_COMMAND_RECORDED,
     CANCEL_REQUESTED,
+    REQUEST_CANCEL_FAILED,
   }
 
   public static final StateMachineDefinition<State, ExplicitEvent, CancelNexusOperationStateMachine>
       STATE_MACHINE_DEFINITION =
           StateMachineDefinition
               .<State, ExplicitEvent, CancelNexusOperationStateMachine>newInstance(
-                  "CancelNexusOperation", State.CREATED, State.CANCEL_REQUESTED)
+                  "CancelNexusOperation",
+                  State.CREATED,
+                  State.CANCEL_REQUESTED,
+                  State.REQUEST_CANCEL_FAILED)
               .add(
                   State.CREATED,
                   ExplicitEvent.SCHEDULE,
@@ -62,8 +75,18 @@ final class CancelNexusOperationStateMachine
               .add(
                   State.REQUEST_CANCEL_NEXUS_OPERATION_COMMAND_CREATED,
                   EventType.EVENT_TYPE_NEXUS_OPERATION_CANCEL_REQUESTED,
+                  State.REQUEST_CANCEL_NEXUS_OPERATION_COMMAND_RECORDED,
+                  EntityStateMachineInitialCommand::setInitialCommandEventId)
+              .add(
+                  State.REQUEST_CANCEL_NEXUS_OPERATION_COMMAND_RECORDED,
+                  EventType.EVENT_TYPE_NEXUS_OPERATION_CANCEL_REQUEST_COMPLETED,
                   State.CANCEL_REQUESTED,
-                  CancelNexusOperationStateMachine::notifyCompleted);
+                  CancelNexusOperationStateMachine::notifyCompleted)
+              .add(
+                  State.REQUEST_CANCEL_NEXUS_OPERATION_COMMAND_RECORDED,
+                  EventType.EVENT_TYPE_NEXUS_OPERATION_CANCEL_REQUEST_FAILED,
+                  State.REQUEST_CANCEL_FAILED,
+                  CancelNexusOperationStateMachine::notifyFailed);
 
   private void createCancelNexusCommand() {
     addCommand(
@@ -74,6 +97,17 @@ final class CancelNexusOperationStateMachine
   }
 
   private void notifyCompleted() {
-    setInitialCommandEventId();
+    Failure canceledFailure =
+        Failure.newBuilder()
+            .setMessage("operation canceled")
+            .setCanceledFailureInfo(CanceledFailureInfo.getDefaultInstance())
+            .build();
+    completionCallback.apply(null, canceledFailure);
+  }
+
+  private void notifyFailed() {
+    Failure failure =
+        currentEvent.getNexusOperationCancelRequestFailedEventAttributes().getFailure();
+    completionCallback.apply(null, failure);
   }
 }
--- a/temporal-sdk/src/main/java/io/temporal/internal/statemachines/NexusOperationStateMachine.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/statemachines/NexusOperationStateMachine.java
@@ -37,6 +37,10 @@ final class NexusOperationStateMachine
   private final String service;
   private final String operation;
 
+  public boolean isAsync() {
+    return async;
+  }
+
   public boolean isCancellable() {
     return State.SCHEDULE_COMMAND_CREATED == getState();
   }
@@ -145,25 +149,12 @@ final class NexusOperationStateMachine
 
   private void cancelNexusOperationCommand() {
     cancelCommand();
-    Failure canceledFailure =
+    Failure cause =
         Failure.newBuilder()
-            .setSource(JAVA_SDK)
             .setMessage("operation canceled before it was started")
             .setCanceledFailureInfo(CanceledFailureInfo.getDefaultInstance())
             .build();
-    NexusOperationFailureInfo nexusFailureInfo =
-        NexusOperationFailureInfo.newBuilder()
-            .setEndpoint(endpoint)
-            .setService(service)
-            .setOperation(operation)
-            .setScheduledEventId(getInitialCommandEventId())
-            .build();
-    Failure failure =
-        Failure.newBuilder()
-            .setNexusOperationExecutionFailureInfo(nexusFailureInfo)
-            .setCause(canceledFailure)
-            .setMessage(NEXUS_OPERATION_CANCELED_MESSAGE)
-            .build();
+    Failure failure = createCancelNexusOperationFailure(cause);
     startedCallback.apply(Optional.empty(), failure);
     completionCallback.apply(Optional.empty(), failure);
   }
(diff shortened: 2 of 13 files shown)


temporal-sdk/src/main/java/io/temporal/workflow/NexusOperationCancellationType.java
```diff
@@ -0,0 +1,29 @@
+package io.temporal.workflow;
+
+import io.temporal.failure.CanceledFailure;
+
+/**
+ * Defines behavior of the parent workflow when {@link CancellationScope} that wraps Nexus operation
+ * is canceled. The result of the cancellation independently of the type is a {@link
+ * CanceledFailure} thrown from the Nexus operation method.
+ */
+public enum NexusOperationCancellationType {
```
Quinn-With-Two-Ns, line 10: Are we releasing this as GA or will it be Pre-release/Public preview?

temporal-sdk/src/main/java/io/temporal/workflow/NexusOperationOptions.java
```diff
@@ -45,6 +46,18 @@ public NexusOperationOptions.Builder setScheduleToCloseTimeout(
       return this;
     }
 
+    /**
+     * Sets the cancellation type for the Nexus operation.
+     *
+     * @param cancellationType the cancellation type for the Nexus operation
+     * @return this
+     */
+    public NexusOperationOptions.Builder setCancellationType(
```
Quinn-With-Two-Ns, line 55: Can we document the default?

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
```diff
@@ -930,32 +932,70 @@ public Functions.Proc startChildWorkflow(
...
+              }
+            });
+        if (cancellationType == NexusOperationCancellationType.TRY_CANCEL) {
+          notifyNexusOperationCanceled(operation, startedCallback, completionCallback);
+        }
       }
     };
   }
 
+  private void notifyNexusOperationCanceled(
```
Quinn-With-Two-Ns, line 974: We need to be careful we don't copy this bug from child workflows to nexus operations https://github.com/temporalio/sdk-java/issues/2160

Quinn-With-Two-Ns: essentially not run the event loop if we are notified by an event

temporal-sdk/src/main/java/io/temporal/workflow/NexusOperationCancellationType.java
pdoerner: It will be GA

Quinn-With-Two-Ns: Shouldn't it be Pre-release/Public preview, we talked about it during the nexus crew stand up ~2 weeks ago?

temporal-sdk/src/main/java/io/temporal/internal/statemachines/NexusOperationStateMachine.java
```diff
@@ -263,4 +254,20 @@ public void createScheduleNexusTaskCommand() {
     scheduleAttributes = null;
     metadata = null;
   }
+
+  public Failure createCancelNexusOperationFailure(Failure cause) {
+    Failure canceledFailure = Failure.newBuilder(cause).setSource(JAVA_SDK).build();
```
Quinn-With-Two-Ns: The source should only be set to the `JAVA_SDK` on the very specific code path where the "operation canceled before it was started", not in general

temporal-sdk/src/main/java/io/temporal/internal/statemachines/CancelNexusOperationStateMachine.java
```diff
@@ -74,6 +97,17 @@ private void createCancelNexusCommand() {
   }
 
   private void notifyCompleted() {
-    setInitialCommandEventId();
+    Failure canceledFailure =
```
Quinn-With-Two-Ns: Why are we passing a failure back on `notifyCompleted`? If you look at `CancelExternalStateMachine` which is basically the same state machine but for workflows we don't pass anything back

temporal-sdk/src/test/java/io/temporal/workflow/nexus/CancelWorkflowAsyncOperationTest.java
```diff
@@ -0,0 +1,299 @@
...
+
+    testWorkflowRule.regenerateHistoryForReplay(
+        execution.getWorkflowId(), "testCancelAsyncOperationWaitCompleted");
+  }
+
+  @Test
+  public void cancelAsyncOperationWaitRequested() {
+    // TODO: remove skip once OSS 1.29.0 is released
+    assumeFalse(
+        "Nexus operation cancel request delivery events are not recorded by default for server versions <1.29.0",
```
Quinn-With-Two-Ns: By default? Can it be enabled in server versions <1.29.0?

pdoerner: Yes, it is under a dynamic config: `component.nexusoperations.recordCancelRequestCompletionEvents`

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
Quinn-With-Two-Ns: I think this point still needs to be addressed

temporal-sdk/src/test/java/io/temporal/workflow/nexus/CancelWorkflowAsyncOperationTest.java
Quinn-With-Two-Ns: Can we just set that dynamic config then so we can enabled the test?

temporal-sdk/src/main/java/io/temporal/internal/statemachines/NexusOperationStateMachine.java
Quinn-With-Two-Ns: Well it applies in any case where the SDK is generating the failure

temporal-sdk/src/main/java/io/temporal/internal/statemachines/CancelNexusOperationStateMachine.java
pdoerner: It is to pass back the specific Nexus operation failure. I think CancelExternal always has the same cause.

Quinn-With-Two-Ns: The `completionCallback` takes two parameters. The intention is the first parameter is for communicating success, the second is for communication failure. `notifyCompleted` is not a failure so it shouldn't pass a value on the second parameter

temporal-sdk/src/test/java/io/temporal/workflow/nexus/CancelWorkflowAsyncOperationTest.java
pdoerner: Well it can ben enabled in 1.28.0 when it was added. Earlier versions still will not record those new events.

pdoerner: I think the dev server is still on 1.27 for now. But once it is updated to 1.28 I can enable those tests

temporal-sdk/src/main/java/io/temporal/internal/statemachines/CancelNexusOperationStateMachine.java
Quinn-With-Two-Ns: otherwise `completionCallback ` should just take one parameter since you are never using the first parameter

temporal-sdk/src/test/java/io/temporal/workflow/nexus/CancelWorkflowAsyncOperationTest.java
Quinn-With-Two-Ns: Dev server is on the same/later version then the Go SDK is and we enabled it for Go

temporal-sdk/src/main/java/io/temporal/internal/statemachines/CancelNexusOperationStateMachine.java
pdoerner: Gotcha. Updated, let me know if that looks better.

temporal-sdk/src/test/java/io/temporal/workflow/nexus/CancelWorkflowAsyncOperationTest.java
pdoerner: Ah I didn't see we were using a custom CLI version for CI/CD. Added the config to the GH workflow.

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
pdoerner: Does it? I just removed the call to `eventLoop()`. Was there some other change required?

temporal-sdk/src/main/java/io/temporal/workflow/NexusOperationCancellationType.java
pdoerner: IIRC we did not mark it as experimental in the Go SDK. I could do so here if you prefer, I forget that conversation.

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
Quinn-With-Two-Ns: Well removing the call to `eventLoop()` doesn't address the issue because if we are **not** notified by an event we do need to run the event loop

temporal-sdk/src/main/java/io/temporal/workflow/NexusOperationCancellationType.java
Quinn-With-Two-Ns: We did end up marking it experimental for Go, lets mark it experimental here as well

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
pdoerner: In what situation would we not be notified by an event? I guess I don't totally understand the event loop logic or how to check where the notification came from.

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowStateMachines.java
```diff
@@ -985,32 +987,71 @@ public Functions.Proc startChildWorkflow(
...
-            metadata,
+            parameters.getAttributes().build(),
+            parameters.getMetadata(),
             startedCallback,
             completionCallback,
             commandSink,
             stateMachineSink);
     return () -> {
       if (operation.isCancellable()) {
         operation.cancel();
```
Quinn-With-Two-Ns: hm, if `NexusOperationCancellationType.ABANDON` should we cancel?

pdoerner: I think if the operation hasn't started, we should still cancel yes. ABANDON indicates that we should not send the RequestCancel command which we do not do if the operation hasn't been started.

Quinn-With-Two-Ns: hm so looking at the child workflow cancellation
```
      if (cancellationType == ChildWorkflowCancellationType.ABANDON) {
        notifyChildCanceled(completionCallback);
        return;
      }
      // The only time child can be canceled directly is before its start command
      // was sent out to the service. After that RequestCancelExternal should be used.
      if (child.isCancellable()) {
        child.cancel();
        return;
      }
```
We don't cancel the child workflow if `ChildWorkflowCancellationType.ABANDON`

pdoerner: If the schedule command hasn't been sent, I mean. Not that the operation itself has not started.

Quinn-With-Two-Ns: I think most users understand `ABANDON` as we don't request cancellation of the entity (child workflow or nexus operation) regardless of what state the entity happens to be in

pdoerner: Oh okay. That makes sense.

temporal-sdk/src/test/java/io/temporal/workflow/nexus/CancelWorkflowAsyncOperationTest.java
```diff
@@ -0,0 +1,282 @@
...
+import io.temporal.testing.WorkflowReplayer;
+import io.temporal.testing.internal.SDKTestWorkflowRule;
+import io.temporal.workflow.*;
+import io.temporal.workflow.shared.TestWorkflows;
+import java.time.Duration;
+import org.junit.Assert;
+import org.junit.Rule;
+import org.junit.Test;
+
+public class CancelWorkflowAsyncOperationTest extends BaseNexusTest {
```
Quinn-With-Two-Ns, line 27: Does this work?
```suggestion
public class CancelWorkflowAsyncOperationTest {
```
`SDKTestWorkflowRule` should do all the setup for this test

temporal-sdk/src/test/java/io/temporal/workflow/nexus/CancelWorkflowAsyncOperationTest.java
```diff
@@ -0,0 +1,282 @@
...
+              .setCancellationType(cancellationType)
+              .build();
+      NexusServiceOptions serviceOptions =
+          NexusServiceOptions.newBuilder().setOperationOptions(options).build();
+      TestNexusService serviceStub =
+          Workflow.newNexusServiceStub(TestNexusService.class, serviceOptions);
+
+      NexusOperationHandle<Void> handle =
+          Workflow.startNexusOperation(serviceStub::operation, 2000L);
+      handle.getExecution().get();
```
Quinn-With-Two-Ns, line 243: Can we also have a test where we cancel with the different cancellation types before calling `handle.getExecution().get();` to trigger the code path when we cancel before the command is sent

Quinn-With-Two-Ns:
LGTM! would like a bit more test coverage on the start then immediately cancel code path, but otherwise look good.

temporal-sdk/src/test/java/io/temporal/workflow/nexus/CancelWorkflowAsyncOperationTest.java
pdoerner: I did it this way only for the replay test since we cannot use the test rule to get the endpoint name from within the static workflow definition

temporal-sdk/src/test/java/io/temporal/workflow/nexus/CancelWorkflowAsyncOperationTest.java
pdoerner: Done

temporal-sdk/src/main/java/io/temporal/workflow/NexusOperationOptions.java
```diff
@@ -45,6 +46,19 @@ public NexusOperationOptions.Builder setScheduleToCloseTimeout(
       return this;
     }
 
+    /**
+     * Sets the cancellation type for the Nexus operation. Defaults to WAIT_COMPLETED. Note:
+     * EXPERIMENTAL
```
Quinn-With-Two-Ns: Oh sorry I missed this, for the Java SDK we have an annotation to mark APIs as experimental you can add `@Experimental` to the API and type 





temporal-sdk/src/main/java/io/temporal/common/Experimental.java

Issue #2328, opened 2024-11-25
Add an interceptor for listExecutions


Quinn-With-Two-Ns:
**Is your feature request related to a problem? Please describe.**
Add an interceptor for `WorkflowClient.listExecutions`. We should also expose the constructor of `WorkflowExecutionMetadata`

**Additional context**
https://github.com/temporalio/sdk-java/pull/2326#discussion_r1854044466

PR #2524, opened 2025-05-14
Add an interceptor for listExecutions

Quinn-With-Two-Ns:
Add an interceptor for listExecutions

closes https://github.com/temporalio/sdk-java/issues/2328

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/client/GetWorkflowExecutionHistoryIterator.java
+++ b/temporal-sdk/src/main/java/io/temporal/client/GetWorkflowExecutionHistoryIterator.java
@@ -5,6 +5,7 @@ import io.temporal.api.common.v1.WorkflowExecution;
 import io.temporal.api.history.v1.HistoryEvent;
 import io.temporal.api.workflowservice.v1.GetWorkflowExecutionHistoryRequest;
 import io.temporal.api.workflowservice.v1.GetWorkflowExecutionHistoryResponse;
+import io.temporal.internal.client.EagerPaginator;
 import io.temporal.internal.client.external.GenericWorkflowClient;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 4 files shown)


temporal-sdk/src/main/java/io/temporal/client/ListWorkflowExecutionIterator.java
```diff
@@ -11,14 +11,14 @@
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
-class ListWorkflowExecutionIterator
+public class ListWorkflowExecutionIterator
```
cretz, line 14: Does this need to be public to support the interceptor? In .NET for example we felt there was no need to expose the underlying machinery of list workflows because creating an async iterator (or Java stream) should be easy enough if they _must_ completely replace the logic and not just delegate to the next interceptor.

Quinn-With-Two-Ns: ideally no, the problem is the [RootWorkflowClientInvoker.java](https://github.com/temporalio/sdk-java/pull/2524/files#diff-7209d2833e3d842e729546850af071d4522573f109e0c979cf21314be7488bba) is in internal so it can't be private, moving this class causes a lot of other issues

cretz: > moving this class causes a lot of other issues

Hrmm. I wonder making public _and_ moving from `io.temporal.client` to `io.temporal.client.internal` would make sense.

Quinn-With-Two-Ns: Let me try, I think I had some issue doing that since some other classes in `io.temporal.client` also have Iterators

Quinn-With-Two-Ns: Moved this class and `EagerPaginator` to internal

Issue #1256, opened 2022-06-09
Last cached activity heartbeat should be discarded or drained with activity completion


Spikhalskiy:
**Is your feature request related to a problem? Please describe.**
Currently, if an Activity heartbeat too often and the heartbeat gets throttled and cached, an activity finishes without draining or canceling the heartbeat.

**Describe the solution you'd like**
The last cached heartbeat should be either drained before the activity completion or discarded/canceled.

PR #2526, opened 2025-05-15
Cancel pending heartbeat when activity completes

Quinn-With-Two-Ns:
Cancel pending heartbeat when activity completes. This prevents the heartbeat from holding up worker shutdown and prevents heartbeating an already completed activity

closes: https://github.com/temporalio/sdk-java/issues/1256

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/activity/ActivityTaskExecutors.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/activity/ActivityTaskExecutors.java
@@ -128,6 +128,12 @@ final class ActivityTaskExecutors {
             metricsScope,
             local,
             dataConverterWithActivityContext);
+      } finally {
+        if (!context.isDoNotCompleteOnReturn()) {
+          // if the activity is not completed, we need to cancel the heartbeat
+          // to avoid sending it after the activity is completed
+          context.cancelOutstandingHeartbeat();
+        }
       }
     }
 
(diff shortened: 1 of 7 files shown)


temporal-sdk/src/main/java/io/temporal/internal/activity/ActivityTaskExecutors.java
```diff
@@ -128,6 +128,12 @@ public ActivityTaskHandler.Result execute(ActivityInfoInternal info, Scope metri
             metricsScope,
             local,
             dataConverterWithActivityContext);
+      } finally {
+        if (!context.isDoNotCompleteOnReturn()) {
+          // if the activity is not completed, we need to cancel the heartbeat
+          // to avoid sending it after the activity is completed
+          context.cancelOutstandingHeartbeat();
```
cretz, line 135: Can you clarify this a bit? So except for async-activity-completion, this is going to prevent a pending heartbeat from being recorded? Is there somewhere else in Java where the last heartbeat of a failing activity is guaranteed to be sent?

Quinn-With-Two-Ns: > Is there somewhere else in Java where the last heartbeat of a failing activity is guaranteed to be sent?

Yes, we send the last heartbeat value as part of the activity failure.

cretz (approved):

Issue #2180, opened 2024-08-11
Workflow.newExternalWorkflowStub


mfateev:
## Expected Behavior
The following works when using Workflow Client:

```
public interface Retryable {
    @SignalMethod
    void retryNow();
}

@WorkflowInterface
public interface FileProcessingWorkflow extends Retryable {

    @WorkflowMethod
    String processFile(Arguments args);

    @QueryMethod(name="history")
    List<String> getHistory();

    @QueryMethod
    String getStatus();

    @SignalMethod
    void abandon();
}

@WorkflowInterface
public interface MediaProcessingWorkflow extends Retryable {

    @WorkflowMethod
    String processBlob(Arguments args);
}

Retryable r1 = client.newWorkflowStub(Retryable.class, firstWorkflowId);
Retryable r2 = client.newWorkflowStub(Retryable.class, secondWorkflowId);
r1.retryNow();
r2.retryNow();
```
The same should work when signaling workflow from an external workflow:
```
Workflow.newExternalWorkflowStub(Retryable.class, firstWorkflowId);

Retryable r1 = Workflow.newExternalWorkflowStub(Retryable.class, firstWorkflowId);
Retryable r2 = Workflow.newExternalWorkflowStub(Retryable.class, secondWorkflowId);
r1.retryNow();
r2.retryNow();
```

## Actual Behavior
The newExternalWorkflowStub fails with:
```
Caused by: java.lang.IllegalArgumentException: Missing required @WorkflowInterface annotation: interface ...Retryable
	at io.temporal.common.metadata.POJOWorkflowInterfaceMetadata.newInstanceInternal(POJOWorkflowInterfaceMetadata.java:179)
	at io.temporal.common.metadata.POJOWorkflowInterfaceMetadata.newInstance(POJOWorkflowInterfaceMetadata.java:108)
	at io.temporal.internal.sync.ExternalWorkflowInvocationHandler.<init>(ExternalWorkflowInvocationHandler.java:43)
	at io.temporal.internal.sync.WorkflowInternal.newExternalWorkflowStub(WorkflowInternal.java:412)
	at io.temporal.workflow.Workflow.newExternalWorkflowStub(Workflow.java:204)
```

cretz:
Arguably it shouldn't even work on the client. It'd make one wonder why a `@WorkflowInterface` is even needed ever if we can just assume it's there. In Python and .NET at least, we require the client class passed in to have this. But I can understand if Java we don't want that.

[...]

PR #2531, opened 2025-05-17
Fix newExternalWorkflowStub on a interfaces without a WorkflowMethod

Quinn-With-Two-Ns:
Fix a bug causing newExternalWorkflowStub to fail on interfaces without WorkflowMethod

closes https://github.com/temporalio/sdk-java/issues/2180

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/ExternalWorkflowInvocationHandler.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/ExternalWorkflowInvocationHandler.java
@@ -20,7 +20,7 @@ class ExternalWorkflowInvocationHandler implements InvocationHandler {
       WorkflowExecution execution,
       WorkflowOutboundCallsInterceptor workflowOutboundCallsInterceptor,
       Functions.Proc1<String> assertReadOnly) {
-    this.workflowMetadata = POJOWorkflowInterfaceMetadata.newInstance(workflowInterface);
+    this.workflowMetadata = POJOWorkflowInterfaceMetadata.newInstance(workflowInterface, false);
     this.stub =
         new ExternalWorkflowStubImpl(execution, workflowOutboundCallsInterceptor, assertReadOnly);
   }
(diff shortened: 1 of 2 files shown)


temporal-sdk/src/main/java/io/temporal/internal/sync/ExternalWorkflowInvocationHandler.java
```diff
@@ -20,7 +20,7 @@ public ExternalWorkflowInvocationHandler(
       WorkflowExecution execution,
       WorkflowOutboundCallsInterceptor workflowOutboundCallsInterceptor,
       Functions.Proc1<String> assertReadOnly) {
-    this.workflowMetadata = POJOWorkflowInterfaceMetadata.newInstance(workflowInterface);
+    this.workflowMetadata = POJOWorkflowInterfaceMetadata.newInstance(workflowInterface, false);
```
cretz, line 23: Got a bit confused on this trailing param reading:

```
if true, methods of the {@code anInterface} that are
annotated with {@link WorkflowMethod}, {@link QueryMethod}, {@link UpdateMethod}, {@link
UpdateValidatorMethod} or {@link SignalMethod} are processed like {@code current} is a
workflow interface even if it is not annotated with {@link WorkflowInterface} itself.
```

That'd make me think the default of `true` should work with non-workflow interfaces. But maybe I'm misunderstanding what was meant there.

cretz (approved):

temporal-sdk/src/main/java/io/temporal/internal/sync/ExternalWorkflowInvocationHandler.java
Quinn-With-Two-Ns: I think you are looking at the [wrong](https://github.com/temporalio/sdk-java/blob/95314a2357cf56efc1d703c657fd937bf6cbbc81/temporal-sdk/src/main/java/io/temporal/common/metadata/POJOWorkflowInterfaceMetadata.java#L131) method, the method I am calling is [here](https://github.com/temporalio/sdk-java/blob/95314a2357cf56efc1d703c657fd937bf6cbbc81/temporal-sdk/src/main/java/io/temporal/common/metadata/POJOWorkflowInterfaceMetadata.java#L108)

PR #2532, opened 2025-05-20
Upgrade otel & otel-shim to 1.50.0

robzienert:
## What was changed
Upgrading `io.opentelemetry:opentelemetry-bom` and `i.o:opentelemetry-opentracing-shim` to `1.50.0`.

## Why?
The opentracing version transitively referenced by `io.opentelemetry` is `1.26.0` which is quite old and causing issues. We could create a resolution rule internally for this, but it seems like a good thing to upgrade regardless.

## Checklist

2. How was this tested:

`./gradlew build check`


dantheman5398 (approved):
Ty!

Issue #1956, opened 2023-12-15
Worker autotuning - Pollers


Quinn-With-Two-Ns:
https://github.com/temporalio/features/issues/334

PR #2535, opened 2025-05-26
Add poller autoscaling

Quinn-With-Two-Ns:
Add poller autoscaling. This PR adds poller auto scaling support support to Java based off the same algorithm in [core](https://github.com/temporalio/sdk-core/commit/93471ac6d8bbf62839148a1bf97ef6dfd49aead0). The autoscaling poller works very differently to how the Java SDK has traditionally polled for new tasks, to reduce risk I split the autoscaling and traditional poller into two separate implementation that the SDK picks from at run time.

* Renamed the old `Poller` -> `MultiThreadedPoller`
* Created a new autoscaling poller called `AsyncPoller`
* Moved any common code to `BasePoller`

closes https://github.com/temporalio/sdk-java/issues/1456

closes https://github.com/temporalio/sdk-java/issues/1956


temporal-sdk/src/main/java/io/temporal/internal/common/GrpcUtils.java
```diff
@@ -14,4 +18,20 @@ public static boolean isChannelShutdownException(StatusRuntimeException ex) {
...
+        () -> {
+          try {
+            result.complete(listenableFuture.get());
+          } catch (ExecutionException e) {
+            result.completeExceptionally(e.getCause());
+          } catch (Exception e) {
+            result.completeExceptionally(e);
+          }
+        },
+        ForkJoinPool.commonPool());
```
Quinn-With-Two-Ns, line 34: This is also how we convert grpc futures to completable futures for the client so I copied the same approach here.

Sushisource (commented):
Overall algorithm looks right to me. Left a few other misc. comments and there's logging I assume you'll clean up for the full PR.

The only thing that looks like a possible gap to me is ensuring that sticky or nonsticky WFTs don't hog all the permits. In Core I have the balancing stuff run before permit acquisition because otherwise it's possible for, EX: all sticky pollers to get to permit acquisition first and thereby starve out the normal pollers or the inverse. I think that was covered with existing code for the non-async poller, but I'm not sure there's anything to prevent that happening with the new code. So, even though they're scaled independently, we want to balance their opportunities to obtain permits & reach the scaling decision.

temporal-sdk/src/main/java/io/temporal/internal/worker/AdjustableSemaphore.java
```diff
@@ -0,0 +1,90 @@
...
+public final class AdjustableSemaphore {
+  private final ResizeableSemaphore semaphore;
+
+  /**
+   * how many permits are allowed as governed by this semaphore. Access must be synchronized on this
+   * object.
+   */
+  private int maxPermits = 0;
+
+  /** New instances should be configured with setMaxPermits(). */
```
Sushisource: Not immediately clear to me what this docstring means

temporal-sdk/src/main/java/io/temporal/internal/worker/AdjustableSemaphore.java
```diff
@@ -0,0 +1,90 @@
...
+
+    if (delta == 0) {
+      return;
+    } else if (delta > 0) {
+      // new max is higher, so release that many permits
+      this.semaphore.release(delta);
+    } else {
+      // delta < 0.
+      // reducePermits needs a positive #, though.
+      this.semaphore.reducePermits(delta * -1);
```
Sushisource: nit: `Math.abs` I think is more obvious for this sort of thing

temporal-sdk/src/main/java/io/temporal/internal/worker/AsyncActivityPollTask.java
```diff
@@ -0,0 +1,141 @@
...
+            .setTaskQueue(TaskQueue.newBuilder().setName(taskQueue));
+    if (activitiesPerSecond > 0) {
+      pollRequest.setTaskQueueMetadata(
+          TaskQueueMetadata.newBuilder()
+              .setMaxTasksPerSecond(DoubleValue.newBuilder().setValue(activitiesPerSecond).build())
+              .build());
+    }
+
+    if (serverCapabilities.get().getBuildIdBasedVersioning()) {
+      pollRequest.setWorkerVersionCapabilities(
```
Sushisource, line 71: You'll want to set deploymentOptions as well here, it applies to activities too not just WFTs. The sync poller might need updating as well if it doesn't have it.

In fact probably wouldn't hurt to de-dupe this stuff across the sync/async pollers

temporal-sdk/src/main/java/io/temporal/internal/worker/AsyncActivityPollTask.java
```diff
@@ -0,0 +1,141 @@
...
+      response =
+          grpcContext.call(
+              () ->
+                  GrpcUtils.toCompletableFuture(
+                      service
+                          .futureStub()
+                          .withOption(METRICS_TAGS_CALL_OPTIONS_KEY, metricsScope)
+                          .pollActivityTaskQueue(pollRequest)));
+    } catch (Exception e) {
+      throw new RuntimeException(e);
```
Sushisource, line 104: This might also need to decrement poll gauge? (applies to all poll types, as well as the deployment options comment)

temporal-sdk/src/main/java/io/temporal/internal/worker/AsyncPoller.java
```diff
@@ -0,0 +1,387 @@
...
+              slotSupplier.releaseSlot(SlotReleaseReason.neverUsed(), permit);
+            }
+            if (pollerSemaphoreAcquired) {
+              pollerSemaphore.release();
+            }
+          }
+
+          if (shouldTerminate()) {
+            pollerBalancer.removePoller(asyncTaskPoller.getLabel());
+            abort = true;
```
Sushisource, line 301: Seems like this abort = true and one above could just be `return`?

temporal-sdk/src/main/java/io/temporal/internal/worker/AsyncPoller.java
```diff
@@ -0,0 +1,387 @@
...
+      log.info(
+          "Poll loop is terminated: {} - {}",
+          AsyncPoller.this.getClass().getSimpleName(),
+          asyncTaskPoller.getLabel());
+    }
+  }
+
+  /**
+   * PollQueueBalancer is used to ensure that at least one poller is running for each task. This is
+   * necessary to avoid one poller from consuming all the slots and starving other pollers.
```
Sushisource: I would just make this comment specific to sticky/nonsticky WFTs since that's the only place it matters.

"at least one poller is running for each task" sounds like the tasks are what create pollers rather than the other way around. Should probably be something like "task kind"

I'd also probably just not bother to instantiate this for non-workflow pollers

temporal-sdk/src/main/java/io/temporal/internal/worker/AsyncPoller.java
```diff
@@ -0,0 +1,387 @@
...
+   */
+  @ThreadSafe
+  class PollQueueBalancer {
+    Map<String, AtomicInteger> taskCounts = new HashMap<>();
```
Sushisource: Seems like this could store normal ints in the values since usage is always protected by the lock?

temporal-sdk/src/main/java/io/temporal/internal/worker/PollScaleReportHandle.java
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  private void updateTarget(Functions.Func1<Integer, Integer> func) {
+    Integer target = targetPollerCount;
+    Integer newTarget = func.apply(target);
+    if (newTarget < minPollerCount) {
+      newTarget = minPollerCount;
+    } else if (newTarget > maxPollerCount) {
+      newTarget = maxPollerCount;
+    }
```
Sushisource, line 80: nit: `Math.max` / `Math.min` maybe slightly more readable

Sushisource (commented):
Looking good! Nothing major.

Check out the tests I have [here](https://github.com/temporalio/sdk-core/blob/93471ac6d8bbf62839148a1bf97ef6dfd49aead0/tests/manual_tests.rs#L25). You may want to make something like that and just manually inspect the metrics around pollers to ensure everything is behaving in a reasonably sane way.

temporal-sdk/src/main/java/io/temporal/internal/worker/AsyncPoller.java
Quinn-With-Two-Ns: ASyncPoller is generic a comment about stick/nonsticky is not appropriate here

temporal-sdk/src/main/java/io/temporal/internal/worker/AsyncPoller.java
Quinn-With-Two-Ns: You can't return inside a finally block

temporal-sdk/src/main/java/io/temporal/internal/worker/AsyncActivityPollTask.java
Quinn-With-Two-Ns: currently `deploymentOptions ` is only set for the WFT if that is a bug lets open an issue to fix that. I don't want to fix that in this PR

temporal-sdk/src/main/java/io/temporal/internal/worker/AsyncPoller.java
Quinn-With-Two-Ns: Ah yep

temporal-sdk/src/main/java/io/temporal/internal/worker/PollScaleReportHandle.java
Quinn-With-Two-Ns: In my personal opinion it is harder to read since it puts a lot of logic in one line

temporal-sdk/src/main/java/io/temporal/internal/worker/AdjustableSemaphore.java
Quinn-With-Two-Ns: Ah yeah I'll fix

temporal-sdk/src/main/java/io/temporal/internal/worker/AsyncPoller.java
Quinn-With-Two-Ns: updated the docs to talk about task kind

temporal-sdk/src/main/java/io/temporal/internal/worker/AsyncPoller.java
Sushisource: Wild, I didn't realize that

temporal-sdk/src/main/java/io/temporal/internal/worker/AsyncActivityPollTask.java
Sushisource: OK, here's the item: https://github.com/temporalio/sdk-java/issues/2550

Sushisource (approved):
LGTM, worth checking out that one other spot the gauge looks like it may need to be deprecated & running the manual tests I linked to, but, nothing blocking.

temporal-sdk/src/main/java/io/temporal/internal/worker/AsyncActivityPollTask.java
Quinn-With-Two-Ns: I am not sure `call` will ever actually throw in my use case, but you are right we should decrement here regardless

PR #2540, opened 2025-05-31
Issue #1585: Added detection for oversized gRPC messages in GrpcRetryerUtils and stopped retries in this case

### LINKED ISSUE #1585: Do not auto-retry gRPC-message-size-too-large errors
## Expected Behavior

When server gets > 4MB message, it returns `ResourceExhausted` error with something like:

> grpc: received message larger than max (10248653 vs. 4194304)

Which is retried by our gRPC interceptor as are all other `ResourceExhausted`. We need to stop retrying this one in particular but still retry others. If we can prevent it from going over the wire without pre-calculating the size that would be best, but otherwise we may need to hack something to change the error type for this particular error before it gets to the retry interceptor (or inside the retry interceptor if you can). Yes that may mean a message string check which is admittedly gross.

Also, add an integration test with a workflow whose activity payload will trigger this and confirm that it's a workflow task failure (can be in sdk-features).

dmytrodanilenkov:
Resolves #1585 

Added detection for oversized gRPC messages in GrpcRetryerUtils and stopped retries in this case

### DIFF
--- a/temporal-serviceclient/src/main/java/io/temporal/internal/retryer/GrpcRetryerUtils.java
+++ b/temporal-serviceclient/src/main/java/io/temporal/internal/retryer/GrpcRetryerUtils.java
@@ -15,6 +15,9 @@ import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
 class GrpcRetryerUtils {
+
+  private static final String MESSAGE_TOO_LARGE_TEXT = "received message larger than max";
+
   /**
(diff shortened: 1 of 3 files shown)


Quinn-With-Two-Ns:
@dmytrodanilenkov Thanks for the PR, but we are already working on a different solution for this problem across our various SDKs. That should hopefully be in a upcoming release soon

PR #2542, opened 2025-06-02
Issue #2057: parsing workflow id from WorkflowExecutionStartedEventAttributes

### LINKED ISSUE #2057: Java SDK seems to not play well with Replay tests. Setting child workflow ID using Temporal library seems to be throwing a non-deterministic error
Details attached on the temporal forum support request - https://community.temporal.io/t/workflowreplayer-replayworkflowexecution-throwing-nondeterministicexception-when-reading-workflow-id-from-workflow/12040/6

Copied from the post above

========

Here is a sample repo which replicates the error.
https://github.com/gauravojha/learningTemporal/tree/master

So what we are trying to do is

(Also mentioned here, hope this helps - [learningTemporal/src/main/java/dev/ojha/learningtemporal/workflows/HelloWorldWorkflowImpl.java at master · gauravojha/learningTemporal · GitHub](https://github.com/gauravojha/learningTemporal/blob/master/src/main/java/dev/ojha/learningtemporal/workflows/HelloWorldWorkflowImpl.java#L35-L40))

* Have a parent workflow trigger a child workflow
* What we are trying to do is, reusing the parent workflows ID (this we get by using the provided Temporal utility method - `Workflow.getInfo().getWorkflowId()`, and then using this is a prefix and just append "-child" to it. So for instance if the parent workflow ID is "helloWorld", we are assigning our child workflow ID as "helloWorld-child", as this simplifies some usecases of tracing and audit for our usecases
  * When we do this, and replay the payload through the replay tests, we get an undeterministic error, I have attached the trace for this specific case at the end
  * `Failure handling event 5 of type 'EVENT_TYPE_START_CHILD_WORKFLOW_EXECUTION_INITIATED' during replay. [TMPRL1100] Command COMMAND_TYPE_START_CHILD_WORKFLOW_EXECUTION doesn't match event EVENT_TYPE_START_CHILD_WORKFLOW_EXECUTION_INITIATED with EventId=5 on check workflowId with an expected value 'workflow_id_in_replay-child' and an actual value 'HelloWorldWorkflow-child'`
* However, if we don't reuse the parent's workflow ID, and let temporal use a random ID as the workflow ID of the child, then the replay tests pass.

dmytrodanilenkov:
fixes #2057

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/common/WorkflowExecutionHistory.java
+++ b/temporal-sdk/src/main/java/io/temporal/common/WorkflowExecutionHistory.java
@@ -40,7 +40,19 @@ public final class WorkflowExecutionHistory
    * @return WorkflowExecutionHistory
    */
   public static WorkflowExecutionHistory fromJson(String serialized) {
-    return fromJson(serialized, DEFAULT_WORKFLOW_ID);
+    String protoJson = HistoryJsonUtils.historyFormatJsonToProtoJson(serialized);
+
+    JsonFormat.Parser parser = JsonFormat.parser().ignoringUnknownFields();
+    History.Builder historyBuilder = History.newBuilder();
+    try {
+      parser.merge(protoJson, historyBuilder);
+    } catch (InvalidProtocolBufferException e) {
+      throw new DataConverterException(e);
+    }
+    History history = historyBuilder.build();
+    String workflowId =
+        io.temporal.internal.common.WorkflowExecutionHistory.extractWorkflowId(history);
+    return new WorkflowExecutionHistory(history, workflowId);
   }
 
   /**
--- a/temporal-sdk/src/main/java/io/temporal/internal/common/WorkflowExecutionHistory.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/common/WorkflowExecutionHistory.java
@@ -35,7 +35,7 @@ public class WorkflowExecutionHistory {
   private final String workflowId;
 
   public WorkflowExecutionHistory(History history) {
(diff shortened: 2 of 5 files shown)


dmytrodanilenkov:
two tests fail locally, not sure how ti is connected to my changes @Quinn-With-Two-Ns can you have a quick look?
[LocalActivityAfterCancelTest](https://github.com/temporalio/sdk-java/pull/classes/io.temporal.workflow.activityTests.LocalActivityAfterCancelTest.html). [testLocalActivityAfterChildWorkflowCanceledReplay](https://github.com/temporalio/sdk-java/pull/classes/io.temporal.workflow.activityTests.LocalActivityAfterCancelTest.html#testLocalActivityAfterChildWorkflowCanceledReplay)
[ExternalWorkflowCancelTest](https://github.com/temporalio/sdk-java/pull/classes/io.temporal.workflow.cancellationTests.ExternalWorkflowCancelTest.html). [cancelExternalWorkflow](https://github.com/temporalio/sdk-java/pull/classes/io.temporal.workflow.cancellationTests.ExternalWorkflowCancelTest.html#cancelExternalWorkflow)

Quinn-With-Two-Ns:
@dmytrodanilenkov Looking at CI I see no failure https://github.com/temporalio/sdk-java/actions/runs/15449813166

dmytrodanilenkov:
I fixed test, now replay is deterministic.  Now both live history and replay now use the same parent WorkflowId for test, that's why it failed

dmytrodanilenkov:
can you run CI again? @Quinn-With-Two-Ns

Quinn-With-Two-Ns (approved):

Issue #2488, opened 2025-04-23
[java] SDK support for activity reset


drewhoskins-temporal:
See https://github.com/temporalio/features/issues/620

PR #2546, opened 2025-06-04
Add support for activity reset

Quinn-With-Two-Ns:
Add support for activity reset

closes https://github.com/temporalio/sdk-java/issues/2488

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/client/ActivityResetException.java
+++ b/temporal-sdk/src/main/java/io/temporal/client/ActivityResetException.java
@@ -4,17 +4,17 @@ import io.temporal.activity.ActivityInfo;
 import io.temporal.common.Experimental;
 
 /***
- * Indicates that the activity was paused by the user.
+ * Indicates that the activity attempt was reset by the user.
  *
  * <p>Catching this exception directly is discouraged and catching the parent class {@link ActivityCompletionException} is recommended instead.<br>
  */
 @Experimental
-public final class ActivityPausedException extends ActivityCompletionException {
-  public ActivityPausedException(ActivityInfo info) {
+public final class ActivityResetException extends ActivityCompletionException {
+  public ActivityResetException(ActivityInfo info) {
     super(info);
   }
 
-  public ActivityPausedException() {
+  public ActivityResetException() {
     super();
   }
 }
--- a/temporal-sdk/src/main/java/io/temporal/internal/activity/HeartbeatContextImpl.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/activity/HeartbeatContextImpl.java
@@ -228,6 +228,8 @@ class HeartbeatContextImpl implements HeartbeatContext {
         lastException = new ActivityCanceledException(info);
       } else if (status.getActivityPaused()) {
         lastException = new ActivityPausedException(info);
+      } else if (status.getActivityReset()) {
+        lastException = new ActivityResetException(info);
       } else {
         lastException = null;
       }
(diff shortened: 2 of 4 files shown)


temporal-sdk/src/main/java/io/temporal/client/ActivityResetException.java
cretz, line 1: With some other SDKs, we have also provided a way to activities to see _all_ states that may affect cancellation even though only one is thrown. We call it "cancellation details". I wonder if we should provide something like that in Java on the activity context (so they can see bools for cancel requested, paused, reset, etc).

Quinn-With-Two-Ns: Yes, we should there is a separate issue for [this](https://github.com/temporalio/sdk-java/issues/1005)

cretz (approved):

temporal-sdk/src/main/java/io/temporal/internal/client/external/ManualActivityCompletionClientImpl.java
```diff
@@ -190,6 +188,10 @@ public void recordHeartbeat(@Nullable Object details) throws CanceledFailure {
                 metricsScope);
         if (status.getCancelRequested()) {
           throw new ActivityCanceledException();
+        } else if (status.getActivityPaused()) {
+          throw new ActivityPausedException();
+        } else if (status.getActivityReset()) {
+          throw new ActivityResetException();
```
THardy98: I think the ordering here matters.
What happens if a client issues a pause and a reset before the next heartbeat?

In principle, I think we would want to give precedence to reset. Reset > pause would be consistent with the priority/ordering in core as well

temporal-sdk/src/main/java/io/temporal/internal/client/external/ManualActivityCompletionClientImpl.java
```diff
@@ -203,6 +205,10 @@ public void recordHeartbeat(@Nullable Object details) throws CanceledFailure {
...
```
THardy98: ditto

temporal-sdk/src/main/java/io/temporal/internal/activity/HeartbeatContextImpl.java
```diff
@@ -228,6 +228,8 @@ private void sendHeartbeatRequest(Object details) {
         lastException = new ActivityCanceledException(info);
       } else if (status.getActivityPaused()) {
         lastException = new ActivityPausedException(info);
+      } else if (status.getActivityReset()) {
+        lastException = new ActivityResetException(info);
```
THardy98, line 230: here as well

temporal-sdk/src/test/java/io/temporal/activity/ActivityResetTest.java
```diff
@@ -0,0 +1,112 @@
...
+      SDKTestWorkflowRule.newBuilder()
+          .setWorkflowTypes(TestWorkflowImpl.class)
+          .setActivityImplementations(new HeartBeatingActivityImpl())
+          .setUseExternalService(true)
+          .build();
+
+  @Test
+  public void activityReset() {
+    //    assumeTrue(
+    //        "Test Server doesn't support activity pause", SDKTestWorkflowRule.useExternalService);
```
THardy98: old comment?

temporal-sdk/src/main/java/io/temporal/internal/client/external/ManualActivityCompletionClientImpl.java
Quinn-With-Two-Ns: Yeah we should be consistent with Core I agree. I I will update this PR once it is testable

PR #2563, opened 2025-06-17
Release Java SDK v1.30.0-RC1

Quinn-With-Two-Ns:
Release Java SDK v1.30.0-RC1

### DIFF
--- a/releases/v1.30.0-RC1
+++ b/releases/v1.30.0-RC1
@@ -2,7 +2,7 @@
 
 ## Workflow Metadata **(Public Preview)**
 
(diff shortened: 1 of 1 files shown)


releases/v1.30.0-RC1
```diff
@@ -0,0 +1,81 @@
+# **💥 BREAKING CHANGES**
+
+## Workflow Metadata **(Public Preview)**
+
+The built in  `__temporal_workflow_metadata`  query now uses `io.temporal.common.converter.RawValue`  for the `return type. This should help address issues where users data converters could not properly convert the `WorkflowMetadata` protobuf type. If you are using a custom data converter that does not support `RawValue`  you will need to update it to support this type.
```
Sushisource:
```suggestion
The built in  `__temporal_workflow_metadata`  query now uses `io.temporal.common.converter.RawValue`  for the return type. This should help address issues where users data converters could not properly convert the `WorkflowMetadata` protobuf type. If you are using a custom data converter that does not support `RawValue`  you will need to update it to support this type.
```

PR #2569, opened 2025-06-23
Make sure PollerBehavior is marked as Experimental

Quinn-With-Two-Ns:
Make sure PollerBehavior is marked as Experimental

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/worker/tuning/PollerBehaviorSimpleMaximum.java
+++ b/temporal-sdk/src/main/java/io/temporal/worker/tuning/PollerBehaviorSimpleMaximum.java
@@ -1,11 +1,13 @@
 package io.temporal.worker.tuning;
 
+import io.temporal.common.Experimental;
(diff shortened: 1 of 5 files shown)


cretz (approved):

temporal-sdk/src/main/java/io/temporal/worker/tuning/PollerBehaviorSimpleMaximum.java
```diff
@@ -1,11 +1,13 @@
 package io.temporal.worker.tuning;
 
+import io.temporal.common.Experimental;
 import java.util.Objects;
 
 /**
  * A poller behavior that will attempt to poll as long as a slot is available, up to the provided
  * maximum. Cannot be less than two for workflow tasks, or one for other tasks.
  */
+@Experimental
```
Sushisource, line 10: I don't think this one really needs to be marked, but, not a big deal either way.

PR #2577, opened 2025-06-27
Update snapshot URL

Quinn-With-Two-Ns:
Update snapshot URL since we moved to sonatype central

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,9 +50,7 @@ We'd love your help in improving the Temporal Java SDK. Please review our [contr
 
 ## Snapshot release
 
-We also publish snapshot releases during SDK development often under the version `1.x.0-SNAPSHOT`. This allows users to test out new SDK features before an official SDK release.
-
-[Snapshot releases](https://central.sonatype.com/service/rest/repository/browse/maven-snapshots/io/temporal/) Find the latest snapsphot release.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -52,15 +52,15 @@ We'd love your help in improving the Temporal Java SDK. Please review our [contr
 
 We also publish snapshot releases during SDK development often under the version `1.x.0-SNAPSHOT`. This allows users to test out new SDK features before an official SDK release.
 
-[Snapshot releases](https://oss.sonatype.org/content/repositories/snapshots/io/temporal/temporal-sdk/) Find the latest snapsphot release.
+[Snapshot releases](https://central.sonatype.com/service/rest/repository/browse/maven-snapshots/io/temporal/) Find the latest snapsphot release.
```
yuandrew: I see a 404 error when trying this URL

Quinn-With-Two-Ns: yeah it looks like they disabled HTTPS browsing

Quinn-With-Two-Ns: seems they have no browsable url anymore so I had to remove this part

cretz (approved):

README.md
cretz: That's unfortunate

yuandrew (approved):

Issue #2579, opened 2025-06-30
Non-root namespace Spring dependencies fail on app startup


robzienert:
## Expected Behavior

Defining non-root namespaces in the Spring Starter should allow beans to be injected. 

## Actual Behavior

The documentation calls out that this is supported, yet the `BeanPostProcessor` that creates these beans run after bean initialization, so application startup will fail.

In order to work around this, one must use an `ObjectProvider` to decouple component autowiring from Temporal bean creation, deferring bean finalization to either `PostConstruct` or app runtime.

## Steps to Reproduce the Problem

  1. Create a non-root namespace configuration 
  1. Create a `@Component` that injects a non-root `WorkflowClient` (using `@Resource`, per docs)
  1. Start the application: The app should fail to start due to the non-root namespace WorkflowClient being unavailable.

## Specifications

  - Version: 1.29.0, 1.30.1
  - Platform: N/A

PR #2600, opened 2025-07-17
Fix non-root namespace bean injection timing issue (#2579)

expanded-for-real:
Fixes timing issue where non-root namespace beans were created too late in the Spring lifecycle as described in [2579](https://github.com/temporalio/sdk-java/issues/2579). This is fixed by modifying `NonRootBeanPostProcessor` to trigger bean creation when the root namespace template is actually ready as opposed to triggering on any bean initialization


Quinn-With-Two-Ns:
Hey thanks for the contribution, I will make sure to review this as soon as I can. Either today or tomorrow. 

Can we make sure we test the specific scenario brought up in the issue as well https://github.com/temporalio/sdk-java/issues/2579? Basically ensure we can inject a non-root bean like a non-root client, Thanks!

expanded-for-real:
Slight change to my original proposed fix - added `shouldStartSuccessfullyWithNonRootNamespaceConfiguration` for verification

Quinn-With-Two-Ns:
@expanded-for-real I don't think `shouldStartSuccessfullyWithNonRootNamespaceConfiguration ` really covers the case in question. I don't think this verifies that constructor or field injection would work since your just getting the beans from the application context. I think we should test those two cases explicitly

expanded-for-real:
@Quinn-With-Two-Ns I was able to create a test that basically mimicked the timing issue and I think it ended up being a bit more subtle than I previously thought. I believe the original `BeanPostProcessor` approach was running too late in the Spring lifecycle to enable @Resource injection, but I was able to resolve it by allowing `@Resource` beans to be registered earlier through a `ImportBeanDefinitionRegistrar` implementation. The existing tests all pass with only one modification to `MultiNamespaceTest` but this will likely require careful review.

PR #2604, opened 2025-07-21
Do not auto-retry gRPC-message-size-too-large errors

### LINKED ISSUE #1585: Do not auto-retry gRPC-message-size-too-large errors

maciejdudko:
## What was changed
- `GrpcRetryer` does not retry on gRPC error if the cause is message size limit, instead wraps the error in the new custom `GrpcMessageTooLargeException` and throws.
- `WorkflowWorker` catches `GrpcMessageTooLargeException` on reporting workflow task completion/failure, and reports task failure with this error. This does NOT prevent server-side retry of the workflow task, but it does record the error in event history for easier debugging.

## Why?

Feature request: https://github.com/temporalio/features/issues/624

## Checklist

1. Closes #1585

2. How was this tested: added tests to `GrpcSyncRetryerTest`, `GrpcAsyncRetryerTest` and `GrpcMessageTooLargeTest`.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/failure/ApplicationFailure.java
+++ b/temporal-sdk/src/main/java/io/temporal/failure/ApplicationFailure.java
@@ -236,9 +236,7 @@ public final class ApplicationFailure extends TemporalFailure {
     private Duration nextRetryDelay;
     private ApplicationErrorCategory category;
 
-    private Builder() {
-      category = ApplicationErrorCategory.UNSPECIFIED;
-    }
+    private Builder() {}
 
     private Builder(ApplicationFailure options) {
       if (options == null) {
--- a/temporal-sdk/src/main/java/io/temporal/failure/ApplicationFailure.java
+++ b/temporal-sdk/src/main/java/io/temporal/failure/ApplicationFailure.java
@@ -236,7 +236,9 @@ public final class ApplicationFailure extends TemporalFailure {
     private Duration nextRetryDelay;
     private ApplicationErrorCategory category;
 
-    private Builder() {}
+    private Builder() {
+      category = ApplicationErrorCategory.UNSPECIFIED;
+    }
 
     private Builder(ApplicationFailure options) {
       if (options == null) {
--- a/temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
@@ -9,7 +9,9 @@ import com.google.protobuf.ByteString;
 import com.uber.m3.tally.Scope;
 import com.uber.m3.tally.Stopwatch;
 import com.uber.m3.util.ImmutableMap;
+import io.grpc.StatusRuntimeException;
 import io.temporal.api.common.v1.WorkflowExecution;
+import io.temporal.api.enums.v1.QueryResultType;
 import io.temporal.api.enums.v1.TaskQueueKind;
 import io.temporal.api.enums.v1.WorkflowTaskFailedCause;
 import io.temporal.api.failure.v1.Failure;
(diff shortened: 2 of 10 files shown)


temporal-sdk/src/main/java/io/temporal/failure/ApplicationFailure.java
```diff
@@ -236,7 +236,9 @@ public static final class Builder {
     private Duration nextRetryDelay;
     private ApplicationErrorCategory category;
 
-    private Builder() {}
```
Quinn-With-Two-Ns: I already fixed this in a different PR :)

temporal-serviceclient/src/main/java/io/temporal/internal/retryer/GrpcMessageTooLargeException.java
```diff
@@ -0,0 +1,7 @@
+package io.temporal.internal.retryer;
+
+public class GrpcMessageTooLargeException extends RuntimeException {
```
Quinn-With-Two-Ns: Is this type really internal? If a user makes a client call with a payload that is to large won't that client call throw this exception?

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
```diff
@@ -394,73 +398,108 @@ public void handle(WorkflowTask task) throws Exception {
...
-              }
-            } else if (taskFailed != null) {
-              sendTaskFailed(
-                  currentTask.getTaskToken(),
-                  taskFailed.toBuilder(),
-                  result.getRequestRetryOptions(),
-                  workflowTypeScope);
-            } else if (queryCompleted != null) {
+            if (queryCompleted != null) {
               sendDirectQueryCompletedResponse(
```
Quinn-With-Two-Ns: The direct query response can also fail due to gRPC message size too large so I think that needs to be under the try as well no?

temporal-sdk/src/main/java/io/temporal/failure/ApplicationFailure.java
maciejdudko: Undid the change.

temporal-serviceclient/src/main/java/io/temporal/internal/retryer/GrpcMessageTooLargeException.java
maciejdudko: As per conversation, the exception is internal, but I changed it so it derives StatusRuntimeException so the user can catch it that way.

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
maciejdudko: Done.

temporal-serviceclient/src/main/java/io/temporal/internal/retryer/GrpcMessageTooLargeException.java
```diff
@@ -0,0 +1,34 @@
...
+        && (status.getDescription().startsWith("grpc: received message larger than max")
+            || status
+                .getDescription()
+                .startsWith("grpc: message after decompression larger than max")
+            || status
+                .getDescription()
+                .startsWith("grpc: received message after decompression larger than max"))) {
+      return new GrpcMessageTooLargeException(status.withCause(exception), exception.getTrailers());
+    } else {
+      return null;
```
Quinn-With-Two-Ns, line 31: nit: could just change the return type and return the original `StatusRuntimeException`. Since all the callers just do that as well.

maciejdudko: The logic is different depending on whether the wrap was successful or not - `GrpcMessageTooLargeException` is sent as a failure to the server, other `StatusRuntimeException` is just rethrown. Technically it could be `instanceof` check, but I think null check is cleaner.

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
```diff
@@ -394,73 +400,125 @@ public void handle(WorkflowTask task) throws Exception {
...
-                        ? Optional.of(response.getWorkflowTask())
-                        : Optional.empty();
-                // TODO we don't have to do this under the runId lock
-                activitySlotsReservation.handleResponse(response);
+                Failure failure =
+                    grpcMessageTooLargeFailure(
+                        workflowExecution.getWorkflowId(),
+                        tooLargeException,
+                        "Failed to send query response");
+                RespondQueryTaskCompletedRequest.Builder queryFailedBuilder =
```
Quinn-With-Two-Ns, line 424: Sorry I missed this in the initial review, we should be failing the workflow task here

Quinn-With-Two-Ns: Talked offline, In Java we seem to always use `RespondQueryTaskCompletedRequest` for queries, even if there was something that would normally fail the workflow task.

Issue #2413, opened 2025-02-11
Adding a new parameter type List to a workflow input causes the data converter to fail


antmendoza:
## Expected Behavior
Adding new parameters to running/closed workflows should be safe for the data converter and converted to null during replay

#### Adding a new parameter of type List to a workflow input causes the data converter to fail.

From: 
```
        @WorkflowMethod
        String execute(String name);
```
to :

```
        @WorkflowMethod
        String execute(String name, List<String> names);
```

throws during replay 

```
Caused by: java.lang.ClassCastException: class sun.reflect.generics.reflectiveObjects.ParameterizedTypeImpl cannot be cast to class java.lang.Class (sun.reflect.generics.reflectiveObjects.ParameterizedTypeImpl and java.lang.Class are in module java.base of loader 'bootstrap')
	at io.temporal.common.converter.DataConverter.fromPayloads(DataConverter.java:162)
	at io.temporal.internal.sync.POJOWorkflowImplementationFactory$POJOWorkflowImplementation.execute(POJOWorkflowImplementationFactory.java:349)
	at io.temporal.internal.sync.WorkflowExecutionHandler.runWorkflowMethod(WorkflowExecutionHandler.java:71)
```

#### Adding a new parameter of type String or Custom object works as expected.

From: 
```
        @WorkflowMethod
        String execute(String name);
```
to :

```
        @WorkflowMethod
        String execute(String name, String names);

        @WorkflowMethod
        String execute(String name, MyObject names);
```

 

## Steps to Reproduce the Problem

I have attached a project that reproduces the issue 

[Archive.zip](https://github.com/user-attachments/files/18751454/Archive.zip)

## Specifications

  - Version: SDK 1.27.1
  - Platform:

antmendoza:
> I believe adding a parameter is a backwards incompatible operation and is why we encourage using objects where fields can be added instead.

Agreed Chad. 
I can update the request to have the data converter behaving consistently, regardless of the parameter type (meaning it should either always fail or never fail).

cretz:
Not following, does the JAva converter not behave consistently today when there is more parameters in the signature than in the input?

cretz:
Oh, I see now, we do allow extra non-generic parameters, we just don't allow generic parameters due to a bug. Disregard my comments, we should fix it to be consistent I agree.

antmendoza:
thanks @cretz

PR #2605, opened 2025-07-22
 Add support for parameterized and generic types in DataConverter (#2413)

expanded-for-real:
Proposal to resolve https://github.com/temporalio/sdk-java/issues/2413

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/common/converter/DataConverter.java
+++ b/temporal-sdk/src/main/java/io/temporal/common/converter/DataConverter.java
@@ -217,7 +217,7 @@ public interface DataConverter {
   }
 
   /**
-   * Extract the raw Class from a Type, handling both regular classes and parameterized types.
+   * Extract the raw Class from a Type; handles both regular classes and parameterized types.
    *
    * @param type the Type to extract from (could be Class or ParameterizedType)
    * @return the raw Class for the type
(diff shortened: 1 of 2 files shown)


temporal-sdk/src/main/java/io/temporal/common/converter/DataConverter.java
```diff
@@ -214,4 +215,20 @@ static Object[] arrayFromPayloads(
...
   }
+
+  /**
+   * Extract the raw Class from a Type; handles both regular classes and parameterized types.
+   *
+   * @param type the Type to extract from (could be Class or ParameterizedType)
+   * @return the raw Class for the type
+   */
+  static Class<?> getRawClass(Type type) {
+    if (type instanceof Class) {
```
Quinn-With-Two-Ns, line 229: Do we need to handle any other `type`? I might just need to test but isn't `GenericArrayType` possible as well?

expanded-for-real:
@Quinn-With-Two-Ns added specific test you mentioned to `GenericParametersWorkflowTest` (seemed like the right place to add it), let me know if that looks correct.

temporal-sdk/src/main/java/io/temporal/common/converter/DataConverter.java
expanded-for-real: Good catch regarding `GenericArrayType`  - I've added support for that as well now.

temporal-sdk/src/main/java/io/temporal/common/converter/DataConverter.java
```diff
@@ -214,4 +217,25 @@ static Object[] arrayFromPayloads(
...
   }
+
+  /**
+   * Extract the raw Class from a Type; handles regular classes, parameterized types, and generic
+   * array types.
+   *
+   * @param type the Type to extract from (could be Class, ParameterizedType, or GenericArrayType)
+   * @return the raw Class for the type
+   */
+  static Class<?> getRawClass(Type type) {
```
Quinn-With-Two-Ns, line 228: Also don't shouldn't expose this on the data converter interface, this is an implementation detail and should not be part of the public API

PR #2607, opened 2025-07-30
Fix using wrong config option for resource controller

vanhtuan0409:
## What was changed

Correct PID controller config for Memory usage. Currently, pid controller for memory is using the config from CPU options

## Why?

Resource base tuner currently running with wrong configuration

## Checklist

1. Closes 

2. How was this tested: Compile test

3. Any docs updates needed? No


Sushisource (approved):
Thank you for this! Pretty sad little mistake.

PR #2609, opened 2025-07-31
Ensure namespace exists on worker startup

### LINKED ISSUE #2060: Call describe namespace on worker start

See https://github.com/temporalio/features/issues/477

yuandrew:
## What was changed

Use `describeNamespace` on worker start

## Why?

to make sure user has access to the namespace and that it exists

## Checklist

1. Closes #2060 

2. How was this tested:
Added new test

3. Any docs updates needed?

### DIFF
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -118,7 +118,7 @@ jobs:
           RH_PASSWORD: ${{ secrets.RH_PASSWORD }}
 
       - name: Publish
-        run: ./gradlew publishToSonatype
+        run: ./gradlew publishToSonatype closeSonatypeStagingRepository
 
   build_native_images:
     name: Build native test server


Quinn-With-Two-Ns (approved):

.github/workflows/ci.yml
```diff
@@ -32,7 +32,7 @@ jobs:
     - name: Run unit tests
       env:
         USER: unittest
-        USE_DOCKER_SERVICE: false
+#        USE_DOCKER_SERVICE: false
```
cretz: Is this change intentional?

yuandrew: I'm trying to figure out why CI fails, but locally tests pass. Some reason `autoconfigure/WorkerVersioningTest` is running in CI, but doesn't run locally. I'm trying to see if it could be some CI configuration that's causing this difference in behavior

PR #2622, opened 2025-08-13
Added retry options to ActivityInfo. Added ActivityInfo tests.

### LINKED ISSUE #2481: Show retry policy on activity info

See https://github.com/temporalio/features/issues/615

maciejdudko:
## What was changed

- Added retry options to ActivityInfo.
- Added ActivityInfo tests that check if it's populated properly.

## Why?

Feature request: https://github.com/temporalio/features/issues/615

## Checklist

1. Closes #2481 

2. How was this tested:
New test suite added.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/activity/ActivityInfo.java
+++ b/temporal-sdk/src/main/java/io/temporal/activity/ActivityInfo.java
@@ -3,6 +3,7 @@ package io.temporal.activity;
 import io.temporal.api.common.v1.Payloads;
 import io.temporal.common.Experimental;
 import io.temporal.common.Priority;
+import io.temporal.common.RetryOptions;
 import java.time.Duration;
 import java.util.Optional;
 import javax.annotation.Nonnull;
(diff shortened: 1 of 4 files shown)


temporal-sdk/src/main/java/io/temporal/activity/ActivityInfo.java
```diff
@@ -131,4 +132,9 @@ public interface ActivityInfo {
   @Experimental
   @Nonnull
   Priority getPriority();
+
+  /**
+   * @return Retry options for the Activity Execution.
+   */
+  RetryOptions getRetryOptions();
```
Quinn-With-Two-Ns, line 139: What is the behaviour on old servers that don't have this issue https://github.com/temporalio/temporal/issues/6502 resolved?

maciejdudko: Default retry options are returned (from `RetryOptions.Builder.validateBuildWithDefaults`) which would look valid but wouldn't be the real retry policy for the activity. I haven't thought of that case; I guess explicit null would be better?

Quinn-With-Two-Ns: So for other fields we returned the default, so I think that is fine here. We should be consistent with what is already here. If we were doing the SDK from scratch I would maybe have returned `null`.

Quinn-With-Two-Ns (approved):

Quinn-With-Two-Ns (approved):

Issue #2644, opened 2025-08-28
`workflow_task_queue_poll_succeed` is not emitted when poller autoscaling is on


Quinn-With-Two-Ns:
## Expected Behavior

`workflow_task_queue_poll_succeed` is emitted when poller autoscaling is on

## Actual Behavior

 `workflow_task_queue_poll_succeed` is not emitted when poller autoscaling is on

PR #2645, opened 2025-08-28
Make sure workflow_task_queue_poll_succeed is emitted

Quinn-With-Two-Ns:
Make sure workflow_task_queue_poll_succeed is emitted for poller autoscaling

closes https://github.com/temporalio/sdk-java/issues/2644


Sushisource (approved):
Nice, glad that was a simple fix

Issue #2657, opened 2025-09-12
Expose Temporal Nexus operation info to Temporal Nexus operation authors


Quinn-With-Two-Ns:
https://github.com/temporalio/features/issues/674

PR #2658, opened 2025-09-12
Add NexusInfo

Quinn-With-Two-Ns:
Add `NexusInfo` to expose Temporal specific information to a Nexus handler

closes https://github.com/temporalio/sdk-java/issues/2657

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/nexus/NexusInfo.java
+++ b/temporal-sdk/src/main/java/io/temporal/nexus/NexusInfo.java
@@ -1,0 +1,17 @@
+package io.temporal.nexus;
+
+/**
+ * Temporal information about the Nexus Operation. Use {@link NexusOperationContext#getInfo()} from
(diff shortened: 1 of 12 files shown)


temporal-sdk/src/main/java/io/temporal/nexus/NexusInfo.java
```diff
@@ -0,0 +1,17 @@
+package io.temporal.nexus;
+
+/**
+ * Temporal information about the Nexus Operation. Use {@link NexusOperationContext#getInfo()} from
+ * a Nexus Operation implementation to access.
+ */
+public interface NexusInfo {
```
cretz: Arguably this could be called `NexusOperationInfo` but meh

Quinn-With-Two-Ns: Yeah I see we called it that in dotnet, fine with me

PR #2660, opened 2025-09-15
Add support for environment config

Quinn-With-Two-Ns:
Add support for environment config to the Java SDK

https://docs.temporal.io/develop/environment-configuration


temporal-envconfig/build.gradle
```diff
@@ -0,0 +1,13 @@
+description = '''Temporal Java SDK Environment Config Module'''
```
Quinn-With-Two-Ns, line 1: In terms of naming it would probably more consistent internally in the Java SDK to call this module `temporal-enviorment-configuration`, but that isn't consistent with the Go SDK

temporal-envconfig/src/main/java/io/temporal/envconfig/ClientConfig.java
```diff
@@ -0,0 +1,92 @@
...
+      }
+      if (file == null || file.isEmpty()) {
+        file = getDefaultConfigFilePath();
+      }
+      ClientConfigToml.TomlClientConfig result = reader.readValue(new File(file));
+      return new ClientConfig(ClientConfigToml.getClientProfiles(result));
+    }
+  }
+
+  public ClientConfig(Map<String, ClientConfigProfile> profiles) {
```
Quinn-With-Two-Ns: May want to use a builder here for consistency

temporal-envconfig/src/main/java/io/temporal/envconfig/ClientConfig.java
```diff
@@ -0,0 +1,92 @@
...
+
+import com.fasterxml.jackson.databind.ObjectReader;
+import com.fasterxml.jackson.dataformat.toml.TomlMapper;
+import io.temporal.common.Experimental;
+import java.io.*;
+import java.util.Map;
+
+/** ClientConfig represents a client config file. */
+@Experimental
+public class ClientConfig {
```
cretz, line 14: Hrmm, ideally we could have one static call on client config in the SDKs that gives you what you need to connect a client. With Java this is admittedly harder because 1) stubs are separate from client, and 2) Java doesn't support a native tuple to return two things at once.

I wonder if there is any creative approach we can think of here.

cretz, line 14: I know it sounds pedantic, but can we also expose static calls for getting the default profile stuff to connect. Specifically, can we have:

```java
public static WorkflowServiceStubsOptions loadWorkflowServiceStubsOptions()
```
and
```java
public static WorkflowClientOptions loadWorkflowClientOptions()
```

on this class? Or maybe named `loadDefaultWorkflowServiceStubsOptions` or something? With the other SDKs, we are putting a static call on the highest level class/place we can to be able to load what is needed from the default profile to make client connections. I would guess these would be the calls used by every SDK sample.

temporal-envconfig/src/main/java/io/temporal/envconfig/ClientConfigToml.java
cretz, line 1: One feature we were adding to other SDKs is the ability to be able to read/write TOML. For languages that didn't want a dependency, we would just give an object that could serialize as such, but Java already has the dependency.

temporal-envconfig/src/main/java/io/temporal/envconfig/ClientConfigTLS.java
```diff
@@ -0,0 +1,184 @@
...
+  public static Builder newBuilder() {
+    return new Builder();
+  }
+
+  /** Create a builder from an existing {@link ClientConfigTLS}. */
+  public static Builder newBuilder(ClientConfigTLS config) {
+    return new Builder(config);
+  }
+
+  private final boolean disabled;
```
cretz: IMO we should match the TOML representation exactly, which IMO I think means this is a `Boolean` because it can be unset and unset has different behavior than just false (default is TLS disabled by default unless API key is present). This is mostly thinking about the ability to convert it back to TOML, but maybe that's not a concern.

temporal-envconfig/build.gradle
cretz: :+1: The `envconfig` shortened term for the modules/packages/namespaces is kinda nice

temporal-envconfig/src/main/java/io/temporal/envconfig/ClientConfig.java
Quinn-With-Two-Ns: The only thing I could think of was have a method that gives you a full client that takes a few options customizers to set additional options on the service stub options and client options.

temporal-envconfig/src/main/java/io/temporal/envconfig/ClientConfigTLS.java
Quinn-With-Two-Ns: does this also apply to `disableHostVerification` as well?

temporal-envconfig/src/main/java/io/temporal/envconfig/ClientConfig.java
Quinn-With-Two-Ns: So we wan't a short cut to call `ClientConfigProfile.load().toWorkflowServiceStubsOptions()`?

temporal-envconfig/src/main/java/io/temporal/envconfig/ClientConfigToml.java
Quinn-With-Two-Ns: So the Go SDK does not have the feature, does any SDK support the ability to read/write the TOML file? Is any user asking for this?

temporal-envconfig/src/main/java/io/temporal/envconfig/ClientConfig.java
cretz: Yeah, we have been adding this to other SDKs. But it's not a big deal if we don't do it.

cretz: I think what you have here is good enough

temporal-envconfig/src/main/java/io/temporal/envconfig/ClientConfigToml.java
cretz: > So the Go SDK does not have the featur

`ClientConfig` does have `FromTOML` and `ToTOML`

> does any SDK support the ability to read/write the TOML file? Is any user asking for this?

Every SDK supports at least the raw TOML structure (even if it's maps/dictionary structures of generic objects). We don't necessarily do the reading/writing of TOML to an actual file for the users, but the ability for users to parse/write our config format in SDKs was something we had originally stated was a goal. Doesn't have to be TOML or even typesafe structures, just any way a user can parse/write it.

temporal-envconfig/src/main/java/io/temporal/envconfig/ClientConfigTLS.java
cretz: Hrmm, up to you since it's only ever true/false (absent is _always_ false). My concern was mostly about disabled's absence having a special meaning, granted in Go we don't use a nullable bool either, hrmm. No strong opinion here. Core-based SDKs do not have a disable-host-verification option.

temporal-envconfig/src/main/java/io/temporal/envconfig/ClientConfigToml.java
Quinn-With-Two-Ns: Okay so we want Okay I added a `fromToml` and `toToml` equivalent

temporal-envconfig/build.gradle
```diff
@@ -0,0 +1,13 @@
...
+
+dependencies {
+    // this module shouldn't carry temporal-sdk with it, especially for situations when users may be using a shaded artifact
+    compileOnly project(':temporal-sdk')
+
+    implementation 'com.fasterxml.jackson.dataformat:jackson-dataformat-toml:2.15.4'
```
cretz:
```suggestion
    implementation "com.fasterxml.jackson.dataformat:jackson-dataformat-toml:${jacksonVersion}"
```

If Jackson is releasing versions together, I wonder if we should try to keep the same Jackson versions for the different JARs across our deps in case they themselves reference common Jackson libraries of a version.

cretz (approved):
LGTM (especially since it's experimental atm). Added small suggestion concerning Jackson version.

temporal-envconfig/build.gradle
Quinn-With-Two-Ns: let me check, not sure if they do since it is a separate project

PR #2669, opened 2025-09-24
Align workflow_task_schedule_to_start_latency

Quinn-With-Two-Ns:
Align how `workflow_task_schedule_to_start_latency` is measured across the Async and Multithreaded Poller

https://github.com/temporalio/sdk-java/blob/4631b16cd251afcf513e174715486befc4ad0f6b/temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowPollTask.java#L200

Now both pollers will use the time from the task to calculate `workflow_task_schedule_to_start_latency`, this also aligns with Core and Go

---

> [!NOTE]
> Compute schedule-to-start latency for async workflow and activity pollers using task-provided timestamps instead of local time.
> 
> - **Metrics**:
>   - **Workflow poller**: `WORKFLOW_TASK_SCHEDULE_TO_START_LATENCY` now computed from task timestamps (`r.getStartedTime()` → `r.getScheduledTime()`).
>   - **Activity poller**: `ACTIVITY_SCHEDULE_TO_START_LATENCY` now computed from task timestamps (`r.getStartedTime()` → `r.getCurrentAttemptScheduledTime()`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 402f814b969b528c35a9d27ae17309b25431624a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/worker/AsyncActivityPollTask.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/worker/AsyncActivityPollTask.java
@@ -3,7 +3,6 @@ package io.temporal.internal.worker;
 import static io.temporal.serviceclient.MetricsTag.METRICS_TAGS_CALL_OPTIONS_KEY;
 
 import com.google.protobuf.DoubleValue;
-import com.google.protobuf.Timestamp;
 import com.uber.m3.tally.Scope;
 import io.grpc.Context;
 import io.temporal.api.common.v1.WorkerVersionCapabilities;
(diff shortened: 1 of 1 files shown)


temporal-sdk/src/main/java/io/temporal/internal/worker/AsyncWorkflowPollTask.java
```diff
@@ -154,10 +153,9 @@ public CompletableFuture<WorkflowTask> poll(SlotPermit permit)
               pollerMetricScope
                   .counter(MetricsType.WORKFLOW_TASK_QUEUE_POLL_SUCCEED_COUNTER)
                   .inc(1);
-              Timestamp startedTime = ProtobufTimeUtils.getCurrentProtoTime();
               pollerMetricScope
                   .timer(MetricsType.WORKFLOW_TASK_SCHEDULE_TO_START_LATENCY)
-                  .record(ProtobufTimeUtils.toM3Duration(startedTime, r.getScheduledTime()));
+                  .record(ProtobufTimeUtils.toM3Duration(r.getStartedTime(), r.getScheduledTime()));
```
cretz, line 158: Any idea of what the practical time difference could be for users that may be monitoring this metric? I guess we can just call it out in release notes

cretz, line 158: I see this same logic for activity, do we want to update that one to use the activity poll response start time too? I also saw this logic with Nexus, but no start time on its response (I also confirmed the activity and workflow started time fields on the response go as far back as Temporal, so we're safe for old server versions)

Quinn-With-Two-Ns: ah good catch, fixed

Quinn-With-Two-Ns: clock skew between the Server and Worker, maybe some unaccounted for network time. The main issue is the async poller was measuring it differently then everywhere else which is a bug.

PR #2683, opened 2025-10-10
Remove experimental status from worker tuners

Sushisource:
## Summary
- remove experimental annotations from worker tuner interfaces, slot info types, and resource-based tuning classes
- allow WorkerOptions to accept WorkerTuner without the experimental guard
- drop the experimental tag from the worker task slots used metric
------
https://chatgpt.com/codex/tasks/task_b_68e97eccfc2883318321fef725cebb09

### DIFF
--- a/temporal-shaded/build.gradle
+++ b/temporal-shaded/build.gradle
@@ -57,6 +57,8 @@ dependencies {
     shadow "io.grpc:grpc-stub:$grpcVersion"
     shadow "io.grpc:grpc-netty-shaded:$grpcVersion"
     shadow "io.grpc:grpc-services:$grpcVersion"
+    shadow "io.grpc:grpc-inprocess:$grpcVersion"
+    shadow "io.grpc:grpc-util:$grpcVersion"
     shadow "com.google.api.grpc:proto-google-common-protos:2.9.0"
 }
 
--- a/temporal-spring-boot-autoconfigure/src/main/java/io/temporal/spring/boot/autoconfigure/properties/WorkerProperties.java
+++ b/temporal-spring-boot-autoconfigure/src/main/java/io/temporal/spring/boot/autoconfigure/properties/WorkerProperties.java
@@ -385,6 +385,8 @@ public class WorkerProperties {
     private final @Nullable String deploymentVersion;
     private final @Nullable Boolean useVersioning;
     private final @Nullable VersioningBehavior defaultVersioningBehavior;
+    private final @Nullable String deploymentName;
+    private final @Nullable String buildId;
 
     /**
      * Sets options that will be passed to {@link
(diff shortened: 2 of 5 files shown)


temporal-sdk/src/main/java/io/temporal/worker/tuning/PollerBehavior.java
```diff
@@ -10,5 +8,4 @@
  * PollerBehaviorSimpleMaximum}. For all intents and purpose this interface should be considered
  * sealed.
  */
-@Experimental
 public interface PollerBehavior {}
```
maciejdudko, line 14: Is it intentional to also move poller behavior out of experimental too? If so, it's still marked experimental inside `WorkerOptions`.

Sushisource: No, this wasn't intentional, good catch. Though, we should probably un-mark it too, but, I'll do that separately for all langs.

temporal-sdk/src/main/java/io/temporal/worker/tuning/PollerBehaviorAutoscaling.java
```diff
@@ -11,7 +10,6 @@
  * <p>If the server does not support autoscaling, then the number of pollers will stay at the
  * initial number of pollers.
  */
-@Experimental
 public final class PollerBehaviorAutoscaling implements PollerBehavior {
```
maciejdudko, line 15: This needs to be restored.

temporal-sdk/src/main/java/io/temporal/worker/tuning/PollerBehaviorSimpleMaximum.java
```diff
@@ -1,13 +1,11 @@
...
 
-import io.temporal.common.Experimental;
 import java.util.Objects;
 
 /**
  * A poller behavior that will attempt to poll as long as a slot is available, up to the provided
  * maximum. Cannot be less than two for workflow tasks, or one for other tasks.
  */
-@Experimental
 public class PollerBehaviorSimpleMaximum implements PollerBehavior {
```
maciejdudko, line 9: This needs to be restored.

Sushisource: This one I'll just keep the removal right now honestly, it was never really experimental since this just delegates to the old behavior.

temporal-sdk/src/main/java/io/temporal/worker/tuning/PollerBehaviorAutoscaling.java
Sushisource: Thanks. 🤦

maciejdudko (approved):

PR #2686, opened 2025-10-14
Add grpc-inprocess, grpc-util to shaded dependencies

FelipeCortez:
## What was changed
This PR adds two dependencies to `temporal-shaded` that are necessary for writing test suites.

## Why?
Tests rely on `InProcessServerBuilder`: 

https://github.com/temporalio/sdk-java/blob/89c427eb3ae76e47042765fb522c36e532b311f3/temporal-serviceclient/src/main/java/io/temporal/internal/testservice/InProcessGRPCServer.java#L5-L6

The ServerBuilder class is part of the `io.grpc:grpc-inprocess` [artifact](https://mvnrepository.com/artifact/io.grpc/grpc-inprocess), which was not currently covered by the dependencies declared for `temporal-shaded`, causing `java.lang.ClassNotFoundException: io.temporal.shaded.io.grpc.inprocess.InProcessServerBuilder` errors at (test executing) runtime.

We also depend on `SecretRoundRobinLoadBalancerProvider`:

https://github.com/temporalio/sdk-java/blob/89c427eb3ae76e47042765fb522c36e532b311f3/temporal-serviceclient/src/main/resources/META-INF/native-image/io/temporal/temporal-serviceclient/reflect-config.json#L338-L339

 I'm not exactly sure on how the reflection mechanism works, but I got similar `ClassNotFoundException`s without declaring a dependency on `grpc-util`.

## How was this tested
I have a separate project that uses the Temporal Java SDK and contains a test suite. No tests passed using the shaded SDK prior to this change. I applied the changes that are commited to this PR, rebuilt `temporal-shaded` with `cd temporal-shaded && gradle build` and installed the updated `temporal-shaded` dependency to my local maven repository by updating the plugins section in `temporal-shaded/build`:

```diff
plugins {
    id 'com.gradleup.shadow' version '8.3.3'
+   id 'maven-publish'
}
   
```

and then running `cd temporal-shaded && gradle publishToMavenLocal`. I changed my project to use the new artifact and all tests passed.

### DIFF
--- a/temporal-spring-boot-autoconfigure/src/main/java/io/temporal/spring/boot/autoconfigure/properties/WorkerProperties.java
+++ b/temporal-spring-boot-autoconfigure/src/main/java/io/temporal/spring/boot/autoconfigure/properties/WorkerProperties.java
@@ -385,6 +385,8 @@ public class WorkerProperties {
     private final @Nullable String deploymentVersion;
     private final @Nullable Boolean useVersioning;
     private final @Nullable VersioningBehavior defaultVersioningBehavior;
+    private final @Nullable String deploymentName;
+    private final @Nullable String buildId;
 
     /**
      * Sets options that will be passed to {@link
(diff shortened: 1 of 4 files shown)


maciejdudko (approved):
Looks good, LGTM!

PR #2692, opened 2025-10-19
Add support for FirstExecutionRunId to cancel and terminate

Quinn-With-Two-Ns:
Add support for FirstExecutionRunId to cancel and terminate to the Time skipping test server.

---

> [!NOTE]
> Enable cancel and terminate operations to target a workflow chain via `firstExecutionRunId`, not just the current `runId`.
> 
> - **Server (test-server)**:
>   - Add lookup/index by `firstExecutionRunId` using `WorkflowChainId` (namespace, `workflowId`, `firstExecutionRunId`).
>   - On start, index executions in `executionsByFirstExecutionRunId`; expose `getFirstExecutionRunId()` on mutable state.
>   - Update `requestCancelWorkflowExecution` and `terminateWorkflowExecution` to resolve executions via `firstExecutionRunId` when provided.
> - **Tests**:
>   - Add `FirstExecutionRunIdSupportTest` covering cancel/terminate by `firstExecutionRunId` and after continue-as-new.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 78562128f96dd3f6ea3add7afe96ee09aa79b454. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
+++ b/temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
@@ -280,7 +280,7 @@ public final class TestWorkflowService extends WorkflowServiceGrpc.WorkflowServi
             : new WorkflowChainId(namespace, execution.getWorkflowId(), firstExecutionRunId);
 
     if (workflowChainId != null) {
-      return getMutableState(workflowChainId, true);
+      return getMutableState(workflowChainId, failNotExists);
     }
     return getMutableState(executionId, failNotExists);
   }


temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
```diff
@@ -205,6 +252,63 @@ private TestWorkflowMutableState getMutableState(WorkflowId workflowId, boolean
...
+    if (mutableState != null) {
+      if (workflowChainId == null) {
+        return mutableState;
+      }
+      WorkflowExecution mutableStateExecution = mutableState.getExecutionId().getExecution();
+      if (mutableStateExecution.getRunId().equals(execution.getRunId())
+          && firstExecutionRunId.equals(mutableState.getFirstExecutionRunId())) {
+        return mutableState;
+      }
+    }
```
maciejdudko: Considering that workflow executions that have `firstRunId` are added to both maps in `startWorkflowExecutionNoRunningCheckLocked`, in what circumstances will the last `if` evaluate true (line 295)? If there is an execution with a matching first run ID, I'd expect it to have already returned inside the earlier if (line 285).

temporal-test-server/src/test/java/io/temporal/testserver/functional/FirstExecutionRunIdSupportTest.java
maciejdudko, line 1: We should add negative tests for what happens when the wrong first execution ID is provided to cancel/terminate request (both when there was and there wasn't continue-as-new).

Quinn-With-Two-Ns: Done

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
Quinn-With-Two-Ns: Yeah good point, I think I overcomplicated the check. I simplified it a lot and should match how the current real server finds the correct mutable state. All the tests run on the real server and test server to validate this.

PR #2693, opened 2025-10-19
Correct Preconditions error message in WorkerFactory

phlego:
## What was changed

This PR correct the error message for the `start` method of `WorkerFactory`.
The fix correctly shows `State.Initial` and `State.Started` as the expected states.
The PR also remove a unused variable and several unused imports.

## Why?

The original code had `State.Initial.name()` duplicated in the error message, which was likely a copy-paste error.

## Checklist

1. Closes 
N/A

2. How was this tested:
N/A

3. Any docs updates needed?
No


phlego:
Done @Quinn-With-Two-Ns 
Thanks a lot for your support 🙏

PR #2694, opened 2025-10-20
Fix activity completion when thread interrupted flag is set

### LINKED ISSUE #731: Activities that return with interrupted flag should be successfully reported as Completed
## Expected Behavior

If Activity 
- never noticed and interrupted flag or
-  caught an interrupted exception, restored the interrupted flag, and continued to run or returned a result
should be reported as Completed with its returned value.

## Actual Behavior

According to https://github.com/temporalio/sdk-java/pull/722#discussion_r707879325
if activity returns with an interrupted flag turned on, subsequent gRPC calls fail and the Activity result never gets reported.

## Proposed solution

- We should write a stable reproducible test close to a test that was changed in https://github.com/temporalio/sdk-java/pull/722
- When Activity returns with interrupted flag = "true", we should hide the flag, report the result of Activity execution to the server, and return the flag back after that allowing the thread pool to stop.

sojanmathew:
Activities can **now successfully report completion** even when they have the 'interrupted' flag set.

Without this fix,  activity results are not successfully reported to server when they have the 'interrupted' flag set( gRPC calls in ActivityWorker.sendReply() would fail with **CancellationException** ).

The fix clears the 'interrupted' flag before making gRPC calls in ActivityWorker.sendReply() and restores it afterward, ensuring:
  - Activity results are successfully reported to the server
  - Thread interruption semantics are preserved for worker shutdown
  - All activity completion scenarios work (success/failure/cancellation)

  Fixes: https://github.com/temporalio/sdk-java/issues/731
  Related: https://github.com/temporalio/sdk-java/pull/722

## What was changed
Fix the bug where gRPC calls fail when the thread's 'interrupted' flag is set, hence unable to report their completion status to the server.

## Why?
The fix ensures that activities  report their completion status to the Temporal server, while  maintaining proper thread interruption semantics for the worker shutdown process.                            
 According to the issue description and  PR [#722](https://github.com/temporalio/sdk-java/pull/722), the issue was that when an activity has  'interrupted' flag set:                                                                                  
 i.e,                                                                                                                       
   1 Catches an InterruptedException                                                                                    
   2 Restores the 'interrupted' flag (Thread.currentThread().interrupt())                                                 
   3 Continues to run and returns a result                                                                              
                                                                                                                        
  The subsequent gRPC calls in ActivityWorker.sendReply() would fail with CancellationException ( gRPC request was   
  cancelled because the 'interrupted' flag was still set), hence activity result were not reported to the server.     

The **CancellationException** is thrown in [GrpcSyncRetryer](https://github.com/temporalio/sdk-java/blob/master/temporal-serviceclient/src/main/java/io/temporal/internal/retryer/GrpcSyncRetryer.java#L52-L54) when reporting activity completion results:

```
      
    Thread.currentThread().interrupt();                                                                               
    throw new CancellationException();                                                                                
  }
```

**Solution**                                                                                                          
Modify ActivityWorker.sendReply() method to:                                        
                                                                                                                        
   1 Check the 'interrupted' flag before making gRPC calls using Thread.interrupted() (which also clears the flag)        
   2 Temporarily clear the flag to allow gRPC calls to succeed                                                          
   3 Make the gRPC call to report the activity result                                                                   
   4 Restore the 'interrupted' flag in a finally block if it was originally set                                           

@Spikhalskiy 
                                                                                                                       
## Checklist
1. Closes https://github.com/temporalio/sdk-java/issues/731

2. How was this tested:
                                                                                                                        
   1 Created and ran a comprehensive test that reproduces the exact scenario described in the bug report                
   2 Verified the test failed before the fix (confirming the bug existed)                                               
   3 Verified the test passes after the fix (confirming the bug is resolved)                                            
   4 Ran existing shutdown tests to ensure no regressions were introduced                                               

3. Any docs updates needed?
No


temporal-sdk/src/main/java/io/temporal/internal/worker/ActivityWorker.java
```diff
@@ -337,6 +337,32 @@ public Throwable wrapFailure(ActivityTask t, Throwable failure) {
...
+     * Activities that return with the 'interrupted' flag set, were unable to report their completion to the server.
+     * We need to clear 'interrupted' flag to allow gRPC calls to succeed,then restore it after reporting completion,
+     * to ensure gRPC calls succeed even when the thread has been interrupted.
+     *
+     * @param grpcCall the gRPC call to execute
+     * @see <a href="https://github.com/temporalio/sdk-java/issues/731">GitHub Issue #731</a>
+     */
+    private void executeGrpcCallWithInterruptHandling(Runnable grpcCall) {
+      // Check if the current thread is interrupted before making gRPC calls
+      // If it is, we need to clear the flag to allow gRPC calls to succeed,then restore it after reporting.
```
Quinn-With-Two-Ns, line 389: What part of the SDK relies on it being restored?

sojanmathew: > What part of the SDK relies on it being restored?
@Quinn-With-Two-Ns 
 The `CancellationException` is thrown in [GrpcSyncRetryer](https://github.com/temporalio/sdk-java/blob/master/temporal-serviceclient/src/main/java/io/temporal/internal/retryer/GrpcSyncRetryer.java#L52-L54) when reporting activity completion results:                        
```
      
    Thread.currentThread().interrupt();                                                                               
    throw new CancellationException();                                                                                
  }      
```

sojanmathew: @Quinn-With-Two-Ns Are you good with above change, if so, could you please approve PR?

Quinn-With-Two-Ns: @sojanmathew no I still don't understand the logic here, why is the flag being restored?

sojanmathew: @Quinn-With-Two-Ns 
- Why Restoration is Critical:                                                                                      
**without restoration**: Activity result never gets reported (if activity returns with an interrupted flag turned on)
**with restoration** : Activity result get reported (Contract Preservation: Activity result reporting should be transparent to interrupt state)

     • Calling Code Expectations: Framework or user code may check interrupt status after completion     

Hope the [test](https://github.com/temporalio/sdk-java/blob/e9584e1eab37efc39fcf6f16a0b84ae823e25bf2/temporal-sdk/src/test/java/io/temporal/worker/InterruptedActivityCompletionValidationTest.java)  helps to explain before / after situation. 

In summary : 
- Activity sets interrupted flag → Store original flag → Clear flag → gRPC call succeeds → Restore original flag .     
     Please let me know if it clarifies.

PR #2704, opened 2025-10-20
Set worker version on eager start requests

Sushisource:
## What was changed
Add worker deployment options to eager workflow start requests

## Why?
Needed so server has versioning info available to it when processing eager start requests

## Checklist

1. Closes 

2. How was this tested:
Added unit tests

3. Any docs updates needed?

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
@@ -10,9 +10,7 @@ import io.grpc.Deadline;
 import io.grpc.Status;
 import io.grpc.StatusRuntimeException;
 import io.temporal.api.common.v1.*;
-import io.temporal.api.deployment.v1.WorkerDeploymentOptions;
 import io.temporal.api.enums.v1.UpdateWorkflowExecutionLifecycleStage;
-import io.temporal.api.enums.v1.WorkerVersioningMode;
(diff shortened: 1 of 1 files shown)


temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
```diff
@@ -743,4 +749,22 @@ private WorkflowTaskDispatchHandle obtainDispatchHandle(WorkflowStartInput input
     }
     return eagerWorkflowTaskDispatcher.tryGetLocalDispatchHandler(input);
   }
+
+  private static WorkerDeploymentOptions toProtoDeploymentOptions(
+      io.temporal.worker.WorkerDeploymentOptions deploymentOptions) {
```
maciejdudko: This function duplicates `io.temporal.internal.worker.WorkerVersioningProtoUtils.deploymentOptionsToProto`

maciejdudko (changes requested):
Looks good, but the `toProtoDeploymentOptions` function should be removed because we already have `WorkerVersioningProtoUtils.deploymentOptionsToProto` that does the same thing.

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
Sushisource: Ah, thanks - didn't see that one

maciejdudko (approved):

PR #2709, opened 2025-10-23
Add new constructor for WorkflowStub to target correct execution

Quinn-With-Two-Ns:
Add new constructor for WorkflowStub to target correct execution. Before the PR `WorkflowStub` was very inconsistent in what workflow execution it targets, sometimes it would ignore the RunID, sometimes it wouldn't. This PR keeps the old inconsistent behaviour if a user uses the old constructor and adds a new constructor that always uses the execution the user specified. It also adds the ability to target by `FirstExecutionRunID` to target a workflow chain.

---

> [!NOTE]
> Add `WorkflowTargetOptions` and new stub APIs that correctly honor runId/firstExecutionRunId, deprecating legacy constructors and updating internals/tests accordingly.
> 
> - **Client API**:
>   - Introduces `WorkflowTargetOptions` for precise workflow targeting (`workflowId`, `runId`, `firstExecutionRunId`).
>   - Adds `newWorkflowStub(Class, WorkflowTargetOptions)` and `newUntypedWorkflowStub(WorkflowTargetOptions[, Optional<String>])`.
>   - Deprecates legacy constructors using `workflowId`/`Optional<String> runId` and `WorkflowExecution`.
> - **Targeting semantics**:
>   - `WorkflowStubImpl` gains `legacyTargeting` flag; new stubs honor provided `runId` while legacy paths follow latest run.
>   - Propagates `firstExecutionRunId` for updates, cancel, and terminate.
> - **Invocation/Interceptor layer**:
>   - `WorkflowInvocationHandler`/`WorkflowClientInternalImpl` updated to build stubs with targeting options.
>   - `RootWorkflowClientInvoker` sets `firstExecutionRunId` on cancel/terminate and update requests.
> - **Docs**: Javadoc updated to reference new APIs and clarify behavior.
> - **Tests**:
>   - Broad refactor to use `WorkflowTargetOptions`.
>   - New `WorkflowStubRespectRunIdTest` validating run-id and first-execution-run-id behavior.
>   - `SDKTestWorkflowRule` gains history helpers with `runId` variants.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit afb504449689e09ec2f4c7e43106e60d2fd76182. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/client/WorkflowStubImpl.java
+++ b/temporal-sdk/src/main/java/io/temporal/client/WorkflowStubImpl.java
@@ -371,6 +371,7 @@ class WorkflowStubImpl implements WorkflowStub {
             .setWaitForStage(waitForStage)
             .setResultClass(resultClass)
             .setResultType(resultClass)
+            .setFirstExecutionRunId(firstExecutionRunId)
             .build();
 
     return startUpdate(options, args);


temporal-sdk/src/main/java/io/temporal/client/WorkflowClient.java
```diff
@@ -158,9 +158,21 @@ static WorkflowClient newInstance(WorkflowServiceStubs service, WorkflowClientOp
...
 
+  /**
+   * Creates workflow client stub for a known execution. Use it to send signals or queries to a
+   * running workflow. Do not call methods annotated with @WorkflowMethod.
+   *
+   * @param workflowInterface interface that given workflow implements.
+   * @param workflowTargetOptions options that specify target workflow execution.
+   * @return Stub that implements workflowInterface and can be used to signal or query it.
+   */
+  <T> T newWorkflowStub(Class<T> workflowInterface, WorkflowTargetOptions workflowTargetOptions);
```
cretz, line 176: Pedantic, but I like `WorkflowStubOptions` name myself in case we ever want an option unrelated to "workflow target"

cretz (approved):
LGTM, especially knowing that no code in use today will break (though users may get a bit annoyed when these deprecation warnings pop up). Usually would leave this PR up to other Java reviewers, but thought I'd put my approval on too since I had discussions on the concerns previously.

temporal-sdk/src/main/java/io/temporal/client/WorkflowClient.java
Quinn-With-Two-Ns: Yeah my reason for not using `WorkflowStubOptions` was because these options are specifically for creating a stub that is attached to a running workflow, instead of a stub that can start a workflow

temporal-sdk/src/main/java/io/temporal/client/WorkflowStubImpl.java
```diff
@@ -589,6 +604,7 @@ private Throwable throwAsWorkflowFailureException(
 
   private void populateExecutionAfterStart(WorkflowExecution startedExecution) {
     this.startedExecution.set(startedExecution);
+    // this.firstExecutionRunId.set(startedExecution.getRunId());
```
maciejdudko: Leftover commented code. Should `firstExecutionRunId` be set here or should it stay null?

Quinn-With-Two-Ns: Should stay `null` for now until https://github.com/temporalio/temporal/issues/8537 is resolved

PR #2717, opened 2025-10-28
add nightly throughput stress

THardy98:
## What was changed
Added nightly throughput stress CI workflow


THardy98:
@Quinn-With-Two-Ns the build step sometimes takes a long time (+1hr), I think it's running tests but I don't know how to fix it (or if it needs fixing)

Quinn-With-Two-Ns:
>Hmm that's what I've done but it still seems to take forever

https://github.com/temporalio/sdk-java/actions/runs/18916658621/job/54001731839?pr=2717 it seems the build here takes only 5 min no?

Quinn-With-Two-Ns:
@THardy98 you probably also want to exclude `virtualThreadTests` when doing the build step to save more time

Issue #2178, opened 2024-08-09
Carry over search attributes on continue-as-new if none are explicitly specified


Sushisource:
This will match the behavior in other SDKs. Currently it looks like nothing is carried over.

PR #2731, opened 2025-11-19
💥  Carry over search attributes on continue-as-new if none are explicitly set

Quinn-With-Two-Ns:
Carry over search attributes on continue-as-new if none are explicitly set, this is the documented behaviour but previously the Java SDK was not actually carrying over search attributes.

closes https://github.com/temporalio/sdk-java/issues/2178

---

> [!NOTE]
> Carry over existing search attributes across continue-as-new and update test server and tests to support and verify it.
> 
> - **SDK (workflow execution)**:
>   - `SyncWorkflowContext.continueAsNew(...)`: if no `searchAttributes`/`typedSearchAttributes` provided, carry over existing `SearchAttributes` from `replayContext`.
>   - Copy `RetryPolicy` from `replayContext` when `ContinueAsNewOptions` is null.
> - **Test Server**:
>   - `StateMachines`: include `searchAttributes` in `WorkflowExecutionContinuedAsNewEventAttributes` when present.
>   - `TestWorkflowService.continueAsNew(...)`: forward continued run `searchAttributes` into `StartWorkflowExecutionRequest`.
> - **Tests**:
>   - `ContinueAsNewTest`: add typed search attributes to initial options; assert they are carried over on first continue-as-new and can be overridden on subsequent runs.
> - **Docs**:
>   - Minor Javadoc grammar tweak in `ContinueAsNewOptions`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a2acbff02b247276cc216e6793712f798fdc6a03. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/temporal-sdk/src/test/java/io/temporal/workflow/ContinueAsNewTest.java
+++ b/temporal-sdk/src/test/java/io/temporal/workflow/ContinueAsNewTest.java
@@ -23,10 +23,7 @@ public class ContinueAsNewTest {
 
   @Rule
   public SDKTestWorkflowRule testWorkflowRule =
-      SDKTestWorkflowRule.newBuilder()
-          .setWorkflowTypes(TestContinueAsNewImpl.class)
-          .setUseExternalService(true)
-          .build();
+      SDKTestWorkflowRule.newBuilder().setWorkflowTypes(TestContinueAsNewImpl.class).build();
 
   @Test
   public void testContinueAsNew() {
--- a/temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
+++ b/temporal-test-server/src/main/java/io/temporal/internal/testservice/StateMachines.java
@@ -1420,6 +1420,9 @@ class StateMachines {
     if (d.hasFailure()) {
       a.setFailure(d.getFailure());
     }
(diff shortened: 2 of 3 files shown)


temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflowContext.java
cretz, line 1: _Technically_ this is a change to behavior incompatibly if people have built expectation on the carryover not occurring. Should we add a 💥 in the title and note it in release notes? I assume if they want to go back to today's behavior they can provide an explicit empty attr collection?

(not reviewing the PR itself, just confirming the relaying of compatibility change)

Quinn-With-Two-Ns: So according to the Javadoc the behaviour has always been to copy if none are specify, the SDK just hasn't been. This is a change in behavior so I have no issue calling it out with 💥 .

And yes if a user wants the previous behaviour they can set an empt collection.

PR #2736, opened 2025-11-24
fix(spring-boot): use in-memory test server for non-root namespaces

### LINKED ISSUE #2735: Non-root namespaces should be able to use the in-memory test server
## Expected Behavior
Non-root namespaces should be able to use the in-memory test server when `spring.temporal.test-server.enabled=true` is configured, just like the root namespace does.

## Actual Behavior
Non-root namespaces cannot use the in-memory test server due to a bean name/type mismatch in Spring autoconfiguration. Instead of using `TestWorkflowEnvironment`, the configuration creates a client connecting to the dev server when multiple namespaces are used.

[Example](https://github.com/bbeck10/temporal-samples-java/pull/1/files)
1. Configure Spring Boot application with `spring.temporal.test-server.enabled=true`
2. Set up multiple namespaces (root and non-root namespaces)
3. Attempt to run tests with non-root namespaces
4. Observe that non-root namespaces connect to dev server instead of in-memory test server

## Specifications
- Version: Temporal Java SDK v1.30.0+
- Platform: Spring Boot with Temporal Spring Boot Autoconfigure

**Root Cause:** The issue appears to be in `NonRootBeanPostProcessor` where there's a name/type mismatch:
```java
testWorkflowEnvironment = findBean("temporalTestWorkflowEnvironment", TestWorkflowEnvironmentAdapter.class);
```

However, `TestServerAutoConfiguration` registers:
- `"temporalTestWorkflowEnvironment"` as a `TestWorkflowEnvironment`
- `"temporalTestWorkflowEnvironmentAdapter"` as a `TestWorkflowEnvironmentAdapter`

The code is looking for a `TestWorkflowEnvironmentAdapter` with the wrong bean name.

billrich2001:
Fixes #2735.

What changed:
- Resolve TestWorkflowEnvironmentAdapter by type, fallback to correct bean name
- Add regression test ensuring non-root namespaces use in-memory env even with connection.target set

Why:
- Non-root namespaces were connecting to dev server due to a bean name/type mismatch in NonRootBeanPostProcessor.

Breaking changes:
- None. Internal autoconfiguration fix only.

Tests:
- New test NonRootNamespacesUseTestServerTest mirrors the minimal repro and validates health against the in-memory test server.

### DIFF
--- a/temporal-spring-boot-autoconfigure/src/test/java/io/temporal/spring/boot/autoconfigure/NonRootNamespacesUseTestServerTest.java
+++ b/temporal-spring-boot-autoconfigure/src/test/java/io/temporal/spring/boot/autoconfigure/NonRootNamespacesUseTestServerTest.java
@@ -5,7 +5,6 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import io.grpc.health.v1.HealthCheckResponse;
 import io.temporal.serviceclient.WorkflowServiceStubs;
 import io.temporal.testing.TestWorkflowEnvironment;
-import javax.annotation.Resource;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.TestInstance;
@@ -35,7 +34,8 @@ public class NonRootNamespacesUseTestServerTest {
   @Qualifier("temporalTestWorkflowEnvironment")
   TestWorkflowEnvironment testWorkflowEnvironment;
 
-  @Resource(name = "pomegranateWorkflowServiceStubs")
+  @Autowired
+  @Qualifier("pomegranateWorkflowServiceStubs")
   WorkflowServiceStubs pomegranateWorkflowServiceStubs;
 
   @BeforeEach


temporal-spring-boot-autoconfigure/src/test/java/io/temporal/spring/boot/autoconfigure/NonRootNamespacesUseTestServerTest.java
```diff
@@ -0,0 +1,63 @@
...
+import org.springframework.context.annotation.ComponentScan;
+import org.springframework.context.annotation.FilterType;
+
+@SpringBootTest(
+    classes = NonRootNamespacesUseTestServerTest.Configuration.class,
+    properties = {
+      "spring.temporal.test-server.enabled=true",
+      "spring.temporal.connection.target=127.0.0.1:7233",
+      "spring.temporal.start-workers=false",
+      "spring.temporal.namespaces[0].namespace=pomegranate",
```
Quinn-With-Two-Ns, line 25: Why "pomegranate"?

billrich2001: Lol, the min reproducible sample used it and I thought it was good ¯\_(ツ)_/¯ happy to change it to something else

Quinn-With-Two-Ns: Yeah other tests used more boring names like `namespace0`. Not something we need to be consistent with.

Quinn-With-Two-Ns (approved):

PR #2740, opened 2025-12-02
💥 [BREAKING] enable TLS if api key provided

THardy98:
##What was changed
DWISOTT

1.    Part of  https://github.com/temporalio/features/issues/687

How was this tested:
Unit tests

Any docs updates needed?
Maybe

### DIFF
--- a/.github/workflows/build-native-image.yml
+++ b/.github/workflows/build-native-image.yml
@@ -102,7 +102,7 @@ jobs:
       # path ends in a wildcard because on windows the file ends in '.exe'
       - name: Upload executable to workflow
         if: ${{ inputs.upload_artifact }}
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v5
         with:
           name: ${{ matrix.musl && format('{0}_{1}_musl', matrix.os_family, matrix.arch) || format('{0}_{1}', matrix.os_family, matrix.arch)}}
           path: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,7 +42,7 @@ jobs:
       run: ./gradlew --no-daemon temporal-sdk:testResourceIndependent -x spotlessCheck -x spotlessApply -x spotlessJava -P edgeDepsTest
 
     - name: Publish Test Report
-      uses: mikepenz/action-junit-report@v5
+      uses: mikepenz/action-junit-report@v6
       if: success() || failure() # always run even if the previous step fails
       with:
         report_paths: '**/build/test-results/test/TEST-*.xml'
@@ -127,7 +127,7 @@ jobs:
       run: ./gradlew --no-daemon :temporal-sdk:virtualThreadTests -x spotlessCheck -x spotlessApply -x spotlessJava
 
     - name: Publish Test Report
-      uses: mikepenz/action-junit-report@v5
+      uses: mikepenz/action-junit-report@v6
       if: success() || failure() # always run even if the previous step fails
       with:
         report_paths: '**/build/test-results/test/TEST-*.xml'
@@ -166,7 +166,7 @@ jobs:
       run: ./gradlew --no-daemon :temporal-sdk:test --tests '*CloudOperationsClientTest'
 
     - name: Publish Test Report
-      uses: mikepenz/action-junit-report@v5
+      uses: mikepenz/action-junit-report@v6
       if: success() || failure() # always run even if the previous step fails
       with:
         report_paths: '**/build/test-results/test/TEST-*.xml'
--- a/.github/workflows/nightly-throughput-stress.yml
+++ b/.github/workflows/nightly-throughput-stress.yml
@@ -4,9 +4,6 @@ on:
   schedule:
     # Run at 3 AM PST (11:00 UTC) - offset from existing nightly
     - cron: '00 11 * * *'
-  push:
-    branches:
-      - add-nightly-throughput-stress-workflow
   workflow_dispatch:
     inputs:
       duration:
@@ -24,8 +21,15 @@ on:
         required: false
         default: 360
         type: number
+      is_experiment:
+        description: 'Mark this run as an experiment (excluded from nightly dashboards)'
+        required: false
+        default: false
+        type: boolean
+
 permissions:
     contents: read
+    id-token: write
 
 env:
   # Workflow configuration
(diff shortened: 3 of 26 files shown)


temporal-serviceclient/src/main/java/io/temporal/serviceclient/ServiceStubsOptions.java
```diff
@@ -851,6 +856,14 @@ public ServiceStubsOptions validateAndBuildWithDefaults() {
       Collection<ClientInterceptor> grpcClientInterceptors =
           MoreObjects.firstNonNull(this.grpcClientInterceptors, Collections.emptyList());
 
+      // Auto-enable TLS when API key is provided and TLS is not explicitly set
+      boolean enableHttps;
+      if (this.enableHttpsExplicitlySet) {
+        enableHttps = this.enableHttps;
+      } else {
+        enableHttps = this.apiKeyProvided;
+      }
```
cretz:
```suggestion
      boolean enableHttps = this.enableHttps;
      if (!this.enableHttpsExplicitlySet && this.apiKeyProvided && this.sslContext != null && this.channel != null) {
        enableHttps = true;
      }
```

I think this is better. IIUC, you should only default `enableHttps` if _no_ SSL/channel-related things are set. I think this way of defining the logic is also clearer that you are setting to true in this one rare situation.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/ServiceStubsOptions.java
```diff
@@ -851,6 +856,14 @@ public ServiceStubsOptions validateAndBuildWithDefaults() {
       Collection<ClientInterceptor> grpcClientInterceptors =
           MoreObjects.firstNonNull(this.grpcClientInterceptors, Collections.emptyList());
 
+      // Auto-enable TLS when API key is provided and TLS is not explicitly set
```
cretz: This PR should be marked as 💥 COMPAT BREAK. Perhaps even more in Java than other languages, there may be users setting API keys and expecting non-TLS by default (e.g. to their proxies). So we need to make very clear in release notes this is a breaking change.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/ServiceStubsOptions.java
```diff
@@ -443,6 +445,7 @@ protected Builder(ServiceStubsOptions options) {
       this.target = options.target;
       this.channelInitializer = options.channelInitializer;
       this.enableHttps = options.enableHttps;
+      this.enableHttpsExplicitlySet = true;
```
Quinn-With-Two-Ns: Why are you assuming this is always true if a builder is created from an options?

temporal-serviceclient/src/main/java/io/temporal/serviceclient/ServiceStubsOptions.java
THardy98: That makes sense to me. I've basically done this but removed the "explicit" flag in favor of changing `enableHttps` to `Boolean`

temporal-serviceclient/src/main/java/io/temporal/serviceclient/ServiceStubsOptions.java
THardy98: Yeah, I figured this would be a safe-ish way to assume `enableHttpsExplicitlySet` but it doesn't make sense.
It's not a good way to represent the tri-state of: not set, explicitly false, set.
I've instead opted to make `enableHttps` a `Boolean` instead of a `boolean`.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/ServiceStubsOptions.java
```diff
@@ -418,7 +418,7 @@ public String toString() {
   public static class Builder<T extends Builder<T>> {
     private ManagedChannel channel;
     private SslContext sslContext;
-    private boolean enableHttps;
+    private Boolean enableHttps;
```
cretz, line 445: I assume we don't want to support a user calling `setEnableHttps(null)` to restore to "default"? I don't mind, just want to confirm

cretz, line 445: Here's a contrived situation:

* I create service stubs options with no API key and having set nothing about TLS whatsoever, thereby making `enableHttps` default to `false` on the service stub options
* I want to create new service stub options build from the existing options, so now `enableHttps` is `false` even though I never set it
* I do a `addApiKey` on this new service stub options I have never set any TLS setting on, but it still won't default to TLS because with the current code, it's impossible to create a builder from existing options and retain default TLS behavior

I don't mind if we break and/or don't-support this scenario, just wanted to mention it

Quinn-With-Two-Ns: I think we should at least document that scenario since it is confusing. Arguably if `enableHttps` wasn't explicitly set when the service stub options was created we should preserve that information when we go back to a builder

THardy98: Hmm, maybe should revert the change:
`this.enableHttps != null ? this.enableHttps : false,`

and make `enableHttps` a `Boolean` in `ServiceStubOptions` as well

The semantics might be clearer this way tbh

cretz: If we don't consider changing the return type of `ServiceStubOptions.getEnableHttps()` to `Boolean` instead of `boolean` to be too big of a compat break, I agree that is the better/clearer approach

Quinn-With-Two-Ns: Hm I think changing the return type to a `Boolean` is a bit dangerous since if we return a `null` and users expect it to be a `bool` then they could get a null pointer exception. Can we just internally keep track of it as a `Boolean`, but keep the return type the same?

cretz: I didn't figure it'd be that common, but meh. This works for me, but what should be the return value of `getEnableHttps` be if `enableHttps` is `null`? Would the logic be `return (this.enableHttps != null && this.enableHttps) || (enableHttps == null && this.apiKeyProvided && this.sslContext == null && this.channel == null)`? I also assume users may never want to differentiate whether it's explicitly set or not? Maybe a different field name/getter and `getEnableHttps` is that more advanced logic but the getter still exists for the wrapper?

Quinn-With-Two-Ns: Yeah I don't think it is that common, my main concern is that a user who does call it doesn't realize the type can now be null now

Quinn-With-Two-Ns:
@THardy98 can you also verify that TLS is automatically enabled for Springboot as well?

temporal-serviceclient/src/main/java/io/temporal/serviceclient/ServiceStubsOptions.java
Quinn-With-Two-Ns: @maciejdudko do you have any thoughts on this point?

THardy98: > Yeah I don't think it is that common, my main concern is that a user who does call it doesn't realize the type can now be null now
I basically used the expression Chad suggested - I think that's pretty reasonable. I've updated the comment to reflect the return value.

Quinn-With-Two-Ns (approved):
Would like @maciejdudko to have a look as well

PR #2742, opened 2025-12-03
Defer updating workflow completion metrics until completion accepted by server

### LINKED ISSUE #1590: workflow_completed counter counts successful completions of workflow method instead of workflow executions

cretz:
## What was changed

Defer updating workflow completion metrics until completion accepted by server. This is done by capturing the metrics to update in the task handler result and then providing a post-completion runnable to be invoked if the gRPC method is a success.

This technically changes behavior to no longer record metrics in situations like unhandled commands, which is a more accurate approach.

## Checklist

1. Closes #1590

### DIFF
--- a/temporal-sdk/src/test/java/io/temporal/internal/worker/WorkflowFailedMetricsTests.java
+++ b/temporal-sdk/src/test/java/io/temporal/internal/worker/WorkflowFailedMetricsTests.java
@@ -1,5 +1,6 @@
 package io.temporal.internal.worker;
 
+import static io.temporal.testUtils.Eventually.assertEventually;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertThrows;
@@ -146,8 +147,11 @@ public class WorkflowFailedMetricsTests {
     triggerNonDeterministicException = true;
     workflow.unblock();
     assertThrows(WorkflowFailedException.class, () -> workflow.workflow());
-    reporter.assertCounter(
-        MetricsType.WORKFLOW_FAILED_COUNTER, getWorkflowTags("TestWorkflowWithSignal"), 1);
+    assertEventually(
+        Duration.ofSeconds(2),
+        () ->
+            reporter.assertCounter(
+                MetricsType.WORKFLOW_FAILED_COUNTER, getWorkflowTags("TestWorkflowWithSignal"), 1));
   }
 
   @Test
@@ -161,7 +165,11 @@ public class WorkflowFailedMetricsTests {
                 .setTaskQueue(testWorkflowRule.getTaskQueue())
                 .validateBuildWithDefaults());
     assertThrows(WorkflowFailedException.class, () -> workflow.workflow(true));
-    reporter.assertCounter(MetricsType.WORKFLOW_FAILED_COUNTER, getWorkflowTags("TestWorkflow"), 1);
+    assertEventually(
+        Duration.ofSeconds(2),
+        () ->
+            reporter.assertCounter(
+                MetricsType.WORKFLOW_FAILED_COUNTER, getWorkflowTags("TestWorkflow"), 1));
   }
 
   @Test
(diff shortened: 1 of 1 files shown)


temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowExecutor.java
```diff
@@ -131,12 +125,8 @@ private void completeWorkflow(@Nullable WorkflowExecutionException failure) {
 
     if (context.isCancelRequested()) {
       workflowStateMachines.cancelWorkflow();
-      metricsScope.counter(MetricsType.WORKFLOW_CANCELED_COUNTER).inc(1);
```
cretz, line 134: These completion counters are now tracked inside workflow state machines object

temporal-sdk/src/test/java/io/temporal/client/functional/MetricsTest.java
```diff
@@ -136,6 +136,120 @@ public void testAsynchronousStartAndGetResult() throws InterruptedException, Exe
...
+            .getWorkflowClient()
+            .streamHistory(workflowId)
+            .anyMatch(
+                evt ->
+                    evt.getWorkflowTaskFailedEventAttributes().getCause()
+                        == WorkflowTaskFailedCause.WORKFLOW_TASK_FAILED_CAUSE_UNHANDLED_COMMAND));
+
+    // Confirm we only got the one workflow completed. Before the code fixes that were added with
+    // this test, this would have returned multiple workflow completed counts.
+    Thread.sleep(REPORTING_FLUSH_TIME);
```
Quinn-With-Two-Ns: Can you just use our `assertEventually` utility?

cretz: This was copied from the other two existing tests that do the same thing in this file. Do you think these new tests are worth updating? Do you think the old tests should also be updated?

Quinn-With-Two-Ns: In principle yes they should, but I don't think you need to as part of this PR. If you want to please do however.

Quinn-With-Two-Ns (approved):

temporal-sdk/src/test/java/io/temporal/client/functional/MetricsTest.java
cretz: :+1: It's easy enough I think, will adjust (though may be a day or two)

cretz: Updated

temporal-sdk/src/test/java/io/temporal/client/functional/MetricsTest.java
```diff
@@ -32,108 +43,228 @@ public class MetricsTest {
...
+              2, counts.get(io.temporal.worker.MetricsType.WORKFLOW_COMPLETED_COUNTER).intValue());
+          assertEquals(
+              1, counts.get(io.temporal.worker.MetricsType.WORKFLOW_FAILED_COUNTER).intValue());
+          assertEquals(
+              1,
+              counts
+                  .get(io.temporal.worker.MetricsType.WORKFLOW_CONTINUE_AS_NEW_COUNTER)
+                  .intValue());
+          assertEquals(
+              1, counts.get(io.temporal.worker.MetricsType.WORKFLOW_CANCELED_COUNTER).intValue());
```
maciejdudko: Nit: splitting each workflow run into separate testcase would more strongly verify that the right counter was incremented in each scenario. Could also add checking presence of `WORKFLOW_E2E_LATENCY`.

cretz: I split the tests. I wasn't going to and was going to add e2e check in here, but it turns out we calculate e2e as current - start event time, and the latter is extremely skewed with time skipping causing negative values (which Tally ignores), so it was easier to check in separate tests.

PR #2749, opened 2025-12-26
Fix UnsupportedOperationException when merging ActivityOptions with i…

### LINKED ISSUE #2482: Merging ActivityOptions fails with UnsupportedOperationException when using immutable lists for context propagators

Merging ActivityOptions fails  [here](https://github.com/temporalio/sdk-java/blob/d326d7ff59cd4ec9f4e422633133283707a57d2d/temporal-sdk/src/main/java/io/temporal/activity/ActivityOptions.java#L272) with UnsupportedOperationException when using immutable lists for context propagators

### Steps to Reproduce the Problem

Create a worker with WorkflowImplementationOptions and set setContextPropagators as an immutable list for one of the activities. 

````
 WorkflowImplementationOptions options =
        WorkflowImplementationOptions.newBuilder()
            .setActivityOptions(
                ImmutableMap.of(
                    "ActivityTypeB",
                    ActivityOptions.newBuilder()
                        // Set activity exec timeout (single run)
                        .setStartToCloseTimeout(Duration.ofSeconds(2))
                        .setRetryOptions(
                            RetryOptions.newBuilder()
                                // ActivityTypeB activity type shouldn't retry on NPE
                                .setDoNotRetry(NullPointerException.class.getName())
                                .build())
                        .setContextPropagators(List.of(new MDCContextPropagator()))
                        .build()))
            .build();

````

for the same activity, set the context propagator as an immutable list in the activityStub
```

  private FailingActivities activities =
      Workflow.newActivityStub(
          FailingActivities.class,
          ActivityOptions.newBuilder()
              .setContextPropagators(List.of(new MDCContextPropagator2()))
              .build());
```

The worker throws an `UnsupportedOperationException` [here](https://github.com/temporalio/sdk-java/blob/d326d7ff59cd4ec9f4e422633133283707a57d2d/temporal-sdk/src/main/java/io/temporal/activity/ActivityOptions.java#L272) when it tries to schedule the activity. 

Create a new list that combines both instead of modifying one of them.

### PR DESCRIPTION
**Pull request by nitishagar**
## What was changed

  Modified `ActivityOptions.Builder.mergeActivityOptions()` to create a new `ArrayList` when combining context propagators instead of calling `addAll()` directly on the existing list. Added a test to verify the fix.

  **Files changed:**
  - `temporal-sdk/src/main/java/io/temporal/activity/ActivityOptions.java`
  - `temporal-sdk/src/test/java/io/temporal/activity/ActivityOptionsTest.java`

  ## Why?

  When users set context propagators using immutable lists (e.g., `List.of()` or `Collections.emptyList()`), merging `ActivityOptions` would throw `UnsupportedOperationException` because the code attempted to call `addAll()` on an immutable list.

  This fix follows the approach suggested by @cretz in the issue: create a new list that combines both instead of modifying one of them.

  ## Checklist

  1. Closes #2482

  2. How was this tested:
     - Added unit test `testActivityOptionsMergeWithImmutableContextPropagators` that creates two `ActivityOptions` with immutable context propagator lists (`Collections.emptyList()`) and verifies they can be merged without throwing `UnsupportedOperationException`
     - All existing `ActivityOptionsTest` tests pass
     - Full build passes (`./gradlew clean build`)

  3. Any docs updates needed?
     - No documentation updates needed - this is a bug fix with no API changes

### DIFF
--- a/temporal-sdk/src/test/java/io/temporal/activity/ActivityOptionsTest.java
+++ b/temporal-sdk/src/test/java/io/temporal/activity/ActivityOptionsTest.java
@@ -2,13 +2,17 @@ package io.temporal.activity;
 
 import static org.junit.Assert.*;
 
+import io.temporal.api.common.v1.Payload;
 import io.temporal.common.MethodRetry;
 import io.temporal.common.RetryOptions;
+import io.temporal.common.context.ContextPropagator;
 import io.temporal.testing.TestActivityEnvironment;
 import io.temporal.workflow.shared.TestActivities.TestActivity;
 import io.temporal.workflow.shared.TestActivities.TestActivityImpl;
 import java.lang.reflect.Method;
 import java.time.Duration;
+import java.util.Collections;
+import java.util.List;
 import java.util.Map;
 import org.junit.*;
 import org.junit.rules.Timeout;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: temporal-sdk/src/test/java/io/temporal/activity/ActivityOptionsTest.java [hunk 1]
```diff
@@ -62,6 +63,31 @@ public void testActivityOptionsMerge() {
...
+        ActivityOptions.newBuilder()
+            .setStartToCloseTimeout(Duration.ofSeconds(1))
+            .setContextPropagators(Collections.emptyList())
+            .build();
+
+    ActivityOptions options2 =
+        ActivityOptions.newBuilder()
+            .setStartToCloseTimeout(Duration.ofSeconds(2))
+            .setContextPropagators(Collections.emptyList())
+            .build();
```
maciejdudko, line 118: The test should use non-empty context propagator lists to verify they get merged properly.

nitishagar: Updated.

nitishagar:
@maciejdudko Thanks for the feedback! Could you please re-review when you get a chance?

temporal-sdk/src/test/java/io/temporal/activity/ActivityOptionsTest.java
```diff
@@ -62,6 +66,87 @@ public void testActivityOptionsMerge() {
     Assert.assertEquals(methodOps1, merged);
   }
 
+  @Test
+  public void testActivityOptionsMergeWithImmutableContextPropagators() {
+    // Create simple test context propagators
+    ContextPropagator propagator1 =
+        new ContextPropagator() {
```
maciejdudko: This is a lot of code duplication. It would be better to have a [local class](https://docs.oracle.com/javase/tutorial/java/javaOO/localclasses.html) inside the test function:

```
public void testActivityOptionsMergeWithImmutableContextPropagators() {
  class MyContextPropagator implements ContextPropagator {
    MyContextPropagator(String name) {
      this.name = name;
    }

      @Override
    public String getName() {
      return name;
    }
  
    // etc.
  }
```

Then instantiate it twice:
```
ContextPropagator propagator1 = new MyContextPropagator("propagator 1");
ContextPropagator propagator2 = new MyContextPropagator("propagator 2");
```

nitishagar:
@maciejdudko Updated based on feedback. Take a look when you get a chance.

PR #2751, opened 2025-12-29
Add Async.await() for non-blocking condition waiting

mfateev:
## Summary
- Add `Async.await(Supplier<Boolean>)` and `Async.await(Duration, Supplier<Boolean>)` methods for non-blocking condition waiting that return a `Promise`
- Add comprehensive tests for cancellation behavior and condition exception handling
- Add cross-reference `@see` Javadoc tags between `Workflow.sleep`, `Workflow.newTimer`, `Workflow.await`, and `Async.await` to help developers discover blocking vs non-blocking alternatives

## Test plan
- [x] Unit tests for `Async.await()` with immediate condition satisfaction
- [x] Unit tests for timed `Async.await()` with timeout expiration
- [x] Unit tests for cancellation behavior (CanceledFailure propagation)
- [x] Unit tests for condition throwing exceptions
- [x] All existing AsyncAwaitTest and WorkflowTest tests pass


temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkflowOutboundCallsInterceptor.java
```diff
@@ -742,6 +742,26 @@ public DynamicUpdateHandler getHandler() {
...
+  /**
+   * Asynchronously wait until unblockCondition evaluates to true or timeout expires.
+   *
+   * @param timeout maximum time to wait for the condition
+   * @param unblockCondition condition that should return true to indicate completion
+   * @return Promise that completes with true if the condition was satisfied, false if the timeout
+   *     expired, or exceptionally with CanceledFailure if the enclosing CancellationScope is
+   *     canceled
+   */
+  Promise<Boolean> awaitAsync(Duration timeout, Supplier<Boolean> unblockCondition);
```
cretz: Will defer to Java SDK reviewers for PR in general, but I wonder if as part of this review we also need to think about an overload that provides a summary to the timer this creates. Also applies to `await` and part of me wonders if we can reuse `reason` in `await` as timer summary, and add `reason` here. Or if we should just have an overload of both `await` and `awaitAsync` that accepts timer options.

mfateev: Added timerSummary to await

maciejdudko (changes requested):
The existing `await` overloads always requires reason. `awaitAsync` should match that.

I don't think it's worth introducing `AwaitOptions`. `await` already takes the reason as direct argument, and it's more ergonomic this way. If there's ever a need for additional options (unlikely), we can add the `AwaitOptions` overloads then. For now, `awaitAsync` should have `String reason` as a direct argument too.

In short - there should be just these two overloads:
```
public static Promise<Void> awaitAsync(String reason, Supplier<Boolean> unblockCondition)
public static Promise<Boolean> awaitAsync(Duration timeout, String reason, Supplier<Boolean> unblockCondition)
```

I'm still reviewing the rest of the code, but wanted to highlight this comment early.

mfateev:
Will remove async options. Hold reviewing the code. I want to change it to execute the condition in a workflow thread for consistency with other callbacks.

temporal-sdk/src/main/java/io/temporal/internal/sync/DeterministicRunner.java
```diff
@@ -92,6 +93,24 @@ static DeterministicRunner newRunner(
   @Nonnull
   WorkflowThread newCallbackThread(Runnable runnable, @Nullable String name);
 
+  /**
+   * Creates a new repeatable workflow thread that re-evaluates its condition on each
+   * runUntilBlocked() call. The thread completes when the condition returns true or when
+   * cancelled/destroyed.
+   *
+   * <p>This is used for async await where the condition can contain blocking operations like
```
mjameswh, line 101: Here, the Javadoc indicates that the condition provided to `newRepeatableThread` can contain blocking operations. But the same method in DeterministicRunnerImpl says the opposite.

Can you confirm which one is in line your your now-currrent design?

PR #2760, opened 2026-01-14
Nexus caller timeouts

bergundy:
## What was changed

- Added support for `ScheduleToStartTimeout` and `StartToCloseTimeout` for Nexus operations in the Java SDK and test server.
- In the test server:
    - Cap secondary timeouts to scheduleToClose timeout
    - Support all three timeout types in `timeoutNexusOperation()`
    - Calculate OPERATION_TIMEOUT header dynamically in `pollNexusTaskQueue()` and use the correct header format
    - Terminate workflows on `GRPC_MESSAGE_TOO_LARGE` workflow task failures with `FORCE_CLOSE_COMMAND` cause, matching real server behavior
    - Updated nexus failure wrapping to match real server: `nexusFailureToAPIFailure` no longer wraps non-temporal failures in `ApplicationFailureInfo`, and `handlerErrorToFailure` no longer sets message at the top level
- Added tests for the new functionality
- Bumped Temporal Server to 1.31.0 (CLI v1.6.1-server-1.31.0-151.0) and proto submodule to v1.61.0
- Updated tests for server behavior changes:
    - Activity timeout retry state is now `RETRY_STATE_TIMEOUT` for schedule-to-start and schedule-to-close timeouts (server fix for [temporalio/temporal#3667](https://github.com/temporalio/temporal/issues/3667))
    - Workflows terminated due to gRPC message too large now fail with `TerminatedFailure` instead of `TimeoutFailure`
    - Nexus operation error failures no longer include `ApplicationFailureInfo` wrapper
    - Nexus handler error failure messages are now in the nested cause (not the top-level failure)

### DIFF
--- a/temporal-sdk/src/test/java/io/temporal/workflow/updateTest/UpdateWithStartTest.java
+++ b/temporal-sdk/src/test/java/io/temporal/workflow/updateTest/UpdateWithStartTest.java
@@ -648,7 +648,6 @@ public class UpdateWithStartTest {
   @SuppressWarnings(
       "deprecation") // Backwards compatibility for WORKFLOW_ID_REUSE_POLICY_TERMINATE_IF_RUNNING
   @Test
-  @SuppressWarnings("deprecation") // Test uses deprecated policy for validation
   public void failServerSideWhenStartIsInvalid() {
     WorkflowClient workflowClient = testWorkflowRule.getWorkflowClient();
 


bergundy:
> Can we add some tests like we have for activity
> 
> https://github.com/temporalio/sdk-java/blob/2322bd0da735fe310b7291f3b7fbeda10bdf5e40/temporal-sdk/src/test/java/io/temporal/workflow/activityTests/ActivityTimeoutTest.java#L51
> 
> to verify the parameter is properly wired through and respected

We have coverage in NexusWorkflowTest.java.

temporal-test-server/src/test/java/io/temporal/testserver/functional/NexusWorkflowTest.java
```diff
@@ -712,6 +712,140 @@ public void testNexusOperationTimeout_AfterCancel() {
...
+          "OPERATION_TIMEOUT should end with 'ms'", operationTimeoutHeader.endsWith("ms"));
+      long operationTimeoutMs =
+          Long.parseLong(operationTimeoutHeader.substring(0, operationTimeoutHeader.length() - 2));
+      // Should be <= schedule-to-start timeout (1 second = 1000ms)
+      Assert.assertTrue(
+          "OPERATION_TIMEOUT should be <= schedule-to-start timeout", operationTimeoutMs <= 1000);
+      Assert.assertTrue("OPERATION_TIMEOUT should be positive", operationTimeoutMs > 0);
+
+      // Sleep longer than schedule-to-start timeout to trigger the timeout
+      Thread.sleep(2000);
```
Quinn-With-Two-Ns: Should be able to use `testWorkflowRule.sleep` here to avoid waiting real time in the time skipping test server, but I would prefer the test not use a sleep and just wait for the condition to be true. Should be possible to wait for the operation to fail by looking at the workflow describe output no?

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
```diff
@@ -955,6 +963,65 @@ public void pollNexusTaskQueue(
...
+            getMutableState(token.getOperationRef().getExecutionId());
+        long scheduledEventId = token.getOperationRef().getScheduledEventId();
+        NexusOperationScheduledEventAttributes scheduledEvent =
+            mutableState.getNexusOperationScheduledEventAttributes(scheduledEventId);
+        boolean isStarted = mutableState.isNexusOperationStarted(scheduledEventId);
+
+        Timestamp scheduledTime = req.getScheduledTime();
+        Timestamp currentTime = store.currentTime();
+        long elapsedSeconds = Timestamps.between(scheduledTime, currentTime).getSeconds();
+        long elapsedMillis = elapsedSeconds * 1000;
```
maciejdudko: This rounds off milliseconds from the difference. At the extreme, a sub-second timeout will be rounded to zero. 

```suggestion
        long elapsedMillis = com.google.protobuf.util.Durations.toMillis(
            Timestamps.between(scheduledTime, currentTime));
```

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
```diff
@@ -955,6 +963,65 @@ public void pollNexusTaskQueue(
...
+            remainingMillis =
+                (remainingMillis == null) ? remaining : Math.min(remainingMillis, remaining);
+          }
+        }
+
+        if (scheduledEvent.hasStartToCloseTimeout()) {
+          long startToCloseMillis =
+              com.google.protobuf.util.Durations.toMillis(scheduledEvent.getStartToCloseTimeout());
+          if (startToCloseMillis > 0) {
+            long remaining = startToCloseMillis - elapsedMillis;
```
maciejdudko: I don't think it's correct to subtract elapsed time from `StartToCloseTimeout` here?

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
```diff
@@ -955,6 +963,65 @@ public void pollNexusTaskQueue(
...
+
+        Timestamp scheduledTime = req.getScheduledTime();
+        Timestamp currentTime = store.currentTime();
+        long elapsedSeconds = Timestamps.between(scheduledTime, currentTime).getSeconds();
+        long elapsedMillis = elapsedSeconds * 1000;
+
+        // Calculate minimum of all applicable timeouts
+        Long remainingMillis = null;
+
+        if (!isStarted && scheduledEvent.hasScheduleToStartTimeout()) {
```
maciejdudko: `ScheduleToStartTimeout` should not be part of the calculation here.

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
```diff
@@ -955,6 +963,65 @@ public void pollNexusTaskQueue(
...
+                (remainingMillis == null) ? remaining : Math.min(remainingMillis, remaining);
+          }
+        }
+
+        if (scheduledEvent.hasScheduleToCloseTimeout()) {
+          long scheduleToCloseMillis =
+              com.google.protobuf.util.Durations.toMillis(
+                  scheduledEvent.getScheduleToCloseTimeout());
+          if (scheduleToCloseMillis > 0) {
+            long remaining = scheduleToCloseMillis - elapsedMillis;
```
maciejdudko: This code should handle the situation where `elapsedMillis` is larger than `scheduleToCloseMillis` for some reason, just for extra safety.
```suggestion
            long remaining = Math.max(1, scheduleToCloseMillis - elapsedMillis);
```

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
```diff
@@ -955,6 +963,65 @@ public void pollNexusTaskQueue(
...
+          }
+        }
+
+        if (scheduledEvent.hasStartToCloseTimeout()) {
+          long startToCloseMillis =
+              com.google.protobuf.util.Durations.toMillis(scheduledEvent.getStartToCloseTimeout());
+          if (startToCloseMillis > 0) {
+            long remaining = startToCloseMillis - elapsedMillis;
+            remainingMillis =
+                (remainingMillis == null) ? remaining : Math.min(remainingMillis, remaining);
```
maciejdudko: `StartToCloseTimeout` should not have elapsed time subtracted.

```suggestion
            remainingMillis =
                (remainingMillis == null) ? startToCloseMillis : Math.min(remainingMillis, startToCloseMillis);
```

maciejdudko (changes requested):
There are two tests failing.

`io.temporal.workflow.nexus.HeaderTest#testOperationHeaders` (test server only): the test expects both `request-timeout` and `operation-timeout` headers to be always set. Currently, test server doesn't send `operation-timeout` if it's not set in options. The real server seems to have a default schedule-to-close timeout of 200 seconds, though I'm not sure where it comes from. The test server should match the real server and have a default timeout too.

`io.temporal.testserver.functional.NexusWorkflowTest#testNexusOperationStartToCloseTimeout` (real server only): there are two problems. One is that the real server sets error message as `"operation timed out"` but the test expects `"operation timed out after starting"`. I think we should have a contains-substring check here rather than equality. The second I was unable to reproduce locally but it failed 2/2 times in CI - `"OPERATION_TIMEOUT should be <= start-to-close timeout"` assertion failed. Might be because of JDK8, or something else.

temporal-test-server/src/test/java/io/temporal/testserver/functional/NexusWorkflowTest.java
bergundy: I think the sleep is reasonable here and I don't expect it to be flaky. If you point me to an example of how to use some "wait for condition" or "eventually" in the Java tests, I can adapt that.

.github/workflows/ci.yml
```diff
@@ -94,8 +94,13 @@ jobs:
             --search-attribute CustomBoolField=Bool \
             --dynamic-config-value system.enableActivityEagerExecution=true \
             --dynamic-config-value frontend.workerVersioningDataAPIs=true \
-            --dynamic-config-value component.nexusoperations.recordCancelRequestCompletionEvents=true \
-            --dynamic-config-value component.callbacks.allowedAddresses='[{"Pattern":"*","AllowInsecure":true}]' \
+            --dynamic-config-value history.MaxBufferedQueryCount=100000 \
```
tconley1428: Why are we adding a bunch of (unrelated?) dynamic configs? Is this a merge conflict with us cleaning them up or are they actually needed?

temporal-test-server/src/test/java/io/temporal/testserver/functional/NexusWorkflowTest.java
```diff
@@ -713,6 +713,144 @@ public void testNexusOperationTimeout_AfterCancel() {
     }
   }
 
+  @Test
+  public void testNexusOperationScheduleToStartTimeout() {
+    assumeTrue(
```
Quinn-With-Two-Ns: Why are we skipping for the real server?

temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflowContext.java
```diff
@@ -799,6 +799,10 @@ public <R> ExecuteNexusOperationOutput<R> executeNexusOperation(
     input.getHeaders().forEach((k, v) -> attributes.putNexusHeader(k.toLowerCase(), v));
     attributes.setScheduleToCloseTimeout(
         ProtobufTimeUtils.toProtoDuration(input.getOptions().getScheduleToCloseTimeout()));
+    attributes.setScheduleToStartTimeout(
+        ProtobufTimeUtils.toProtoDuration(input.getOptions().getScheduleToStartTimeout()));
```
Quinn-With-Two-Ns, line 803: note: `ProtobufTimeUtils.toProtoDuration` will return a zero duration here not a null duration, sometimes the server treats a zero duration different then a empty duration so we need to be careful that is not the case here.

temporal-test-server/src/test/java/io/temporal/testserver/functional/NexusWorkflowTest.java
```diff
@@ -753,8 +891,6 @@ public void testNexusOperationError() {
       Assert.assertEquals("nexus operation completed unsuccessfully", failure.getMessage());
       io.temporal.api.failure.v1.Failure cause = failure.getCause();
       Assert.assertEquals("deliberate test failure", cause.getMessage());
-      Assert.assertTrue(cause.hasApplicationFailureInfo());
```
Quinn-With-Two-Ns, line 756: What type of failure info does the cause have now?

.github/workflows/ci.yml
bergundy: Looks like a merge conflict that wasn't properly resolved. Thanks.

temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflowContext.java
bergundy: Server treats zero and null the same here.

temporal-test-server/src/test/java/io/temporal/testserver/functional/NexusWorkflowTest.java
bergundy: This is a mistake. Let me remove this.

temporal-test-server/src/test/java/io/temporal/testserver/functional/WorkflowIdConflictPolicyTest.java
```diff
@@ -41,6 +41,9 @@ public class WorkflowIdConflictPolicyTest {
 
   @Test
   public void conflictPolicyUseExisting() {
+    org.junit.Assume.assumeTrue(
```
bergundy: This also shouldn't have been committed.

PR #2761, opened 2026-01-14
Plugin Support for Java SDK

### LINKED ISSUE #2626: Plugin support

See https://github.com/temporalio/features/issues/652

donald-pinckney:
# Plugin System for Java Temporal SDK

## What was changed

Added a plugin system for the Java Temporal SDK, modeled after the Python SDK's plugin architecture but adapted to Java idioms.

### New Plugin Interfaces

- **`WorkflowServiceStubsPlugin`** (`temporal-serviceclient`) - Configuration and connection lifecycle for service stubs
- **`WorkflowClientPlugin`** (`temporal-sdk`) - Configuration for workflow clients
- **`ScheduleClientPlugin`** (`temporal-sdk`) - Configuration for schedule clients
- **`WorkerPlugin`** (`temporal-sdk`) - Configuration and lifecycle for worker factories and workers

### WorkerPlugin Lifecycle Methods

`WorkerPlugin` includes lifecycle methods that use a chain pattern where `next` accepts the same parameters as the method:

- `startWorkerFactory(factory, next)` - Wraps factory startup, `next` is `Consumer<WorkerFactory>`
- `shutdownWorkerFactory(factory, next)` - Wraps factory shutdown, `next` is `Consumer<WorkerFactory>`
- `startWorker(taskQueue, worker, next)` - Wraps individual worker startup, `next` is `BiConsumer<String, Worker>`
- `shutdownWorker(taskQueue, worker, next)` - Wraps individual worker shutdown, `next` is `BiConsumer<String, Worker>`
- `replayWorkflowExecution(worker, history, next)` - Wraps workflow replay, `next` is `ReplayCallback` (custom interface that throws `Exception`)

Configuration methods are called in forward (registration) order. Lifecycle methods are called in reverse order for proper nesting (first plugin wraps all others)

### SimplePlugin

- **`SimplePlugin`** (`temporal-sdk`) - Abstract convenience class implementing all plugin interfaces
  - Can be used via builder: `SimplePlugin.newBuilder("name").build()`
  - Can be subclassed for custom behavior
  - Provides convenience methods for common operations (interceptors, data converters, workflow/activity registration)

### Modified Files

- `WorkflowServiceStubsOptions` - Added `plugins` field
- `WorkflowServiceStubs` - Applies plugin configuration and connection lifecycle
- `WorkflowClientOptions` - Added `plugins` field
- `WorkflowClientInternalImpl` - Applies plugin configuration, propagates from service stubs
- `ScheduleClientOptions` - Added `plugins` field
- `ScheduleClientImpl` - Applies plugin configuration, propagates from service stubs
- `WorkerFactoryOptions` - Added `plugins` field
- `WorkerFactory` - Full plugin lifecycle (configuration, worker init, start/shutdown)
- `Worker` - Added plugins field and replay plugin chain support

### Test Files

- `PluginTest.java` - Tests subclassing pattern and lifecycle ordering
- `SimplePluginBuilderTest.java` - Tests builder API comprehensively
- `PluginPropagationTest.java` - Integration tests for plugin propagation chain
- `WorkflowClientOptionsPluginTest.java` - Tests options class plugin field handling

## Why?

The plugin system provides a higher-level abstraction over the existing interceptor infrastructure, enabling users to:

- Modify configuration during client/worker creation
- Wrap execution lifecycles with setup/teardown logic
- Auto-propagate plugins through the creation chain
- Bundle multiple customizations (interceptors, context propagators, data converters, workflow/activity registrations) into reusable units

## Checklist

1. Closes #2626, tracked in [https://github.com/temporalio/features/issues/652](https://github.com/temporalio/features/issues/652)

2. How was this tested:
   - Unit tests for plugin interfaces and SimplePlugin builder
   - Integration tests for plugin propagation through full chain
   - All tests pass locally

3. Any docs updates needed?
   - [ ] Documentation for the new plugin API should be added

---

### Plugin Interfaces

Each level in the creation chain has its own plugin interface:

```
WorkflowServiceStubsPlugin     →  WorkflowClientPlugin  →  WorkerPlugin
                               →  ScheduleClientPlugin
```

Plugins set at higher levels propagate down automatically. For example, a plugin set on `WorkflowServiceStubsOptions` that also implements `WorkflowClientPlugin` will have its `configureWorkflowClient()` method called when creating a `WorkflowClient`.

### SimplePlugin

`SimplePlugin` is an abstract class that implements all plugin interfaces with sensible defaults. It can be used two ways:

**Builder pattern** for declarative configuration:
```java
SimplePlugin plugin = SimplePlugin.newBuilder("my-org.tracing")
    .setDataConverter(myConverter)
    .addWorkerInterceptors(new TracingInterceptor())
    .addClientInterceptors(new LoggingInterceptor())
    .registerWorkflowImplementationTypes(MyWorkflow.class)
    .registerActivitiesImplementations(new MyActivityImpl())
    .onWorkerStart((taskQueue, worker) -> logger.info("Worker started: {}", taskQueue))
    .build();
```

**Subclassing** for custom behavior:
```java
public class TracingPlugin extends SimplePlugin {
    private final Tracer tracer;

    public TracingPlugin(Tracer tracer) {
        super("my-org.tracing");
        this.tracer = tracer;
    }

    @Override
    public void configureWorkflowClient(WorkflowClientOptions.Builder builder) {
        builder.setInterceptors(new TracingClientInterceptor(tracer));
    }

    @Override
    public void configureWorkerFactory(WorkerFactoryOptions.Builder builder) {
        builder.setWorkerInterceptors(new TracingWorkerInterceptor(tracer));
    }
}
```

### Usage Example

```java
// Create plugin
SimplePlugin metricsPlugin = SimplePlugin.newBuilder("my-org.metrics")
    .customizeServiceStubs(b -> b.setMetricsScope(myScope))
    .addWorkerInterceptors(new MetricsInterceptor())
    .build();

// Set on service stubs - will propagate to client and workers
WorkflowServiceStubsOptions stubsOptions = WorkflowServiceStubsOptions.newBuilder()
    .setTarget("localhost:7233")
    .setPlugins(metricsPlugin)
    .build();

WorkflowServiceStubs stubs = WorkflowServiceStubs.newServiceStubs(stubsOptions);
WorkflowClient client = WorkflowClient.newInstance(stubs);  // plugin propagates here
WorkerFactory factory = WorkerFactory.newInstance(client);   // and here
```


temporal-sdk/src/main/java/io/temporal/common/plugin/ClientPlugin.java
```diff
@@ -0,0 +1,141 @@
...
+ *         return stubs;
+ *     }
+ * }
+ * }</pre>
+ *
+ * @see WorkerPlugin
+ * @see PluginBase
+ */
+@Experimental
+public interface ClientPlugin extends ClientPluginCallback {
```
cretz: Is this `WorkflowClientPlugin` or is it meant to work for `ScheduleClient` too (and the upcoming standalone activity client and Nexus operation client)?

temporal-sdk/src/main/java/io/temporal/client/ClientPlugin.java
cretz, line 1: In other SDKs like Go, Python, .NET, and Ruby, we have put this in the same package/module/namespace alongside the client and we have put the worker plugin alongside the worker. I think in Java's case we should do the same and just have `SimplePlugin` be in `common` similar to what we did in .NET, we don't need a whole new package.

temporal-sdk/src/main/java/io/temporal/client/ClientPlugin.java
```diff
@@ -0,0 +1,141 @@
...
+  /**
+   * Allows the plugin to modify service stubs options before the service stubs are created. Called
+   * during configuration phase in forward (registration) order.
+   *
+   * @param builder the options builder to modify
+   * @return the modified builder (may return same instance or new builder)
+   */
+  @Override
+  @Nonnull
+  default WorkflowServiceStubsOptions.Builder configureServiceStubs(
```
cretz: > default

Even in places where we could provide default implementations (e.g. Python and Ruby), we intentionally chose not to because we want to force implementers to implement these (even if they choose no-op). Granted in the simple plugin it makes sense to have default implementations.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubs.java
```diff
@@ -122,5 +126,97 @@ static WorkflowServiceStubs newInstance(
...
+    } catch (Exception e) {
+      throw new RuntimeException("Failed to create service stubs with plugins", e);
+    }
+  }
+
+  /**
+   * Callback interface for client plugins to participate in service stubs creation. This interface
+   * is implemented by {@code ClientPlugin} in the temporal-sdk module.
+   */
+  interface ClientPluginCallback {
```
cretz: I think we should just consider making a `ServiceStubPlugin` separate from a `WorkflowClientPlugin` and if someone wants to implement both, they can. We can discuss whether a plugin here automatically propagates to clients if it implements the proper interface (I think it should, same way client plugins should automatically propagate to workers if the they implement the right interfaces).

temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubs.java
```diff
@@ -122,5 +126,97 @@ static WorkflowServiceStubs newInstance(
...
+   *
+   * <p>This method creates stubs with "lazy" connectivity. The connection is not performed during
+   * the creation time and happens on the first request.
+   *
+   * @param options stub options to use
+   * @param plugins list of plugins to apply (plugins implementing ClientPlugin are processed)
+   * @return the workflow service stubs
+   */
+  static WorkflowServiceStubs newServiceStubs(
+      @Nonnull WorkflowServiceStubsOptions options, @Nonnull List<?> plugins) {
```
cretz: I think we should accept the plugins on the options and therefore only mutate existing stub construction methods instead of adding a new one. All other SDK implementations of plugins do (acknowledging that plugins should not mutate the plugin list or it's undefined behavior). This makes it easy for users building options separate from where they call this. It also makes it easy to properly apply to the existing overloads like `newConnectedServiceStubs`. This was done for client, I think it's worth doing here.

temporal-sdk/src/main/java/io/temporal/client/WorkflowClientOptions.java
```diff
@@ -132,6 +136,48 @@ public Builder setQueryRejectCondition(QueryRejectCondition queryRejectCondition
       return this;
     }
 
+    /**
+     * Sets the plugins to use with this client. Plugins can modify client and worker configuration,
+     * intercept connection, and wrap execution lifecycle.
+     *
+     * <p>Each plugin should implement {@link io.temporal.common.plugin.ClientPlugin} and/or {@link
+     * io.temporal.common.plugin.WorkerPlugin}. Plugins that implement both interfaces are
```
cretz: I think we should require all plugins here to implement `ClientPlugin` as we do in other SDKs. Plugins that only implement `WorkerPlugin` can be configured on worker options. This is similar to interceptors.

temporal-sdk/src/main/java/io/temporal/client/WorkflowClientOptions.java
```diff
@@ -132,6 +136,48 @@ public Builder setQueryRejectCondition(QueryRejectCondition queryRejectCondition
...
+     * automatically propagated to workers created from this client.
+     *
+     * @param plugins the list of plugins to use (each should implement ClientPlugin and/or
+     *     WorkerPlugin)
+     * @return this builder for chaining
+     * @see io.temporal.common.plugin.ClientPlugin
+     * @see io.temporal.common.plugin.WorkerPlugin
+     */
+    @Experimental
+    public Builder setPlugins(List<?> plugins) {
```
cretz: I think we should consider matching our approach for interceptors (just a varargs setter) instead of doing something different here. How interceptors are configured is an ideal model for plugins throughout the SDK even if just for consistency reasons IMO (e.g. as arrays, etc).

temporal-sdk/src/main/java/io/temporal/client/ClientPlugin.java
cretz: Arguably this shouldn't return a value since builder is mutable, though it's mostly harmless to do so

temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubs.java
cretz: Hrmm, I wonder if we want to consider the pattern of having plugins that operate on options do it at build time instead of construction time. For users that might themselves use `validateAndBuildWithDefaults()`, the plugins won't get to participate pre-validation.

Though I think it is probably ok to do it this way as well as I think it may be rare for users to use the `validateAndBuildWithDefaults()` approach. But we should at least document above `validateAndBuildWithDefaults()` that it validates and sets defaults before plugins get a chance to update options.

temporal-sdk/src/main/java/io/temporal/client/WorkflowClientInternalImpl.java
```diff
@@ -65,6 +66,8 @@ public static WorkflowClient newInstance(
 
   WorkflowClientInternalImpl(
       WorkflowServiceStubs workflowServiceStubs, WorkflowClientOptions options) {
+    // Apply plugin configuration phase (forward order)
+    options = applyClientPluginConfiguration(options);
     options = WorkflowClientOptions.newBuilder(options).validateAndBuildWithDefaults();
```
cretz: > WorkflowClientOptions.newBuilder(options)

Should just pass this builder into applying plugin configuration (or just inline applying plugin configuration), no need to make the builder twice.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubs.java
```diff
@@ -122,5 +126,97 @@ static WorkflowServiceStubs newInstance(
...
+     *
+     * @param options the final options being used for connection
+     * @param next supplier that creates the service stubs
+     * @return the service stubs
+     * @throws Exception if connection fails
+     */
+    @Nonnull
+    WorkflowServiceStubs connectServiceClient(
+        @Nonnull WorkflowServiceStubsOptions options, @Nonnull ServiceStubsSupplier next)
+        throws Exception;
```
cretz: Not sure this should throw a checked exception (our other connection calls don't IIUC)

temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubs.java
```diff
@@ -122,5 +126,97 @@ static WorkflowServiceStubs newInstance(
...
+
+    /** Functional interface for the connection chain. */
+    @FunctionalInterface
+    interface ServiceStubsSupplier {
```
cretz: Not much benefit to this separate interface vs just changing param type to `Supplier<WorkflowServiceStubs>` (we don't have `WorkerPluginRunnable` for worker). I also don't think it should throw a checked exception.

temporal-sdk/src/main/java/io/temporal/common/SimplePluginBuilder.java
```diff
@@ -0,0 +1,296 @@
...
+        new ArrayList<>(clientCustomizers),
+        new ArrayList<>(factoryCustomizers),
+        new ArrayList<>(workerCustomizers),
+        new ArrayList<>(workerInterceptors),
+        new ArrayList<>(clientInterceptors),
+        new ArrayList<>(contextPropagators));
+  }
+
+  /** Internal implementation of the simple plugin. */
+  private static final class SimplePlugin extends PluginBase {
```
cretz: IMO this needs to be a public concept. People need to be able to use simple plugins and still be able to override maybe one method if they still need to. I think therefore `PluginBase` does not need to exist, it can be replaced by `SimplePlugin` and `Builder` can be an inner class of `SimplePlugin` (as we do elsewhere). Can still have the (public) constructor accept options though, but I think a public no-arg (or name-only arg) constructor should also be allowed for some use cases.

temporal-sdk/src/main/java/io/temporal/common/plugin/WorkerPlugin.java
```diff
@@ -0,0 +1,145 @@
...
+   *     next.run();
+   *     logger.info("Workers started");
+   * }
+   * }</pre>
+   *
+   * @param factory the worker factory being started
+   * @param next runnable that starts the next in chain (eventually starts actual workers)
+   * @throws Exception if startup fails
+   */
+  default void runWorkerFactory(@Nonnull WorkerFactory factory, @Nonnull Runnable next)
```
cretz: This should be called `startWorkerFactory` IMO

temporal-sdk/src/main/java/io/temporal/common/plugin/WorkerPlugin.java
```diff
@@ -0,0 +1,145 @@
...
+  /**
+   * Called when the worker factory is shutting down. Plugins are notified in forward (registration)
+   * order during shutdown.
+   *
+   * <p>This is called during both {@link WorkerFactory#shutdown()} and {@link
+   * WorkerFactory#shutdownNow()}.
+   *
+   * @param factory the worker factory being shut down
+   */
+  default void onWorkerFactoryShutdown(@Nonnull WorkerFactory factory) {
```
cretz: IMO this should be called `shutdownWorkerFactory` and have a `next()` too that surrounds the shutdown calls instead of just reacts to them. This may be valuable if they want to do something before vs after shutdown.

temporal-sdk/src/main/java/io/temporal/common/plugin/WorkerPlugin.java
cretz: This abstraction isn't enough. Plugins need to be able to register activities/workflows/etc on a worker and shutdown things on a worker. Arguably this one of the primary reasons for plugins. I think you need `startWorker` and `shutdownWorker` methods too.

donald-pinckney: I've added:
- initializeWorker: you should register workflows/etc. in here. Called when workers are created.
- startWorker, with a next() chain
- shutdownWorker, with a next() chain

I think this should be sufficient.

temporal-sdk/src/main/java/io/temporal/common/plugin/ClientPlugin.java
donald-pinckney: Let's keep it scoped to just workflows. I think it should be clear now, after I moved it to the appropriate package.

cretz: This means that schedule clients can't have plugins? Clients doing schedule things can have plugins in every other SDK (same concern for standalone activity clients coming and Nexus operation clients coming)

temporal-sdk/src/main/java/io/temporal/client/ClientPlugin.java
cretz, line 1: Unlike some langs like Go/Python/Ruby where qualifying an ambiguous name is normal, it's not normal in Java (or .NET). I would recommend calling this `ClientPlugin` and the other `WorkerPlugin`.

temporal-sdk/src/main/java/io/temporal/common/plugin/WorkerPlugin.java
cretz: Works for me. I think `startWorker` could effectively do what `initializeWorker` does and therefore we only need `startWorker`, but I don't strongly object to the separate methods.

temporal-sdk/src/main/java/io/temporal/common/plugin/ClientPlugin.java
donald-pinckney: Ah i see, i was missing that in Python the big huge `Client` also includes scheduling stuff, whereass in java they are separate. I'll take a closer look at that.

temporal-sdk/src/main/java/io/temporal/common/SimplePluginBuilder.java
donald-pinckney: I've replaced `PluginBase` with `SimplePlugin`, with `Builder` now an inner class. `SimplePlugin` now has 2 constructors:
- Just a `name`.
- Or, a full `Builder`.

temporal-sdk/src/main/java/io/temporal/common/plugin/ClientPlugin.java
cretz: 👍 And note, there are more clients expected to come with standalone activity client and Nexus operation client, but they all will use the same service stubs

temporal-sdk/src/main/java/io/temporal/common/SimplePlugin.java
```diff
@@ -0,0 +1,471 @@
...
+ *         worker.registerWorkflowImplementationTypes(MyWorkflow.class);
+ *     }
+ * }
+ * }</pre>
+ *
+ * @see ClientPlugin
+ * @see WorkerPlugin
+ */
+@Experimental
+public class SimplePlugin implements ClientPlugin, WorkerPlugin {
```
cretz: Pedantic, but if we're not expecting this to be user-instantiable directly and only via builder or extension, could mark it as `abstract` (not worried about us adding future client/worker plugin methods we forget to implement here, because to be backwards compatible, all future client/worker plugin methods we add will have to have a `default`)

temporal-sdk/src/main/java/io/temporal/common/SimplePlugin.java
```diff
@@ -0,0 +1,471 @@
...
+    next.run();
+  }
+
+  @Override
+  public String toString() {
+    return getClass().getSimpleName() + "{name='" + name + "'}";
+  }
+
+  /** Builder for creating {@link SimplePlugin} instances with declarative configuration. */
+  public static final class Builder {
```
cretz, line 334: For a "simple plugin", in addition to interceptors and context propagators, I think users should also be able to directly/simply set data converters, activities, workflows, and Nexus services

temporal-sdk/src/main/java/io/temporal/client/WorkflowClientOptions.java
donald-pinckney: In`WorkflowClientOptions`, I've changed it to now be this signature:
```
public Builder setPlugins(ClientPlugin... plugins) { ... }
```

And in `WorkerFactorOptions`, I've added:
```
public Builder setPlugins(WorkerPlugin... plugins) { ... }
```

temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubs.java
donald-pinckney: @cretz I've pushed some commits to separate out `ServiceStubsPlugin` and `WorkflowClientPlugin`. We now have these interfaces:

- `ServiceStubsPlugin`
  - `WorkflowServiceStubsPlugin extends ServiceStubsPlugin` (fills in type parameters)
  - In theory could add a `CloudServiceStubsPlugin` and `OperatorServiceStubsPlugin`
- `WorkflowClientPlugin`
- `WorkerPlugin`


With these plugin installation points:

1. `WorkflowServiceStubsOptions.setPlugins(WorkflowServiceStubsPlugin...)`
2. `WorkflowClientOptions.setPlugins(WorkflowClientPlugin...)`
3. `WorkerFactoryOptions.setPlugins(WorkerPlugin...)`
  
And the plugins propagate through:

```
WorkflowServiceStubsOptions.setPlugins([A, B])
                │
                ▼ extract WorkflowClientPlugin instances
WorkflowClientOptions  →  merged: [A, B] + [C]  →  [A, B, C]
                │
                ▼ extract WorkerPlugin instances
WorkerFactoryOptions   →  merged: [A, B, C] + [D]  →  [A, B, C, D]
```

When you have time could you TAL at this updated design? 

Now that service stub stuff is separated out from `WorkflowClientPlugin` I should be able to add in a`ScheduleClientPlugin` that is parallel to `WorkflowClientPlugin`.

cretz: > WorkflowServiceStubsPlugin extends ServiceStubsPlugin (fills in type parameters)

IMO you don't need the super interface at this time (we can tease it out later if there ever is another service we want to support plugins for, but for the most part, I think just workflow service is fine)

> When you have time could you TAL at this updated design?

This looks great! Makes perfect sense. It is unfortunate that Java has this separate layer in a completely separate JAR preventing you from combining service stubs plugins w/ client plugins (which is effectively what we do in other SDKs, combine connection plugins w/ client plugins), but that is the unique nature of our Java SDK setup.

> Now that service stub stuff is separated out from WorkflowClientPlugin I should be able to add in a ScheduleClientPlugin that is parallel to WorkflowClientPlugin.

Agreed, and is also an unfortunate Java design decision that is not in other SDKs. If you'd like, you can do the alternative, which is a single `ClientPlugin` interface with `configureWorkflowClient` and `configureScheduleClient` methods, and that plugin type can be accepted by both client options. Your choice, though I admit I prefer the combined `ClientPlugin` used by all clients w/ just different configure calls (knowing when activity client and Nexus client come about, they'll be new method calls too, but if they come out after this, we'll have to give them `default` impls, so maybe that's a point against this approach).

Also even with the separate iface approach, you'll probably want to change `configureClient` to `configureWorkflowClient` for users that build plugins that implement multiple of these interfaces (including ourselves with `SimplePlugin`).

Let me know when ready for full re-review once decided about the schedule client thing.

donald-pinckney: I'd rather keep them as separate interfaces, mostly because of the `default` issue. Will ping you for full re-review when I've added in `ScheduleClientPlugin`, and addressed the other comments.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/WorkflowServiceStubs.java
donald-pinckney: Yeah, I'm going to leave it as is, with added documentation.

temporal-sdk/src/main/java/io/temporal/client/WorkflowClientInternalImpl.java
```diff
@@ -65,7 +70,18 @@ public static WorkflowClient newInstance(
...
+    WorkflowClientPlugin[] propagatedPlugins =
+        extractClientPlugins(workflowServiceStubs.getOptions().getPlugins());
+
+    // Merge propagated plugins with client-specified plugins
+    WorkflowClientPlugin[] mergedPlugins = mergePlugins(propagatedPlugins, options.getPlugins());
+
+    // Apply plugin configuration phase (forward order), then validate
+    WorkflowClientOptions.Builder builder = WorkflowClientOptions.newBuilder(options);
+    builder.setPlugins(mergedPlugins);
+    applyClientPluginConfiguration(builder, mergedPlugins);
```
donald-pinckney: @cretz One subtle thing here with plugin propagation is what the plugins see, for which plugins are set on the builder when `plugin.configureWorkflowClient(builder)` is called on the plugin.

In this Java implementation, the plugins will all see the `mergedPlugins` (propagated + explicit).

In Python, the plugins only see the plugins that are in the explicit config, and **not** the propagated ones:

```python
plugins_from_client = cast(
            list[Plugin],
            [p for p in client.config()["plugins"] if isinstance(p, Plugin)],
        )
        for client_plugin in plugins_from_client:
            if type(client_plugin) in [type(p) for p in plugins]:
                warnings.warn(
                    f"The same plugin type {type(client_plugin)} is present from both client and worker. It may run twice and may not be the intended behavior."
                )
        plugins = plugins_from_client + list(plugins)
        self._initial_config = config.copy()

        self._plugins = plugins
        for plugin in plugins:
            config = plugin.configure_worker(config)
```

(that code is for _worker.py, but point still stands).

Do you have an intuition for which is the correct behavior and why? I would think that the Java behavior is more natural.

temporal-sdk/src/main/java/io/temporal/client/WorkflowClientInternalImpl.java
```diff
@@ -771,4 +787,73 @@ public NexusStartWorkflowResponse startNexus(
...
+      propagatedTypes.add(p.getClass());
+    }
+    for (WorkflowClientPlugin p : explicit) {
+      if (propagatedTypes.contains(p.getClass())) {
+        log.warn(
+            "Plugin type {} is present in both propagated plugins (from service stubs) and "
+                + "explicit plugins. It may run twice which may not be the intended behavior.",
+            p.getClass().getName());
+      }
+    }
```
cretz: We can warn about duplicate instances, but I don't think we should warn about duplicate types. For instance, it may be very normal to provide a `SimplePlugin` that sets up data converter at client level and another that sets up endpoint/address at service stub level.

temporal-sdk/src/main/java/io/temporal/client/WorkflowClientInternalImpl.java
cretz: I don't have a _strong_ opinion here. While there is a use case technically for one plugin to know whether another is configured, in general we discourage plugins from caring about each other.

In my head though, configure should be called on the user-provided options only. This includes _before_ even defaults are put on the options (so a plugin can differentiate unset from default set like for namespace).

So IMO this should be done after the validate and build step, but it doesn't matter too much.

temporal-sdk/src/main/java/io/temporal/client/WorkflowClientInternalImpl.java
```diff
@@ -65,7 +70,18 @@ public static WorkflowClient newInstance(
 
   WorkflowClientInternalImpl(
       WorkflowServiceStubs workflowServiceStubs, WorkflowClientOptions options) {
-    options = WorkflowClientOptions.newBuilder(options).validateAndBuildWithDefaults();
+    // Extract WorkflowClientPlugins from service stubs plugins (propagation)
+    WorkflowClientPlugin[] propagatedPlugins =
+        extractClientPlugins(workflowServiceStubs.getOptions().getPlugins());
+
+    // Merge propagated plugins with client-specified plugins
+    WorkflowClientPlugin[] mergedPlugins = mergePlugins(propagatedPlugins, options.getPlugins());
```
cretz: Pedantic, but feel free to use lists and convert back to arrays, and therefore can probably just inline most of this vs two new private single-use methods. Not too concerned with eeking out every bit of array-only performance here, IMO we only use arrays for user-facing reasons because interceptors are. But don't have to change anything if you don't want.

temporal-sdk/src/main/java/io/temporal/worker/WorkerPlugin.java
```diff
@@ -0,0 +1,293 @@
...
+   * needs replay-specific configuration that differs from normal worker configuration.
+   *
+   * <p>This is useful when a plugin needs to apply the same settings to replay that it applies to
+   * normal workers (e.g., data converters, interceptors) to ensure replay behavior matches
+   * execution behavior.
+   *
+   * @param taskQueue the task queue name for the replay worker
+   * @param builder the options builder to modify
+   */
+  default void configureReplayWorker(
```
cretz: Not sure configuring replay worker is any different from configuring worker in Java, I think `configureWorker` is enough. Same for `initializeReplayWorker`.

temporal-testing/src/main/java/io/temporal/testing/WorkflowReplayer.java
cretz, line 1: I don't think this file should change. For instance, a data converter may be needed for a replayer too, how is that configured today? I'm thinking plugins are configured similarly. And they can participate in regular worker lifecycle things. Only Java supports replay on workers, so only Java plugins support general configure worker also for configure replayer.

temporal-sdk/src/main/java/io/temporal/client/WorkflowClientInternalImpl.java
donald-pinckney: Hhhmm, that does make sense to me to change it to check instances. But, we should re-think this for Python then as well, since it also checks types: https://github.com/temporalio/sdk-python/blob/23244e028074a37ac167333fc23f8d25f5e91c1b/temporalio/worker/_worker.py#L366

donald-pinckney: Changed it to check instances.

temporal-sdk/src/main/java/io/temporal/worker/WorkerPlugin.java
donald-pinckney: So in Python the Plugin interface configure methods are:
- `def configure_worker(self, config: WorkerConfig) -> WorkerConfig:`
- `def configure_replayer(self, config: ReplayerConfig) -> ReplayerConfig:` 

In Java we don't have different config types, so its just same signatures:
- `void configureWorker(@Nonnull String taskQueue, @Nonnull WorkerOptions.Builder builder)`
- `void configureReplayWorker(@Nonnull String taskQueue, @Nonnull WorkerOptions.Builder builder)`

In that sense it seems like they are duplicates. My main question though is if plugins need to ability to configure themselves differently in the case of execution vs. replay. For example, a metrics plugin might want to:                                                                               
  - During normal execution: record workflow start/completion metrics                                                                                                                        
  - During replay: skip metrics (since replaying old histories shouldn't count as new executions)

Though perhaps `replayWorkflowExecution` is enough for that?

temporal-testing/src/main/java/io/temporal/testing/WorkflowReplayer.java
donald-pinckney: The only need for the changes to `WorkflowReplayer.java` is because of `configureReplayWorker` and `initializeReplayWorker` (see other thread). So if we axe those, then these changes would also go away.

temporal-sdk/src/main/java/io/temporal/client/WorkflowClientInternalImpl.java
cretz: Agreed, cc @tconley1428. Or maybe we can just exclude `SimplePlugin` specifically from this check?

temporal-sdk/src/main/java/io/temporal/worker/WorkerPlugin.java
cretz: > So in Python the Plugin interface configure methods are:

This is because replayers are completely divorced from workers, unlike Java.

> In that sense it seems like they are duplicates. My main question though is if plugins need to ability to configure themselves differently in the case of execution vs. replay

We have chosen in Java to have workers used for both replay and non-replay IIUC, which is unlike Python or any other SDK. We don't know the use case of a worker when the worker is configured on purpose. Not every use of replayer goes through `io.temporal.testing.WorkflowReplayer`, many may use the literal `replayWorkflowExecution` on the worker (that they create from the test environment or otherwise).

IMO a worker used for running or replaying is the same as client used for starting a workflow or listing workflows. I do think as far as someone needing to know which they configure, if that is needed, maybe there is some way to know the worker options are coming from `TestWorkflowEnvironment.newWorker`? But we may be able to table that at first.

donald-pinckney: Then I would think this could make sense: remove `configureReplayWorker` and `initializeReplayWorker`. Instead, we would just call the normal `configureWorker` and `initialzieWorker`, and merge this PR.

If we have a future use case where the plugin needs to know which setting they are in, that could be added as a flag to `WorkerOptions` or something.

cretz: Completely agree.

temporal-sdk/src/main/java/io/temporal/client/WorkflowClientInternalImpl.java
cretz: Yeah, I think merging plugins _after_ plugin configure makes the most sense. Doesn't matter much, but I think it makes sense for plugins to operate on the configuration _given by the user_, not with adjusted defaults.

temporal-sdk/src/main/java/io/temporal/worker/WorkerPlugin.java
```diff
@@ -0,0 +1,258 @@
...
+   *     next.run();
+   *     logger.info("Workers started");
+   * }
+   * }</pre>
+   *
+   * @param factory the worker factory being started
+   * @param next runnable that starts the next in chain (eventually starts actual workers)
+   * @throws Exception if startup fails
+   */
+  void startWorkerFactory(@Nonnull WorkerFactory factory, @Nonnull Runnable next) throws Exception;
```
cretz: I don't think this should throw a checked exception since WorkerFactory.start doesn't (and therefore we'll just wrap in unchecked anyways in code)

temporal-sdk/src/main/java/io/temporal/worker/WorkerPlugin.java
```diff
@@ -0,0 +1,258 @@
...
+   *     logger.info("Shutting down workers...");
+   *     next.run();
+   *     logger.info("Workers shut down");
+   * }
+   * }</pre>
+   *
+   * @param factory the worker factory being shut down
+   * @param next runnable that shuts down the next in chain (eventually shuts down actual workers)
+   */
+  void shutdownWorkerFactory(@Nonnull WorkerFactory factory, @Nonnull Runnable next)
```
cretz: I don't think this should throw a checked exception since WorkerFactory.shutdown doesn't (and therefore we'll just wrap in unchecked anyways in code)

temporal-sdk/src/main/java/io/temporal/worker/Worker.java
```diff
@@ -469,12 +472,57 @@ public String toString() {
...
-        String.class,
-        new Object[] {});
+    // Convert to public history type for plugin API
+    WorkflowExecutionHistory publicHistory =
+        new WorkflowExecutionHistory(
+            history.getHistory(), history.getWorkflowExecution().getWorkflowId());
+
+    // Build plugin chain in reverse order (first plugin wraps all others)
+    // Wrap checked exception in RuntimeException for Runnable compatibility
+    Runnable chain =
```
cretz: I think you should use `Callable<Void>` and let the pieces throw where they do and bubble out so you get the nice stack traces and such. I don't think you should do any special exception handling with plugins (same with other places where plugins are used in this PR).

temporal-sdk/src/main/java/io/temporal/worker/WorkerPlugin.java
```diff
@@ -0,0 +1,258 @@
...
+   *     next.run();
+   * }
+   * }</pre>
+   *
+   * @param taskQueue the task queue name for the worker
+   * @param worker the worker being started
+   * @param next runnable that starts the next in chain (eventually starts the actual worker)
+   * @throws Exception if startup fails
+   */
+  void startWorker(@Nonnull String taskQueue, @Nonnull Worker worker, @Nonnull Runnable next)
```
cretz: IMO, `next` should accept the task queue and worker. Same for all chain pieces in this PR, the "next" callable the user should call should accept the same params this accepts. This is a clearer chaining model and _technically_ allows customization by users on some things (though not practically). This also matches our other SDKs' plugins.

temporal-sdk/src/main/java/io/temporal/worker/Worker.java
```diff
@@ -469,12 +472,57 @@ public String toString() {
...
   public void replayWorkflowExecution(io.temporal.internal.common.WorkflowExecutionHistory history)
       throws Exception {
-    workflowWorker.queryWorkflowExecution(
-        history,
-        WorkflowClient.QUERY_TYPE_REPLAY_ONLY,
-        String.class,
-        String.class,
-        new Object[] {});
+    // Convert to public history type for plugin API
+    WorkflowExecutionHistory publicHistory =
```
cretz, line 476: Pedantic, but we should consider an overload of `replayWorkflowExecution` that accepts this public history type

temporal-sdk/src/main/java/io/temporal/client/WorkflowClientInternalImpl.java
donald-pinckney: I've updated it now so its:

0. Start with user-provided options
1. Run plugin configures
2. `setPlugins`
3. `validateAndBuildWithDefaults`

I think this should be good.

I also inlined those apply configure methods.

temporal-sdk/src/main/java/io/temporal/worker/WorkerPlugin.java
donald-pinckney: I've removed checked exceptions from:

  - `startWorker`                                                                                                                                                                             
  - `startWorkerFactory`                                                                                                                                                                  
  - `shutdownWorkerFactory` 

I kept `replayWorkflowExecution` as checked, since `Worker.replayWorkflowExecution` throws a checked exception.

I've also removed the special exception handling and across the different places.

temporal-sdk/src/main/java/io/temporal/worker/Worker.java
donald-pinckney: Changed to Callable, and removed special exception handling.

temporal-sdk/src/main/java/io/temporal/worker/WorkerPlugin.java
donald-pinckney: They've all been changed to Consumers / BiConsumers. Except for the replay one, this had to be changed to a custom functional interface (for the checked exception).

temporal-sdk/src/main/java/io/temporal/worker/Worker.java
donald-pinckney: Added

temporal-sdk/src/main/java/io/temporal/worker/WorkerFactory.java
```diff
@@ -286,12 +354,73 @@ public synchronized void shutdownNow() {
...
+      final WorkerPlugin workerPlugin = plugins.get(i);
+      shutdownChain =
+          (factory) -> {
+            try {
+              workerPlugin.shutdownWorkerFactory(factory, next);
+            } catch (RuntimeException e) {
+              log.warn("Plugin {} failed during shutdown", workerPlugin.getName(), e);
+              // Still try to continue shutdown
+              next.accept(factory);
+            }
```
cretz: I think plugins need to be sophisticated enough tor recognize that their exceptions in shutdown can halt the process if they aren't calling next (same as any other plugin call). Also, this causes an issue where they may have called next and _then_ threw an exception, but you'll re-call next.

I don't think we have to try to be too smart here. In API docs, can just document concerns around shutdown throwing for plugin authors. Same for worker shutdown.

There is _technically_ a use case where someone would want to halt shutdown (though not sure this is how it would work and not sure that we can support successive calls here anyways).

donald-pinckney: I've opted to remove this specialized handling here, and add in to the API docs the issues around shutdown exceptions.

cretz (commented):
Looks great, thanks! Approved from my POV, but not marking approved because you'll want @maciejdudko or @tconley1428 approval to merge (owners of this SDK).

temporal-sdk/src/main/java/io/temporal/client/WorkflowClientInternalImpl.java
```diff
@@ -771,4 +791,56 @@ public NexusStartWorkflowResponse startNexus(
...
+      }
+    }
+    return clientPlugins.toArray(new WorkflowClientPlugin[0]);
+  }
+
+  /**
+   * Merges propagated plugins with explicitly specified plugins. Propagated plugins come first
+   * (from service stubs), followed by client-specific plugins.
+   */
+  private static WorkflowClientPlugin[] mergePlugins(
```
tconley1428: Is it possible to make a generic version of the merges? Seems like the same for all.

temporal-sdk/src/main/java/io/temporal/common/SimplePlugin.java
```diff
@@ -0,0 +1,730 @@
...
+  public void configureServiceStubs(@Nonnull WorkflowServiceStubsOptions.Builder builder) {
+    for (Consumer<WorkflowServiceStubsOptions.Builder> customizer : stubsCustomizers) {
+      customizer.accept(builder);
+    }
+  }
+
+  @Override
+  public void configureWorkflowClient(@Nonnull WorkflowClientOptions.Builder builder) {
+    // Apply customizers
+    for (Consumer<WorkflowClientOptions.Builder> customizer : clientCustomizers) {
```
tconley1428: I'm not sure I would bother with the customizers versus just letting them subclass simpleplugin if they want to, but I don't object strenuously.

temporal-sdk/src/main/java/io/temporal/common/SimplePlugin.java
```diff
@@ -0,0 +1,730 @@
...
+ * }
+ * }</pre>
+ *
+ * @see WorkflowServiceStubsPlugin
+ * @see WorkflowClientPlugin
+ * @see ScheduleClientPlugin
+ * @see WorkerPlugin
+ */
+@Experimental
+public abstract class SimplePlugin
```
tconley1428, line 110: The usability of this seems limited compared to at least the other languages I wrote given the lack of ability to see and modify existing configs. For instance, if I want my plugin to add a codec to a dataconverter or change the converter, I can't do that. In other languages I allow you to provide a function as well as a raw value. Not sure how easy that is in Java though.

tconley1428: I guess some of the builder stuff makes that tricky. DataConverter in particular is common to need to modify, so maybe just something special for that.

temporal-sdk/src/main/java/io/temporal/common/SimplePlugin.java
cretz: IIUC, in all of our SDKs we decided to have such a concept where they don't have to subclass to provide a lambda to customize existing values. While I disagreed with this (and disagreed with the general aversion to requiring subclassing to do advanced customization), it was overruled.

Granted we could change this to have "customizers" for the specific fields we offer instead of the options as a whole, that'd make sense to me.

temporal-sdk/src/main/java/io/temporal/common/SimplePlugin.java
cretz: > For instance, if I want my plugin to add a codec to a dataconverter or change the converter, I can't do that

Why not? You have an options customizer you can do whatever. But if you're saying you want a data converter customizer instead of an options customizer, that can make sense and more aligns with what other SDKs do.

> In other languages I allow you to provide a function as well as a raw value. Not sure how easy that is in Java though.

.NET you couldn't either, but we allowed callables as alternatives for fixed values for every specific simple-plugin option we offered, I think we can do the same here and get rid of the options-level customizers.

donald-pinckney: The current support for customizers (`customizeClient`) sort of let's you do this, but it's a bit janky:

```java
SimplePlugin.newBuilder("my-plugin")                                                                                                                                                                           
    .customizeClient(builder -> {                                                                                                                                                                              
        DataConverter existing = builder.build().getDataConverter();                                                                                                                                           
        builder.setDataConverter(wrapWithCodec(existing, myCodec));                                                                                                                                            
    })                                                                                                                                                                                                         
    .build();
```

Which isn't a whole lot better than subclassing `SimplePlugin` with this override:

```java
@Override                                                                                                                                                                                                  
public void configureWorkflowClient(WorkflowClientOptions.Builder builder) {                                                                                                                               
    // Call super first to apply any base configuration                                                                                                                                                    
    super.configureWorkflowClient(builder);                                                                                                                                                                
                                                                                                                                                                                                            
    // Now see what's configured and wrap it                                                                                                                                                               
    DataConverter existing = builder.build().getDataConverter();                                                                                                                                           
    builder.setDataConverter(                                                                                                                                                                              
        new CodecDataConverter(existing, List.of(myCodec)));                                                                                                                                               
}
```

Here's couple of options:

### Option 0: Remove the customizers

Just remove these customizer things like `customizeClient`. If they want to customize, just have them override `configureWorkflowClient`. Builder-style supports full replacement only.

### Option 1: Add `customizeX` methods alongside `setX` methods     

```java                                                                                                                                                                                                            
// Existing - replaces                                                                                                                                                                                         
public Builder setDataConverter(DataConverter dataConverter)                                                                                                                                                   
                                                                                                                                                                                                                
// New - transforms existing                                                                                                                                                                                   
public Builder customizeDataConverter(UnaryOperator<DataConverter> customizer)                                                                                                                                 
                                                                                                                                                                                                                
// Usage:                                                                                                                                                                                                         
.setDataConverter(myConverter)  // Replace                                                                                                                                                                     
.customizeDataConverter(existing -> wrapWithCodec(existing, myCodec))  // Transform 
```

### Option 2: Only accept functions                                                                                                                                                                          
          
```java                                                                                                                                                                                                       
public Builder setDataConverter(UnaryOperator<DataConverter> customizer)                                                                                                                                       
                                                                                                                                                                                                                
// Usage:                                                                                                                                                                                                         
.setDataConverter(existing -> myConverter)  // Replace (ignores existing)                                                                                                                                      
.setDataConverter(existing -> wrapWithCodec(existing, myCodec))  // Transform 
```

**I don't really like this, I think it makes call sites hard/weird to read for full replacements**.

### Option 3: Custom `PluginValue<T>` wrapper (closer to Python)                                                                                                                                                          
                                                                                                                                                                                                                 
```java
@FunctionalInterface                                                                                                                                                                                           
public interface PluginValue<T> {                                                                                                                                                                              
    T resolve(@Nullable T existing);                                                                                                                                                                           
                                                                                                                                                                                                                
    static <T> PluginValue<T> set(T value) { return existing -> value; }                                                                                                                                       
    static <T> PluginValue<T> customize(UnaryOperator<T> fn) { return fn::apply; }                                                                                                                             
}                                                                                                                                                                                                              
                                                                                                                                                                                                                
// Usage                                                                                                                                                                                                       
.dataConverter(PluginValue.set(myConverter))                                                                                                                                                                   
.dataConverter(PluginValue.customize(existing -> wrap(existing)))                                                                                                                                              
.dataConverter(existing -> wrap(existing))  // Works since it's a functional interface 
```

**Seems like a more complicated version of Option 1**.

donald-pinckney: (FYI i just renamed `customizeClient` -> `customizeWorkflowClient`, so its consistent with`configureWorkflowClient`).

temporal-sdk/src/main/java/io/temporal/client/WorkflowClientInternalImpl.java
donald-pinckney: Done

Issue #2754, opened 2026-01-04
Environment Configuration does not read the correct file path on macOS


pvsone:
## Expected Behavior
Per [the documentation](https://docs.temporal.io/develop/environment-configuration), the SDKs should read the toml config file on a mac from `$HOME/Library/Application Support/temporalio/temporal.toml`.  However this path is not being read (the Linux path is being read instead `~/.config/temporalio/temporal.toml`)

## Actual Behavior
On a mac I expect the SDK to read the Environment Config profile from $HOME/Library/Application Support/temporalio/temporal.toml

## Steps to Reproduce the Problem

  1. Create a Profile TOML file on a Mac using the CLI, which writes a file to: $HOME/Library/Application Support/temporalio/temporal.toml
  1. Create a Java program that attempts to read the Profile using the Env Config APIs.  An error will be thrown that the profile cannot be found
  1. For fun, copy the profile file from the default, documented location ($HOME/Library/Application Support/temporalio/temporal.toml) to ~/.config/temporalio/temporal.toml.  Run the Java program again.  The profile will now be successfully found and loaded

## Specifications

  - Version: Java SDK 1.32.1
  - Platform: Mac

PR #2762, opened 2026-01-15
fix: envconfig, use OS-specific config file paths

THardy98:
## What was changed
Envconfig fix, to use OS-specific config file paths

## Why?
This is the correct behaviour (as documented)

1. Closes https://github.com/temporalio/sdk-java/issues/2754

2. How was this tested:
Added test

3. Any docs updates needed?
No


temporal-envconfig/src/main/java/io/temporal/envconfig/ClientConfig.java
```diff
@@ -32,12 +32,39 @@ public static ClientConfig getDefaultInstance() {
...
     }
+    return getDefaultConfigFilePath(userDir, System.getProperty("os.name"), System.getenv());
+  }
+
+  static String getDefaultConfigFilePath(
+      String userDir, String osName, Map<String, String> environment) {
+    if (osName != null) {
+      String osNameLower = osName.toLowerCase();
+      if (osNameLower.contains("mac")) {
+        return userDir + "/Library/Application Support/temporalio/temporal.toml";
```
mjameswh: nit: Not a big deal, but it's preferable to avoid building paths by simply concatenating raw strings.

```suggestion
        return Paths.get(userDir, "Library", "Application", "Support", "temporalio", "temporal.toml").toString();
```

Same for other below.

temporal-envconfig/src/main/java/io/temporal/envconfig/ClientConfig.java
```diff
@@ -32,12 +32,39 @@ public static ClientConfig getDefaultInstance() {
     return new ClientConfig.Builder().build();
   }
 
-  /** Get the default config file path: $HOME/.config/temporalio/temporal.toml */
+  /**
+   * Get the default config file path based on the operating system:
+   *
+   * <ul>
+   *   <li>macOS: $HOME/Library/Application Support/temporalio/temporal.toml
```
mjameswh: This comment is on a private method, where it is less likely to be found by a user that wonders where `ClientConfig` is storing its data, which is a typical and legitimate question for a user. I'd suggest moving this bullet list to the class's javadoc instead, then just add a reference to that here.

PR #2764, opened 2026-01-21
Add sleep method with options for timer

kevinrobayna:
## What was changed

Adds interface method to WorkflowOutboundCallsInterceptor to accept options for sleep

## Why?

The sleep function in a workflow is syntactic sugar over the timer method. Today I was going to do a change in one of my services to sleep for a few mins but wanted to add a summary and found out this is only
  available on the timer api.

Let me know what you think of this change!

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


tconley1428 (changes requested):
Please add some test coverage for your new methods.

temporal-sdk/src/test/java/io/temporal/workflow/SleepTest.java
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  public static class TestTimerWorkflowImpl implements TestTraceWorkflow {
+
+    @Override
+    public String execute() {
+      boolean useExternalService = SDKTestWorkflowRule.useExternalService;
+      Duration timeout1 = useExternalService ? Duration.ofMillis(700) : Duration.ofSeconds(700);
+      long time = Workflow.currentTimeMillis();
+      Workflow.sleep(timeout1, TimerOptions.newBuilder().setSummary("timer1").build());
```
tconley1428: There doesn't appear to be any validation that this summary has actually taken effect.

kevinrobayna: sorry @tconley1428! busy couple of weeks i think now it should be there

kevinrobayna:
Since there's no interest on your side for this change i'm closing this.

PR #2766, opened 2026-01-27
Support plugins in spring-boot-autoconfigure

donald-pinckney:
## What was changed

Added plugin support to the Spring Boot autoconfigure module. This allows users to define plugin beans that are automatically wired into the Temporal client and worker configuration.

**Key changes:**

1. **Plugin auto-wiring**: Plugin beans (`WorkflowServiceStubsPlugin`, `WorkflowClientPlugin`, `ScheduleClientPlugin`, `WorkerPlugin`) defined in the Spring context are automatically picked up and registered at the appropriate level.

2. **Plugin filtering**: Plugins that implement multiple interfaces (e.g., a plugin implementing both `WorkflowServiceStubsPlugin` and `WorkflowClientPlugin`) are only registered at their highest applicable level to avoid double-registration. 

3. **Plugin ordering**: Plugins support `@Order` and `@Priority` annotations for consistent ordering via `AnnotationAwareOrderComparator`.

4. **Test mode handling**: In test mode (`TestWorkflowEnvironment`), `WorkflowServiceStubsPlugin` beans are not supported since the test environment creates its own internal server. A warning is logged when such plugins are present.

5. **Backwards compatibility**: Added `@Deprecated` constructor overloads to all template classes to maintain backwards compatibility.

**Files modified:**
- `AutoConfigurationUtils.java` - Added `filterPlugins()` and `sortPlugins()` utility methods
- `ServiceStubsAutoConfiguration.java` - Wire `WorkflowServiceStubsPlugin` beans
- `RootNamespaceAutoConfiguration.java` - Wire client and worker plugins with filtering
- `TestServerAutoConfiguration.java` - Wire plugins with test-mode filtering and warning
- `NonRootBeanPostProcessor.java` - Wire plugins for non-root namespaces
- Template classes (`ServiceStubsTemplate`, `ClientTemplate`, `WorkersTemplate`, etc.) - Accept plugin parameters

## Why?

Users need a way to configure Temporal plugins (for tracing, metrics, custom interceptors, etc.) through Spring Boot's dependency injection. Previously, there was no way to automatically wire plugin beans into the Temporal configuration.

This change enables:
- Defining plugin beans with `@Bean` annotations
- Automatic discovery and registration of plugins

## Checklist

1. How was this tested:
   - Added `PluginAutoWiringTest.java` with tests for:
     - Plugin auto-wiring (beans are registered in context)
     - Plugin configuration callbacks are invoked
     - Plugin filtering works correctly (combo plugins filtered at appropriate levels)
     - Plugin ordering respects `@Order` annotation

2. Any docs updates needed?
   - Consider adding documentation for:
     - How to define plugin beans in Spring Boot applications
     - Samples showing this.

### DIFF
--- a/temporal-spring-boot-autoconfigure/src/main/java/io/temporal/spring/boot/autoconfigure/AutoConfigurationUtils.java
+++ b/temporal-spring-boot-autoconfigure/src/main/java/io/temporal/spring/boot/autoconfigure/AutoConfigurationUtils.java
@@ -181,8 +181,7 @@ class AutoConfigurationUtils {
    * @param excludeTypes the plugin interfaces to exclude
    * @return the filtered list, or null if input was null or all plugins were filtered out
    */
-  static <T> @Nullable List<T> filterPlugins(
-      @Nullable List<T> plugins, Class<?>... excludeTypes) {
+  static <T> @Nullable List<T> filterPlugins(@Nullable List<T> plugins, Class<?>... excludeTypes) {
     if (plugins == null || plugins.isEmpty()) {
       return null;
     }
--- a/temporal-spring-boot-autoconfigure/src/main/java/io/temporal/spring/boot/autoconfigure/RootNamespaceAutoConfiguration.java
+++ b/temporal-spring-boot-autoconfigure/src/main/java/io/temporal/spring/boot/autoconfigure/RootNamespaceAutoConfiguration.java
@@ -149,9 +149,7 @@ public class RootNamespaceAutoConfiguration {
             sortedSchedulePlugins, WorkflowServiceStubsPlugin.class);
     List<WorkerPlugin> filteredWorkerPlugins =
         AutoConfigurationUtils.filterPlugins(
-            sortedWorkerPlugins,
-            WorkflowServiceStubsPlugin.class,
-            WorkflowClientPlugin.class);
+            sortedWorkerPlugins, WorkflowServiceStubsPlugin.class, WorkflowClientPlugin.class);
 
     return new NamespaceTemplate(
         properties,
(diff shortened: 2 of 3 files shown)


temporal-spring-boot-autoconfigure/src/main/java/io/temporal/spring/boot/autoconfigure/template/NamespaceTemplate.java
```diff
@@ -71,6 +80,49 @@ public NamespaceTemplate(
     this.clientCustomizers = clientCustomizers;
     this.scheduleCustomizers = scheduleCustomizers;
     this.workflowImplementationCustomizers = workflowImplementationCustomizers;
+    this.workflowClientPlugins = workflowClientPlugins;
+    this.scheduleClientPlugins = scheduleClientPlugins;
+    this.workerPlugins = workerPlugins;
+  }
+
+  /**
+   * @deprecated Use constructor with plugins parameters
```
Quinn-With-Two-Ns: Don't think we need to deprecate , I think we can just remove it. Not sure why a user would ever manually construct this class

donald-pinckney: removed it

temporal-spring-boot-autoconfigure/src/main/java/io/temporal/spring/boot/autoconfigure/NonRootBeanPostProcessor.java
```diff
@@ -78,6 +87,20 @@ public Object postProcessAfterInitialization(@Nonnull Object bean, @Nonnull Stri
...
+        serviceStubsPlugins = findAllBeans(WorkflowServiceStubsPlugin.class);
+        // Filter plugins so each is only registered at its highest applicable level
+        workflowClientPlugins =
+            AutoConfigurationUtils.filterPlugins(
+                findAllBeans(WorkflowClientPlugin.class), WorkflowServiceStubsPlugin.class);
+        scheduleClientPlugins =
+            AutoConfigurationUtils.filterPlugins(
+                findAllBeans(ScheduleClientPlugin.class), WorkflowServiceStubsPlugin.class);
+        workerPlugins =
+            AutoConfigurationUtils.filterPlugins(
```
tconley1428, line 100: Maybe worth just taking a list?

donald-pinckney: 👍 . Fixed

PR #2767, opened 2026-01-28
workflowcheck - initial support for Java static analyzer

tconley1428:
## What was changed

Added initial `temporal-workflowcheck` project. This is currently considered "beta". See the README for more about the project.

TODO in the near future:

* Turn the TODO at the bottom of temporal-workflowcheck/README.md into a set of issues
* After we have published our first release of this...
  * Update primary README with link to the project in the child dir
  * Update the 3 sample projects herein to reference the project
  * Update samples-java to incorporate workflowcheck

### DIFF
--- a/temporal-workflowcheck/build.gradle
+++ b/temporal-workflowcheck/build.gradle
@@ -7,7 +7,7 @@ plugins {
 description = 'Temporal Java WorkflowCheck Static Analyzer'
 
 dependencies {
-    implementation 'org.ow2.asm:asm:9.6'
+    implementation 'org.ow2.asm:asm:9.7'
     compileOnly 'com.google.code.findbugs:jsr305:3.0.2'
     testImplementation project(":temporal-sdk")
     testImplementation "junit:junit:${junitVersion}"
@@ -46,14 +46,16 @@ tasks.named('generateMetadataFileForShadowPublication').configure {
     dependsOn 'jar'
 }
 
-// Treat test Java sources as resources so .java and adjacent files are on the test classpath
-sourceSets {
-    test {
-        resources {
-            srcDir 'src/test/java'
-        }
+// Copy Java test source files to resources so they can be loaded at runtime
+tasks.register('copyJavaSourcesToResources', Copy) {
+    from('src/test/java') {
+        include '**/*.java'
+        include '**/*.properties'
     }
+    into 'build/resources/test'
 }
+processTestResources.dependsOn copyJavaSourcesToResources
+
 
 spotless {
     java {


temporal-serviceclient/src/main/protocloud
cretz, line 1: None of this should have changed as part of this PR I don't suspect. Is the main merge/rebase showing something strange here?

temporal-workflowcheck/README.md
```diff
@@ -0,0 +1,325 @@
...
+* Module support
+* Prevent field mutation in queries and update validators
+* Config prebuilding where you can give a set of packages and it will generate a `.properties` set of invalid methods
+  and save from having to reread the class files of that package at runtime
+  * Also consider shipping with prebuilt config for Java standard library through Java 21
+* Support SARIF output for better integration with tooling like GitHub actions
+* Change output to work with IntelliJ's console linking better (see
+  [this SO answer](https://stackoverflow.com/questions/7930844/is-it-possible-to-have-clickable-class-names-in-console-output-in-intellij))
+* Support an HTML-formatted result with collapsible hierarchy
+* For very deep trees, support `[...]` by default to replace all but the two beginning and two end entries (with CLI
```
cretz: I think we should remove this section and open issues relating to these bullets. I'm thinking issues for:

* workflowcheck - Gradle plugin
* workflowcheck - Maven plugin
* workflowcheck - External configurability improvements
* workflowcheck - Confirm/add module support
* workflowcheck - Lambda support
* workflowcheck - Extra query and update validator checks
* workflowcheck - Improved output (SARIF support and better IntelliJ formatting)

I can create these if they seem right to you. Can add a link to https://github.com/temporalio/sdk-java/issues?q=is%3Aissue%20state%3Aopen%20workflowcheck in place of this TODO instead of just removing the TODO if you want, or just remove it all

temporal-workflowcheck/build.gradle
```diff
@@ -0,0 +1,60 @@
...
+}
+
+// Need all-in-one JAR
+shadowJar {
+    relocate 'org.objectweb.asm', 'io.temporal.workflowcheck.shaded.org.objectweb.asm'
+    archiveClassifier = ''
+}
+build.dependsOn shadowJar
+
+// Configure publishing to publish the shadow jar instead of regular jar
```
cretz: > instead of

I am not sure we want to configure publishing the shadow jar _instead of_ the regular jar. Ideally we can publish both, I admit I am not sure if that's what's happening here, probably

tconley1428: Yeah I'm not following best practices here and knowing what Claude is doing. Java publishing is new to me.

temporal-workflowcheck/README.md
tconley1428: I think maybe make one issue for now, we can split them off when we take them on.

temporal-serviceclient/src/main/protocloud
tconley1428: Sigh. Probably came from the old branch.

temporal-workflowcheck/build.gradle
tconley1428: Courtesy of Claude. 

The shadow jar publishing configuration is correct and necessary for this specific project. Here's why:

  Why Shadow Jar Instead of Regular Jar

  1. All-in-one executable: Lines 22-26 show this is a static analyzer tool that needs to be an all-in-one JAR (shadowJar task with archiveClassifier = '')
  2. Dependency shading: Line 24 shows ASM library is relocated to avoid conflicts: org.objectweb.asm → io.temporal.workflowcheck.shaded.org.objectweb.asm
  3. Tool distribution: This is a command-line static analysis tool (line 19: mainClass = 'io.temporal.workflowcheck.Main') that needs all dependencies bundled

  Should Both Be Published?

  No, only the shadow jar should be published because:

  - The regular JAR would be incomplete (missing dependencies) and unusable
  - The shadow JAR is the actual distributable artifact users need
  - Publishing both would create confusion about which artifact to use
  - Line 25 sets archiveClassifier = '' making the shadow JAR the primary artifact

  This configuration follows the correct pattern for tools that need to be distributed as standalone executables with all dependencies included.

temporal-workflowcheck/README.md
cretz: I don't think it's reasonable to make one overall issue because we have already had a request for at least the plugins individually. But maybe 3 issues:

* Gradle plugin for workflowcheck
* Maven plugin for workflowcheck
* Improved workflowcheck ergonomics

That acceptable

temporal-workflowcheck/build.gradle
cretz: > The regular JAR would be incomplete (missing dependencies) and unusable

Claude is incorrect or at least missing some context. Make sure to inform Claude that this is _also_ used as a library for two reasons: 1) for people to programmatically invoke it (including with their own ASM version if wanted), and 2) at compile-time only for `@WorkflowCheck.SuppressWarnings` annotation.

It is very intentional that the `io.temporal.workflowcheck.WorkflowCheck` class is public and has javadoc for API use. It is a user's choice, when using this as a library (or executable JAR), whether to use their own ASM version or not.

temporal-workflowcheck/README.md
tconley1428: Sure

cretz: Opened #2768, #2769, and #2770.

temporal-workflowcheck/build.gradle
```diff
@@ -0,0 +1,66 @@
...
+    doLast {
+        copy {
+            from('src/test/java') {
+                include '**/*.*'
+            }
+            into 'build/resources/test'
+        }
+    }
+}
+processTestResources.dependsOn copyJavaSourcesToResources
```
cretz:
```suggestion
// Treat test Java sources as resources so .java and adjacent files are on the test classpath
sourceSets {
    test {
        resources {
            srcDir 'src/test/java'
        }
    }
}
```

I think we may have to make this change for CI to pass

PR #2771, opened 2026-02-04
Remove experimental warnings for worker versioning and envconfig

THardy98:
## What was changed
DWISOTT:
- remove experimental warnings for worker deployment API
- remove experimental warnings for (client) environment configuration
- add deprecated warnings/markings to old worker deployment APIs

## Why?
Worker deployment versioning and envconfig are GA

1. Part of https://github.com/temporalio/features/issues/744

3. Any docs updates needed?
Yes?


temporal-sdk/src/main/java/io/temporal/client/BuildIdOperation.java
```diff
@@ -11,6 +11,7 @@
  * <p>See each public static method to learn about and construct the available operations.
  */
 @Experimental
+@Deprecated
```
tconley1428, line 14: Missing the comment here

PR #2773, opened 2026-02-05
💥 Use Temporal Failures for Nexus Error Serialization

Quinn-With-Two-Ns:
💥 Use Temporal Failures for Nexus Error Serialization. The Java SDK now responds to Nexus operation failures and task failures with plain Temporal Failures. This change also allows `OperationException` and `HandlerException` to have their own message and stacktrace independent of their cause. If the Server does not support the new format the SDK converts back to the old format before sending the response. 

Requires:
* https://github.com/temporalio/api/pull/682
* https://github.com/temporalio/temporal/pull/9290

---

> [!NOTE]
> **Medium Risk**
> Changes how Nexus failures are serialized/deserialized and sent over the wire, including compatibility branching based on server capabilities, so mismatches could affect error propagation and retry/cancel semantics.
> 
> **Overview**
> Switches Nexus error reporting to the *new failure format*: Nexus operation failures and handler task failures are returned using Temporal `Failure` protos directly (instead of legacy `UnsuccessfulOperationError`/`HandlerError`), allowing independent message/stacktrace on `OperationException`/`HandlerException` while preserving causes.
> 
> Adds conversion helpers in `NexusUtil` for round-tripping Temporal failures through Nexus `FailureInfo` and for emitting legacy Nexus `HandlerError` when needed; `NexusWorker` now gates behavior on server capability (`temporalFailureResponses`) with a test-only override system property, and the test server is updated to advertise/handle both formats. Tests are expanded/updated to cover failed vs canceled operation metrics, handler errors without causes, stacktrace preservation, and backward-compat behavior; bumps `nexusVersion` to `0.5.0-alpha`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f2baa570206357eff1e3036681b4ea14e6409863. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -30,7 +30,7 @@ ext {
     // Platforms
     grpcVersion = '1.75.0' // [1.38.0,) Needed for io.grpc.protobuf.services.HealthStatusManager
     jacksonVersion = '2.15.4' // [2.9.0,)
-    nexusVersion = '0.4.0-alpha'
+    nexusVersion = '0.5.0-alpha'
     // we don't upgrade to 1.10.x because it requires kotlin 1.6. Users may use 1.10.x in their environments though.
     micrometerVersion = project.hasProperty("edgeDepsTest") ? '1.13.6' : '1.9.9' // [1.0.0,)
 
--- a/temporal-sdk/src/main/java/io/temporal/failure/DefaultFailureConverter.java
+++ b/temporal-sdk/src/main/java/io/temporal/failure/DefaultFailureConverter.java
@@ -15,6 +15,7 @@ import io.temporal.common.converter.EncodedValues;
 import io.temporal.common.converter.FailureConverter;
 import io.temporal.internal.activity.ActivityTaskHandlerImpl;
 import io.temporal.internal.common.FailureUtils;
+import io.temporal.internal.common.NexusUtil;
 import io.temporal.internal.common.ProtobufTimeUtils;
 import io.temporal.internal.sync.POJOWorkflowImplementationFactory;
 import io.temporal.serviceclient.CheckedExceptionWrapper;
@@ -192,7 +193,19 @@ public final class DefaultFailureConverter implements FailureConverter {
               retryBehavior = HandlerException.RetryBehavior.NON_RETRYABLE;
               break;
           }
-          return new HandlerException(info.getType(), cause, retryBehavior);
+          if (failure.getMessage().isEmpty()
+              || failure
+                  .getMessage()
+                  .startsWith(String.format("handler error (%s)", info.getType()))) {
+            return new HandlerException(info.getType(), cause, retryBehavior);
+          } else {
+            return new HandlerException(
+                info.getType(),
+                failure.getMessage(),
+                cause,
+                retryBehavior,
+                NexusUtil.temporalFailureToNexusFailureInfo(failure));
+          }
         }
       case FAILUREINFO_NOT_SET:
       default:
@@ -324,6 +337,9 @@ public final class DefaultFailureConverter implements FailureConverter {
       failure.setNexusOperationExecutionFailureInfo(op);
     } else if (throwable instanceof HandlerException) {
       HandlerException he = (HandlerException) throwable;
+      if (he.getOriginalFailure() != null) {
+        return NexusUtil.nexusFailureToTemporalFailure(he.getOriginalFailure(), true);
+      }
       NexusHandlerErrorRetryBehavior retryBehavior =
           NexusHandlerErrorRetryBehavior.NEXUS_HANDLER_ERROR_RETRY_BEHAVIOR_UNSPECIFIED;
       switch (he.getRetryBehavior()) {
--- a/temporal-sdk/src/main/java/io/temporal/internal/common/NexusUtil.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/common/NexusUtil.java
@@ -1,10 +1,18 @@
 package io.temporal.internal.common;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.ObjectWriter;
 import com.google.protobuf.ByteString;
 import com.google.protobuf.InvalidProtocolBufferException;
 import com.google.protobuf.util.JsonFormat;
+import io.nexusrpc.FailureInfo;
 import io.nexusrpc.Link;
+import io.nexusrpc.handler.HandlerException;
+import io.temporal.api.common.v1.Payload;
+import io.temporal.api.enums.v1.NexusHandlerErrorRetryBehavior;
 import io.temporal.api.nexus.v1.Failure;
+import io.temporal.api.nexus.v1.HandlerError;
 import io.temporal.common.converter.DataConverter;
 import java.net.URI;
 import java.net.URISyntaxException;
@@ -13,7 +21,8 @@ import java.util.Collections;
 import java.util.Map;
 
 public class NexusUtil {
-  private static final JsonFormat.Printer JSON_PRINTER =
+  private static final ObjectWriter JSON_OBJECT_WRITER = new ObjectMapper().writer();
+  private static final JsonFormat.Printer PROTO_JSON_PRINTER =
       JsonFormat.printer().omittingInsignificantWhitespace();
   private static final String TEMPORAL_FAILURE_TYPE_STRING =
       io.temporal.api.failure.v1.Failure.getDescriptor().getFullName();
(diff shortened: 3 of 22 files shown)


build.gradle
```diff
@@ -22,6 +22,7 @@ plugins {
 
 allprojects {
     repositories {
+        mavenLocal()
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: Yes I mention this in the PR description

temporal-sdk/src/main/java/io/temporal/internal/worker/NexusWorker.java
```diff
@@ -344,10 +373,55 @@ private void logExceptionDuringResultReporting(
...
+            return;
+          }
+          taskResponse =
+              b.setStartOperation(
+                      taskResponse.getStartOperation().toBuilder()
+                          .setOperationError(
+                              UnsuccessfulOperationError.newBuilder()
+                                  .setOperationState(operationState)
+                                  .setFailure(
+                                      NexusUtil.temporalFailureToNexusFailure(failure.getCause()))))
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: Yes the old format loses information, this is why we have the new format

temporal-sdk/src/main/java/io/temporal/failure/DefaultFailureConverter.java
```diff
@@ -192,7 +193,18 @@ private RuntimeException failureToExceptionImpl(Failure failure, DataConverter d
               retryBehavior = HandlerException.RetryBehavior.NON_RETRYABLE;
               break;
           }
-          return new HandlerException(info.getType(), cause, retryBehavior);
+          if (failure
+              .getMessage()
+              .startsWith(String.format("handler error (%s)", info.getType()))) {
```
mjameswh: I understand this is here to allow proper decoding of legacy errors, but for which languages exactly? If we mean to support legacy errors coming from Go or others, are we sure the casing returned by `.getType()` here would match the error type of other SDKs?, or should we make this condition a little bit more resilient?

temporal-sdk/src/main/java/io/temporal/internal/worker/NexusWorker.java
```diff
@@ -319,7 +338,17 @@ private void handleNexusTask(NexusTask task, Scope metricsScope) {
...
+        // Check if the server supports using the Failure directly in responses
+        boolean supportTemporalFailure =
+            task.getResponse().getRequest().getCapabilities().getTemporalFailureResponses();
+
+        // Allow tests to force old format for backward compatibility testing
+        String forceOldFormat = System.getProperty("temporal.nexus.forceOldFailureFormat");
+        if ("true".equalsIgnoreCase(forceOldFormat)) {
+          supportTemporalFailure = false;
+        }
+
```
mjameswh: A bit ugly to have this `System.getProperty` + `equalsIgnoreCase` check on the main code path.

I suppose using a different approach to configure that flag would be uselessly complicated, but can we at least move that to the Worker's construtor?

temporal-sdk/src/main/java/io/temporal/internal/worker/NexusWorker.java
```diff
@@ -344,10 +373,55 @@ private void logExceptionDuringResultReporting(
       }
     }
 
+    @SuppressWarnings("deprecation")
     private void sendReply(
```
mjameswh, line 399: The structure of this function is very hard to reason about.

temporal-sdk/src/main/java/io/temporal/internal/worker/NexusWorker.java
```diff
@@ -344,10 +373,55 @@ private void logExceptionDuringResultReporting(
...
+        if (!supportTemporalFailure && taskResponse.getStartOperation().hasFailure()) {
+          Response.Builder b = taskResponse.toBuilder();
+          Failure failure = taskResponse.getStartOperation().getFailure();
+          String operationState;
+          if (failure.hasApplicationFailureInfo()) {
+            operationState = OperationState.FAILED.toString().toLowerCase();
+          } else if (failure.hasCanceledFailureInfo()) {
+            operationState = OperationState.CANCELED.toString().toLowerCase();
+          } else {
+            RespondNexusTaskFailedRequest.Builder request =
```
mjameswh: Is this a case that is expected to happen in real life? Can't we simply throw an `IllegalArgumentException` like we do below?

Quinn-With-Two-Ns: This is some sort of SDK bug, yeah we can treat it the same as below

temporal-sdk/src/main/java/io/temporal/internal/worker/NexusWorker.java
Quinn-With-Two-Ns: If we do [this](https://github.com/temporalio/sdk-java/pull/2773/changes/BASE..3d44cf513058280fd5d7fea404fbb6df3c951f5f#r2854829084), are you still concerned about the structure?

mjameswh: That one is certainly the most patent culprit, but I'd also try to restructure this to avoid mixing happy paths and failure cases.

Like, instead of:

```
      if (taskResponse != null) {
        if (!supportTemporalFailure && taskResponse.getStartOperation().hasFailure()) {   
            // some failure case on old server
            reuturn
        }
        // happy path AND some failure case on newer servers
      }
      // other failure cases
```

I'd suggest restructuring the function along the line of:

```
if (taskResponse != null && !taskResponse.getStartOperation().hasFailure()) {
    // Operation succeeded

} else if (taskResponse != null) {
    // Operation failed - OperationError

} else if (response.getHandlerException()) {
    // Operation failed - HandlerError

} else {
    // throw ...
}
```

Quinn-With-Two-Ns: I'm not sure that will really be clearer since we use the same gRPC method to respond to the server ion the first and second case here.

temporal-sdk/src/main/java/io/temporal/internal/worker/NexusWorker.java
Quinn-With-Two-Ns: Moved to a the constructor

temporal-sdk/src/main/java/io/temporal/failure/DefaultFailureConverter.java
Quinn-With-Two-Ns: Any old failure serialized by Go or Java. This is the handler error type so that should match the Nexus spec and  consistent across Go and Java

temporal-sdk/src/main/java/io/temporal/internal/worker/NexusWorker.java
Quinn-With-Two-Ns: I have an attempt to refactor to make it clearer let me push and we can discuss

temporal-sdk/src/main/java/io/temporal/internal/common/NexusUtil.java
```diff
@@ -47,23 +56,134 @@ public static Link nexusProtoLinkToLink(io.temporal.api.nexus.v1.Link nexusLink)
...
   }
 
-  public static Failure exceptionToNexusFailure(Throwable exception, DataConverter dataConverter) {
-    io.temporal.api.failure.v1.Failure failure = dataConverter.exceptionToFailure(exception);
+  public static Failure temporalFailureToNexusFailure(
+      io.temporal.api.failure.v1.Failure temporalFailure) {
     String details;
     try {
-      details = JSON_PRINTER.print(failure.toBuilder().setMessage("").build());
+      details = PROTO_JSON_PRINTER.print(temporalFailure.toBuilder().setMessage("").build());
```
VegetarianOrc: In Go & Python, we're clearing the stack trace out of the details as well as the message. Seems like it'd make sense to do it here as well.

temporal-sdk/src/main/java/io/temporal/internal/common/NexusUtil.java
```diff
@@ -47,23 +56,134 @@ public static Link nexusProtoLinkToLink(io.temporal.api.nexus.v1.Link nexusLink)
...
+    return temporalFailureToNexusFailure(failure);
+  }
+
+  public static HandlerError handlerErrorToNexusError(
+      HandlerException e, DataConverter dataConverter) {
+    HandlerError.Builder handlerError =
+        HandlerError.newBuilder()
+            .setErrorType(e.getErrorType().toString())
+            .setRetryBehavior(mapRetryBehavior(e.getRetryBehavior()));
+    // TODO: check if this works on old server
```
VegetarianOrc: Have we checked this? If so, let's remove this todo

VegetarianOrc (commented):
LGTM, just a couple of minor things

temporal-sdk/src/main/java/io/temporal/internal/common/NexusUtil.java
Quinn-With-Two-Ns: Yep verified it works

temporal-sdk/src/main/java/io/temporal/failure/DefaultFailureConverter.java
```diff
@@ -324,6 +336,9 @@ private Failure exceptionToFailure(Throwable throwable) {
       failure.setNexusOperationExecutionFailureInfo(op);
     } else if (throwable instanceof HandlerException) {
       HandlerException he = (HandlerException) throwable;
+      if (he.getOriginalFailure() != null) {
+        return NexusUtil.nexusFailureToAPIFailure(he.getOriginalFailure(), true);
```
bergundy: Seems like a better name for this

```suggestion
        return NexusUtil.nexusFailureToTemporalFailure(he.getOriginalFailure(), true);
```

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
```diff
@@ -475,6 +476,7 @@ private StartWorkflowExecutionResponse throwDuplicatedWorkflow(
         WorkflowExecutionAlreadyStartedFailure.getDescriptor());
   }
 
+  @SuppressWarnings("deprecation")
```
bergundy: Would you please add comments in all places that deprecation warnings are suppressed to explain why?

temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java
```diff
@@ -1028,21 +1040,30 @@ public void respondNexusTaskCompleted(
...
   @Override
   public void respondNexusTaskFailed(
       RespondNexusTaskFailedRequest request,
       StreamObserver<RespondNexusTaskFailedResponse> responseObserver) {
     try {
-      if (!request.hasError()) {
+      Failure failure;
+      if (request.hasFailure()) {
```
bergundy, line 1101: This can only be a failure with a handlerFailureInfo: https://github.com/temporalio/temporal/blob/45617a882c4f403aade3490807b13a759e65a1c7/service/frontend/workflow_handler.go#L5739

PR #2779, opened 2026-02-09
Fix deep recursion in ResourceBasedSlotSupplier causing futures to never resolve

tconley1428:
## Summary

- Fixed a critical bug where ResourceBasedSlotSupplier futures would never complete due to deep recursion in CompletableFuture chains
- Replaced recursive `thenCompose()` chaining with iterative polling using `ScheduledExecutorService`
- Added comprehensive test suite to verify the fix and prevent regressions

## Problem

The previous implementation in `scheduleSlotAcquisition()` used recursive CompletableFuture chaining that could create very deep call stacks. When resource starvation lasted for extended periods, this would result in chains of thousands of nested futures that would never resolve even after resources became available, due to stack depth issues and recursive completion problems.

## Solution

Replaced the recursive approach with:

1. **Iterative polling mechanism** using `ScheduledExecutorService.schedule()`
2. **Proper cancellation handling** with `AtomicReference<ScheduledFuture<?>>`  
3. **Maintained timing contracts** - still respects ramp throttle delays
4. **Added minimum 10ms delay** to prevent tight spinning
5. **Preserved all existing API behavior** and contracts

## Test Results

All tests in `ResourceBasedSlotSupplierNonRecursiveTest` now pass:

- ✅ `testFutureNeverResolvesEvenAfterResourcesBecomeAvailable` - **FIXED**: Future now completes in ~1-9ms instead of never completing
- ✅ `testFutureCancellationWithRecursiveChain` - Proper cancellation still works
- ✅ `testMemoryAndPerformanceImpactOfDeepRecursion` - Performance maintained
- ✅ All other existing behavior preserved

## Test plan

- [x] Run `ResourceBasedSlotSupplierNonRecursiveTest` - all tests pass
- [x] Verify the main bug fix: futures complete immediately when resources become available  
- [x] Verify no regressions in cancellation, memory usage, or timing behavior
- [x] Code compiles and builds successfully

🤖 Generated with [Claude Code](https://claude.ai/code)


temporal-sdk/src/main/java/io/temporal/worker/tuning/ResourceBasedSlotSupplier.java
```diff
@@ -190,35 +191,62 @@ public SlotSupplierFuture reserveSlot(SlotReserveContext<SI> ctx) throws Excepti
...
+              return; // Already completed or cancelled
+            }
+
+            try {
+              Optional<SlotPermit> permit = tryReserveSlot(ctx);
+              if (permit.isPresent()) {
+                resultFuture.complete(permit.get());
+              } else {
+                // Calculate delay respecting ramp throttle on each retry
+                Duration mustWaitFor;
```
mjameswh: We should not apply ramp throttle again here. Only the 10ms.

PR #2783, opened 2026-02-11
Add opt-in Jackson 3 support via multi-release JAR

### LINKED ISSUE #2746: Jackson 3 support
Add support for Jackson 3, the next major version of the library the SDK uses for JSON serialization.  

Add support for Jackson 3 to the Java SDK. This would mainly involve creating a new Jackson 3 JSON data converter, and any other place the SDK calls into Jackson. Since Jackson 3 requires Java 17 at least we will need to support both for the foreseeable future.

**Additional context**
* Spring Boot 4 only has plans to support Jackson 3 long term https://spring.io/blog/2025/10/07/introducing-jackson-3-support-in-spring
* https://github.com/FasterXML/jackson/blob/main/jackson3/MIGRATING_TO_JACKSON_3.md

cretz:
## What was changed

- Added `Jackson3JsonPayloadConverter` as a multi-release JAR entry (Java 8 stub in `src/main/java/`, real implementation in `src/main/java17/`) that uses Jackson 3.x `JsonMapper` for JSON serialization
- Added `JacksonJsonPayloadConverter.setDefaultAsJackson3(boolean, boolean)` static method to globally opt in to Jackson 3 delegation, including for the default converter in `DefaultDataConverter.STANDARD_PAYLOAD_CONVERTERS`
- Jackson 3 converter supports two modes: native Jackson 3 defaults (`jackson2Compat=false`) and Jackson 2 compatibility mode (`jackson2Compat=true` via `JsonMapper.builderWithJackson2Defaults()`)
- Converters with an explicit `ObjectMapper` passed to the constructor are not affected by the global Jackson 3 delegate
- Both converters produce identical `"json/plain"` encoding type and are wire-compatible in both directions
- Jackson 3 is published as an optional dependency with version range `[3.0.4,)` in the POM
- Added `jackson3Tests` test suite (8 tests, Java 17 launcher) covering direct converter usage, wire compatibility, global delegation via `GlobalDataConverter.get()`, and explicit-mapper opt-out
- Added Jackson-3-absent tests to the main test suite verifying correct error behavior on Java 8 (`UnsupportedOperationException`) and Java 17+ without Jackson 3 (`NoClassDefFoundError`)
- Added `jackson3Tests` CI step to the `unit_test_jdk8` job
- Added `@Documented` to the `@Experimental` annotation

## Checklist

1. Closes #2746

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/common/converter/JacksonJsonPayloadConverter.java
+++ b/temporal-sdk/src/main/java/io/temporal/common/converter/JacksonJsonPayloadConverter.java
@@ -11,13 +11,56 @@ import com.fasterxml.jackson.datatype.jdk8.Jdk8Module;
 import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
 import com.google.protobuf.ByteString;
 import io.temporal.api.common.v1.Payload;
+import io.temporal.common.Experimental;
 import java.io.IOException;
 import java.lang.reflect.Type;
 import java.util.Optional;
 
 public class JacksonJsonPayloadConverter implements PayloadConverter {
 
+  private static volatile PayloadConverter jackson3Delegate;
+
+  /**
+   * Opts in to or out of using Jackson 3.x as the default JSON payload converter. When enabled,
+   * instances created via the default constructor will delegate all serialization/deserialization
+   * to a {@link Jackson3JsonPayloadConverter}.
+   *
+   * <p>This applies globally, including to the converter in {@link
+   * DefaultDataConverter#STANDARD_PAYLOAD_CONVERTERS}. Call this method early in your application,
+   * before creating any Temporal clients.
+   *
+   * <p>Requires Java 17+ and {@code tools.jackson.core:jackson-databind:3.x} on the classpath.
+   *
+   * @param defaultAsJackson3 {@code true} to delegate to Jackson 3, {@code false} to revert to
+   *     Jackson 2
+   * @param jackson2Compat if {@code true}, the Jackson 3 converter is configured with Jackson 2.x
+   *     default behaviors for maximum wire compatibility. If {@code false}, Jackson 3.x native
+   *     defaults are used. Only relevant when {@code defaultAsJackson3} is {@code true}.
+   * @throws IllegalStateException if Jackson 3 is not available
+   * @see Jackson3JsonPayloadConverter
+   */
+  @Experimental
+  public static void setDefaultAsJackson3(boolean defaultAsJackson3, boolean jackson2Compat) {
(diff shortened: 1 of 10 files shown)


Quinn-With-Two-Ns:
What about the SDK's internal uses of jackson like for Nexus?

Quinn-With-Two-Ns:
So to be clear this PR doesn't let users opt out of including Jackson 2. It just lets them include Jackson 2 and Jackson 3?

mjameswh (approved):

temporal-sdk/src/main/java/io/temporal/common/converter/JacksonJsonPayloadConverter.java
```diff
@@ -39,11 +82,13 @@ public static ObjectMapper newDefaultObjectMapper() {
   }
 
   public JacksonJsonPayloadConverter() {
-    this(newDefaultObjectMapper());
+    this.mapper = newDefaultObjectMapper();
+    this.useDefaultJackson3Delegate = true;
```
tconley1428, line 86: This seems like an odd discrepancy with the below constructor.

mjameswh: That seems ok to me, though I have to admit the control flow isn't immediately obvious.

Essentially:
- If an object mapper was explicitly provided to this converter, then we must use that mapper, rather than the default Jackson 3 object mapper. That's imposed by `useDefaultJackson3Delegate==false`.
- If an object mapper was not explicitly provided, then this converter should use either the Jackson 2 or the Jackson 3 mapper, depending on whether `setDefaultAsJackson3(true, bool)`  was called at init time. That's controlled by having both `useDefaultJackson3Delegate==true` and `jackson3Delegate!=null`.

tconley1428 (approved):

jrpedrianes:
Hi!

Are there any plans to support Jackson 3 exclusively in the future?

Although it would require a small breaking change, it would allow users to remove Jackson 2 from their classpath.

mjameswh:
> Are there any plans to support Jackson 3 exclusively in the future?

We'll certainly have to cross that bridge at some point in the future, but there's no plan to do that _soon_. There are a few challenges and risks involved, and we do not consider breaking changes lightly. Our Java user base is particularly inclined to retain their deployment environment for extended periods of time (e.g. Java 8).

Out of curiosity, are there particular reasons you want to _completely_ drop the Jackson 2 dependency, rather than what we do today? If the presence of the Jackson 2 library on the class path is causing unwanted side effects on other libraries (e.g. hibernate?), would using Temporal's shaded JAR be sufficient?

Also, are you using our Spring Boot integration?

PR #2786, opened 2026-02-13
[DO NOT MERGE] Spring 4 support

### LINKED ISSUE #2738: Spring Boot 4 Support

Currently, Temporal SDK [supports Spring Boot 2 and 3](https://docs.temporal.io/develop/java/spring-boot-integration). Spring Boot 4 has [already come out](https://spring.io/blog/2025/11/20/spring-boot-4-0-0-available-now), so it would be great if Temporal SDK adds support for it as well.

I haven't tested it, so maybe it already works as-is, but I'm almost sure some changes will be required to support Spring Boot 4 due to [its modularization](https://spring.io/blog/2025/10/28/modularizing-spring-boot).

Of course, it's possible to use a custom Spring Boot integration layer based on the core Temporal SDK which supports Spring Boot 4, but since it was released only a week ago, it doesn't make sense to rush and upgrade instantly, so we can wait.

**Additional context**

Since the open-source support of Spring Boot 2 ended in 2023, I think it would be reasonable to drop support for this version in Temporal SDK and only support the latest two versions of Spring Boot. Of course, it's up to you to decide, this is just a suggestion.

cretz:
## What was changed

- Added `temporal-spring-boot-autoconfigure-4` and `temporal-spring-boot-starter-4` modules for Spring Boot 4.0.2 support (Java 17+)
- Extracted 28 shared source files and all tests into `temporal-spring-boot-shared/src/` (not a Gradle module — both `temporal-spring-boot-autoconfigure` and `temporal-spring-boot-autoconfigure-4` compile it via `srcDir`)
- Only the 7 `@ConfigurationProperties` classes in `io.temporal.spring.boot.autoconfigure.properties` are version-specific (`@ConstructorBinding` removed for SB4)
- SB4 modules use `--release 17` and `AutoConfiguration.imports` only (no `spring.factories`)
- Added `springBoot4Version = '4.0.2'` to root `build.gradle`
- Added both new modules to `settings.gradle` and `temporal-bom/build.gradle`
- Updated `gradle/linting.gradle` to run spotless on `temporal-spring-boot-shared/src/` at the root project level (spotless requires targets within the project dir)
- Updated `.github/workflows/ci.yml` to exclude SB4 test tasks from the `unit_test_jdk8` job (`-x :temporal-spring-boot-autoconfigure-4:test -x :temporal-spring-boot-starter-4:test`) since that job runs on JDK 11
- Bumped `.github/workflows/publish-snapshot.yml` and `.github/workflows/prepare-release.yml` from JDK 11 to JDK 17
- Added READMEs with cross-references between SB 2/3 and SB4 modules

Leans on #2783 (since both are in same fork, hard to target it without forcing PR to be on the fork too, will update when other merged, for now ignore the Jackson 3 parts).

We're going to have to pay attention to publish workflows to make sure they work at release time.

DO NOT MERGE - we are going to wait on Jackson 3 support in #2783 to be merged first.

## Checklist

1. Closes #2738


temporal-spring-boot-starter-4/build.gradle
```diff
@@ -0,0 +1,36 @@
...
+
+dependencies {
+    // This platform is the same defined in temporal-spring-boot-autoconfigure-4 and it doesn't need to be here
+    // for gradle, but it is needed for maven and maven pom. Looks like Maven doesn't transitively propagate BOMs.
+    // https://github.com/temporalio/sdk-java/issues/1479
+    api(platform("org.springframework.boot:spring-boot-dependencies:$springBoot4Version"))
+    implementation "org.springframework.boot:spring-boot-starter"
+
+    api project(':temporal-spring-boot-autoconfigure-4')
+    api project(':temporal-sdk')
```
Quinn-With-Two-Ns, line 11: These should be compileOnly like the original springboot auto-configure is no?

temporal-spring-boot-starter-4/build.gradle
```diff
@@ -0,0 +1,36 @@
...
+    sb4JavaVersion = project.hasProperty("edgeDepsTest") || project.hasProperty("nativeBuild") ? JavaVersion.VERSION_21 : JavaVersion.VERSION_17
+    sb4Release = project.hasProperty("edgeDepsTest") || project.hasProperty("nativeBuild") ? '21' : '17'
+}
+
+java {
+    sourceCompatibility = sb4JavaVersion
+    targetCompatibility = sb4JavaVersion
+}
+
+// Override the --release 8 flag set by gradle/java.gradle
```
Quinn-With-Two-Ns, line 27: Could probably just not set these for this subproject , but this is fine to

temporal-spring-boot-autoconfigure/README.md
```diff
@@ -1,4 +1,7 @@
-# Temporal Spring Boot
+# Temporal Spring Boot (Spring Boot 2/3)
+
+> **This module supports Spring Boot 2.7.x and 3.x only.**
```
Quinn-With-Two-Ns, line 3: Can we also have an issue to update the docs as well https://docs.temporal.io/develop/java/spring-boot-integration

PR #2787, opened 2026-02-17
Add Spring Boot 4 support

### LINKED ISSUE #2738: Spring Boot 4 Support

### LINKED ISSUE #2758: Spring Metrics Configuration fails with Spring Boot 4
## Expected Behavior

`temporal_` metrics emitted properly.

## Actual Behavior

Metrics get emitted to a no-op scope, and not to the underlying Micrometer MeterRegistry.

This occurs because metrics scope bean fails to configure properly before being supplied to other temporal configuration. It waits to run after `CompositeMeterRegistryAutoConfiguration` (https://github.com/temporalio/sdk-java/blob/2322bd0da735fe310b7291f3b7fbeda10bdf5e40/temporal-spring-boot-autoconfigure/src/main/java/io/temporal/spring/boot/autoconfigure/MetricsScopeAutoConfiguration.java#L18), however the package for this class changed to `org.springframework.boot.micrometer.metrics.autoconfigure.CompositeMeterRegistryAutoConfiguration` in Spring Boot 4.

  1. Update application to Spring Boot 4
  2. Check if metrics are emitted to Micrometer properly

## Specifications

  - Version: 1.32.1
  - Platform: Mac/Linux

cretz:
## What was changed

* Updated Gradle build files and added CI test to ensure Spring 4 works
  * Did not have to change code
* Updated `MetricsScopeAutoConfiguration` and `OpenTracingAutoConfiguration` to point to the newer Spring 4 class names
  * Also added test to that would have broken on Spring 4 without these changes

## Checklist

1. Closes #2738
2. Closes #2758

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,181 +13,193 @@ jobs:
     runs-on: ubuntu-latest-16-cores
     timeout-minutes: 30
     steps:
-    - name: Checkout repo
-      uses: actions/checkout@v5
-      with:
-        fetch-depth: 0
-        submodules: recursive
-        ref: ${{ github.event.pull_request.head.sha }}
+      - name: Checkout repo
+        uses: actions/checkout@v5
+        with:
+          fetch-depth: 0
+          submodules: recursive
+          ref: ${{ github.event.pull_request.head.sha }}
 
-    - name: Set up Java
-      uses: actions/setup-java@v5
-      with:
-        java-version: "23"
-        distribution: "temurin"
+      - name: Set up Java
+        uses: actions/setup-java@v5
+        with:
+          java-version: "23"
+          distribution: "temurin"
 
-    - name: Set up Gradle
-      uses: gradle/actions/setup-gradle@v5
-    
-    - name: Run unit tests
-      env:
-        USER: unittest
-        USE_DOCKER_SERVICE: false
-      run: ./gradlew --no-daemon test -x spotlessCheck -x spotlessApply -x spotlessJava -P edgeDepsTest
+      - name: Set up Gradle
+        uses: gradle/actions/setup-gradle@v5
 
-    - name: Run independent resource tuner test
-      env:
-        USER: unittest
-        USE_DOCKER_SERVICE: false
-      run: ./gradlew --no-daemon temporal-sdk:testResourceIndependent -x spotlessCheck -x spotlessApply -x spotlessJava -P edgeDepsTest
+      - name: Run unit tests
+        env:
+          USER: unittest
+          USE_DOCKER_SERVICE: false
+        run: ./gradlew --no-daemon test -x spotlessCheck -x spotlessApply -x spotlessJava -P edgeDepsTest
 
-    - name: Publish Test Report
-      uses: mikepenz/action-junit-report@v6
-      if: success() || failure() # always run even if the previous step fails
-      with:
-        report_paths: '**/build/test-results/test/TEST-*.xml'
+      - name: Run independent resource tuner test
+        env:
+          USER: unittest
+          USE_DOCKER_SERVICE: false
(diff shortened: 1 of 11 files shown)


.github/workflows/ci.yml
```diff
@@ -13,181 +13,193 @@ jobs:
...
+            23
+            11
+          distribution: "temurin"
+
+      - name: Set up Gradle
+        uses: gradle/actions/setup-gradle@v5
+
+      - name: Start containerized server and dependencies
+        env:
+          TEMPORAL_CLI_VERSION: 1.6.1-server-1.31.0-151.0
```
mjameswh, line 87: [unrelated] Seems like removing this line is causing test flakes in Nexus tests, not sure why.

gradle/versioning.gradle
```diff
@@ -55,8 +55,6 @@ group = 'io.temporal'
 version = getVersionName()
 
 subprojects {
-    apply plugin: "com.palantir.git-version"
```
mjameswh, line 58: Unrelated, but this was causing Gradle error messages in my environment because `jgit` used internally by `com.palantir.git-version` doesn't know how to deal with Git worktrees. AFAICS, we're anyway relying on CLI calls to `git`, rather than use APIs provided by that plugin.

temporal-spring-boot-autoconfigure/src/test/java/io/temporal/spring/boot/autoconfigure/AutoConfigOrderingTest.java
```diff
@@ -0,0 +1,109 @@
...
+ * Tests that {@link MetricsScopeAutoConfiguration} and {@link OpenTracingAutoConfiguration}
+ * correctly order themselves after their dependency auto-configurations via
+ * {@code @AutoConfiguration(afterName=...)}.
+ *
+ * <p>Without correct ordering, {@code @ConditionalOnBean} evaluates before the dependency beans
+ * exist, silently skipping our entire auto-configuration. This happens because our FQCNs ({@code
+ * io.temporal...}) sort alphabetically before Spring's ({@code org.springframework...}), which is
+ * the default processing order when no ordering constraints are specified.
+ */
+class AutoConfigOrderingTest {
```
Quinn-With-Two-Ns, line 27: These tests just seem to confirm that the dependency was create and ordered correctly. I don't think this test verifies that metrics or traces will actually be emitted properly?

mjameswh: I enriched the tests a bit.

gianny82:
Hi,

I still see version 2.7.18 of `org.springframework.boot` `spring-boot-starter` in `temporal-spring-boot-starter` and `temporal-spring-boot-autoconfigure`, is that correct or should we update the version there?

[temporal-spring-boot-starter](https://mvnrepository.com/artifact/io.temporal/temporal-spring-boot-starter/1.34.0/dependencies)
[temporal-spring-boot-autoconfigure](https://mvnrepository.com/artifact/io.temporal/temporal-spring-boot-autoconfigure/1.34.0/dependencies)

PR #2789, opened 2026-02-18
fix nexus-service-bean explicitConfig support

mnichols:
It is not possible to register Nexus services via explicit config today (`nexus-service-beans`).

## Expected Behavior
`@Service` implementations should be registered on a Temporal Worker without using AutoConfiguration (aka explicit configuration).

The problem is, Nexus Spring registration is not fully tested.
This config fails in `ExplicitConfigTest`

```
spring:
  config:
    activate:
      on-profile: explicit-config
  temporal:
    workers:
      - task-queue: UnitTest
        workflow-classes:
          - io.temporal.spring.boot.autoconfigure.bytaskqueue.TestWorkflowImpl
        activity-beans:
          - TestActivityImpl
       # this should register the service impl but fails. See stack trace
        nexus-service-beans:
          - TestNexusServiceImpl
```

## Actual Behavior

When bringing up a worker doing explicit config only (spring), the `@ServiceImpl` attribute is not found while registering the service even when the attribute is present and correct.

```
Caused by:
                java.lang.IllegalArgumentException: Missing @ServiceImpl annotation
                    at io.nexusrpc.handler.ServiceImplInstance.fromInstance(ServiceImplInstance.java:25)
                    at io.temporal.spring.boot.autoconfigure.template.WorkersTemplate.lambda$createWorkerFromAnExplicitConfig$7(WorkersTemplate.java:400)
```

  1. Actually use explicit config with Nexus somewhere in tests that has `nexus-service-beans`.
  
  **The problem** is we are using AopUtils to unpack the bean before getting the @ServiceImpl definition but that is passing the class, not the implementation.
  
  The fix is to just pass the bean in directly and tests pass.

## Specifications

  - Version:
  - Platform:
[stack.txt](https://github.com/user-attachments/files/25397135/stack.txt)

See tests

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/common/SearchAttributePayloadConverter.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/common/SearchAttributePayloadConverter.java
@@ -331,13 +331,19 @@ final class SearchAttributePayloadConverter {
 
   @Nullable
   private static IndexedValueType encodedValueToIndexValueType(String encodedValue) {
+    // The type metadata is usually in PascalCase (e.g. "KeywordList") but in rare cases may be in
+    // SCREAMING_SNAKE_CASE (e.g. "INDEXED_VALUE_TYPE_KEYWORD_LIST").
     try {
       return IndexedValueType.valueOf(
           ProtoEnumNameUtils.simplifiedToUniqueName(
               encodedValue, ProtoEnumNameUtils.INDEXED_VALUE_TYPE_PREFIX));
     } catch (IllegalArgumentException e) {
-      log.warn("[BUG] No IndexedValueType mapping for {} value exist", encodedValue);
-      return null;
+      try {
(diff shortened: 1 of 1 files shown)


mnichols:
By the way, if you need to workaround this bug right now you can just autodiscover on the `services` package (if you've isolated your Nexus services) and do explicit config for the rest. Eg

```yaml
  temporal:
    connection:
      # Connection settings
      target: ${TEMPORAL_ADDRESS:localhost:7233}

      # API Key for Temporal Cloud
      api-key: ${TEMPORAL_API_KEY:}

    start-workers: true
    namespace: ${TEMPORAL_NAMESPACE:default}
    workers:
      - task-queue: processing
        workflow-classes:
          - com.acme.processing.workflows.OrderImpl
#        nexus-service-beans:
#          - processing-service
    workers-auto-discovery:
      packages: com.acme.processing.services
      ```

PR #2799, opened 2026-02-24
Cancel timer when Workflow.await condition is satisfied

### LINKED ISSUE #2312: `Workflow.await(duration, condition)` does not automatically cancel the timer if the condition is resolved
`Workflow.await(duration, condition)` should automatically cancel the timer when the condition is resolved

User can currently do this manually with a cancellation scope around the `await`

**Additional context**
This will need to be versioned with SDK flags to avoid breaking determinism.

mfateev:
## Summary
- `Workflow.await(duration, condition)` now cancels the timer when the condition is satisfied before the timeout expires
- Prevents unnecessary workflow tasks caused by timer fired events
- If condition is already true when called, returns immediately without creating a timer

## Background
Fixes #2312. Port of the same fix from the Go SDK: https://github.com/temporalio/sdk-go/pull/2153

## Implementation
- Added `CANCEL_AWAIT_TIMER_ON_CONDITION(4)` SDK flag for replay compatibility
- When the flag is enabled, the timer is created inside a `CancellationScope` so it can be cancelled when the condition resolves
- Uses `checkSdkFlag` (not `tryUseSdkFlag`) so the flag is **not auto-enabled** for new workflows in this release — it must wait at least 1 release before being enabled by default, per SDK flag rollout policy
- TODO comment marks where to switch to `tryUseSdkFlag` in the next release

## Test plan
- [x] `testTimerCancelledWhenFlagEnabled` — explicitly enables the flag, verifies `TIMER_CANCELED` in history
- [x] `testTimerNotCancelledWhenFlagDisabled` — default (flag off), verifies timer is NOT cancelled (old behavior)
- [x] `testNoTimerWhenConditionImmediatelySatisfiedWithFlag` — verifies no timer created when condition is already true
- [x] `testAwaitReturnValue` — verifies return value semantics (true=condition met, false=timeout)
- [x] `testReplayOldHistoryWithoutFlag` — replay compatibility with old workflow histories recorded without the flag

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflowContext.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflowContext.java
@@ -1322,14 +1322,12 @@ final class SyncWorkflowContext implements WorkflowContext, WorkflowOutboundCall
     boolean cancelTimerOnCondition =
         replayContext.checkSdkFlag(SdkFlag.CANCEL_AWAIT_TIMER_ON_CONDITION);
 
-    // If new behavior is enabled and condition is already satisfied, skip creating timer
-    if (cancelTimerOnCondition && unblockCondition.get()) {
-      return true;
-    }
-
     if (cancelTimerOnCondition) {
-      // New behavior: create timer in a cancellation scope so we can cancel it when condition is
-      // satisfied
+      // If condition is already satisfied, skip creating timer
+      if (unblockCondition.get()) {
+        return true;
+      }
+      // Create timer in a cancellation scope so we can cancel it when condition is satisfied
       CompletablePromise<Void> timer = Workflow.newPromise();
       CancellationScope timerScope =
           Workflow.newCancellationScope(() -> timer.completeFrom(newTimer(timeout)));
--- a/temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflowContext.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflowContext.java
@@ -1317,9 +1317,35 @@ final class SyncWorkflowContext implements WorkflowContext, WorkflowOutboundCall
 
   @Override
   public boolean await(Duration timeout, String reason, Supplier<Boolean> unblockCondition) {
-    Promise<Void> timer = newTimer(timeout);
-    WorkflowThread.await(reason, () -> (timer.isCompleted() || unblockCondition.get()));
-    return !timer.isCompleted();
+    boolean cancelTimerOnCondition =
+        replayContext.tryUseSdkFlag(SdkFlag.CANCEL_AWAIT_TIMER_ON_CONDITION);
+
+    // If new behavior is enabled and condition is already satisfied, skip creating timer
+    if (cancelTimerOnCondition && unblockCondition.get()) {
+      return true;
+    }
+
+    if (cancelTimerOnCondition) {
+      // New behavior: create timer in a cancellation scope so we can cancel it when condition is
+      // satisfied
+      CompletablePromise<Void> timer = Workflow.newPromise();
+      CancellationScope timerScope =
+          Workflow.newCancellationScope(() -> timer.completeFrom(newTimer(timeout)));
+      timerScope.run();
+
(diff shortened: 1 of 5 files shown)


temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflowContext.java
```diff
@@ -1317,9 +1317,37 @@ public void sleep(Duration duration) {
...
-    Promise<Void> timer = newTimer(timeout);
-    WorkflowThread.await(reason, () -> (timer.isCompleted() || unblockCondition.get()));
-    return !timer.isCompleted();
+    // TODO: Change checkSdkFlag to tryUseSdkFlag in the next release to enable this flag by
+    // default.
+    boolean cancelTimerOnCondition =
+        replayContext.checkSdkFlag(SdkFlag.CANCEL_AWAIT_TIMER_ON_CONDITION);
+
+    // If new behavior is enabled and condition is already satisfied, skip creating timer
+    if (cancelTimerOnCondition && unblockCondition.get()) {
```
mjameswh: nit: I'd move this block into the first branch of the `if (cancelTimerOnCondition) {` below

mfateev: Done

temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflowContext.java
```diff
@@ -1317,9 +1317,35 @@ public void sleep(Duration duration) {
...
+      CompletablePromise<Void> timer = Workflow.newPromise();
+      CancellationScope timerScope =
+          Workflow.newCancellationScope(() -> timer.completeFrom(newTimer(timeout)));
+      timerScope.run();
+
+      WorkflowThread.await(reason, () -> (timer.isCompleted() || unblockCondition.get()));
+
+      boolean conditionSatisfied = !timer.isCompleted();
+      if (conditionSatisfied) {
+        timerScope.cancel("await condition resolved");
```
mjameswh, line 1344: I'm not sure about providing a reason message here. AFAICS, we're not currently providing reason messages in similar context anywhere in the Java SDK, and we don't do that either in the Go SDK, which was used as reference for this PR.

But no strong opinion either way.

temporal-sdk/src/test/java/io/temporal/workflow/cancellationTests/WorkflowAwaitCancelTimerOnConditionTest.java
```diff
@@ -0,0 +1,228 @@
...
+   * CANCEL_AWAIT_TIMER_ON_CONDITION flag.
+   */
+  @Test
+  public void testReplayOldHistoryWithoutFlag() throws Exception {
+    WorkflowReplayer.replayWorkflowExecutionFromResource(
+        "awaitTimerConditionOldBehavior.json", TestAwaitCancelTimerWorkflowImpl.class);
+  }
+
+  @WorkflowInterface
+  public interface TestAwaitWorkflow {
```
mjameswh, line 167: nit: I believe we generally try to deduplicate workflow interfacess in our tests.

mjameswh (approved):
Two very minor comments, but overall looks correct to me.

EugenePopov:
Hi @mjameswh @mfateev — `CANCEL_AWAIT_TIMER_ON_CONDITION` from this PR shipped gated-off in 1.34.0 and is still wired with `checkSdkFlag` (not `tryUseSdkFlag`) on `master` as of 1.36.0, with the `// TODO: Change checkSdkFlag to tryUseSdkFlag in the next release` still present in `SyncWorkflowContext.await`.

The PR notes the flag should "wait at least 1 release before being enabled by default, per SDK flag rollout policy." Since two releases have now shipped since it landed (1.34 → 1.36), a couple of questions:

- Is the switch to `tryUseSdkFlag` (default-on for new executions) already scheduled for an upcoming release, or tracked somewhere I can follow/subscribe to?
- Is enablement gated on anything else?

rocketraman:
Created follow-up issue https://github.com/temporalio/sdk-java/issues/3012.

PR #2821, opened 2026-03-30
Wait for MARKER_RECORDED to fire version callback on replay

### LINKED ISSUE #2796: Replay fails: UpdateCompleted events between version markers breaks replay
## Expected Behavior
When an Update handler contains a SideEffect or LocalActivity and completes during the same workflow task as getVersion markers, replay should succeed deterministically without errors.

## Actual Behavior
A workflow task failure occurs during replay with the following error:

```
[TMPRL1100] getVersion call before the existing version marker event. The most probable cause is retroactive addition of a getVersion call with an existing 'changeId'
```
```
Version: failure executing RESULT_NOTIFIED_REPLAYING->NON_MATCHING_EVENT, transition history is [CREATED->CHECK_EXECUTION_STATE, REPLAYING->SCHEDULE, MARKER_COMMAND_CREATED_REPLAYING->RECORD_MARKER]
```
```
Failure handling event 26 of type 'EVENT_TYPE_WORKFLOW_EXECUTION_UPDATE_COMPLETED' during replay. {WorkflowTaskStartedEventId=41, CurrentStartedEventId=3}
The UpdateCompleted events are interleaved between two version markers in history, causing a mismatch during replay.
```

- Start a workflow that uses getVersion markers (patching).
- Send Updates simultaneously while the workflow task is in flight.
- Update handler blocks on a SideEffect or LocalActivity.
- Observe that UpdateCompleted events are written between the two version marker events in history.
- Trigger replay — workflow task fails with the error above.

## Additional Context

Workaround: using `Workflow.await(Workflow::isEveryHandlerFinished)` before getVersion calls prevents the race, but this feels like a workaround rather than a fix.
The SDK should replay in the same order it initially executed — this failure suggests a bug in the SDK state machine when Update completions interleave with version markers.

Image

some sample code here - https://github.com/gauravthadani/samples-kotlin/tree/main/app/src/main/kotlin/io/temporal/samples/update_nde

## Specifications

  - Version:
  - Platform:

eamsden:
## What was changed

- Changed `VersionStateMachine` replay behavior behind a new SDK flag `VERSION_WAIT_FOR_MARKER` so `getVersion()` no longer resumes workflow code when the fake `RECORD_MARKER` command is created during replay.
- Replay now waits until the real `MARKER_RECORDED` event is matched before firing the version callback.

## Why?

Replay was resuming `getVersion()` too early relative to update completion protocol handling. In the interleaved update history from issue `#2796`, that allowed replay to queue the second version marker before the update completion work had been matched, which produced a replay-only nondeterminism error:

`[TMPRL1100] getVersion call before the existing version marker event`

Delaying the replay callback until the actual marker event is matched makes version replay ordering consistent with replayed side effects and fixes replay for the recorded history.

## Checklist

1. Closes #2796 **but does not fix replay of existing histories**.

2. How was this tested:

- `./gradlew --offline :temporal-sdk:test --tests "io.temporal.workflow.versionTests.GetVersionMultithreadingRemoveTest" --tests "io.temporal.workflow.versionTests.GetVersionRemovedInReplayTest" --tests "io.temporal.workflow.versionTests.GetVersionWithoutCommandEventTest" --tests "io.temporal.workflow.versionTests.GetVersionAndTimerTest" --tests "io.temporal.workflow.versionTests.GetVersionMultipleCallsTest" --tests "io.temporal.workflow.versionTests.GetVersionInSignalTest" --tests "io.temporal.workflow.versionTests.GetVersionMultithreadingTest" --tests "io.temporal.workflow.versionTests.GetVersionInterleavedUpdateReplayTest" --tests "io.temporal.internal.replay.GetVersionInterleavedUpdateReplayTaskHandlerTest"`
- `./gradlew --offline :temporal-sdk:test --tests "io.temporal.workflow.versionTests.GetVersionInterleavedUpdateReplayTest" --tests "io.temporal.internal.replay.GetVersionInterleavedUpdateReplayTaskHandlerTest"`
- Verified that the original repro history now replays successfully through both `WorkflowReplayer` and the lower-level direct-query replay task handler.

3. Any docs updates needed?

- No docs updates needed. This is an internal replay-ordering fix with regression tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,5 @@
 name: Continuous Integration
 permissions:
-  checks: write
   contents: read
 on:
   pull_request:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,6 @@
 name: Continuous Integration
 permissions:
+  checks: write
   contents: read
 on:
   pull_request:
--- a/temporal-sdk/src/main/java/io/temporal/internal/common/SdkFlag.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/common/SdkFlag.java
@@ -28,6 +28,12 @@ public enum SdkFlag {
   /*
    * Changes replay behavior of GetVersion to wait for the matching marker event before executing
    * the callback.
+   *
+   * Introduced: 1.34.0
+   *
+   * Enabled: (pending)
+   *
+   * Bug: https://github.com/temporalio/sdk-java/issues/2796
    */
   VERSION_WAIT_FOR_MARKER(5),
   UNKNOWN(Integer.MAX_VALUE);
--- a/temporal-sdk/src/main/java/io/temporal/internal/common/SdkFlag.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/common/SdkFlag.java
@@ -25,6 +25,11 @@ public enum SdkFlag {
    * condition is resolved before the timeout.
    */
   CANCEL_AWAIT_TIMER_ON_CONDITION(4),
+  /*
+   * Changes replay behavior of GetVersion to wait for the matching marker event before executing
+   * the callback.
+   */
+  VERSION_WAIT_FOR_MARKER(5),
   UNKNOWN(Integer.MAX_VALUE);
 
   private final int value;
--- a/temporal-sdk/src/main/java/io/temporal/internal/common/SdkFlag.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/common/SdkFlag.java
@@ -29,7 +29,7 @@ public enum SdkFlag {
    * Changes replay behavior of GetVersion to wait for the matching marker event before executing
    * the callback.
    *
-   * Introduced: 1.34.0
+   * Introduced: 1.36.0
    *
    * Enabled: (pending)
    *
--- a/temporal-sdk/src/test/java/io/temporal/workflow/versionTests/GetVersionInterleavedUpdateReplayTest.java
+++ b/temporal-sdk/src/test/java/io/temporal/workflow/versionTests/GetVersionInterleavedUpdateReplayTest.java
@@ -19,6 +19,7 @@ import io.temporal.internal.common.SdkFlag;
 import io.temporal.internal.history.VersionMarkerUtils;
 import io.temporal.internal.statemachines.WorkflowStateMachines;
 import io.temporal.testing.TestWorkflowEnvironment;
+import io.temporal.testing.WorkflowHistoryLoader;
 import io.temporal.testing.WorkflowReplayer;
 import io.temporal.testing.internal.SDKTestWorkflowRule;
 import io.temporal.worker.Worker;
@@ -44,6 +45,8 @@ import org.slf4j.Logger;
 public class GetVersionInterleavedUpdateReplayTest {
   private static final String HISTORY_RESOURCE =
       "testGetVersionInterleavedUpdateReplayHistory.json";
+  private static final String WAIT_FOR_MARKER_HISTORY_RESOURCE =
+      "testGetVersionInterleavedUpdateReplayWaitForMarkerHistory.json";
   public static final String TASK_QUEUE = "get-version-interleaved-update-replay";
   private static final String EXPECTED_FIRST_CHANGE_ID = "ChangeId1";
   private static final String EXPECTED_SECOND_CHANGE_ID = "ChangeId2";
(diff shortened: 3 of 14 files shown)


temporal-sdk/src/test/java/io/temporal/workflow/versionTests/GetVersionInterleavedUpdateReplayTest.java
```diff
@@ -0,0 +1,232 @@
...
+   * the histories produced by the current branch.
+   *
+   * <p>We keep the fixture around for reference, but do not execute it as part of the suite. Making
+   * this exact history replay again would require changing replay behavior for histories that did
+   * not record the newer flags, which may break other existing replays. The fix is to put the
+   * state-machine behavior change behind an SDK flag {@link SdkFlag#VERSION_WAIT_FOR_MARKER}, and
+   * to make sure new workflows run with {@link SdkFlag#SKIP_YIELD_ON_VERSION} by default to avoid
+   * interleaved histories.
+   */
+  @Ignore("Recorded history predates SKIP_YIELD_ON_VERSION. Use the live-history replay test.")
```
mjameswh: Why ignore _this_ test? Given that the history doesn't contain the flag, it should replay exactly as it did originally. That's specifically the reason for such replay test.

temporal-sdk/src/main/java/io/temporal/internal/common/SdkFlag.java
```diff
@@ -23,6 +23,11 @@ public enum SdkFlag {
    * condition is resolved before the timeout.
    */
   CANCEL_AWAIT_TIMER_ON_CONDITION(4),
+  /*
+   * Changes replay behavior of GetVersion to wait for the matching marker event before executing
+   * the callback.
+   */
+  VERSION_WAIT_FOR_MARKER(5),
```
mjameswh, line 38: We're getting more of these. We should think ahead at how to make sure this remains maintainable in the future.

Would you mind adding some comments on this flag and others above, indicating:
- Which release introduced support for the flag (assume next patch release for new flags)
- Which release turned it on (or "pending" if not yet turned on)
- Ticket to the bug that the flag resolves.

temporal-sdk/src/test/java/io/temporal/workflow/versionTests/GetVersionInterleavedUpdateReplayTest.java
mjameswh: Also add a replay test where the history contains the flag and consequent events.

That is:
1. change the code on your local machine to make the feature flag turned on by default
2. run the test workflow
3. capture that workflow history to a JSON file in the resources directory
4. revert back the change you did in 1
5. add a replay test that uses the history file captured in 3

That test sequence basically correspond to a user that would execute a workflow on a version of the SDK where the flag is enabled, but then rollback their deployment to a version of the SDK where the flag is _supported but not enabled by default_.

temporal-sdk/src/main/java/io/temporal/internal/common/SdkFlag.java
eamsden: I have done this.

temporal-sdk/src/test/java/io/temporal/workflow/versionTests/GetVersionInterleavedUpdateReplayTest.java
eamsden: This is done in 515c90dd95c2b37e1e4a0227920eeadbfc32c812

.github/workflows/ci.yml
```diff
@@ -1,5 +1,6 @@
 name: Continuous Integration
 permissions:
+  checks: write
```
mjameswh: Why?

temporal-sdk/src/main/java/io/temporal/internal/common/SdkFlag.java
```diff
@@ -25,6 +25,17 @@ public enum SdkFlag {
    * condition is resolved before the timeout.
    */
   CANCEL_AWAIT_TIMER_ON_CONDITION(4),
+  /*
+   * Changes replay behavior of GetVersion to wait for the matching marker event before executing
+   * the callback.
+   *
+   * Introduced: 1.34.0
```
mjameswh:
```suggestion
   * Introduced: 1.36.0
```

.github/workflows/ci.yml
eamsden: It was part of fixing CI flakes. I think in this case It was for 

https://github.com/temporalio/sdk-java/pull/2821/changes/BASE..ff8f61ae3cd4cbc030ec2469d1f63eb65458b39d#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fL56

to reliably succeed, but I'm checking that recall.

eamsden: It turns out not to be necessary, sorry this slipped through.

mjameswh (approved):

PR #2829, opened 2026-04-06
Add temporal-spring-ai module

donald-pinckney:
## Summary

Adds `temporal-spring-ai`, a new module that integrates Spring AI with Temporal workflows. AI model calls become durable activities; tools can be activity stubs, `@SideEffectTool`-wrapped, or plain workflow code.

## Quick Start

```java
@WorkflowInit
public MyWorkflowImpl(String goal) {
    ActivityChatModel chatModel = ActivityChatModel.forDefault();
    WeatherActivity weather = Workflow.newActivityStub(WeatherActivity.class, opts);

    this.chatClient = TemporalChatClient.builder(chatModel)
            .defaultSystem("You are a helpful assistant.")
            .defaultTools(weather, new MyTools())
            .build();
}

@Override
public String run(String goal) {
    return chatClient.prompt().user(goal).call().content();
}
```

## Tool Types

Tools passed to `defaultTools()` are handled based on their type:

- **Activity stubs** — interfaces with `@ActivityInterface` + `@Tool`. Auto-detected, executed as durable activities.
- **`@SideEffectTool`** — wrapped in `Workflow.sideEffect()`. Use for timestamps, UUIDs, etc.
- **Plain tools** — any class with `@Tool` methods. Executes directly in workflow context; user is responsible for determinism.
- **Nexus service stubs** — auto-detected, executed as Nexus operations.

## Optional Integrations

Auto-configured when their dependencies are on the classpath. Can also be created programmatically.

| Feature | Dependency | What it registers |
|---|---|---|
| Vector Store | `spring-ai-rag` | `VectorStoreActivity` |
| Embeddings | `spring-ai-rag` | `EmbeddingModelActivity` |
| MCP | `spring-ai-mcp` | `McpClientActivity` |

## Compatibility

| Dependency | Minimum Version |
|---|---|
| Java | 17 |
| Spring Boot | 3.x |
| Spring AI | 1.1.0 |
| Temporal Java SDK | 1.33.0 |

The Java 17 floor is transitive: Spring AI 1.1 depends on Spring Boot 3.x (the 1.1.2 upgrade notes explicitly call out Spring Boot 3.3.x / 3.4.2 as supported — [upgrade-notes.adoc](https://github.com/spring-projects/spring-ai/blob/v1.1.2/spring-ai-docs/src/main/antora/modules/ROOT/pages/upgrade-notes.adoc)), and Spring Boot 3.x has a hard Java 17 minimum ([Spring Boot 3.0 release notes](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.0-Release-Notes)).

## Build infrastructure changes

Spring AI requires Java 17+, but the rest of the SDK targets Java 8 bytecode and the release/snapshot publish jobs run on JDK 11. An earlier iteration of this PR conditionally excluded the whole `temporal-spring-ai` subproject when Gradle ran on <JDK 17, which silently dropped the artifact from both publish jobs and omitted it from the BOM.

Replaced with the same pattern `temporal-sdk` already uses for its Java 17/21 source sets — resolve a JDK 17+ compiler/javadoc/launcher via Gradle toolchains when the running JVM is older. Concretely:

- **`temporal-spring-ai/build.gradle`** — `compileJava`, `compileTestJava`, `javadoc`, and `test` fall back to `javaToolchains.{compilerFor, javadocToolFor, launcherFor}` with `languageVersion = 17` when `JavaVersion.current()` is older. Uses `options.release = 17` instead of mutating `compilerArgs`.
- **`settings.gradle` / `temporal-bom/build.gradle`** — dropped the `if (JavaVersion.current().isCompatibleWith(VERSION_17))` guards. `temporal-spring-ai` is now always included in the build and the BOM.
- **`gradle/java.gradle`** — the shared `javadoc` block used to pick flags (`html5`, `--no-module-directories`) based on `JavaVersion.current()`. That's wrong for any module that overrides `javadocTool` via a toolchain (`temporal-spring-ai` on JDK 11), because it would pass JDK 9–12-only flags to a JDK 17 javadoc tool and fail. Switched to `javadocTool.get().metadata.languageVersion` so the decision tracks the tool actually being invoked. Behavior is unchanged for every other module (their tool equals the current JVM).
- **`.github/workflows/prepare-release.yml`, `.github/workflows/publish-snapshot.yml`** — both publish jobs now install JDK 17 alongside JDK 11. JDK 11 remains the default `JAVA_HOME` (last entry in the multi-line `java-version` list), so Gradle still runs on 11; JDK 17 is present only so the toolchain resolver can locate it. Gradle toolchains match the major version exactly, so without an installed JDK 17 the `languageVersion = 17` lookup fails — this was reproduced locally on a box that had only JDK 11 and 21 installed.

Verified locally by running `JAVA_HOME=<jdk11> ./gradlew :temporal-spring-ai:build` and `:test`: compile, javadoc, javadocJar, sourcesJar, and the full test suite all succeed on JDK 11 with a JDK 17 sidecar.

## Related
- Samples: temporalio/samples-java#775
- Workshop demos: temporal-community/springio-agents-springai-temporal#1


temporal-spring-ai/src/main/java/io/temporal/springai/tool/LocalActivityToolCallbackWrapper.java
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  @Override
+  public ToolMetadata getToolMetadata() {
+    return delegate.getToolMetadata();
+  }
+
+  @Override
+  public String call(String toolInput) {
+    String callbackId = UUID.randomUUID().toString();
```
donald-pinckney: **High severity — non-deterministic call in workflow context.**

`UUID.randomUUID()` is called from the `call()` method, which runs on the workflow thread. This violates Temporal's replay constraints — on replay a different UUID is generated.

In practice this may work because the local activity replays from markers rather than re-executing, but the intent is unclear and it depends on local activity replay semantics. Use `Workflow.randomUUID()` instead.

temporal-spring-ai/src/main/java/io/temporal/springai/tool/LocalActivityToolCallbackWrapper.java
```diff
@@ -0,0 +1,128 @@
...
+ * <p><b>Memory Management:</b> Callbacks are automatically removed from the map after execution to
+ * prevent memory leaks.
+ *
+ * <p>This class is primarily used by {@code SandboxingAdvisor} to wrap unsafe tools.
+ *
+ * @see ExecuteToolLocalActivity
+ */
+public class LocalActivityToolCallbackWrapper implements ToolCallback {
+
+  private static final Map<String, ToolCallback> CALLBACK_REGISTRY = new ConcurrentHashMap<>();
```
donald-pinckney: **Medium severity — static registry lifecycle risk.**

This static `ConcurrentHashMap` holds live `ToolCallback` references. Callbacks are removed in a `finally` block after the local activity completes, but if the workflow is evicted from the worker cache mid-execution (before `finally` runs), callbacks leak. Worth documenting or adding periodic cleanup.

temporal-spring-ai/src/main/java/io/temporal/springai/util/TemporalStubUtil.java
```diff
@@ -0,0 +1,81 @@
...
+   * @param object the object to check
+   * @return true if the object is an activity stub
+   */
+  public static boolean isActivityStub(Object object) {
+    return object != null
+        && Proxy.isProxyClass(object.getClass())
+        && Proxy.getInvocationHandler(object)
+            .getClass()
+            .getName()
+            .contains("ActivityInvocationHandler")
```
donald-pinckney: **Medium severity — fragile string matching.**

Checking `getInvocationHandler().getClass().getName().contains("ActivityInvocationHandler")` is brittle. If the SDK renames or refactors these internal handler classes, detection silently breaks and tools fall through to the "unknown type" error. Since this plugin now lives in the SDK repo, consider using internal SDK APIs directly, or at minimum add a comment acknowledging the coupling.

temporal-spring-ai/src/main/java/io/temporal/springai/model/ActivityChatModel.java
```diff
@@ -0,0 +1,376 @@
...
+
+      if (toolExecutionResult.returnDirect()) {
+        // Return tool execution result directly
+        return ChatResponse.builder()
+            .from(response)
+            .generations(ToolExecutionResult.buildGenerations(toolExecutionResult))
+            .build();
+      } else {
+        // Send tool results back to the model (recursive call)
+        return call(new Prompt(toolExecutionResult.conversationHistory(), prompt.getOptions()));
```
donald-pinckney: **Medium severity — unbounded recursive tool loop.**

This recursively calls `call()` when the model keeps requesting tools. A misbehaving model could cause infinite recursion / stack overflow. Consider adding a max iteration count. Spring AI's default `ToolCallingManager` has one internally, but since the loop is driven manually here, that limit is bypassed.

temporal-spring-ai/src/main/java/io/temporal/springai/plugin/SpringAiPlugin.java
```diff
@@ -0,0 +1,406 @@
...
+    }
+    pendingMcpWorkers.clear();
+  }
+
+  /** Registers McpClientActivity with a worker using reflection to avoid MCP class dependencies. */
+  private void registerMcpActivity(
+      Worker worker, List<?> clients, List<String> registeredActivities) {
+    try {
+      // Use reflection to avoid loading MCP classes when not on classpath
+      Class<?> mcpActivityClass = Class.forName("io.temporal.springai.mcp.McpClientActivityImpl");
```
donald-pinckney: **Low severity — unnecessary reflection.**

`Class.forName("io.temporal.springai.mcp.McpClientActivityImpl")` uses reflection to avoid a compile-time dependency, but `McpClientActivityImpl` is in the same module. This reflection is avoiding a dependency that already exists. If MCP is truly optional, consider moving it to a separate module. If not, just use direct references.

temporal-spring-ai/build.gradle
```diff
@@ -0,0 +1,54 @@
...
+dependencies {
+    api(platform("org.springframework.boot:spring-boot-dependencies:$springBootVersionForSpringAi"))
+    api(platform("org.springframework.ai:spring-ai-bom:$springAiVersion"))
+
+    // this module shouldn't carry temporal-sdk with it, especially for situations when users may be using a shaded artifact
+    compileOnly project(':temporal-sdk')
+    compileOnly project(':temporal-spring-boot-autoconfigure')
+
+    api 'org.springframework.boot:spring-boot-autoconfigure'
+    api 'org.springframework.ai:spring-ai-client-chat'
```
donald-pinckney, line 34: **High severity — `compileOnly` deps needed at runtime.**

`VectorStore`, `EmbeddingModel`, and MCP are `compileOnly` here, but `SpringAiPlugin` directly references these classes in field declarations and method signatures. When they're not on the runtime classpath, Spring fails to introspect the class at all (`ClassNotFoundException: org.springframework.ai.vectorstore.VectorStore`).

Either:
1. Change to `implementation` (simplest, but makes them transitive)
2. Split `SpringAiPlugin` so VectorStore/Embedding/MCP handling is in separate classes loaded conditionally
3. Use `runtimeOnly` in consumer builds (what we had to do in the samples)

temporal-spring-ai/src/main/java/io/temporal/springai/model/ActivityChatModel.java
```diff
@@ -0,0 +1,376 @@
...
+ *     // Use different models for different purposes
+ *     this.fastClient = TemporalChatClient.builder(openAiModel).build();
+ *     this.smartClient = TemporalChatClient.builder(anthropicModel).build();
+ * }
+ * }</pre>
+ *
+ * @see #forDefault()
+ * @see #forModel(String)
+ */
+public class ActivityChatModel implements ChatModel {
```
donald-pinckney, line 78: **Low severity — no streaming support.**

`ActivityChatModel` only implements synchronous `call()`. Streaming through activities doesn't make sense, but the class should override `stream()` to throw a clear `UnsupportedOperationException` rather than silently inheriting a default that may behave unexpectedly.

donald-pinckney (commented):
## Self-review: temporal-spring-ai plugin

### What's done well

**Determinism architecture is sound.** The core design — routing all non-deterministic operations (LLM calls, vector store ops, embeddings, MCP tools) through Temporal activities — is exactly right. The three-tier tool system (`@DeterministicTool` for pure functions, `@SideEffectTool` for cheap non-determinism, activity stubs for durable I/O) maps cleanly onto Temporal's primitives.

**Tool execution stays in the workflow.** `ChatModelActivityImpl` sets `internalToolExecutionEnabled(false)` and only passes tool *definitions* to the model. The actual tool dispatch happens back in the workflow via `ToolCallingManager`, which means tool calls respect their Temporal wrapping (activity, sideEffect, etc.).

**`SideEffectToolCallback`** correctly wraps each call in `Workflow.sideEffect()`, recording the result in history on first execution and replaying the stored value.

**`ActivityChatModel.call()` handles the tool loop correctly** — it recursively calls itself when the model requests tools that don't `returnDirect`, maintaining proper conversation history.

### Issues flagged inline below

temporal-spring-ai/build.gradle
donald-pinckney: TODO: go with option 2 (split into conditional auto-configuration classes).

Confirmed that Spring AI itself does NOT transitively pull in RAG, vector store, embeddings, or MCP — those are separate opt-in starters. We shouldn't be heavier than the framework we're integrating with.

Plan: keep `SpringAiPlugin` with only `ChatModel` (required). Add separate `@ConditionalOnClass`-guarded config classes for VectorStore, EmbeddingModel, and MCP that register their own activities. The `compileOnly` scope then stays correct — Spring skips the conditional classes entirely when the dep isn't present.

temporal-spring-ai/src/main/java/io/temporal/springai/tool/LocalActivityToolCallbackWrapper.java
donald-pinckney: Done — added javadoc documenting the eviction leak risk and pointing to `getRegisteredCallbackCount()` for monitoring.

temporal-spring-ai/src/main/java/io/temporal/springai/model/ActivityChatModel.java
donald-pinckney: Done — `stream(Prompt)` now throws `UnsupportedOperationException` with a clear message explaining that Temporal activities are request/response based.

temporal-spring-ai/src/main/java/io/temporal/springai/tool/LocalActivityToolCallbackWrapper.java
donald-pinckney: Test coverage added (WorkflowDeterminismTest). Ready to fix — next up.

temporal-spring-ai/src/main/java/io/temporal/springai/util/TemporalStubUtil.java
donald-pinckney: Test coverage added (TemporalToolUtilTest, 22 tests). Ready to refactor — next up.

temporal-spring-ai/src/main/java/io/temporal/springai/model/ActivityChatModel.java
donald-pinckney: Test coverage added (ChatModelActivityImplTest, 10 tests). Ready to add loop limit — next up.

temporal-spring-ai/src/main/java/io/temporal/springai/plugin/SpringAiPlugin.java
donald-pinckney: Test coverage added (SpringAiPluginTest, 10 tests). Ready for the conditional auto-config split (T6) — next up.

donald-pinckney (commented):
Found an additional bug during testing.

temporal-spring-ai/src/main/java/io/temporal/springai/model/ActivityChatModel.java
```diff
@@ -0,0 +1,390 @@
...
+        output.generations().stream()
+            .map(gen -> new Generation(toAssistantMessage(gen.message())))
+            .collect(Collectors.toList());
+
+    ChatResponseMetadata metadata = null;
+    if (output.metadata() != null) {
+      metadata = ChatResponseMetadata.builder().model(output.metadata().model()).build();
+    }
+
+    return ChatResponse.builder().generations(generations).metadata(metadata).build();
```
donald-pinckney: **Bug: NPE when metadata is null.**

When `output.metadata()` is null, `metadata` stays null and `.metadata(null)` is passed to `ChatResponse.builder()`. Spring AI's builder throws an NPE on null metadata.

Fix: only call `.metadata()` on the builder when metadata is non-null, or pass an empty `ChatResponseMetadata.builder().build()`.

temporal-spring-ai/src/main/java/io/temporal/springai/tool/LocalActivityToolCallbackWrapper.java
donald-pinckney: Done — replaced with `Workflow.randomUUID()`.

temporal-spring-ai/src/main/java/io/temporal/springai/model/ActivityChatModel.java
donald-pinckney: Done — converted recursive call to iterative loop with `MAX_TOOL_CALL_ITERATIONS = 10` limit.

temporal-spring-ai/build.gradle
donald-pinckney: Done — split into 4 auto-config classes: `SpringAiTemporalAutoConfiguration` (core), `SpringAiVectorStoreAutoConfiguration`, `SpringAiEmbeddingAutoConfiguration`, `SpringAiMcpAutoConfiguration`. Each guarded by `@ConditionalOnClass`. compileOnly scopes now work correctly.

temporal-spring-ai/src/main/java/io/temporal/springai/plugin/SpringAiPlugin.java
donald-pinckney: Done — resolved by T6. `McpPlugin` directly references `McpClientActivityImpl`, no more reflection.

temporal-spring-ai/src/main/java/io/temporal/springai/model/ActivityChatModel.java
donald-pinckney: Done — builder now only calls `.metadata()` when metadata is non-null.

temporal-spring-ai/src/main/java/io/temporal/springai/util/TemporalStubUtil.java
donald-pinckney: Done — replaced with `instanceof ActivityInvocationHandler`, `instanceof LocalActivityInvocationHandler`, `instanceof NexusServiceInvocationHandler`. ChildWorkflowInvocationHandler is package-private so it uses `endsWith()` on the class name (more precise than `contains()`).

TASK_QUEUE.json
donald-pinckney, line 1: Obviously will delete this prior to merging ;)

temporal-spring-ai/src/main/java/io/temporal/springai/model/ActivityChatModel.java
donald-pinckney: Reverted — removed the iteration limit and switched back to recursive `internalCall()` matching Spring AI's `OpenAiChatModel` pattern. Temporal's activity timeouts and workflow execution timeout already bound runaway tool loops.

temporal-spring-ai/src/main/java/io/temporal/springai/plugin/VectorStorePlugin.java
```diff
@@ -0,0 +1,34 @@
...
+import org.springframework.ai.vectorstore.VectorStore;
+
+/**
+ * Temporal plugin that registers {@link io.temporal.springai.activity.VectorStoreActivity} with
+ * workers.
+ *
+ * <p>This plugin is conditionally created by auto-configuration when Spring AI's {@link
+ * VectorStore} is on the classpath and a VectorStore bean is available.
+ */
+public class VectorStorePlugin extends SimplePlugin {
```
tconley1428, line 15: Is the primary reason for these being separate plugins because of spring configuration interaction? Otherwise I would think it makes more sense to have one with configurations.

temporal-spring-ai/src/main/java/io/temporal/springai/plugin/VectorStorePlugin.java
```diff
@@ -0,0 +1,34 @@
...
+
+  private static final Logger log = LoggerFactory.getLogger(VectorStorePlugin.class);
+
+  private final VectorStore vectorStore;
+
+  public VectorStorePlugin(VectorStore vectorStore) {
+    super("io.temporal.spring-ai-vectorstore");
```
tconley1428: Would it make more sense to use a SimplePlugin builder to create this? In other languages we would not need to manually override `initializeWorker`, I would expect there to be some way to use the simple plugin's `registerActivitiesImplementations` rather than the underlying worker.

temporal-spring-ai/src/main/java/io/temporal/springai/tool/LocalActivityToolCallbackWrapper.java
```diff
@@ -0,0 +1,132 @@
...
+ * reference will leak until the worker is restarted. This is bounded by the number of concurrent
+ * in-flight tool calls and is unlikely to be a practical issue, but callers should be aware that
+ * the registry size ({@link #getRegisteredCallbackCount()}) may drift above zero under heavy
+ * eviction pressure.
+ *
+ * <p>This class is primarily used by {@code SandboxingAdvisor} to wrap unsafe tools.
+ *
+ * @see ExecuteToolLocalActivity
+ */
+public class LocalActivityToolCallbackWrapper implements ToolCallback {
```
tconley1428: Is there a good reason we are making local activities so primary?

temporal-spring-ai/src/main/java/io/temporal/springai/advisor/SandboxingAdvisor.java
```diff
@@ -0,0 +1,119 @@
...
+ * An advisor that automatically wraps unsafe tool callbacks in local activities.
+ *
+ * <p>This advisor inspects all tool callbacks in a chat request and ensures they are safe for
+ * workflow execution:
+ *
+ * <ul>
+ *   <li>{@link ActivityToolCallback} - Already safe (executes as activity)
+ *   <li>{@link NexusToolCallback} - Already safe (executes as Nexus operation)
+ *   <li>{@link SideEffectToolCallback} - Already safe (wrapped in sideEffect())
+ *   <li>Other callbacks - Wrapped in {@link LocalActivityToolCallbackWrapper} with warning
```
tconley1428: Mmmm... this is not the approach I have taken elsewhere. I'm of the opinion that tools should execute in workflow context by default, where the user could define a tool which does various workflow operations without us needing to define special tool types for them. Child workflow, multiple operations, etc.

temporal-spring-ai/build.gradle
```diff
@@ -0,0 +1,56 @@
...
+    options.compilerArgs.removeAll(['--release', '8'])
+    options.compilerArgs.addAll(['--release', '17'])
+}
+
+dependencies {
+    api(platform("org.springframework.boot:spring-boot-dependencies:$springBootVersionForSpringAi"))
+    api(platform("org.springframework.ai:spring-ai-bom:$springAiVersion"))
+
+    // this module shouldn't carry temporal-sdk with it, especially for situations when users may be using a shaded artifact
+    compileOnly project(':temporal-sdk')
```
DABH, line 30: Could this increase friction though if a Spring user is just trying to get started and doesn't realize they need to import the sdk as well?  Is there a `temporal-spring-ai-starter` planned that would pull in the right transitive dependencies to help with easy onboarding?  Maybe not necessary but just a thought

temporal-spring-ai/build.gradle
```diff
@@ -0,0 +1,56 @@
+description = '''Temporal Java SDK Spring AI Plugin'''
+
+ext {
+    springAiVersion = '1.1.0'
```
DABH, line 4: Probably clarify in the docs/readme the compatibility matrix for the module?  (So - maybe not in this PR - but at least in the forthcoming docs PR)

temporal-spring-ai/src/main/java/io/temporal/springai/tool/LocalActivityToolCallbackWrapper.java
DABH: Would a bounded size or TTL-based eviction strategy for the hash map make more sense, as opposed to just giving users a method to monitor the count?  Even some (imo very high) limit like 10,000?  Or maybe log a warning in case a user hits some huge number like that?  Wdyt?

temporal-spring-ai/src/main/java/io/temporal/springai/advisor/SandboxingAdvisor.java
```diff
@@ -0,0 +1,119 @@
...
+ *
+ * <p>Wrapping tools in local activities adds overhead compared to properly annotated tools. For
+ * production, annotate your tools with {@code @DeterministicTool} or {@code @SideEffectTool}, or
+ * use activity stubs.
+ *
+ * @see io.temporal.springai.tool.DeterministicTool
+ * @see io.temporal.springai.tool.SideEffectTool
+ * @see LocalActivityToolCallbackWrapper
+ */
+public class SandboxingAdvisor implements CallAdvisor {
```
DABH: If you do go ahead with this route, it looks like tests are lacking at the moment, is there a meaningful way to explicitly test `adviseCall`?

temporal-spring-ai/src/main/java/io/temporal/springai/mcp/ActivityMcpClient.java
```diff
@@ -0,0 +1,141 @@
...
+   *
+   * <p>Results are cached after the first call.
+   *
+   * @return map of client name to server capabilities
+   */
+  public Map<String, McpSchema.ServerCapabilities> getServerCapabilities() {
+    if (serverCapabilities == null) {
+      serverCapabilities = activity.getServerCapabilities();
+    }
+    return serverCapabilities;
```
DABH, line 105: Does it make sense to cache this in terms of replayability?  If the workflow is replayed in some way where the activity's server capabilities have changed, we'd get a stale value?  Is that a realistic thing we need to worry about?

temporal-spring-ai/src/main/java/io/temporal/springai/mcp/McpClientActivityImpl.java
```diff
@@ -0,0 +1,64 @@
...
+  private final Map<String, McpSyncClient> mcpClients;
+
+  /**
+   * Creates an activity implementation with the given MCP clients.
+   *
+   * @param mcpClients list of MCP sync clients from Spring context
+   */
+  public McpClientActivityImpl(List<McpSyncClient> mcpClients) {
+    this.mcpClients =
+        mcpClients.stream().collect(Collectors.toMap(c -> c.getClientInfo().name(), c -> c));
```
DABH: Can two MCP clients have the same `name()` in their `ClientInfo`?  That would throw an `IllegalStateException` so we should handle that or validate or something

temporal-spring-ai/src/main/java/io/temporal/springai/tool/LocalActivityToolCallbackWrapper.java
```diff
@@ -0,0 +1,132 @@
...
+      return stub.call(callbackId, toolInput);
+    } finally {
+      CALLBACK_REGISTRY.remove(callbackId);
+    }
+  }
+
+  @Override
+  public String call(String toolInput, ToolContext toolContext) {
+    // Note: ToolContext cannot be passed through the activity, so we ignore it here.
+    // If context is needed, consider using activity parameters or workflow state.
```
DABH: Should we warn users about this?

temporal-spring-ai/src/main/java/io/temporal/springai/model/ChatModelTypes.java
```diff
@@ -0,0 +1,192 @@
...
+   * @param role the role of the message author
+   * @param name optional name for the participant
+   * @param toolCallId tool call ID this message responds to (for TOOL role)
+   * @param toolCalls tool calls requested by the model (for ASSISTANT role)
+   * @param mediaContents optional media attachments
+   */
+  @JsonInclude(JsonInclude.Include.NON_NULL)
+  @JsonIgnoreProperties(ignoreUnknown = true)
+  public record Message(
+      @JsonProperty("content") Object rawContent,
```
DABH: You always cast this to String (ChatModelActivityImpl.toSpringMessage() and ActivityChatModel.toAssistantMessage()).  Should we just make this a String?  If we tried to deserialize something into a non-String, we'd get a `ClassCastException` I think?

temporal-spring-ai/src/main/java/io/temporal/springai/activity/EmbeddingModelActivityImpl.java
```diff
@@ -0,0 +1,72 @@
...
+
+    return new EmbeddingModelTypes.EmbedBatchOutput(results, metadata);
+  }
+
+  @Override
+  public EmbeddingModelTypes.DimensionsOutput dimensions() {
+    return new EmbeddingModelTypes.DimensionsOutput(embeddingModel.dimensions());
+  }
+
+  private List<Double> toDoubleList(float[] floats) {
```
DABH: Nitpick here but technically this creates a ton of Double objects (1536+ dimensions for OpenAI)... consider using `float[]` or `double[]` in `EmbeddingModelTypes` to avoid boxing and reduce serialization payload size?

DABH (dismissed):
Looks like a ton of progress in the right direction! 🚀 Just a few questions/comments (on top of what others have already pointed out).  Take them or leave them :)

temporal-spring-ai/src/main/java/io/temporal/springai/plugin/VectorStorePlugin.java
donald-pinckney: it's not just Spring config interaction, it's about not forcing transitive dependencies on users. A single plugin would either need all deps as `implementation` or use reflection/Object types to avoid class references (fragile/yucky). Given that normal (non-Temporal) Spring AI already ships things as separate artifacts, this seemed like the most natural spring-ai-native design.

Other option is to have implementation dependencies on Spring AI's various artifacts (for MCP, RAG, etc.), which is simpler but aligns less with Spring from what I've seen. Others could weigh in here though.

To be clear, you can't just have `compileOnly` deps on those optional dependencies, because Spring needs to load and introspect the plugin class at startup, so any field types that don't exist at runtime cause a `ClassNotFoundException`.

temporal-spring-ai/src/main/java/io/temporal/springai/advisor/SandboxingAdvisor.java
donald-pinckney: To be clear, `SandboxingAdvisor` isn't enabled by default. Regardless, after discussion with @jssmith I think it makes sense to change it to how it is in other AI integrations. The current semantics were lifted from @Quinn-With-Two-Ns's original code, and there are some merits (because Java doesn't give you any runtime determinism protection), but I think consistency and simplicity are more important. So the new proposed tool execution mapping strategy will be as follows.

## Activity stub tools

User defines an interface with both `@ActivityInterface` and `@Tool` annotations:

```java
@ActivityInterface
public interface WeatherActivity {
    @Tool(description = "Get weather") @ActivityMethod
    String getWeather(String city);
}
```

In the workflow, they create a stub and pass it to `defaultTools()`:

```java
WeatherActivity weather = Workflow.newActivityStub(WeatherActivity.class, opts);
chatClient = TemporalChatClient.builder(model).defaultTools(weather).build();
```

Detection: we check `instanceof ActivityInvocationHandler` on the proxy. Each `@Tool` method becomes an ActivityToolCallback.

##  Nexus service stub tools

Same idea but with `@NexusServiceStub`. Detection via `instanceof NexusServiceInvocationHandler`.

## Side effect tools

User annotates a class with `@SideEffectTool`. Methods run via `Workflow.sideEffect()`:

```java
@SideEffectTool
public class TimestampTools {
    @Tool(description = "Current time")
    public String now() { return Instant.now().toString(); }
}
```

Passed via `defaultTools(new TimestampTools())`. Detection: check for `@SideEffectTool` annotation. Each method wrapped in `SideEffectToolCallback`.

## Plain workflow tools (the new default)

User writes a class with `@Tool` methods. No special annotation. Methods execute directly in the workflow thread:
```java
public class MyTools {
    @Tool(description = "Do something complex")
    public String doStuff(String input) {
        // User's responsibility — can call activities, child workflows, sideEffect, whatever
        SomeActivity act = Workflow.newActivityStub(SomeActivity.class, opts);
        return act.process(input);
    }
}
```
Passed via `defaultTools(new MyTools())`. Detection: falls through all other checks. We just call `ToolCallbacks.from()` and run them directly.

This is where the current code throws `IllegalArgumentException`, or wraps into `LocalActivityToolCallbackWrapper` if `SandboxingAdvisor` is enabled.\

## What gets removed:
`@DeterministicTool` (replaced by "plain tools just work"), `SandboxingAdvisor`, `LocalActivityToolCallbackWrapper`.

What stays: `@SideEffectTool`, activity stub detection, nexus stub detection.


How does this sound?

temporal-spring-ai/src/main/java/io/temporal/springai/tool/LocalActivityToolCallbackWrapper.java
donald-pinckney: would be resolved following other discussion below

temporal-spring-ai/src/main/java/io/temporal/springai/plugin/VectorStorePlugin.java
donald-pinckney: fixed

temporal-spring-ai/src/main/java/io/temporal/springai/advisor/SandboxingAdvisor.java
tconley1428: @Quinn-With-Two-Ns if you want to comment on this. It sounds clear to me, but perhaps we should ensure the java static analyzer can still provide feedback. And make sure it is clearly documented

temporal-spring-ai/src/main/java/io/temporal/springai/plugin/SpringAiPlugin.java
```diff
@@ -22,8 +22,8 @@
  * auto-configuration:
  *
  * <ul>
- *   <li>{@link VectorStorePlugin} - when {@code spring-ai-rag} is on the classpath
- *   <li>{@link EmbeddingModelPlugin} - when {@code spring-ai-rag} is on the classpath
+ *   <li>VectorStore plugin - when {@code spring-ai-rag} is on the classpath
```
tconley1428: I don't think we want it to be available _only_ to auto configuration, do we?

temporal-spring-ai/src/main/java/io/temporal/springai/autoconfigure/SpringAiVectorStoreAutoConfiguration.java
```diff
@@ -10,16 +11,19 @@
...
 @AutoConfiguration(after = SpringAiTemporalAutoConfiguration.class)
 @ConditionalOnClass(name = "org.springframework.ai.vectorstore.VectorStore")
 @ConditionalOnBean(VectorStore.class)
 public class SpringAiVectorStoreAutoConfiguration {
 
   @Bean
-  public VectorStorePlugin vectorStorePlugin(VectorStore vectorStore) {
-    return new VectorStorePlugin(vectorStore);
+  public SimplePlugin vectorStorePlugin(VectorStore vectorStore) {
+    return SimplePlugin.newBuilder("io.temporal.spring-ai-vectorstore")
```
tconley1428: I would still have the plugin exist so a user could create it

temporal-spring-ai/src/main/java/io/temporal/springai/advisor/SandboxingAdvisor.java
donald-pinckney: I don't think there is any reason the java static analyzer couldn't provide feedback, but it might not currently (haven't tested). I'd lean towards merging the plugin with this design, and improve static analyzer on the backend, but @Quinn-With-Two-Ns let me know if you have other thoughts.

temporal-spring-ai/src/main/java/io/temporal/springai/autoconfigure/SpringAiVectorStoreAutoConfiguration.java
donald-pinckney: Put them back as actual plugins, but now using SimplePlugin builder.

temporal-spring-ai/src/main/java/io/temporal/springai/mcp/ActivityMcpClient.java
donald-pinckney: This doesn't matter in terms of replayability. In fact, this cache *won't* be hit (initially) during replay, but Temporal itself will return the cached `getServerCapabilities` activity result.

When this cache will be hit is during multiple calls to `getServerCapabilities` throughout the life of the workflow: in those cases we don't do the whole activity and thus MCP remote call, we just return the cached thing.

IMO its pretty common practice to cache server capabilities with MCP, and it would be I think a bit confusing if the MCP client were able to react to new capabilities during the execution of a workflow, but not during later replay of a workflow.

temporal-spring-ai/src/main/java/io/temporal/springai/advisor/SandboxingAdvisor.java
donald-pinckney: most likely not going ahead with SandboxingAdvisor

temporal-spring-ai/build.gradle
donald-pinckney: Hadn't planned that, but I'd be open to it. Let's not do it in this PR, but at least make it clear in docs.

temporal-spring-ai/build.gradle
donald-pinckney: Good call — will add a compatibility matrix to the module README. Tracking as a follow-up task.

- 🤖 Generated with Claude Code

temporal-spring-ai/src/main/java/io/temporal/springai/tool/LocalActivityToolCallbackWrapper.java
donald-pinckney: most likely not going ahead with SandboxingAdvisor, which would also kill this.

temporal-spring-ai/src/main/java/io/temporal/springai/tool/LocalActivityToolCallbackWrapper.java
donald-pinckney: also would go away

temporal-spring-ai/src/main/java/io/temporal/springai/advisor/SandboxingAdvisor.java
Quinn-With-Two-Ns: If we took the approach in our other integrations that "tools should execute in workflow context by default" then I think we should be consistent here.

I wrote this part before we had the static analyzer so I was trying to be defensive here, I agree the static analyzer is a better way to catch stuff like this. I did think the `LocalActivityToolCallbackWrapper` was a nice feature but we can always add something like that later.

maciejdudko (approved):

temporal-spring-ai/README.md
```diff
@@ -0,0 +1,111 @@
+# temporal-spring-ai
+
+Integrates [Spring AI](https://docs.spring.io/spring-ai/reference/) with [Temporal](https://temporal.io/) workflows, making AI model calls, tool execution, vector store operations, embeddings, and MCP tool calls durable Temporal primitives.
+
+## Compatibility
+
+| Dependency | Minimum Version |
```
mjameswh: nit: I don't know if the formatter will allow that, but see if you can add spaces/dashes to align columns in plain text. A lot of people prefer to read markdown as source rather than the generated output.

temporal-spring-ai/README.md
```diff
@@ -0,0 +1,111 @@
...
+
+### Nexus service stubs
+
+Auto-detected and executed as Nexus operations, similar to activity stubs.
+
+## Optional Integrations
+
+Auto-configured when their dependencies are on the classpath:
+
+| Feature | Dependency | What it registers |
```
mjameswh: Same, see if you can align the table.

PR #2830, opened 2026-04-07
Add explicit opt-in flags for activity/nexus bean auto-discovery in Spring Boot

### LINKED ISSUE #2780: Spring boot `workers-auto-discovery` - Workflow vs Activities
Context:
```
The intention is that a user generally has control over the activity beans they are creating, 
but has less control over the classes that may be in their path, thus justifying a way to exclude them.

the config name should have been maybe workflow-packages to represent the intention
```

config sample:
```
  temporal:
    workers-auto-discovery:
      packages:
        - io.temporal.spring.boot.autoconfigure.bytaskqueue
```

- While all Activities are loaded as Beans, they depend on the boolean `workers-auto-discovery` 
- And then the filter for packages only applied Workflows. 

Essentially, if no `packages` are declared, then no Activities/Beans are registered either

## Expected Behavior
Activity beans to be loaded when
- `workers-auto-discovery` present in config despite the presence of `packages`

Preference to rename `packages` to `workflow-packages`

## Actual Behavior
No Activities registered, if no packages specified - [link](https://github.com/temporalio/sdk-java/blob/master/temporal-spring-boot-autoconfigure/src/main/java/io/temporal/spring/boot/autoconfigure/template/WorkersTemplate.java#L147-L152)

## Additional notes
- thought issue is worthwhile given single use Activity feature.

  1.
  1.
  1.

## Specifications

  - Version:
  - Platform:

mjameswh:
Fixes #2780.

## Problem

`@ActivityImpl`- and `@NexusServiceImpl`-annotated Spring beans were silently not registered with Temporal workers unless `spring.temporal.workers-auto-discovery.packages` was configured — even though that setting is only supposed to control workflow *classpath scanning*. The entire auto-discovery block (activities, nexus services, and workflows) was gated on `packages != null` in `WorkersTemplate.createWorkers()`.

## Solution

Separate the concerns with explicit opt-in flags under `spring.temporal.workers-auto-discovery`:

| Property | Type | Default | Description |
|---|---|---|---|
| `register-activity-beans` | `boolean` | `false` | Register `@ActivityImpl`-annotated Spring beans with matching workers |
| `register-nexus-service-beans` | `boolean` | `false` | Register `@NexusServiceImpl`-annotated Spring beans with matching workers |
| `workflow-packages` | `list` | — | Packages to scan for `@WorkflowImpl` classes (replaces `packages`) |
| `enable` | `boolean` | `false` | Convenience flag: registers all `@WorkflowImpl`, `@ActivityImpl`, and `@NexusServiceImpl` Spring-managed beans without package scanning |

The `enable: true` flag covers the common case where all implementations are Spring-managed beans. For more control, `register-activities` and `register-nexus-operations` can be set independently, and `workflow-packages` handles classpath scanning for workflow classes that are not Spring beans.

## Backward compatibility

The existing `packages` property is deprecated but fully preserved:
- If set and non-empty, it implies `register-activities: true`, `register-nexus-operations: true`, and its entries are treated as `workflow-packages`.
- Combining `packages` with any of the new properties throws `IllegalStateException` at startup with a clear migration message.

## Changes

- **`WorkersAutoDiscoveryProperties`** — new fields, deprecation logic, validation, computed accessors (`isRegisterActivities()`, `isRegisterNexusOperations()`, `isRegisterWorkflowBeans()`, `getEffectiveWorkflowPackages()`)
- **`WorkersTemplate`** — auto-discovery block rewritten to use the new property accessors independently; added `autoDiscoverWorkflowBeans()` for bean-based workflow discovery
- **`WorkersPresentCondition`** — binds `WorkersAutoDiscoveryProperties` as a whole (was binding only `packages`), so the condition fires for any `workers-auto-discovery` configuration
- **`application.yml`** (tests) — existing profiles migrated from `packages:` to `workflow-packages:` + explicit flags; legacy `packages` behavior tested via the new `auto-discovery-by-task-queue-legacy` profile
- **New tests**: `AutoDiscoveryActivitiesOnlyTest`, `AutoDiscoveryEnableTest`, `AutoDiscoveryByTaskQueueLegacyTest`, `AutoDiscoveryMixedPropertiesRejectedTest`

### DIFF
--- a/temporal-spring-boot-autoconfigure/src/main/java/io/temporal/spring/boot/autoconfigure/properties/WorkersAutoDiscoveryProperties.java
+++ b/temporal-spring-boot-autoconfigure/src/main/java/io/temporal/spring/boot/autoconfigure/properties/WorkersAutoDiscoveryProperties.java
@@ -1,17 +1,121 @@
 package io.temporal.spring.boot.autoconfigure.properties;
 
+import java.util.ArrayList;
 import java.util.List;
 import javax.annotation.Nullable;
 import org.springframework.boot.context.properties.ConstructorBinding;
+import org.springframework.boot.context.properties.DeprecatedConfigurationProperty;
 
 public class WorkersAutoDiscoveryProperties {
-  private final @Nullable List<String> packages;
+  /**
+   * When {@code true}, enables auto-registration of all {@code @ActivityImpl}-,
+   * {@code @NexusServiceImpl}-, and {@code @WorkflowImpl}-annotated classes that are Spring-managed
+   * beans. This is the recommended option for simple use cases where all implementations are
+   * Spring-managed beans. It implies {@link #registerActivityBeans} and {@link
+   * #registerNexusServiceBeans} default to {@code true}, and also enables auto-registration of
+   * {@code @WorkflowImpl}-annotated classes that are Spring-managed beans, without needing {@link
+   * #workflowPackages}.
+   *
+   * <p>Classpath-scanning via {@link #workflowPackages} (for non-bean workflow classes) still
+   * requires explicit package configuration regardless of this flag.
+   */
+  private final @Nullable Boolean enable;
+
+  private final @Nullable List<String> workflowPackages;
+  private final @Nullable Boolean registerActivityBeans;
+  private final @Nullable Boolean registerNexusServiceBeans;
+
+  /**
+   * @deprecated Use {@link #workflowPackages} instead. If set and non-empty, this property causes
+   *     {@link #registerActivityBeans} and {@link #registerNexusServiceBeans} to default to {@code
+   *     true}, and its entries to be considered as if they were provided through {@link
+   *     #workflowPackages}. Setting both {@link #packages} and any of the other new properties is
+   *     unsupported and will result in an exception.
+   */
+  @Deprecated private final @Nullable List<String> packages;
 
   @ConstructorBinding
-  public WorkersAutoDiscoveryProperties(@Nullable List<String> packages) {
+  public WorkersAutoDiscoveryProperties(
+      @Nullable Boolean enable,
+      @Nullable List<String> workflowPackages,
+      @Nullable Boolean registerActivityBeans,
+      @Nullable Boolean registerNexusServiceBeans,
+      @Nullable List<String> packages) {
+    if (packages != null
+        && !packages.isEmpty()
+        && (enable != null
+            || workflowPackages != null
+            || registerActivityBeans != null
+            || registerNexusServiceBeans != null)) {
+      throw new IllegalStateException(
+          "spring.temporal.workers-auto-discovery.packages is deprecated and cannot be combined "
+              + "with enable, workflow-packages, register-activity-beans, or register-nexus-service-beans. "
+              + "Migrate to the new properties and remove packages.");
+    }
+    this.enable = enable;
+    this.workflowPackages = workflowPackages;
+    this.registerActivityBeans = registerActivityBeans;
+    this.registerNexusServiceBeans = registerNexusServiceBeans;
     this.packages = packages;
   }
 
+  /**
+   * Returns whether {@code @WorkflowImpl}-annotated classes that are also Spring-managed beans
+   * should be automatically registered with matching workers (without classpath scanning). Defaults
+   * to {@code true} when {@link #enable} is {@code true}, {@code false} otherwise. Workflow
+   * implementation classes that are not Spring-managed beans still require {@link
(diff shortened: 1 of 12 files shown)


temporal-spring-boot-autoconfigure/src/main/java/io/temporal/spring/boot/autoconfigure/properties/WorkersAutoDiscoveryProperties.java
```diff
@@ -1,17 +1,113 @@
...
+import java.util.ArrayList;
 import java.util.List;
 import javax.annotation.Nullable;
 import org.springframework.boot.context.properties.ConstructorBinding;
 
 public class WorkersAutoDiscoveryProperties {
-  private final @Nullable List<String> packages;
+  /**
+   * When {@code true}, enables auto-registration of all {@code @ActivityImpl},
+   * {@code @NexusServiceImpl}, and {@code @WorkflowImpl}-annotated Spring beans. This is the
```
Quinn-With-Two-Ns: NIT: `@WorkflowImpl` is applied on classes not beans

temporal-spring-boot-autoconfigure/src/main/java/io/temporal/spring/boot/autoconfigure/properties/WorkersAutoDiscoveryProperties.java
```diff
@@ -1,17 +1,113 @@
...
     this.packages = packages;
   }
 
+  /**
+   * Returns whether {@code @WorkflowImpl}-annotated Spring beans should be automatically registered
+   * with matching workers (without classpath scanning). Defaults to {@code true} when {@link
+   * #enable} is {@code true}, and to {@code false} otherwise. Note: non-bean workflow classes still
+   * require {@link #workflowPackages} for classpath scanning.
+   */
+  public boolean isRegisterWorkflowBeans() {
```
Quinn-With-Two-Ns: NIT: We register workflow classes not beans

Quinn-With-Two-Ns:
For manual config the properties are like
```
spring.temporal:
  workers:
    - task-queue: your-task-queue-name
      name: your-worker-name # unique name of the Worker. If not specified, Task Queue is used as the Worker name.
      workflow-classes:
        - your.package.YourWorkflowImpl
      activity-beans:
        - activity-bean-name1
```
Given that I might change some of the names here like `register-activities` -> `register-activitiy-beans`

Quinn-With-Two-Ns:
How will a user know `packages` is deprecated? Could we log a warning to let users know?

mjameswh:
> How will a user know packages is deprecated? Could we log a warning to let users know?

Turns out Spring Boot has its own annotation for that purpose. I added it.

temporal-spring-boot-autoconfigure/src/main/java/io/temporal/spring/boot/autoconfigure/properties/WorkersAutoDiscoveryProperties.java
```diff
@@ -1,17 +1,121 @@
...
+   * beans. This is the recommended option for simple use cases where all implementations are
+   * Spring-managed beans. It implies {@link #registerActivityBeans} and {@link
+   * #registerNexusServiceBeans} default to {@code true}, and also enables auto-registration of
+   * {@code @WorkflowImpl}-annotated classes that are Spring-managed beans, without needing {@link
+   * #workflowPackages}.
+   *
+   * <p>Classpath-scanning via {@link #workflowPackages} (for non-bean workflow classes) still
+   * requires explicit package configuration regardless of this flag.
+   */
+  private final @Nullable Boolean enable;
```
robzienert: nit: Should probably choose `enabled` instead of `enable` to follow more closely to general Spring ecosystem convention. This is just my Spring exposure, so if you've seen it more commonly expressed `enable` that'd be interesting to learn!

robzienert (approved):

temporal-spring-boot-autoconfigure/src/main/java/io/temporal/spring/boot/autoconfigure/properties/WorkersAutoDiscoveryProperties.java
mjameswh: Thanks a lot for your input! After some search, `enabled` indeed seems more idiomatic to the Spring Boot community.

Quinn-With-Two-Ns (approved):

maciejdudko (approved):

PR #2835, opened 2026-04-09
chore: bump workflowcheck asm version for latest JDK version support

robzienert:
## What was changed
Bumps the asm dependency for `temporal-workflowcheck` to latest.

## Why?
The existing dependency only supports up to JDK22, whereas we have uses for JDK24.

## Checklist

How was this tested:

`./gradlew :temporal-workflowcheck:check`

### DIFF
--- a/temporal-workflowcheck/build.gradle
+++ b/temporal-workflowcheck/build.gradle
@@ -7,7 +7,7 @@ plugins {
 description = 'Temporal Java WorkflowCheck Static Analyzer'
 
 dependencies {
-    implementation 'org.ow2.asm:asm:9.7'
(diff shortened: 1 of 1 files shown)


mjameswh (approved):

mjameswh:
@robzienert Out of curiosity, are you using workflowcheck's `-all` build? If not, I think you should be able to get gradle to use a different version of asm.

robzienert:
I don't think so - we're just importing `temporal-workflowcheck` as a dependency and then programmatically calling it. I know I could've just changed the resolved transitive dep, but thought it would be good to fix upstream instead.

PR #2848, opened 2026-04-20
Fix race in AsyncPollerTest.testSuspendPolling

donald-pinckney:
## What was changed
Remove the race: override reserveSlot on the slot supplier to count invocations, and wait for reserveCalls == 2 before suspending. That's a direct signal that iter 2 has passed the suspend check and called reserveSlot (its future can't complete until iter 1's slot is released, but the call itself happens immediately). After this signal, iter 2 is pinned past the suspend check, so the subsequent suspendPolling() only affects iter 3+.

Also drop the now-redundant first assertEventually: once pollLatch counts down (inside the mocked poll()), iter 1's slot has already been issued, so reservedCount == 1 is synchronously visible.

## Why?
The test relied on the poller's second iteration racing past its suspend-latch check before the test thread called suspendPolling(). If the test thread won the race, iter 2 blocked at the suspend latch and never reached reserveSlot — so reservedCount stayed at 1 and the later `expected:<2> but was:<1>` assertion failed. Bumping the assertEventually timeout didn't help; the race either wins or loses deterministically on a given run.

## AI Transparency
This PR was fully generated by Claude, and I have not yet reviewed it for accuracy with regards to fixing the flakiness. Don't yet review until I review it more.

### DIFF
--- a/temporal-sdk/src/test/java/io/temporal/internal/worker/AsyncPollerTest.java
+++ b/temporal-sdk/src/test/java/io/temporal/internal/worker/AsyncPollerTest.java
@@ -336,7 +336,15 @@ public class AsyncPollerTest {
   @Test
   public void testSuspendPolling()
       throws InterruptedException, ExecutionException, AsyncPoller.PollTaskAsyncAbort {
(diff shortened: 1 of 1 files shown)


brianstrauch:
Ran into this flake today

PR #2852, opened 2026-04-21
temporal-spring-ai: attach activity summaries for chat and MCP calls

donald-pinckney:
## What was changed
- `ActivityChatModel` attaches an activity Summary of the form `chat: <model>` (model label only — user prompts may contain PII, so they are deliberately not included) to each chat-model activity call.
- `ActivityMcpClient` gains a `callTool(clientName, request, summary)` overload; `McpToolCallback` uses it to attach `mcp: <client>.<tool>`.
- Summaries are applied only when the plugin itself created the activity stub (via the existing `forModel` / `create` factories). When the user supplies their own pre-built stub to the public constructors, behavior is unchanged, as they are non-generic tools that are up to the user to annotate.
- New `ActivitySummaryTest` runs a workflow, fetches history, and asserts the `ActivityTaskScheduled` event for `CallChatModel` carries the expected `userMetadata.summary`.

Activity-backed tool calls, Nexus tool calls, and `@SideEffectTool` calls are intentionally out of scope — either the stub is user-owned with no clean per-call option override (activity and Nexus tools), or there is no Summary slot on the resulting history event (`@SideEffectTool` writes a `MarkerRecorded`, not an `ActivityTaskScheduled`).

## Why?
Without Summaries, every chat and MCP step in a Spring AI workflow shows up as an opaque `callChatModel` or `callTool` row in the Temporal UI, making multi-step agents hard to follow. Chat and MCP are the cases where the plugin owns stub creation end-to-end, so adding the Summary is a cheap, low-risk overlay; broader coverage would require reaching into user-owned stubs and was deliberately left out to keep this change small and review-friendly.


temporal-spring-ai/src/main/java/io/temporal/springai/mcp/ActivityMcpClient.java
```diff
@@ -57,7 +58,17 @@ public class ActivityMcpClient {
    * @param activity the activity stub for MCP operations
    */
   public ActivityMcpClient(McpClientActivity activity) {
+    this(activity, null);
```
JasonSteving99, line 62: So I've never been a fan of throwing around raw `null`s in Java. This is likely a hangover of mine from Google days where Java style is very heavily anti-`null` (e.g. [Guava suggests avoiding null](https://github.com/google/guava/wiki/UsingAndAvoidingNullExplained) though after Java 8 Guava's `Optional` type isn't needed, can just use builtin `java.util.Optional`). 

But I'd prefer not using nullable params or fields, and instead explicitly marking them as `Optional<Foo>`. Barring that, you should probably look into marking things with `@Nullable` if you plan to allow that. See this [blog on Spring Null Safety](https://www.baeldung.com/spring-null-safety-annotations)

I'm not sure how standard this type of `null` usage has been across the Spring codebase though, so in the end, this can be a judgement call you make.

Same comment applies across all the files in this PR.

donald-pinckney: Code style wise, sdk-java uses a mix of Nullable annotations and Optional<T>. And my preference would be Optional<T> as well. PR has been updated, all nulls have been changed to Optionals, except for the `ChatModelActivityInput` which I'll defer a decision on for a later PR.

JasonSteving99: awesome, thanks. LGTM

JasonSteving99 (approved):

temporal-spring-ai/src/main/java/io/temporal/springai/mcp/ActivityMcpClient.java
```diff
@@ -57,7 +59,18 @@ public class ActivityMcpClient {
...
+    this(activity, Optional.empty());
+  }
+
+  /**
+   * Creates a new ActivityMcpClient. When {@code baseOptions} is present, {@link #callTool(String,
+   * McpSchema.CallToolRequest, String)} rebuilds the activity stub with a per-call Summary on top
+   * of those options. When empty, the caller supplied a pre-built stub whose options we don't know,
+   * so we call through it as-is and drop any requested summary.
+   */
+  private ActivityMcpClient(McpClientActivity activity, Optional<ActivityOptions> baseOptions) {
```
Quinn-With-Two-Ns: This should not be optional 
```suggestion
  private ActivityMcpClient(McpClientActivity activity, @nullable ActivityOptions baseOptions) {
```

Quinn-With-Two-Ns (dismissed):

temporal-spring-ai/src/main/java/io/temporal/springai/mcp/ActivityMcpClient.java
Quinn-With-Two-Ns: Please revert all the `Optional` changes

donald-pinckney: all reverted, and @Nullable added

mjameswh (approved):

PR #2853, opened 2026-04-21
temporal-spring-ai: accept ActivityOptions and classify non-retryable AI errors

donald-pinckney:
## Based on

- #2852 — `spring-ai/activity-summaries` — adds the chat/MCP Summary plumbing this PR builds on

Merge that first; this PR will auto-retarget to `master`.

---

## What was changed
- `ActivityChatModel.forDefault(ActivityOptions)` and `forModel(String, ActivityOptions)` overloads added. New public `defaultActivityOptions()` returns the plugin's default bundle so callers can tweak one field (e.g., task queue) without losing the other sensible defaults.
- `ActivityMcpClient.create(ActivityOptions)` + `defaultActivityOptions()` added, mirroring the chat side.
- Default `RetryOptions` for chat calls now mark `org.springframework.ai.retry.NonTransientAiException` and `java.lang.IllegalArgumentException` non-retryable. Default options for MCP calls mark `IllegalArgumentException` non-retryable. User-supplied `ActivityOptions` pass through verbatim — the plugin does not augment them.
- The public `new ActivityChatModel(...)` and `new ActivityMcpClient(activity)` constructors are **removed**. The factories are now the sole construction path. The module is pre-release so there are no pinned callers to migrate; the deprecate-then-remove cycle would be pure noise.
- README: new "Activity options and retry behavior" section documents the defaults and how to customize via the `ActivityOptions` overload.
- Tests: two new suites — `ActivityOptionsAndRetryTest` covers non-retryable classification (1 attempt for `NonTransientAiException`, 3 attempts for transient `RuntimeException`, custom task queue landing on the scheduled activity); `ActivitySummaryTest` gains a regression test asserting `forDefault(customOptions)` still emits UI Summaries.
- `build.gradle`: `spring-ai-retry` added as a `testImplementation` so tests can reference `NonTransientAiException` directly.

## Why?
Two gaps the integration guide calls out:

1. **Retry classification.** Without a `doNotRetry` list, a bad API key or invalid-prompt error churned through retries + backoff before failing. The new defaults fail fast on clearly permanent failures while still retrying transient ones (network, 5xx, rate-limits).
2. **Customization surface.** The prior factory API only exposed `(timeout, maxAttempts)` — no heartbeats, task queue override, priority, or `doNotRetry`. Users hitting those needs had to call the public constructor with a hand-built stub. That path silently dropped the UI Summary labels the summaries branch (#2852) added, so users were forced to choose between customization and readability.

The new `ActivityOptions` overloads give users both. Removing the public constructors outright means users can't accidentally end up in a code path that skips UI labels — there's one way in, and it does the right thing.

Depends on #2852 (activity-summaries). The summary regression test in this PR only passes because that branch already landed the `baseOptions` plumbing the new overloads thread through.


temporal-spring-ai/src/main/java/io/temporal/springai/model/ActivityChatModel.java
```diff
@@ -154,30 +137,53 @@ public static ActivityChatModel forDefault() {
...
+   * point when you want to customize one or two fields without losing the sensible defaults:
+   *
+   * <pre>{@code
+   * ActivityChatModel.forDefault(
+   *     ActivityOptions.newBuilder(ActivityChatModel.defaultActivityOptions())
+   *         .setTaskQueue("chat-heavy")
+   *         .build());
+   * }</pre>
+   */
+  public static ActivityOptions defaultActivityOptions() {
```
brianstrauch, line 172: It seems like providing an opinionated "defaultActivityOptions()" function in the first place differs from the regular Temporal model, where the user is required to pass a start-to-close timeout, which essentially makes it a required option.

brianstrauch (changes requested):

temporal-spring-ai/src/main/java/io/temporal/springai/model/ActivityChatModel.java
donald-pinckney: Hhhmm, OpenAI agents plugin does the same thing: https://github.com/temporalio/sdk-python/blob/cfedf83430547a622bdbc6b3cdb8d6e34698029a/temporalio/contrib/openai_agents/_temporal_openai_agents.py#L156

donald-pinckney: It looks like in OAI, the user is required to pass a timeout for tool activities, but not for the LLM activities (uses a default if not passed).

donald-pinckney: Thinking about it a bit more, the regular Temporal model requires explicit timeouts because the SDKs do not know anything at all about the nature of the user-owned activities, and can't possibly guess an even approximately correct timeout.

But for the LLM-calling case, we own the activity definition, not the user, and thus we do know what a reasonable timeout should be.

On the other hand, for activities-as-tools in the Spring AI plugin, we don't know anything about the underlying user-provided activity, so the user must provide all the ActivityOptions in that case.

So I do actually think this is right, though I could be persuaded otherwise. This is probably a recurring question for AI framework integration reviews (including 3rd party ones), so we should make sure we as a team agree.

brianstrauch: Makes sense, especially for the timeout option. The retry option is more specific to the user's application though, but 3 seems reasonable I guess.

brianstrauch (approved):

mjameswh (approved):

PR #2854, opened 2026-04-21
temporal-spring-ai: preserve Usage and RateLimit in ChatResponse metadata

donald-pinckney:
## What was changed
- `ActivityChatModel.toResponse` now rehydrates `Usage` and `RateLimit` onto the `ChatResponseMetadata` returned to workflow code. Previously only `model` was copied across; token counts and rate-limit headers were silently discarded.
- `Usage` is rehydrated as a Spring AI `DefaultUsage(prompt, completion, total)`. `RateLimit` is an interface with no public default impl in `spring-ai-model`, so we return an anonymous implementation backed by the fields from the serialized activity output record.
- New `ResponseMetadataTest` drives a workflow that calls a stub model populating both `Usage` and `RateLimit`, then flattens the resulting metadata to primitives inside the workflow (the interfaces don't Jackson-round-trip across the workflow result) and asserts every field.

No changes to `ChatModelTypes` (the records already carry all fields) or `ChatModelActivityImpl` (already populates them).

## Why?
Users of the plugin couldn't read token counts or rate-limit headers from Spring AI responses even though the underlying `ChatModel` returned them. Cost tracking, observability integrations, and rate-limit-aware advisors were all broken. This is a pure bugfix — the activity payload was already carrying the data, just not being put back on the response. Independent of #2852 and #2853 (no rebase needed, no coupling).


temporal-spring-ai/src/test/java/io/temporal/springai/ResponseMetadataTest.java
```diff
@@ -0,0 +1,193 @@
...
+
+  public static class MetadataWorkflowImpl implements MetadataWorkflow {
+    @Override
+    public MetadataSnapshot collect() {
+      ActivityChatModel chatModel = ActivityChatModel.forDefault();
+      ChatResponse response = chatModel.call(new Prompt("ping"));
+      ChatResponseMetadata md = response.getMetadata();
+      if (md == null) {
+        return new MetadataSnapshot(
+            null, false, null, null, null, false, null, null, null, null, null, null);
```
brianstrauch, line 119: Not a Java expert but would it make sense to do a builder pattern here so it's obvious what each of these values represents?

donald-pinckney: This is test code only, so I don't think its needed.

brianstrauch (approved):

maciejdudko (approved):

PR #2856, opened 2026-04-21
temporal-spring-ai: add side-effect replay tests for chat, activity tools, and @SideEffectTool

donald-pinckney:
## What was changed
Three new tests under `src/test/.../replay/`:

- **`ChatModelSideEffectTest`** — register a `ChatModel` with an `AtomicInteger` counter. Run a workflow that makes one chat call, assert counter=1. Replay the captured history, assert counter still 1 — the activity result comes from history, not from re-invoking the `ChatModel`.
- **`ActivityToolSideEffectTest`** — activity-backed `@Tool` whose impl increments a counter. A tool-calling stub `ChatModel` asks for the tool on the first call and returns final text on the second. Same shape: counter=1 after run, counter=1 after replay.
- **`SideEffectToolReplayTest`** — `@SideEffectTool` body increments a file-scope counter. The workflow drives one tool call. Assertion proves that `Workflow.sideEffect`'s marker is consulted on replay rather than re-invoking the `@Tool` method.

MCP is intentionally omitted for this PR. `spring-ai-mcp` is `compileOnly`, and adding it just for one replay test isn't worth the dep weight — MCP tool calls go through the same Temporal activity machinery as `ChatModel`, which `ChatModelSideEffectTest` already exercises. The plan leaves this as an acceptable substitution.

## Why?
The existing `WorkflowDeterminismTest` catches history-vs-command mismatches but does not prove the plugin isn't re-invoking user side effects on replay. Temporal's AI partner review standards require side-effect safety tests, and a regression — someone adding an in-workflow cache or dropping the `Workflow.sideEffect` wrap — would be easy to slip in without these.

I verified the `SideEffectToolReplayTest` actually catches a real regression: temporarily dropping the `Workflow.sideEffect(...)` wrap in `SideEffectToolCallback.call(...)` produces `expected: <1> but was: <2>` exactly as intended. Restored before this commit.


brianstrauch (approved):

temporal-spring-ai/src/test/java/io/temporal/springai/replay/ChatModelSideEffectTest.java
```diff
@@ -0,0 +1,116 @@
...
+        1, model.callCount.get(), "ChatModel should be called once during the initial run");
+
+    WorkflowExecutionHistory history =
+        client.fetchHistory(WorkflowStub.fromTyped(workflow).getExecution().getWorkflowId());
+    WorkflowReplayer.replayWorkflowExecution(history, ChatWorkflowImpl.class);
+
+    assertEquals(
+        1,
+        model.callCount.get(),
+        "ChatModel must not be re-invoked during replay — activity results come from history");
```
mjameswh: That sounds like you're testing Temporal itself rather than the plugin's code.

I understand that's not really your intent, but the line is thin, and future readers of this code (both humans and agents) might easily misunderstand this.

temporal-spring-ai/src/test/java/io/temporal/springai/replay/ChatModelSideEffectTest.java
```diff
@@ -0,0 +1,116 @@
...
+    worker.registerWorkflowImplementationTypes(ChatWorkflowImpl.class);
+    worker.registerActivitiesImplementations(new ChatModelActivityImpl(model));
+    testEnv.start();
+
+    ChatWorkflow workflow =
+        client.newWorkflowStub(
+            ChatWorkflow.class, WorkflowOptions.newBuilder().setTaskQueue(TASK_QUEUE).build());
+    assertEquals("pong", workflow.chat("ping"));
+    assertEquals(
+        1, model.callCount.get(), "ChatModel should be called once during the initial run");
```
mjameswh: The test methodology isn't exact, which may result in flakes. It is actually possible that the activity function might end up being executed more than once, due to retry, and so the counter might be higher than one here.

A scheduled activity with `maxAttempts > 1` will run exactly once _to completion as observed by the Temporal server_, but the activity itself has no way to know whether the server will have effectively observed the task completion.

Instead of this atomic counter, you can use a concurrency-safe `Set` that records the activity ID of all activity tasks that were executed. The size of that set is what you are looking for.

Alternatively, you could use a very different test approach: look at the workflow history and filter on `ActivityTaskScheduled` events. The expectation is that there should be one `ActivityTaskScheduled` event for this specific activity type.

donald-pinckney:
This PR definitely needs more work on the tests. Will re-work it some.

temporal-spring-ai/src/test/java/io/temporal/springai/replay/ChatModelSideEffectTest.java
donald-pinckney: Yeah, agreed. Assertion comments have been re-worked to make it more clear.

temporal-spring-ai/src/test/java/io/temporal/springai/replay/ChatModelSideEffectTest.java
donald-pinckney: Thats a very good point. I'll fix this, but I'll also work on updating the docs, since they are really vague about the right way to test this:

> Check for duplicate side effects or other types of failures.

https://docs.temporal.io/develop/plugins-guide#side-effects-tests

donald-pinckney: Ok, `ChatModelSideEffectTest` and `ActivityToolSideEffectTest` have been changed to inspecting workflow history.

`SideEffectToolReplayTest` has been left as-is, I think its fine for Workflow.sideEffect.

donald-pinckney:
@mjameswh PR has been improved, and I also opened a PR on docs to explain more about how to write these types of tests: https://github.com/temporalio/documentation/pull/4481

maciejdudko (approved):

PR #2858, opened 2026-04-22
💥 Standalone Activities for Java

GregoryTravis:
Add standalone activity API (ActivityClient)

Introduces support for standalone activities — activities that execute independently of any workflow.

💥 Breaking Changes:
- `ActivityInfo.getWorkflowId`, `.getRunId`, `.getWorkflowNamespace` can now return null, in the case of standalone activities
- `.getRunId` now deprecated in favor of `.getWorkflowRunId` and `.getActivityRunId`.

New public API surface:
- `ActivityClient` — top-level client for starting, describing,
  listing, counting, cancelling, and terminating standalone activities
- `ActivityHandle<R>` / `UntypedActivityHandle` — typed and untyped
  handles returned by `ActivityClient.start()`; provide `getResult()`,
  `getResult(long, TimeUnit)`, `getResultAsync()`, `describe()`,
  `cancel()`, and `terminate()`
- `StartActivityOptions` — builder-based options for starting an
  activity (id, task queue, timeouts, retry, priority, etc.)
- `ActivityClientOptions` — namespace / data converter / interceptor
  configuration for `ActivityClient`
- `ActivityExecutionDescription` — rich descriptor returned by
  `describe()` and `list*()`; extends `ActivityExecutionMetadata`
- `ActivityExecutionMetadata` — lightweight metadata used in list
  results
- `ActivityExecutionCount` — result of `countActivities()`
- `ActivityAlreadyStartedException` — thrown when a duplicate activity
  id is rejected by the server
- `ActivityFailedException` — thrown from `getResult()` when the
  activity fails

New interceptor API:
- `ActivityClientCallsInterceptor` — per-call interceptor for all
  `ActivityClient` operations
- `ActivityClientCallsInterceptorBase` — pass-through base
  implementation (delegates every method to the next interceptor)
- `ActivityClientInterceptor` — factory interceptor that wraps the
  client-level invoker
- `ActivityClientInterceptorBase` — no-op base implementation

`ActivityInfo` additions:
- `getActivityRunId()` — run-scoped id assigned by the server to each
  activity execution
- `isInWorkflow()` — distinguishes workflow-dispatched activities
  from standalone ones

`ActivityCompletionClient` additions:
- New overloads taking `(String activityId, Optional<String> runId, …)`
  for completing, failing, sending heartbeats, and cancelling standalone
  activities without a workflow id

`ActivitySerializationContext` fix:
- `workflowId` and `workflowType` are now `@Nullable`; removed
  `requireNonNull` guards that caused NPEs for standalone activities

`Functions.java`:
- Added `Func` and `VFunc` (zero-arg typed/void functional interfaces)
  needed by `ActivityClient.start()` method-reference overloads

CI:
- Enable standalone-activity server feature flags in the Temporal CLI
  dev server used by unit tests:
  `frontend.activityAPIsEnabled`, `activity.enableStandalone`,
  `history.enableChasm`, `history.enableTransitionHistory`

Tests:
- `StandaloneActivityTest` — integration tests against a real server
  covering the full activity lifecycle (start, poll, complete, cancel,
  terminate, describe, list, heartbeat, async completion)
- `ActivityClientCallsInterceptorBaseTest` — delegation tests for the
  base interceptor
- `ActivityClientCallsInterceptorChainTest` — chain ordering tests
  using a real server with two interceptors
- `ActivityHandleImplTest` — unit tests for handle dispatch
- `ActivityCompletionClientImplTest` — unit tests for completion client
- `ActivityInfoImplTest` — unit tests for `ActivityInfo` in the
  standalone context
- `ActivitySerializationContextTest` — confirms nullable workflow fields
- `StartActivityOptionsTest`, `ActivityClientOptionsTest`,
  `ActivityExecutionDescriptionTest`, `ActivityExecutionMetadataTest`,
  `ActivityAlreadyStartedExceptionTest` — options and value-type tests


temporal-sdk/src/main/java/io/temporal/internal/client/ActivityHandleImpl.java
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  @Override
+  public <R> CompletableFuture<R> getResultAsync(Class<R> resultClass) {
+    return getResultAsync(resultClass, null);
+  }
+
+  @Override
+  public <R> CompletableFuture<R> getResultAsync(Class<R> resultClass, @Nullable Type resultType) {
+    return CompletableFuture.supplyAsync(() -> getResult(resultClass, resultType));
```
GregoryTravis: Do we want to set up a separate Executor for this?

temporal-sdk/src/main/java/io/temporal/client/StartActivityOptions.java
```diff
@@ -0,0 +1,337 @@
...
+import javax.annotation.Nullable;
+
+/**
+ * Options for starting a standalone activity via {@link ActivityClient#start}.
+ *
+ * <p>At least one of {@link #getScheduleToCloseTimeout()} or {@link #getStartToCloseTimeout()} must
+ * be set.
+ */
+@Experimental
+public final class StartActivityOptions {
```
GregoryTravis, line 22: Do we prefer 'StandaloneActivityOptions' for this?

temporal-sdk/src/main/java/io/temporal/client/ActivityClientImpl.java
```diff
@@ -0,0 +1,564 @@
...
+  }
+
+  // ---- Type-sniffing helpers ----
+
+  /**
+   * Creates a dynamic proxy of {@code activityInterface} that records the invoked {@link Method} in
+   * {@code captured[0]} when any method is called on it.
+   */
+  @SuppressWarnings("unchecked")
+  private static <I> I createTypeProbe(Class<I> activityInterface, Method[] captured) {
```
Sushisource: Why is captured an array here? Seems like only one value can ever be written? I guess this is just a workaround for needing some wrapper type, but I wonder if there's something more explicit for that purpose.

temporal-sdk/src/main/java/io/temporal/client/ActivityClientImpl.java
```diff
@@ -0,0 +1,564 @@
...
+      Class<I> activityInterface, Functions.Proc1<I> activity, StartActivityOptions options)
+      throws ActivityAlreadyStartedException {
+    Method[] captured = {null};
+    I probe = createTypeProbe(activityInterface, captured);
+    try {
+      activity.apply(probe);
+    } catch (Throwable ignored) {
+    }
+    UntypedActivityHandle untyped =
+        start(extractActivityType(activityInterface, captured[0]), options, new Object[0]);
```
Sushisource: It seems like, in general with these `createTypeProbe` call sites, we could be doing error handling a bit more robustly. If they pass a method reference that doesn't exist on the class, for example, what kind of error are we gonna throw here?

Will we just end up with an NPE from dereffing `captured[0]`? We want to provide at least some kind of reasonable user-facing error in that case.

temporal-sdk/src/main/java/io/temporal/client/ActivityClientImpl.java
```diff
@@ -0,0 +1,564 @@
...
+  public UntypedActivityHandle start(
+      String activityType, StartActivityOptions options, @Nullable Object... args)
+      throws ActivityAlreadyStartedException {
+    ActivityClientCallsInterceptor.StartActivityOutput output =
+        invoker.startActivity(
+            new ActivityClientCallsInterceptor.StartActivityInput(
+                activityType,
+                Arrays.asList(args != null ? args : new Object[0]),
+                options,
+                Header.empty()));
```
Sushisource: Header is always empty here. I think headers need to be propagated from the contextPropagators in `ActivityClientOptions`.

temporal-sdk/src/main/java/io/temporal/client/ActivityCompletionClient.java
```diff
@@ -37,6 +37,20 @@ public interface ActivityCompletionClient {
...
+   *
+   * <p>Use this overload for standalone activities (not started from a workflow). For
+   * workflow-scheduled activities, use {@link #complete(String, Optional, String, Object)}.
+   *
+   * @param activityId id of the standalone activity
+   * @param activityRunId optional run id of the standalone activity, or {@code Optional.empty()}
+   * @param result of the activity execution
+   */
+  @Experimental
+  <R> void complete(String activityId, Optional<String> activityRunId, R result)
```
Sushisource: I wonder if we want these to be `standaloneComplete`, etc. Slightly worried that the overloads are so similar it'll be easy to get confused. I suppose this was already in the design though so we don't need to change it now unless others agree.

temporal-sdk/src/main/java/io/temporal/client/ActivityExecutionDescription.java
```diff
@@ -0,0 +1,267 @@
...
+import java.util.Optional;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+/**
+ * Detailed information about a standalone activity execution, returned by {@link
+ * ActivityHandle#describe()}.
+ */
+@Experimental
+public final class ActivityExecutionDescription extends ActivityExecutionMetadata {
```
Sushisource, line 28: This guy is missing `lastFailure()` which exists in the other SDKs. Probably also good to expose the raw info too, which I think we typically do?

temporal-sdk/src/main/java/io/temporal/client/StartActivityOptions.java
Sushisource: I'd say probably not since we're eliding that for the client generally.

temporal-sdk/src/main/java/io/temporal/internal/client/ActivityHandleImpl.java
Sushisource: We probably do. I think you might need to do something similar to what happens in the workflow client for getting results (and we probably also want overloads that you can give timeouts to, like it has).

@Quinn-With-Two-Ns I'm sure would be able to say something much more intelligent than me about this. I'd ask him.

Sushisource: We also probably want to cache results here rather than re-polling. Looks like the other SDKs are doing that.

temporal-sdk/src/main/java/io/temporal/internal/client/RootActivityClientInvoker.java
```diff
@@ -0,0 +1,281 @@
...
+                  + "' for activityId='"
+                  + input.getActivityId()
+                  + "'",
+              null);
+      }
+    }
+  }
+
+  @Override
+  public DescribeActivityOutput describeActivity(DescribeActivityInput input) {
```
Sushisource, line 286: I think this will need to support the long poll token, which probably also means the `describe()` on the handle needs an overload to take one too

Sushisource (commented):
Overall looking pretty good to me! I didn't review the tests (it's a long one!). Just a few things to address (and we'll want more than my stamp).

temporal-sdk/src/main/java/io/temporal/client/ActivityClientImpl.java
GregoryTravis: Seems like AtomicReference is the standard approach, but I think that has as semantic connotation I don't want to give, so I'm just adding a Box.

GregoryTravis: Got rid of Box and moved all the logic to MethodExtractor, to reduce boilerplate here.

temporal-sdk/src/main/java/io/temporal/client/ActivityClientImpl.java
GregoryTravis: Moved the logic into MethodExtractor, which throws an information RTE if the method doesn't match the activity.

temporal-sdk/src/main/java/io/temporal/client/ActivityCompletionClient.java
GregoryTravis: Done, after discussion with @maciejdudko

temporal-sdk/src/main/java/io/temporal/client/ActivityClient.java
```diff
@@ -0,0 +1,1419 @@
...
+
+  /** Creates a new {@code ActivityClient} with the given options. */
+  static ActivityClient newInstance(WorkflowServiceStubs stubs, ActivityClientOptions options) {
+    return new ActivityClientImpl(stubs, options);
+  }
+
+  /**
+   * Starts a standalone activity using a typed interface and an unbound method reference, and
+   * returns a typed handle. The activity type name is inferred from the method reference at runtime
+   * via a reflection proxy; the result type is captured from the generic type parameter.
```
Quinn-With-Two-Ns:
```suggestion
```

IMO this is just an implementation detail, I don't think the average user would benefit from that info and might even confuse them since a "reflection proxy" isn't a standard term. The only argument I could see is for GRAAL users, since they do need to know that.

temporal-sdk/src/main/java/io/temporal/client/ActivityClient.java
```diff
@@ -0,0 +1,1419 @@
...
+   *
+   * <p>Example:
+   *
+   * <pre>{@code
```
Quinn-With-Two-Ns, line 55: nit: I like the examples a lot, but I would make 2 small changes. I would add one for each overload, or ask AI to write them, and have just one example per method.

temporal-sdk/src/main/java/io/temporal/client/ActivityClient.java
```diff
@@ -0,0 +1,1419 @@
...
+
+  /**
+   * Returns a single page of activity executions matching the given query.
+   *
+   * @param query Temporal visibility query string
+   * @param nextPageToken token from a previous page, or {@code null} for the first page
+   * @param options pagination options such as page size
+   * @return a page of results and a token for the next page
+   */
+  ActivityListPage listExecutionsPaginated(
```
Quinn-With-Two-Ns: I don't think we need to expose this? We didn't for other list methods.

temporal-sdk/src/main/java/io/temporal/client/ActivityClientImpl.java
```diff
@@ -0,0 +1,474 @@
...
+  @Override
+  public <I, A1, A2, A3, R> ActivityHandle<R> start(
+      Class<I> activityInterface,
+      Functions.Func4<I, A1, A2, A3, R> activity,
+      StartActivityOptions options,
+      A1 arg1,
+      A2 arg2,
+      A3 arg3)
+      throws ActivityAlreadyStartedException {
+    Method method = MethodExtractor.extract(activityInterface, activity);
```
Quinn-With-Two-Ns, line 218: @Evanthx FYI your SANO code could copy this pattern.

temporal-sdk/src/main/java/io/temporal/client/ActivityClient.java
```diff
@@ -0,0 +1,1419 @@
...
+   * @param activityInterface the activity interface class
+   * @param activity unbound method reference with no activity arguments
+   * @param options scheduling options
+   * @param <I> the activity interface type
+   * @return a typed handle to the started activity
+   * @throws ActivityAlreadyStartedException if an activity with the same ID is already running
+   */
+  <I> ActivityHandle<Void> start(
+      Class<I> activityInterface, Functions.Proc1<I> activity, StartActivityOptions options)
+      throws ActivityAlreadyStartedException;
```
Quinn-With-Two-Ns:
```suggestion
```

Generally the Java SDK does NOT used checked exceptions. I think we should be consistent with other client calls.

temporal-sdk/src/main/java/io/temporal/client/ActivityAlreadyStartedException.java
```diff
@@ -0,0 +1,52 @@
...
+import javax.annotation.Nullable;
+
+/**
+ * Thrown by {@link ActivityClient#start} when the server returns an ALREADY_EXISTS error because an
+ * activity with the same ID is already running (or has a completed run that conflicts with the
+ * requested {@link StartActivityOptions#getIdReusePolicy()} / {@link
+ * StartActivityOptions#getIdConflictPolicy()}).
+ */
+@Experimental
+public final class ActivityAlreadyStartedException extends TemporalException {
```
Quinn-With-Two-Ns: For `WorkflowExecutionAlreadyStarted` we inherit from `WorkflowExecution` I don't see anything equivalent in this PR.

temporal-sdk/src/main/java/io/temporal/client/UntypedActivityHandle.java
```diff
@@ -0,0 +1,96 @@
...
+
+  /**
+   * Blocks until the standalone activity completes and returns the typed result. Use this overload
+   * for generic return types (e.g. {@code List<String>}).
+   *
+   * @param resultClass the class to deserialize the result into
+   * @param resultType the generic type to use for deserialization; may be {@code null}
+   * @throws ActivityFailedException if the activity failed, timed out, or was cancelled
+   */
+  <R> R getResult(Class<R> resultClass, @Nullable Type resultType) throws ActivityFailedException;
```
Quinn-With-Two-Ns:
```suggestion
  <R> R getResult(Class<R> resultClass, @Nullable Type resultType);
```

Same comment as above about checked exception

temporal-sdk/src/main/java/io/temporal/client/ActivityFailedException.java
```diff
@@ -0,0 +1,16 @@
...
+
+import io.temporal.common.Experimental;
+import io.temporal.failure.TemporalException;
+
+/**
+ * Thrown by {@link ActivityHandle#getResult()} when the standalone activity fails, times out, or is
+ * cancelled. The original cause can be retrieved via {@link #getCause()}.
+ */
+@Experimental
+public final class ActivityFailedException extends TemporalException {
```
Quinn-With-Two-Ns: I think we are missing a common `ActivityException` base class here that  `ActivityFailedException` and `ActivityAlreadyStartedException`

temporal-sdk/src/main/java/io/temporal/common/interceptors/ActivityClientCallsInterceptor.java
```diff
@@ -0,0 +1,352 @@
...
+ *
+ * <p>Prefer extending {@link ActivityClientCallsInterceptorBase} and overriding only the methods
+ * you need instead of implementing this interface directly. {@link
+ * ActivityClientCallsInterceptorBase} provides correct default implementations to all the methods
+ * of this interface.
+ */
+@Experimental
+public interface ActivityClientCallsInterceptor {
+
+  StartActivityOutput startActivity(StartActivityInput input)
```
Quinn-With-Two-Ns: nit: we should have Java docs for these. I know some interceptors don't but that is more tech debt then intentional IMO.

temporal-sdk/src/main/java/io/temporal/client/ActivityHandleWrapper.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.concurrent.CompletableFuture;
+import javax.annotation.Nullable;
+
+/**
+ * Package-private wrapper that adds typed result methods to an {@link UntypedActivityHandle},
+ * implementing {@link ActivityHandle}{@code <R>}. Created via {@link
+ * ActivityHandle#fromUntyped(UntypedActivityHandle, Class)} or {@link
+ * ActivityHandle#fromUntyped(UntypedActivityHandle, Class, Type)}.
+ */
+final class ActivityHandleWrapper<R> implements ActivityHandle<R> {
```
Quinn-With-Two-Ns:
```suggestion
final class ActivityHandleImpl<R> implements ActivityHandle<R> {
```
Was there a reason we used a fifferent naming convention then other classes?

temporal-sdk/src/main/java/io/temporal/internal/client/ActivityHandleImpl.java
Quinn-With-Two-Ns: Yeah we shouldn't just be calling `getResult` in `supplyAsync` , look at how `getResultAsync` on `WorkflowStubImpl` works you need to wire the async down to the GRPC layer

Quinn-With-Two-Ns: and yeah if we cahced results in other SDK's we should here as well

temporal-sdk/src/main/java/io/temporal/client/UntypedActivityHandle.java
```diff
@@ -0,0 +1,96 @@
...
+   */
+  <R> CompletableFuture<R> getResultAsync(Class<R> resultClass);
+
+  /**
+   * Returns a future for generic return types (e.g. {@code List<String>}).
+   *
+   * @param resultClass the class to deserialize the result into
+   * @param resultType the generic type to use for deserialization; may be {@code null}
+   */
+  <R> CompletableFuture<R> getResultAsync(Class<R> resultClass, @Nullable Type resultType);
```
Quinn-With-Two-Ns, line 97: We are missing the timeour overloads for `getResultAsync` on `WorkflowStub`

temporal-sdk/src/main/java/io/temporal/internal/client/RootActivityClientInvoker.java
Quinn-With-Two-Ns: I thought the SDK wasn't/didn't need to expose the long poll token

Quinn-With-Two-Ns (dismissed):
I didn't review the test or check more subtle behaviour, overall I think it looks really good! The `MethodExtractor` was a good idea. A few changes that if I owned the Java SDK I would want changed is:

* Mirror the exception hierarchy with `WorkflowException`
* Mirror the `getResultAsync` overloads in `WorkflowStub`
* Remove the use of checked exceptions, they are not used in the Java SDK like this
* Mirror the implementation of `getResultAsync` from `WorkflowStub`. `getResultAsync` should **not** but be calling `getResult` in an executor it should flow down to an async gRPC call.

temporal-sdk/src/main/java/io/temporal/internal/client/RootActivityClientInvoker.java
Sushisource: Bleh, I forgot that. I was just going on what existed in Go/Python. We can probably leave it out then.

temporal-sdk/src/main/java/io/temporal/client/ActivityClient.java
GregoryTravis: This is an unchecked exception, so I could remove the throws clause, but there are already a lot of throw clauses for unchecked exceptions in this repo, so I thought that was the convention.

temporal-sdk/src/main/java/io/temporal/client/ActivityAlreadyStartedException.java
GregoryTravis: This PR introduces ActivityAlreadyStartedException, but there are already pre-existing Activity.*Exception classes that extend TemporalException without an activity-specific base class. Do we want a base ActivityException, and would it be the base for all Activity.*Exception , or just for ActivityAlreadyStartedException?

temporal-sdk/src/main/java/io/temporal/client/ActivityClient.java
Quinn-With-Two-Ns: Ah yes that is my bad, I thought you added the `throw` clause because it was a checked exception. I actually wasn't aware there as a lot of throw clauses for unchecked exceptions. Not saying you are wrong just I didn't realize that. Doing quick search on the Java SDK just now I didn't see many, most uses of the `throws` clause I can see are for `TimeoutException` or interrupt. Generally I have just documented with `@throws` in the javadocs.

temporal-sdk/src/main/java/io/temporal/client/ActivityAlreadyStartedException.java
Quinn-With-Two-Ns: I think the existing exceptions are for in workflow activities so I don't think they make sense to share a common base class with the client side exceptions. I was imaging the common base class for just the new client side exceptions you added like how `WorkflowExecution` is used today.

temporal-sdk/src/main/java/io/temporal/client/ActivityHandleWrapper.java
GregoryTravis: No good reason; renamed.

temporal-sdk/src/main/java/io/temporal/client/ActivityAlreadyStartedException.java
GregoryTravis: Ah I see -- done.

temporal-sdk/src/main/java/io/temporal/client/ActivityClient.java
GregoryTravis: Yeah, I think it's best not to use throws clauses for unchecked exceptions. There are a bunch of such throws on master but I can remove them in a separate PR.

GregoryTravis:
> I didn't review the test or check more subtle behaviour, overall I think it looks really good! The `MethodExtractor` was a good idea. A few changes that if I owned the Java SDK I would want changed is:
> 
> * Mirror the exception hierarchy with `WorkflowException`
> * Mirror the `getResultAsync` overloads in `WorkflowStub`
> * Remove the use of checked exceptions, they are not used in the Java SDK like this
> * Mirror the implementation of `getResultAsync` from `WorkflowStub`. `getResultAsync` should **not** but be calling `getResult` in an executor it should flow down to an async gRPC call.

I believe I have addressed all these but please let me know if I missed anything.

temporal-sdk/src/main/java/io/temporal/activity/ActivityInfo.java
```diff
@@ -108,10 +133,19 @@ public interface ActivityInfo {
    * @deprecated use {@link #getNamespace()}
    */
   @Deprecated
+  @Nullable
   String getActivityNamespace();
```
maciejdudko, line 136: `getActivityNamespace` is not nullable.

temporal-sdk/src/main/java/io/temporal/activity/ActivityInfo.java
```diff
@@ -108,10 +133,19 @@ public interface ActivityInfo {
...
 
   String getNamespace();
```
maciejdudko, line 141: Missing docstring. Can copy from `getActivityNamespace`.

temporal-sdk/src/main/java/io/temporal/activity/ActivityInfo.java
```diff
@@ -108,10 +133,19 @@ public interface ActivityInfo {
...
 
+  /**
+   * @return {@code true} if this activity was scheduled by a workflow execution; {@code false} for
+   *     standalone activities started via {@link io.temporal.client.ActivityClient#start}.
+   */
```
maciejdudko, line 146: There are potentially other ways to start standalone activities (CLI, other SDKs, schedule), so the "started via" part is not quite correct. Better to remove it.

temporal-sdk/src/main/java/io/temporal/client/ActivityFailedException.java
```diff
@@ -0,0 +1,17 @@
+package io.temporal.client;
+
+import io.temporal.common.Experimental;
+import javax.annotation.Nullable;
+
+/**
+ * Thrown by {@link ActivityHandle#getResult()} when the standalone activity fails, times out, or is
+ * cancelled. The original cause can be retrieved via {@link #getCause()}.
```
maciejdudko: "When the activity was not successful" avoids the need to enumerate every possible cause.

Sushisource (approved):
This is looking good to me!

I think the only remaining question is if we actually should or should not be exposing the long poll token for describes to users: https://github.com/temporalio/sdk-java/pull/2858#discussion_r3140724994

From what I can tell looking in Slack I'm not sure we ever came to a firm conclusion there. Would be good to discuss w/ @Quinn-With-Two-Ns and @maciejdudko briefly.

temporal-sdk/src/main/java/io/temporal/client/ActivityExecutionDescription.java
```diff
@@ -0,0 +1,293 @@
...
+    return info().hasLastStartedTime()
+        ? ProtobufTimeUtils.toJavaInstant(info().getLastStartedTime())
+        : null;
+  }
+
+  /** Failure details from the last failed attempt. {@code null} if no failure has occurred. */
+  @Nullable
+  public io.temporal.api.failure.v1.Failure getLastFailure() {
+    return info().hasLastFailure() ? info().getLastFailure() : null;
+  }
```
maciejdudko, line 133: Should deserialize with data converter (inside getter, not ahead of time).

temporal-sdk/src/main/java/io/temporal/client/ActivityClient.java
```diff
@@ -0,0 +1,1618 @@
...
+
+  /**
+   * Starts a standalone activity and blocks until it completes, discarding the result.
+   *
+   * @param activityType the registered activity type name
+   * @param options scheduling options
+   * @param args activity arguments
+   * @throws ActivityFailedException if the activity failed, timed out, or was cancelled
+   */
+  void execute(String activityType, StartActivityOptions options, @Nullable Object... args);
```
maciejdudko: This can have a default implementation like the typed `execute` overloads. Same for the other untyped overloads.

temporal-sdk/src/main/java/io/temporal/client/ActivityClient.java
```diff
@@ -0,0 +1,1618 @@
...
+  ActivityExecutionCount countExecutions(String query);
+
+  /**
+   * Returns the count of activity executions matching the given query, with options.
+   *
+   * @param query Temporal visibility query string
+   * @param options count options
+   * @return execution count, optionally with aggregation groups
+   */
+  ActivityExecutionCount countExecutions(String query, ActivityCountOptions options);
```
maciejdudko: `ActivityCountOptions` is empty at the moment, let's not add this overload. It can always be added later. Remove class `ActivityCountOptions` too.

temporal-sdk/src/main/java/io/temporal/client/ActivityClient.java
```diff
@@ -0,0 +1,1618 @@
...
+  Stream<ActivityExecutionMetadata> listExecutions(String query);
+
+  /**
+   * Returns a stream of activity executions matching the given query, with options.
+   *
+   * @param query Temporal visibility query string
+   * @param options list options such as result limit
+   * @return stream of matching activity executions
+   */
+  Stream<ActivityExecutionMetadata> listExecutions(String query, ActivityListOptions options);
```
maciejdudko: Adding page size limit is not part of the current design (we're still figuring out the best API to expose this feature). Remove the `ActivityListOptions` class, this overload of `listExecutions`, and the ability to change `pageSize` in the interceptor (keeping the two interceptor methods is fine).

maciejdudko (changes requested):

GregoryTravis:
> I think the only remaining question is if we actually should or should not be exposing the long poll token for describes to users: [#2858 (comment)](https://github.com/temporalio/sdk-java/pull/2858#discussion_r3140724994)
> 
> From what I can tell looking in Slack I'm not sure we ever came to a firm conclusion there. Would be good to discuss w/ @Quinn-With-Two-Ns and @maciejdudko briefly.

Reverted, after discussion with @maciejdudko

temporal-sdk/src/main/java/io/temporal/internal/client/RootActivityClientInvoker.java
```diff
@@ -0,0 +1,382 @@
...
+          throw new ActivityFailedException(
+              "Interrupted while waiting for activity result for activityId='"
+                  + input.getActivityId()
+                  + "'",
+              input.getActivityId(),
+              input.getRunId(),
+              new InterruptedException());
+        }
+        continue;
+      }
```
dandavison, line 164: Similar to https://github.com/temporalio/sdk-typescript/pull/2029/changes/BASE..0086bdaf3f5b304703d8a5a21c25c3d3792522f1#r3156288887, I don't know the Java SDK very well but, if they don't exist already, can you add integration tests that demonstrate that it can wait for a result for longer than the server timeout, and for longer than the per-attempt client-side deadline, and that the poll loop can be aborted by some client-side technique? And also that it responds correctly when the activity is cancelled on the server.

It might help to compare the long-polling loop implementations in Go and Python.

I'm thinking that `PollActivityExecution` needs to be added to `LongPollUtil.isLongPoll`? (Are there any other locations that implement the client-side polling?)

temporal-sdk/src/main/java/io/temporal/client/ActivityExecutionDescription.java
GregoryTravis: Do you mean using dataConverter.failureToException() on the failure?

temporal-sdk/src/main/java/io/temporal/client/ActivityClient.java
GregoryTravis: I added default impls to four of the six untyped .execute/executeAsync methods -- two of them I could not, because they need a `resultClass` which isn't available, like the one you tagged here.

temporal-sdk/src/main/java/io/temporal/client/ActivityClient.java
GregoryTravis: What interceptor code do you mean? I don't see any place to change page size in interceptors. Should I also remove ActivityListPaginatedOptions?

GregoryTravis: Removed ActivityListOptions and ActivityListPaginatedOptions; keep both listActivities and listActivitiesPaginated in the interceptor.

temporal-sdk/src/main/java/io/temporal/client/ActivityCompletionException.java
```diff
@@ -25,7 +25,7 @@ protected ActivityCompletionException(ActivityInfo info, Throwable cause) {
             ? "WorkflowId="
                 + info.getWorkflowId()
                 + ", RunId="
```
mjameswh, line 27: I wonder if we should change the property name. Any down side to doing that?

```suggestion
                + ", WorkflowRunId="
```

If we do that, there are likely other places where should do the same.

temporal-sdk/src/main/java/io/temporal/internal/activity/ActivityInfoImpl.java
```diff
@@ -172,7 +203,7 @@ public Optional<Header> getHeader() {
 
   @Override
   public String toString() {
-    return "WorkflowInfo{"
+    return "ActivityInfo{"
         + "workflowId="
         + getWorkflowId()
         + ", runId="
```
mjameswh, line 213: We should no longer be using `getRunId` here. Either `workflowRunId` or `activityRunId`.

temporal-sdk/src/main/java/io/temporal/internal/client/RootActivityClientInvoker.java
GregoryTravis: Added tests for both client-side (RPC) timeout and server long-poll timeouts:

https://github.com/temporalio/sdk-java/pull/2858/changes#diff-b2feb080ae5a56f3b8000cd3e73f63a0d2d87ff7b06125f02e5f6573b4cf0fa8R27
https://github.com/temporalio/sdk-java/pull/2858/changes#diff-ed18cdf252f55abef848fbbda0a8cc01540f7c7ff48a3723229f0b87d82df6d2R26
https://github.com/temporalio/sdk-java/pull/2858/changes#diff-6b05493b308a0f1931ad2abf7c436330ffde09e8f443b50b5d778901722ab129R27

temporal-sdk/src/test/java/io/temporal/client/functional/GetActivityResultAsyncOverServerLongPollWaitTest.java
```diff
@@ -0,0 +1,73 @@
...
+ * Verifies that {@link ActivityHandle#getResultAsync()} continues polling after the server cuts a
+ * long poll and returns an empty response. The server cuts activity long polls after approximately
+ * {@value ACTIVITY_LONG_POLL_TIMEOUT_SECONDS} seconds; the activity here runs for 1.5× that so the
+ * server-side cut fires at least once before the activity completes.
+ *
+ * <p>It has a sync version in {@link GetActivityResultSyncOverServerLongPollWaitTest} that is split
+ * to reduce the total execution time.
+ */
+public class GetActivityResultAsyncOverServerLongPollWaitTest {
+  private static final int ACTIVITY_LONG_POLL_TIMEOUT_SECONDS = 20;
```
mjameswh, line 28: Where does this "20 sec" come from?

GregoryTravis: It's the server-side long poll time; see also GetResultsAsyncOverMaximumLongPollWaitTest.

These tests could be made more robust by checking that the retry count is at least two, but that would require more machinery in the test so I'm not sure it's a good idea.

temporal-sdk/src/main/java/io/temporal/client/ActivityCompletionException.java
GregoryTravis: Will do in a followup PR.

temporal-sdk/src/main/java/io/temporal/internal/activity/ActivityInfoImpl.java
GregoryTravis: Will do in a followup PR.

temporal-sdk/src/test/java/io/temporal/client/functional/GetActivityResultOverLongPollWaitTest.java
```diff
@@ -0,0 +1,96 @@
...
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+
+/**
+ * Verifies that {@link ActivityHandle#getResult()} and {@link ActivityHandle#getResultAsync()}
+ * continue polling across the server's long-poll boundary and complete correctly for a standalone
+ * activity that runs longer than one per-RPC timeout.
+ */
+public class GetActivityResultOverLongPollWaitTest {
```
mjameswh, line 27: If I read correctly, we have:
1. Activity Execution requested at 0s
2. Poll Activity Execution request times out at 5s; new Poll request
3. Activity completes at 7s (note: not 7.5s, because of rounding)
4. Tests fail at 10s

That means the test can fail if things get out of alignment by only 2s. That's really not much for our CI environments. I'd expect this to be flaky.

GregoryTravis: Increased to 5s.

temporal-sdk/src/main/java/io/temporal/client/ActivityHandle.java
```diff
@@ -0,0 +1,68 @@
...
+ */
+@Experimental
+public interface ActivityHandle<R> extends UntypedActivityHandle {
+
+  /**
+   * Blocks until the standalone activity completes and returns the typed result.
+   *
+   * @throws ActivityFailedException if the activity failed, timed out, or was cancelled
+   */
+  R getResult();
```
mjameswh, line 29: We also need a variant of this function that takes a timeout, no?

temporal-sdk/src/test/java/io/temporal/client/functional/StandaloneActivityTest.java
```diff
@@ -0,0 +1,980 @@
...
+              .setHeartbeatTimeout(Duration.ofSeconds(10))
+              .build();
+      ActivityHandle<Void> handle =
+          client.start(WaitForCancelActivity.class, WaitForCancelActivity::waitForCancel, opts);
+
+      assertTrue("Activity did not start within 30s", cancelLatch.await(30, TimeUnit.SECONDS));
+
+      CompletableFuture<Void> future = handle.getResultAsync(2, TimeUnit.SECONDS);
+
+      java.util.concurrent.ExecutionException ex =
```
mjameswh, line 917: I think this test could give false negative, i.e. a `TimeoutExceptino` is thrown but not for the timeout we expect to trigger.

Please add a stopwatch validation, like we have in https://github.com/temporalio/sdk-java/blob/d0cc4b45ce49f33b54b0f5625bf9839c8b9aee05/temporal-sdk/src/test/java/io/temporal/client/functional/GetResultsTimeoutTest.java#L65-L71.

temporal-sdk/src/test/java/io/temporal/client/functional/GetActivityResultOverLongPollWaitTest.java
```diff
@@ -0,0 +1,96 @@
...
+  @Test(timeout = 3 * LONG_POLL_TIMEOUT_SECONDS * 1000)
+  public void testGetResult() {
+    assumeTrue(SDKTestWorkflowRule.useExternalService);
+    activityClient.execute(SlowActivity.class, SlowActivity::run, slowOpts());
+  }
+
+  @Test(timeout = 3 * LONG_POLL_TIMEOUT_SECONDS * 1000)
+  public void testGetResultAsync() throws ExecutionException, InterruptedException {
+    assumeTrue(SDKTestWorkflowRule.useExternalService);
+    activityClient.executeAsync(SlowActivity.class, SlowActivity::run, slowOpts()).get();
```
mjameswh: nit: I'd use `start`+`getResultAsync` instead of `executeAsync`, to be very clear on which API we are really testing here.

temporal-sdk/src/test/java/io/temporal/client/functional/GetActivityResultOverLongPollWaitTest.java
```diff
@@ -0,0 +1,96 @@
...
+        .setId("slow-act-" + UUID.randomUUID())
+        .setTaskQueue(testWorkflowRule.getTaskQueue())
+        .setScheduleToCloseTimeout(Duration.ofMinutes(1))
+        .build();
+  }
+
+  @Test(timeout = 3 * LONG_POLL_TIMEOUT_SECONDS * 1000)
+  public void testGetResult() {
+    assumeTrue(SDKTestWorkflowRule.useExternalService);
+    activityClient.execute(SlowActivity.class, SlowActivity::run, slowOpts());
```
mjameswh: nit: prefer `start` + `getResult`, to be very clear on which API is getting tested.

temporal-sdk/src/test/java/io/temporal/client/functional/GetActivityResultSyncOverServerLongPollWaitTest.java
```diff
@@ -0,0 +1,72 @@
...
+        .setId("slow-act-" + UUID.randomUUID())
+        .setTaskQueue(testWorkflowRule.getTaskQueue())
+        .setScheduleToCloseTimeout(Duration.ofMinutes(2))
+        .build();
+  }
+
+  @Test(timeout = 2 * ACTIVITY_LONG_POLL_TIMEOUT_SECONDS * 1000)
+  public void testGetResult() {
+    assumeTrue(SDKTestWorkflowRule.useExternalService);
+    newActivityClient().execute(SlowActivity.class, SlowActivity::run, slowOpts());
```
mjameswh: nit: prefer `start`+`getResult`

temporal-sdk/src/main/java/io/temporal/client/ActivityCompletionClientImpl.java
```diff
@@ -108,10 +146,26 @@ public ActivityCompletionClient withContext(@Nonnull ActivitySerializationContex
...
         .setRunId(runId.orElse(""))
         .build();
   }
+
+  private static WorkflowExecution toStandaloneExecution(Optional<String> activityRunId) {
+    return WorkflowExecution.newBuilder()
+        .setWorkflowId("") // empty workflowId = standalone activity
+        .setRunId(activityRunId.orElse(""))
+        .build();
+  }
```
maciejdudko, line 172: This is a bad pattern. Instead, `ManualActivityCompletionClientFactory` should have a `getClient` overload that does not take `WorkflowExecution`, but activity ID and run ID as strings. You can change `ManualActivityCompletionClientImpl` so that it decomposes `WorkflowExecution` into separate workflow ID and run ID fields in the constructor, that way we can avoid an if/else in the individual methods for standalone activities.

temporal-sdk/src/main/java/io/temporal/client/ActivityExecutionDescription.java
```diff
@@ -0,0 +1,281 @@
...
+    this.namespace = namespace;
+  }
+
+  private static @Nullable String nullIfEmpty(String s) {
+    return s == null || s.isEmpty() ? null : s;
+  }
+
+  private ActivityExecutionInfo info() {
+    return info;
+  }
```
maciejdudko: This method is unnecessary.

temporal-sdk/src/main/java/io/temporal/common/interceptors/ActivityClientCallsInterceptor.java
```diff
@@ -0,0 +1,444 @@
...
+      GetActivityResultInput<R> input) {
+    return CompletableFuture.supplyAsync(
+        () -> {
+          try {
+            return getActivityResult(input);
+          } catch (ActivityFailedException e) {
+            throw new java.util.concurrent.CompletionException(e);
+          }
+        });
+  }
```
maciejdudko: This implementation breaks interceptor chaining. Instead, this method should be abstract here, and implemented in `RootActivityClientInvoker`. You can use the same code as here - this would mean that `getActivityResultAsync` will not be calling `getActivityResult` interceptors, but this is how workflow interceptors already work, and can be easier to reason about.

Workflow interceptor docstrings have this line:
> If you implement this method, {@link #getResult} most likely needs to be implemented too.
If you implement this, please add a similar line to both `getActivityResult` and `getActivityResultAsync`.

temporal-sdk/src/main/java/io/temporal/internal/activity/ActivityInfoImpl.java
maciejdudko, line 148: `getWorkflowNamespace` should return null if there's no workflow execution.

temporal-sdk/src/main/java/io/temporal/activity/ActivityInfo.java
```diff
@@ -92,10 +115,12 @@ public interface ActivityInfo {
...
    * #getActivityNamespace()} methods.
    *
-   * @return the Namespace of Workflow Execution that scheduled the Activity.
+   * @return the Namespace of Workflow Execution that scheduled the Activity, or the activity
+   *     namespace for standalone activities.
    * @deprecated use {@link #getNamespace()}
    */
   @Deprecated
+  @Nullable
   String getWorkflowNamespace();
```
maciejdudko, line 124: Revert this doc change. getWorkflowNamespace should return null for standalone activities.

temporal-sdk/src/main/java/io/temporal/internal/client/ActivityHandleImpl.java
```diff
@@ -0,0 +1,144 @@
...
+  public <R> CompletableFuture<R> getResultAsync(
+      long timeout, TimeUnit unit, Class<R> resultClass, @Nullable Type resultType) {
+    boolean noTimeout = timeout == Long.MAX_VALUE && unit == TimeUnit.MILLISECONDS;
+
+    CompletableFuture<?> cached = noTimeoutResult.get();
+    if (cached != null && (noTimeout || cached.isDone())) {
+      @SuppressWarnings("unchecked")
+      CompletableFuture<R> typed = (CompletableFuture<R>) cached;
+      return typed;
+    }
```
maciejdudko: This will result in invalid cast if `getResultAsync` is called multiple times with different `resultClass`. Better to remove caching for now and just fetch a fresh result every time, we can revisit caching later.

temporal-sdk/src/main/java/io/temporal/workflow/Functions.java
```diff
@@ -81,4 +81,16 @@ public interface Proc6<T1, T2, T3, T4, T5, T6>
       extends TemporalFunctionalInterfaceMarker, Serializable {
     void apply(T1 t1, T2 t2, T3 t3, T4 t4, T5 t5, T6 t6);
   }
+
+  @FunctionalInterface
+  public interface Func7<T1, T2, T3, T4, T5, T6, T7, R>
+      extends TemporalFunctionalInterfaceMarker, Serializable {
+    R apply(T1 t1, T2 t2, T3 t3, T4 t4, T5 t5, T6 t6, T7 t7);
+  }
```
maciejdudko: Nit: this could be moved up with the other `Func` interfaces.

temporal-sdk/src/main/java/io/temporal/activity/ActivityInfo.java
```diff
@@ -110,8 +135,19 @@ public interface ActivityInfo {
...
+  /**
+   * @return the Namespace of this Activity Execution.
+   */
   String getNamespace();
 
+  /**
+   * @return {@code true} if this activity was scheduled by a workflow execution; {@code false} for
+   *     standalone activities.
+   */
+  default boolean isWorkflowActivity() {
```
maciejdudko: Rename to `isInWorkflow` for consistency with other SDKs.

temporal-sdk/src/test/java/io/temporal/client/StartActivityOptionsTest.java
maciejdudko, line 1: Nit: we should add a test or 2 tests that check that both only `scheduleToCloseTimeout` and only `startToCloseTimeout` are fine.

temporal-sdk/src/test/java/io/temporal/common/interceptors/ActivityClientCallsInterceptorChainTest.java
maciejdudko, line 1: This test suite does not meaningfully test anything in real implementation. Instead, the tests should create `ActivityClient` with 2 interceptors and verify that they are being called in order when the client method is called.

temporal-sdk/src/test/java/io/temporal/internal/activity/ActivityInfoImplTest.java
maciejdudko, line 1: Nit: personal taste, but I would rename this suite to `ActivityInfoImplTest`.

temporal-sdk/src/test/java/io/temporal/internal/client/ActivityHandleImplTest.java
```diff
@@ -0,0 +1,227 @@
...
+    when(interceptor.getActivityResultAsync(any(GetActivityResultInput.class)))
+        .thenReturn(CompletableFuture.completedFuture(output));
+
+    UntypedActivityHandle handle = new ActivityHandleImpl("id", null, interceptor);
+    CompletableFuture<String> future = handle.getResultAsync(String.class);
+    assertEquals("async-result", future.get());
+  }
+
+  @Test
+  public void testDescribeNoToken() {
```
maciejdudko: Since we don't support long poll describe, rename to `testDescribe`.

temporal-sdk/src/test/java/io/temporal/client/functional/StandaloneActivityTest.java
```diff
@@ -0,0 +1,980 @@
...
+  public void testGetActivityHandleExistingActivitySucceeds() {
+    assumeTrue(SDKTestWorkflowRule.useExternalService);
+    ActivityClient client = newActivityClient();
+    String activityId = uniqueId();
+
+    ActivityHandle<String> handle =
+        client.start(SimpleActivity.class, SimpleActivity::execute, simpleOpts(activityId), "test");
+    assertEquals("echo:test", handle.getResult());
+
+    UntypedActivityHandle handle2 = client.getHandle(activityId, handle.getActivityRunId());
```
maciejdudko, line 335: We should have a test for `getHandle` with null run ID.

temporal-sdk/src/test/java/io/temporal/client/functional/StandaloneActivityTest.java
```diff
@@ -0,0 +1,980 @@
...
+            .execute(
+                SimpleActivity.class, SimpleActivity::execute, simpleOpts(uniqueId()), "hello");
+    assertEquals("echo:hello", result);
+  }
+
+  @Test
+  public void testExecuteActivityVoidResult() {
+    assumeTrue(SDKTestWorkflowRule.useExternalService);
+    newActivityClient().execute(VoidActivity.class, VoidActivity::execute, simpleOpts(uniqueId()));
+  }
```
maciejdudko, line 257: We should have a test that checks if the void `execute` overload properly discards result of non-void activity.

maciejdudko (changes requested):
Looks good, but there are a couple things that need to be fixed before merging, most notably caching inside `ActivityHandleImpl` and how `AsyncCompletionClient` creates `ManualActivityCompletionClient` for standalone activities. See comments in code.

temporal-sdk/src/main/java/io/temporal/client/ActivityCompletionClientImpl.java
GregoryTravis: Added TODO, after discussion about the best apporach for this.

temporal-sdk/src/main/java/io/temporal/activity/ActivityInfo.java
GregoryTravis: Changed to `@return the Namespace of Workflow Execution that scheduled the Activity, or null for standalone activities.`

temporal-sdk/src/main/java/io/temporal/common/interceptors/ActivityClientCallsInterceptor.java
GregoryTravis: Removed default ActivityClientCallsInterceptor.getActivityResultAsync.

maciejdudko (approved):
LGTM!

PR #2859, opened 2026-04-22
temporal-spring-ai: discover MCP clients by type, not by bean name

donald-pinckney:
## What was changed
`McpPlugin.getMcpClients()` now calls `ApplicationContext.getBeansOfType(McpSyncClient.class)` instead of looking up the hard-coded bean name `"mcpSyncClients"`. The unchecked cast from the old `(List<McpSyncClient>) someObject` path is gone; discovered bean names are logged alongside the count for easier debugging.

New `McpPluginTest` covers three cases via Mockito:
- Two distinct `McpSyncClient` beans discovered by type both reach `McpClientActivityImpl`.
- Zero beans leaves the worker queued for deferred registration; `afterSingletonsInstantiated` handles "still no beans" without crashing.
- Beans that appear between `initializeWorker` and `afterSingletonsInstantiated` are picked up on the deferred attempt.

`build.gradle` gains `spring-ai-mcp` as a `testImplementation` so the tests can reference `McpSyncClient` directly for mocking. The main-scope dep stays `compileOnly`; nothing changes for users who don't use MCP.

## Why?
The old bean-name lookup broke silently whenever Spring AI changed the name its MCP auto-configuration used (e.g. if `mcpSyncClients` ever got renamed, or if sync-vs-async support got refactored). It also ignored user-defined `McpSyncClient` beans declared outside Spring AI's auto-configuration. Type-based discovery is more robust, works with arbitrary user configuration, and removes an unchecked cast that could have misbehaved at runtime if someone had registered a differently-typed `List` bean at that name.


temporal-spring-ai/src/test/java/io/temporal/springai/plugin/McpPluginTest.java
```diff
@@ -0,0 +1,100 @@
...
+    Worker worker = mock(Worker.class);
+    plugin.initializeWorker("mcp-tq", worker);
+
+    ArgumentCaptor<Object> captor = ArgumentCaptor.forClass(Object.class);
+    verify(worker, atLeastOnce()).registerActivitiesImplementations(captor.capture());
+    Object registered = captor.getValue();
+    assertEquals(McpClientActivityImpl.class, registered.getClass());
+
+    // Duplicate-name protection in McpClientActivityImpl still fires if two clients share a
+    // clientInfo().name(); here they differ ("alpha" vs "beta") so construction succeeds.
```
brianstrauch: Should this be a separate test case instead of a comment?

donald-pinckney: Yes it should be. Added!

brianstrauch (approved):

maciejdudko (approved):

PR #2861, opened 2026-04-22
temporal-spring-ai: README expansion and default-model-name consolidation

donald-pinckney:
## What was changed
- Moved the `"default"` model-name constant from `SpringAiPlugin.DEFAULT_MODEL_NAME` to `ChatModelTypes.DEFAULT_MODEL_NAME` and replaced `ChatModelActivityImpl`'s hard-coded literals with the shared constant. `ChatModelTypes` is the right home — the activity package already imports it, so this avoids an activity → plugin import direction that would've been a package cycle.
- README gains four new sections:
  - **Migrating from plain Spring AI** — a three-row table showing which three substitutions get an existing `ChatClient`-based service running on Temporal.
  - **Activity options and retry behavior** — documents the default timeouts / non-retryable classification and the `forModel(name, ActivityOptions)` escape hatch.
  - **Known limitations** — streaming, `defaultToolContext`, child-workflow stubs as tools, the 1 MiB media byte[] cap, and the `ChatOptions.copy()` constraint that affects `ChatClient.defaultOptions` users.
  - **Observability** — pointer to the Temporal Java SDK docs plus a note that `TemporalChatClient` accepts a Micrometer `ObservationRegistry`.

## Why?
The magic string had already drifted once: `SpringAiPlugin.DEFAULT_MODEL_NAME` existed but `ChatModelActivityImpl` kept its own copy of the literal `"default"`. Consolidating prevents the next drift, and moving it to `ChatModelTypes` keeps package directions sensible.

The README covered quick-start and tool types but left adopters guessing about migration steps, failure modes, and what isn't supported. Each new section answers a question we've seen or anticipate from early users of the plugin.


temporal-spring-ai/README.md
```diff
@@ -97,6 +97,93 @@ public class MyTools {
...
+                .build();
+    }
+
+    String respond(String goal) {
+        return chatClient.prompt().user(goal).call().content();
+    }
+}
+```
+
+Inside a workflow it becomes:
```
brianstrauch:
```suggestion
Inside a Temporal Workflow it becomes:
```

temporal-spring-ai/README.md
```diff
@@ -97,6 +97,93 @@ public class MyTools {
...
+ActivityChatModel chatModel = ActivityChatModel.forModel("reasoning", opts);
+```
+
+For repeated per-model overrides, declare a `ChatModelActivityOptions` bean and auto-configuration wires the map into the plugin. See that class's javadoc for the pattern.
+
+`ActivityMcpClient.create()` / `create(ActivityOptions)` behave the same way with a 30-second default timeout.
+
+## Known limitations
+
+- **Streaming (`chatClient.stream(...)`)** — not supported. Activity results are unary; a streaming API doesn't fit Temporal's durability model without buffering. Use `.call()` instead.
```
brianstrauch: Will this change when streaming is released and should we edit this accordingly?

donald-pinckney: Yes, very good catch. I've simplified it to just say its not currently supported, none of that nonsense explanation needed.

donald-pinckney: Later with streaming we will support it then.

brianstrauch (approved):

maciejdudko (approved):

PR #2862, opened 2026-04-27
Setup `contrib/` directory with CODEOWNERS that is extensible for non-SDK team extensions.

donald-pinckney:
## What was changed
Following initial discussion last week, we'd like to separate out extensions from core language SDKs, and allow extensions to be code owned by other teams. This PR:
- Adds an `contrib/` directory.
- Moves `temporal-spring-ai` to `contrib/temporal-spring-ai`.
- Moves `temporal-opentracing` to `contrib/temporal-opentracing`.
- Adds AI Foundations team as a code owner to `contrib/temporal-spring-ai`.

Obviously more things *could* move to `contrib/` than is what is in this PR, but that can be determined by SDK team's preference at any point. Even though `temporal-opentracing` is NOT owned by AI Foundations team, I moved it to `contrib/` because a) it is very clearly an extension, and b) as a simple example of something moved to `extensions/` that retains full SDK team code ownership.

## Why?
Other teams are hard-blocked on getting approvals to merge from the SDK team, for PRs on content that the SDK team does not conceptually own.

## Checklist
1. Add AI SDK team as contributors on this repository (otherwise CODEOWNERS file is invalid).
2. In branch protection, make sure to enable the option "Require review from Code Owners".


donald-pinckney:
`extensions/` has been renamed to `contrib/`

maciejdudko (approved):

PR #2870, opened 2026-05-05
Add ContinueAsNew search attribute tests for test server

### LINKED ISSUE #2655: Testing server continue-as-new doesn't carry memo/search attributes
## Expected Behavior
When using ContinueAsNew with the testing server, the requested Memo and Search Attributes should be propagated into the new workflow execution.

This is what the production Temporal server does.

## Actual Behavior
The memo and search attributes in the command are dropped. See [here](https://github.com/temporalio/sdk-java/blob/31d044cfb67a29f456d0f699296e40cd4f05eb1e/temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowService.java?plain=1#L1531-L1571) for where the start request is constructed—the memo and search attributes are not extracted from the `ContinueAsNewWorkflowExecutionCommandAttributes`.

  1. Start a testing server
  2. Run a workflow that uses continue-as-new with search attributes / memo

## Specifications
  - Version: v1.31.0
  - Platform: all

baekgyu-kim:
## What was changed
Added regression test coverage for search-attribute behavior on ContinueAsNew in the in-memory test server. 
No production code is changed.

The correct behavior already exists on master: on ContinueAsNew the server propagates the command's search attributes to the new run, and the SDK carries the previous run's search attributes into the command when none are specified. 
This PR adds the tests that were missing to lock that behavior in.

Note: an earlier version of this PR also made the test server inherit memo and search attributes from the previous run. 
That does not match the real Temporal service, which builds the new run's StartWorkflowExecutionRequest directly from the command (`Memo: command.Memo`, `SearchAttributes: command.SearchAttributes`) and inherits neither.
Search attributes only survive because the SDK carries them into the command, while memos are not carried over and are effectively cleared. Those production changes were removed.

## Why?
To verify that ContinueAsNew carries search attributes in the test server, matching the real Temporal service, and to guard against regressions. Memo is intentionally not preserved, consistent with the real service.

## Checklist

1. Closes #2655.

2. How was this tested:
- Added two regression tests in ContinueAsNewTest: search-attribute carry-over when none are specified, and explicit override.
- Verified on both the in-process test server and a real dev server.
```bash
./gradlew :temporal-test-server:test --tests io.temporal.testserver.functional.ContinueAsNewTest
```

3. Any docs updates needed?
No. This is test-only coverage for existing test server behavior; no public API or behavior change.

### DIFF
--- a/temporal-test-server/src/test/java/io/temporal/testserver/functional/ContinueAsNewTest.java
+++ b/temporal-test-server/src/test/java/io/temporal/testserver/functional/ContinueAsNewTest.java
@@ -1,15 +1,24 @@
 package io.temporal.testserver.functional;
 
+import io.temporal.api.common.v1.Payload;
 import io.temporal.api.common.v1.WorkflowExecution;
+import io.temporal.api.enums.v1.EventType;
+import io.temporal.api.history.v1.HistoryEvent;
+import io.temporal.api.history.v1.WorkflowExecutionStartedEventAttributes;
 import io.temporal.client.WorkflowOptions;
 import io.temporal.client.WorkflowStub;
+import io.temporal.common.SearchAttributeKey;
+import io.temporal.common.SearchAttributes;
 import io.temporal.common.WorkflowExecutionHistory;
+import io.temporal.common.converter.DefaultDataConverter;
 import io.temporal.common.interceptors.*;
 import io.temporal.testing.internal.SDKTestWorkflowRule;
 import io.temporal.testserver.functional.common.TestWorkflows;
 import io.temporal.worker.WorkerFactoryOptions;
 import io.temporal.workflow.ContinueAsNewOptions;
 import io.temporal.workflow.Workflow;
+import io.temporal.workflow.WorkflowInterface;
+import io.temporal.workflow.WorkflowMethod;
 import java.time.Duration;
 import org.junit.Assert;
 import org.junit.Rule;
@@ -24,7 +33,7 @@ public class ContinueAsNewTest {
               WorkerFactoryOptions.newBuilder()
                   .setWorkerInterceptors(new StripsTqFromCanInterceptor())
                   .build())
-          .setWorkflowTypes(TestWorkflow.class)
+          .setWorkflowTypes(TestWorkflow.class, OverridingWorkflow.class)
           .build();
 
   @Test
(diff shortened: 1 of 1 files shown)


maciejdudko (changes requested):
Hi @baekgyu-kim, thank you for your contribution. Unfortunately there was a mistake in the referenced issue - the real server only preserves Search Attributes on Continue-as-New; Memos are cleared. Which means the PR in its current form does not match the real server behavior. Please remove the changes to Memo handling and only keep the changes to Search Attributes handling. Make sure all tests pass both when run on test server and on real server. You can make the tests use a local dev server by setting environment variables `USE_DOCKER_SERVICE=true` and `USE_EXTERNAL_SERVICE=true`. You can start a local dev server by using Temporal CLI: `temporal server start-dev --namespace UnitTest --search-attribute CustomKeywordField=Keyword`

baekgyu-kim:
Hello @maciejdudko , 
Thank you for taking the time to review this, and for catching the issue. 

I checked the server code and confirmed the mismatch. 
On Continue-as-New, the server builds the new run's `StartWorkflowExecutionRequest` directly from the command (`Memo: command.Memo`, `SearchAttributes: command.SearchAttributes`), inheriting nothing from the previous run. (`mutable_state_impl.go`)
The difference is on the SDK side. It carries the previous run's search attributes into the command, but not the memo. 
So search attributes survive, while the memo arrives empty and is effectively cleared.

The correct search-attribute propagation already existed on master.
**So I left the existing code unchanged and added test coverage for search-attribute behavior on Continue-as-New: SDK carry-over when none are specified, and explicit override.**

I verified that the ContinueAsNewTest cases pass on both the in-process test server and a real dev server (`USE_DOCKER_SERVICE=true`, `USE_EXTERNAL_SERVICE=true`).

Whenever you have a moment, I would appreciate it if you could take another look. 
Thank you again for your guidance.

maciejdudko:
Apologies again. After looking more into both test server and real server implementation, it looks like the server doesn't preserve Search Attributes across CaN either. We have client-side logic in the SDK to automatically transfer SA when none are set in the request ([SyncWorkflowContext.java](https://github.com/temporalio/sdk-java/blob/1386d4b39f6f2e3d7e3bb69825c0ddb89be01a4a/temporal-sdk/src/main/java/io/temporal/internal/sync/SyncWorkflowContext.java#L1399)), and this gives the illusion it's the server doing this work.

The bottom line is - the test server preserves neither Memos nor Search Attributes, this behavior matches the real server, the tests in this PR exercise SDK behavior rather than test server behavior, and issue #2655 was invalid to begin with.

I'm sorry for wasting your time on this. Still, I really appreciate your contribution and you're welcome to look at other open issues and submit more PRs.

baekgyu-kim:
Hi @maciejdudko ,
Thank you for looking into this so carefully and for the follow-up.

That makes sense, and it matches what I found as well. The test server already behaves the same as the real server here, and the preservation actually comes from the SDK rather than the server. 
So there is nothing to fix on the test server side, and these tests end up covering SDK behavior.

Thank you again for your time and guidance. 
I'll take a look at other open issues.
Have a good day.

PR #2871, opened 2026-05-05
Improve CONTRIBUTING guide and streamline local dev requirements

dplyukhin:
## What was changed

- Cleaned up `CONTRIBUTING.md`:
  - Lowered minimum JDK from 23 down to 21
  - Explained how to test against an external Temporal service
- Renamed the `USE_DOCKER_SERVICE` environment variable to `USE_EXTERNAL_SERVICE` because the external service no longer requires Docker
- Trimmed the JDK requirements in `temporal-sdk/build.gradle`: `virtualThreadTests`, `jackson3Tests`, and `newJavaTests` no longer hardcode JDK 17/21 launchers. Locally, a single JDK 21+ install is enough.
- Updated `mise.toml` to pin `temurin-21` (down from `temurin-23`)

## Why?
The previous contributor setup was unnecessarily heavy and sometimes misleading. New contributors had to install three different JDKs and a Temporal server before they could run `./gradlew test`. The `USE_DOCKER_SERVICE` name was also misleading since nothing about the flag is Docker-specific.


CONTRIBUTING.md
```diff
@@ -41,26 +47,24 @@ fatal: No names found, cannot describe anything.
 This can be done resolved by running `git fetch --tags` on your branch. Note, make sure your fork has tags copied from
 the main repo.
 
-## Test and Build
+## Testing
 
-Run a local temporal server with the [temporal CLI](https://docs.temporal.io/cli#installation):
+Some tests assume you have the temporal server running, and others don't. By default, Gradle only runs the tests that don't require a server:
```
Quinn-With-Two-Ns: This paragraph isn't accurate, or is at least misleading. The tests by default use the built in test server, and you can optionally run against a real server instead

CONTRIBUTING.md
```diff
@@ -41,26 +47,24 @@ fatal: No names found, cannot describe anything.
...
-./gradlew test
+temporal server start-dev
 ```
 
-Build with:
+Run the tests that require a temporal server:
 
 ```bash
-./gradlew build
+USE_DOCKER_SERVICE=true ./gradlew test
```
Quinn-With-Two-Ns: The environment variable name predate being able to start the Temporal server with the CLI, so docker was the main way to run the Temporal server locally, but we should probably update it now (doesn't need to be in this PR just adding some historical context)

CONTRIBUTING.md
dplyukhin: Thanks for the review :) I'm trying to say that some tests get skipped if you don't have an external server running, like [ActivityPauseTest](https://github.com/temporalio/sdk-java/blob/master/temporal-sdk/src/test/java/io/temporal/activity/ActivityPauseTest.java). But on second glance, a lot of those tests won't even pass unless you run the server with a [million flags](https://github.com/temporalio/sdk-java/blob/master/.github/workflows/ci.yml#L92-L117). Do you think I should paste the flags into contributing.md, or just remove any mention about running tests with an external server?

CONTRIBUTING.md
dplyukhin: Sure, I can easily add that to the PR. You're not worried about Hyrum's law? https://xkcd.com/1172/

CONTRIBUTING.md
Quinn-With-Two-Ns: Yeah I understand , I thought it was a bit misleading because most test do require a "server" its just the SDK will also start a type of server for you. I know that sounds like a bit of a nitpick. I would personally say something like:

"By default most integration tests run against the built in time skipping test server. Some tests may be skipped if the time skipping test server does not support the required feature. You can run tests against a real server by setting ..."

Yeah you need the right set of flags and version of the server to actually pass CI, Maybe for now just link to the GH action where it is defined? Long term it would be nice if the Java SDK started the sever for you as well like we do in some other SDKs.

CONTRIBUTING.md
Quinn-With-Two-Ns: Not very worried, generally we at Temporal are the only people doing consistent active dev work on the Java SDK.

CONTRIBUTING.md
```diff
@@ -41,26 +47,32 @@ fatal: No names found, cannot describe anything.
...
+1. Install the [temporal CLI](https://docs.temporal.io/cli#installation), which comes with a built-in dev server.
+2. Find the flags that the dev server will need to run the tests by grepping for `temporal server` in [./github/workflows/ci.yml](./github/workflows/ci.yml).
+3. Start the server: 
+```bash
+temporal server start-dev --YOUR-FLAGS-HERE
+```
+4. Set the `USE_EXTERNAL_SERVICE` environment variable and run the tests: 
 ```bash
-./gradlew build
+USE_EXTERNAL_SERVICE=true ./gradlew test
```
dplyukhin, line 72: @Quinn-With-Two-Ns Totally agree with the nitpick. How's this?

CONTRIBUTING.md
dplyukhin: I've gone ahead and renamed it in this PR.

Quinn-With-Two-Ns (approved):

temporal-sdk/src/test/java/io/temporal/activity/ActivityPauseTest.java
```diff
@@ -62,6 +62,7 @@ public String execute() {
   }
 
   public static class HeartBeatingActivityImpl implements TestActivities.TestActivity1 {
+
```
maciejdudko: I assume this is an accidental new line. Undo this change as to not pollute file-level history.

temporal-sdk/src/test/java/io/temporal/worker/StickyWorkerTest.java
```diff
@@ -536,7 +519,6 @@ public TestEnvironmentWrapper(WorkerFactoryOptions options) {
               .setMetricsScope(metricsScope)
               .setWorkflowClientOptions(clientOptions)
               .setWorkerFactoryOptions(options)
-              .setUseExternalService(useExternalService)
```
maciejdudko, line 539: I agree with the other changes, but this line should not be removed. Without it, the test will unconditionally use the in-memory server.

CONTRIBUTING.md
```diff
@@ -8,8 +8,14 @@ before we can merge in any of your changes
 
 ## Development Environment
 
-- Java 23+
-- Docker to run Temporal Server
+- **Java 23+** is required to run Gradle and to compile the project.
+- Some tests assume you also have **Java 17 and 21** installed.
```
maciejdudko: I don't think this is correct? Some tests require a minimum of 17 or 21, but all of them will run fine on 23 and later too. Is there any test that doesn't?

mise.toml
```diff
@@ -1,15 +1,2 @@
...
-#   - `java@temurin-23`
-# 
-# Java 21+ is required for anything that requires gradle.
-# 
-# If you find out that gradle isn't automatically picking it up older JDKs as
-# toolchains, see https://mise.jdx.dev/lang/java.html#gradle-toolchains-detection.
-
 [tools]
-java = "temurin-23"
+java = ["temurin-23", "temurin-21", "temurin-17"]
```
maciejdudko: This might be an unnecessary change, see my comment in CONTRIBUTING.md.

maciejdudko (changes requested):

CONTRIBUTING.md
dplyukhin: [These lines](https://github.com/temporalio/sdk-java/blob/master/temporal-sdk/build.gradle#L175-L239) cause the build to fail unless JDK 17 and 21 toolchains are installed. It looks like we really just wanted *at least* 17 and 21, so I've updated those lines in the latest commit.

temporal-sdk/src/test/java/io/temporal/worker/StickyWorkerTest.java
dplyukhin: Nice catch!!

maciejdudko (approved):

PR #2874, opened 2026-05-11
Fixed a bug with spaces in WorkflowIds when creating links in the UI.

Evanthx:
I used a workflow id that had a space in it, and realized that the links to it in the UI workflow step were broken. They should have linked to "workflow%20id" for example, but instead linked to "workflow%2Bid". This will only show up if you have a space in the workflow ID, so that is why it was missed.

The root issue turns out to be that the URLEncoder is meant for forms, where this is evidently the standard. 

The solution used here is to encode the string, then replace any "+" values with "%20". If the string has a plus, it will be converted to "%2B" by the call, so will not be affected by the replace (that is being double checked by the unit test.)

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/common/LinkConverter.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/common/LinkConverter.java
@@ -32,11 +32,19 @@ public class LinkConverter {
 
   public static io.temporal.api.nexus.v1.Link workflowEventToNexusLink(Link.WorkflowEvent we) {
     try {
+
       String url =
           String.format(
               linkPathFormat,
               URLEncoder.encode(we.getNamespace(), StandardCharsets.UTF_8.toString()),
-              URLEncoder.encode(we.getWorkflowId(), StandardCharsets.UTF_8.toString()),
+              // The 'replace' below handles spaces - the encoder will convert them to a plus,
+              // which the UI then handles as a plus, thus breaking the link as the
+              // space is lost.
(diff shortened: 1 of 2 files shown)


Quinn-With-Two-Ns (approved):

mjameswh (approved):

temporal-sdk/src/main/java/io/temporal/internal/common/LinkConverter.java
```diff
@@ -32,11 +32,19 @@ public class LinkConverter {
...
-              URLEncoder.encode(we.getWorkflowId(), StandardCharsets.UTF_8.toString()),
+              // The 'replace' below handles spaces - the encoder will convert them to a plus,
+              // which the UI then handles as a plus, thus breaking the link as the
+              // space is lost.
+              // It's a known quirk with the URLEncoder as it encodes for forms, not general URIs.
+              // Only done for the WorkflowId as the other two are values we control,
+              // and will never have spaces.
+              URLEncoder.encode(we.getWorkflowId(), StandardCharsets.UTF_8.toString())
+                  .replace("+", "%20"),
               URLEncoder.encode(we.getRunId(), StandardCharsets.UTF_8.toString()));
```
maciejdudko, line 48: Non-blocking: I realize our run IDs don't contain spaces, but for consistency it'd be good to do the replace in both.

Evanthx: I get the consistency idea but hate to do extra string operations when they aren't needed!

PR #2886, opened 2026-05-22
Fixed flaky test WorkerFactoryRegistryTest.testRandomOrder

maciejdudko:
## What was changed
Rewrote `WorkerFactoryRegistryTest.testRandomOrder` in a way that does not depend on distribution of random numbers. The new implementation instead checks that every possible ordering is eventually chosen, which provides a similar guarantee without depending on exact random distribution.

## Why?
Previous implementation of this test explicitly checked that each factory has roughly equal probability of appearing first in the randomized list. The test sometimes failed because the distribution is not actually guaranteed by implementation.

## Checklist
1. How was this tested:
The new test passes.


cconstable (approved):
+1 to removing 3000 iteration loop

PR #2887, opened 2026-05-25
Add ChildWorkflowOptions support to WorkflowImplementationOptions (#2)

### LINKED ISSUE #2790: Extend WorkflowImplementationOptions with ChildWorkflowOptions
Currently, when building `WorkflowImplementationOptions` it is possible to set `ActivityOptions`, `LocalActivityOptions`, and `NexusServiceOptions`.
It would be great if there was a method to set `ChildWorkflowOptions` in the same way using the methods:
- `setChildWorkflowOptions(Map<String, ChildWorkflowOptions> options)`
- `setDefaultChildWorkflowOptions(ChildWorkflowOptions options)`

porunov:
Fixes #2790

Tests are added to: `ChildWorkflowOptionsInWorkflowImplementationOptionsTest.java`

- Add childWorkflowOptions map and defaultChildWorkflowOptions fields
- Add setChildWorkflowOptions() and setDefaultChildWorkflowOptions() builder methods
- Add getChildWorkflowOptions() and getDefaultChildWorkflowOptions() getters
- Update SyncWorkflowContext to store and expose child workflow options
- Update WorkflowInternal.newChildWorkflowStub() to merge predefined options
- Add mergeChildWorkflowOptions() method to ChildWorkflowOptions.Builder
- Add integration and unit tests

Potential documentation place: https://docs.temporal.io/develop/java/workflows/child-workflows (it could be something similar as [activity options documentation](https://docs.temporal.io/develop/java/activities/execution#activity-options-reference)).

Disclaimer: Code was Generated by Opus 4.6 in Copilot. Reviewed manually.

### DIFF
--- a/temporal-sdk/src/test/java/io/temporal/workflow/DefaultChildWorkflowOptionsSetOnWorkflowTest.java
+++ b/temporal-sdk/src/test/java/io/temporal/workflow/DefaultChildWorkflowOptionsSetOnWorkflowTest.java
@@ -1,0 +1,199 @@
+package io.temporal.workflow;
+
+import static org.junit.Assert.assertEquals;
+
+import io.temporal.client.WorkflowOptions;
+import io.temporal.testing.internal.SDKTestWorkflowRule;
+import io.temporal.worker.WorkflowImplementationOptions;
+import java.time.Duration;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import org.junit.Rule;
+import org.junit.Test;
+
+/**
+ * Verifies that the {@link ChildWorkflowOptions} configured on {@link
+ * WorkflowImplementationOptions} are actually applied to child workflows created through both typed
+ * and untyped stubs, and that the precedence between the per-type options ({@link
+ * WorkflowImplementationOptions.Builder#setChildWorkflowOptions(Map)}), the default options ({@link
+ * WorkflowImplementationOptions.Builder#setDefaultChildWorkflowOptions}) and the options passed to
+ * the stub creation method is correct.
+ *
+ * <p>Each scenario is verified by reading the child's memo from inside the child workflow, so a
+ * test only passes if the expected options object was the one that actually took effect.
+ */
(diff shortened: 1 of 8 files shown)


temporal-sdk/src/test/java/io/temporal/workflow/DefaultChildWorkflowOptionsSetOnWorkflowTest.java
```diff
@@ -0,0 +1,74 @@
...
+              WorkflowImplementationOptions.newBuilder()
+                  .setDefaultChildWorkflowOptions(defaultChildWorkflowOptions)
+                  .setChildWorkflowOptions(childWorkflowOptionsMap)
+                  .build(),
+              ParentWorkflowImpl.class,
+              ChildWorkflowImpl.class)
+          .build();
+
+  @Test
+  public void testDefaultChildWorkflowOptionsApplied() {
```
maciejdudko: This test does not actually assert that the right options were used. Please rewrite it so that the child workflow only succeeds if the default options were correctly applied. I think checking its memo would be most straight forward.

Please also add a test that verifies default, not per-type child workflow options work too.

maciejdudko (dismissed):
Hi @porunov, thank you for your contribution. The code looks good, but the tests need to be improved before we can merge this - see comment.

temporal-sdk/src/test/java/io/temporal/workflow/DefaultChildWorkflowOptionsSetOnWorkflowTest.java
porunov: Thank you @maciejdudko for pointing at it. There was a bug when the default options would overwrite per type options, but it's now fixed. Also added the tests as you requested. All tests are passing now. Let me know please if there are any other concerns with this PR.

porunov:
@maciejdudko I've rebased in from `main` branch. Would be great if you could check the PR again. Thank you!

Quinn-With-Two-Ns (approved):

PR #2889, opened 2026-05-26
Nexus Signal links

Evanthx:
Sending signals through Nexus now populates a link from the sender to the receiver as well as one from the receiver back to the sender.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


temporal-sdk/src/main/java/io/temporal/internal/nexus/InternalNexusOperationContext.java
```diff
@@ -15,6 +18,21 @@ public class InternalNexusOperationContext {
   private final WorkflowClient client;
   NexusOperationOutboundCallsInterceptor outboundCalls;
   Link startWorkflowResponseLink;
+  // Links extracted from the inbound Nexus task. Stored once at the task-handler boundary so the
+  // workflow client (signal, signalWithStart) can attach them to outgoing requests via
+  // SignalWorkflowExecutionRequest.links, matching the Go SDK's NexusOperationLinksKey ctx value.
```
VegetarianOrc: IMO don't mention the Go SDK here
```suggestion
  // SignalWorkflowExecutionRequest.links.
```

temporal-sdk/src/main/java/io/temporal/internal/nexus/InternalNexusOperationContext.java
```diff
@@ -68,6 +86,35 @@ public Link getStartWorkflowResponseLink() {
     return startWorkflowResponseLink;
   }
 
+  /**
+   * Set the {@code common.v1.Link}s extracted from the inbound Nexus task so they can be attached
+   * to any signal RPCs issued by the operation handler.
+   */
```
VegetarianOrc, line 88: This is for signal RPCs currently, but will be other RPCs in the near-ish™️ future right?

temporal-sdk/src/main/java/io/temporal/internal/nexus/InternalNexusOperationContext.java
```diff
@@ -68,6 +86,35 @@ public Link getStartWorkflowResponseLink() {
...
+  /**
+   * Append a backlink returned by a signal-class RPC (signal or signalWithStart). Each signal the
+   * operation handler issues should add one entry; the task handler drains the list when building
+   * the operation's StartOperationResponse.
+   */
+  public void addSignalWorkflowResponseLink(Link link) {
+    if (link != null) {
+      this.signalWorkflowResponseLinks.add(link);
+    }
+  }
```
VegetarianOrc: I'm wondering if this is too operation specific. As more operations include links, we'll keep adding new getters/setters here and then have to ensure they're all consumed when building the StartOperationResponse. A perhaps contrived example would be a nexus operation that signals a workflow, updates a workflow, then starts a workflow would have backlinks for all of those if my understanding is correct.

temporal-sdk/src/main/java/io/temporal/internal/nexus/NexusTaskHandlerImpl.java
```diff
@@ -296,7 +301,23 @@ private StartOperationResponse handleStartOperation(
...
+                  LinkConverter.nexusLinkToWorkflowEvent(link);
+              if (commonLink != null) {
+                inboundCommonLinks.add(commonLink);
+              } else {
+                log.warn(
+                    "Dropping inbound Nexus link from outbound signal propagation: type='{}',"
+                        + " url='{}' (not a parseable temporal WorkflowEvent link)",
+                    link.getType(),
+                    link.getUrl());
+              }
```
VegetarianOrc, line 318: This seems incorrect to me. A SANO will have a link to a Nexus operation populated and it should be sent to signal (and other) RPCs. I think we'll need to include a more generic helper on the LinkConverter that converts a Nexus link based on the type field to the appropriate Temporal proto link type.

temporal-sdk/src/main/java/io/temporal/internal/nexus/NexusTaskHandlerImpl.java
```diff
@@ -307,10 +328,27 @@ private StartOperationResponse handleStartOperation(
...
+            CurrentNexusOperationContext.get().getSignalWorkflowResponseLinks()) {
+          if (!signalResponseLink.hasWorkflowEvent()) {
+            continue;
+          }
+          io.temporal.api.nexus.v1.Link converted =
+              LinkConverter.workflowEventToNexusLink(signalResponseLink.getWorkflowEvent());
+          if (converted != null) {
+            signalBacklinks.add(converted);
+          }
+        }
```
VegetarianOrc: Similar to my consideration of RPC specific accessors above, we'd need to check every link accessor do processing here. I think it'd be cleaner here as well to avoid details of the specific RPC that the handler may have invoked.

temporal-sdk/src/main/java/io/temporal/internal/nexus/InternalNexusOperationContext.java
Quinn-With-Two-Ns: yeah I would tend to agree, I don't think the context should care what operation decided to attach a link

temporal-sdk/src/main/java/io/temporal/internal/nexus/InternalNexusOperationContext.java
Evanthx: I don't know! :-) So fixed the language so that we don't have to worry about it

temporal-sdk/src/main/java/io/temporal/internal/nexus/InternalNexusOperationContext.java
Evanthx: Renamed to backlinks

temporal-sdk/src/main/java/io/temporal/internal/nexus/NexusTaskHandlerImpl.java
Evanthx: Generic helper added

temporal-sdk/src/main/java/io/temporal/internal/nexus/NexusTaskHandlerImpl.java
Evanthx: Agreed

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
```diff
@@ -120,6 +120,19 @@ public WorkflowSignalOutput signal(WorkflowSignalInput input) {
...
+    // Nexus task links so the SignalWorkflowExecution history event links back to the caller.
+    boolean inNexusContext = CurrentNexusOperationContext.isNexusContext();
+    if (inNexusContext) {
+      request.addAllLinks(CurrentNexusOperationContext.get().getNexusOperationLinks());
+    } else {
+      // Most signal calls (from a regular client or workflow) won't be in a Nexus context — this
+      // is normal. The log helps a debugger when a Nexus operation handler "mysteriously" lacks
+      // link propagation because it spawned a thread to issue the signal (the thread-local
+      // CurrentNexusOperationContext is invisible from that thread).
+      log.debug("signal RPC issued outside a Nexus operation context; no link propagation");
```
Quinn-With-Two-Ns: I don't think we should add a debug log for 99% of user using signal.

temporal-sdk/src/main/java/io/temporal/internal/common/LinkConverter.java
```diff
@@ -160,6 +169,142 @@ public static Link nexusLinkToWorkflowEvent(io.temporal.api.nexus.v1.Link nexusL
     return link.build();
   }
 
+  /**
+   * Encode a {@link Link.NexusOperation} (a link to a standalone Nexus operation) into the (url,
+   * type) form used on the Nexus wire. URL format matches the canonical server implementation:
+   * {@code temporal:///namespaces/{ns}/nexus-operations/{op_id}/{run_id}/details}.
+   */
+  public static io.temporal.api.nexus.v1.Link nexusOperationToNexusLink(Link.NexusOperation no) {
```
Quinn-With-Two-Ns: I don't think this belongs in this PR, this probably belongs in the SANO PR, BUT the server doesn't support SANO links right now so the SDK can't send them yet.

temporal-sdk/src/main/java/io/temporal/internal/nexus/InternalNexusOperationContext.java
```diff
@@ -15,6 +18,23 @@ public class InternalNexusOperationContext {
   private final WorkflowClient client;
   NexusOperationOutboundCallsInterceptor outboundCalls;
   Link startWorkflowResponseLink;
```
Quinn-With-Two-Ns, line 25: Is `startWorkflowResponseLink` still relevant now that we have `responseBacklinks`

temporal-sdk/src/main/java/io/temporal/internal/nexus/InternalNexusOperationContext.java
```diff
@@ -27,6 +47,18 @@ public InternalNexusOperationContext(
     this.endpoint = endpoint;
     this.metricScope = metricScope;
     this.client = client;
+    this.ownerThread = Thread.currentThread();
+  }
+
+  private void assertOwnerThread() {
```
Quinn-With-Two-Ns: IMO since `InternalNexusOperationContext` is propagated as a threadlocal this is not a serious concern that jsutifilties this unique guard

temporal-sdk/src/main/java/io/temporal/internal/nexus/NexusTaskHandlerImpl.java
```diff
@@ -296,42 +301,67 @@ private StartOperationResponse handleStartOperation(
                     "Invalid link URL: " + link.getUrl(),
                     e);
               }
+              // Convert each inbound nexus.v1.Link to common.v1.Link, dispatching on the link's
+              // type field (WorkflowEvent, NexusOperation, etc.). LinkConverter logs the warn for
+              // any unknown type and returns null.
+              io.temporal.api.common.v1.Link commonLink = LinkConverter.nexusLinkToCommonLink(link);
```
Quinn-With-Two-Ns: Why are these links treated differently then `operationStartDetails.addLink(nexusProtoLinkToLink(link));`? It seems like if there is parsing error the above will fail the operation but here we will just log? Why is there a difference?

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
Evanthx: Ah, good point. I was thinking debug logs would only happen in debug mode - I'm now guessing that is not the case!

temporal-sdk/src/main/java/io/temporal/internal/nexus/NexusTaskHandlerImpl.java
```diff
@@ -296,42 +301,67 @@ private StartOperationResponse handleStartOperation(
...
+          // The SDK stashed this backlink itself in RootWorkflowClientInvoker; failing to re-encode
+          // it now means a LinkConverter regression or a malformed link from the server. Either is
+          // an SDK invariant violation worth shouting about (warn is too quiet — the caller's
+          // history event will be missing a link with no other diagnostic).
+          log.error(
+              "SDK-stashed backlink failed to re-encode as nexus.v1.Link; caller history will be"
+                  + " missing a link. backlink={}",
+              backlink);
         }
-      } catch (OperationException e) {
```
Quinn-With-Two-Ns, line 331: Why did we need to move the exception handling? Quickly, It looks like this will change how we handle exceptions before `startOperation`

temporal-sdk/src/test/java/io/temporal/workflow/nexus/SignalOperationLinkingTest.java
```diff
@@ -0,0 +1,516 @@
...
+
+  // Caller workflow + Nexus handler register here (namespace UnitTest).
+  @Rule
+  public SDKTestWorkflowRule testWorkflowRule =
+      SDKTestWorkflowRule.newBuilder()
+          .setWorkflowTypes(SignalCallerWorkflow.class, SignalCalleeWorkflowImpl.class)
+          .setNexusServiceImplementation(new SignalingNexusServiceImpl())
+          .build();
+
+  // Separate worker/client on the callee namespace, used by the cross-namespace test. No
```
Quinn-With-Two-Ns: I don't think we need a cross-namespace test here . It is not exercising any unique behaviour in the SDK. There is a reason no other test has needed to do it.

temporal-sdk/src/test/java/io/temporal/workflow/nexus/SignalOperationLinkingTest.java
```diff
@@ -0,0 +1,516 @@
...
+ *   <li>{@link #testAsyncSignalOperationLinks()} — handler returns an async result after signaling;
+ *       verifies the backlink lands on {@code NexusOperationStarted} (the async response path in
+ *       {@link io.temporal.internal.nexus.NexusTaskHandlerImpl}).
+ * </ul>
+ *
+ * <p>All four tests require Temporal server &ge; 1.31 with {@code EnableCHASMSignalBacklinks=true};
+ * the in-memory test server does not implement this path so the class is skipped unless a real
+ * server is in use.
+ */
+public class SignalOperationLinkingTest extends BaseNexusTest {
```
Quinn-With-Two-Ns:
```suggestion
public class SignalOperationLinkingTest {
```
Shouldn't be necessary? `BaseNexusTest` is only needed in a few special test cases, and even there ideally we would remove it

temporal-sdk/src/test/java/io/temporal/internal/common/InternalUtilsTest.java
```diff
@@ -0,0 +1,103 @@
...
+            NAMESPACE, "tq", "endpoint", metricsScope, mock(WorkflowClient.class)));
+  }
+
+  @After
+  public void tearDown() {
+    CurrentNexusOperationContext.unset();
+  }
+
+  /**
+   * Regression guard for the {@code LinkConverter} generalization: inbound nexus links of every
```
Quinn-With-Two-Ns: As I said above, this isn't really a regression, and I think is out of scope of this PR

temporal-sdk/src/main/java/io/temporal/internal/common/LinkConverter.java
Quinn-With-Two-Ns: You may not have detected the lack of server support because this change doesn't exercise SANO, another reason this change doesn't belong in this PR

temporal-sdk/src/main/java/io/temporal/internal/nexus/InternalNexusOperationContext.java
Evanthx: Good point - removed and some cleanup done around it

temporal-sdk/src/main/java/io/temporal/internal/nexus/InternalNexusOperationContext.java
Evanthx: Removed

temporal-sdk/src/main/java/io/temporal/internal/nexus/NexusTaskHandlerImpl.java
Evanthx: operationStartDetails.addLink is right above this, so if it fails then this code is never reached. This would only fail if it was given a link type we don't know about, and I was worried that throwing on that would break future compatibility.

temporal-sdk/src/main/java/io/temporal/internal/nexus/NexusTaskHandlerImpl.java
Evanthx: Honestly I read the code and saw try { try { catch() catch() and assumed that the double try was an AI generated code error so I fixed it. I am pretty sure behavior was kept - but also ok, fair, it is adding an unrelated change to the PR so I restored it.

temporal-sdk/src/test/java/io/temporal/workflow/nexus/SignalOperationLinkingTest.java
Evanthx: The test sets up an endpoint to send the link back and forth, so that's why it's extending this

temporal-sdk/src/test/java/io/temporal/internal/common/InternalUtilsTest.java
Evanthx: I can remove it, but I did want to test behavior here - regression guard is probably not the best wording though. I can change the wording but figured I'd double check what you were wanting before i started changing things!

temporal-sdk/src/test/java/io/temporal/workflow/nexus/SignalOperationLinkingTest.java
Evanthx: I figured cross-namespace would have some differences even if not much so was worried about not having any testing there! Removed

temporal-sdk/src/main/java/io/temporal/internal/common/LinkConverter.java
Evanthx: Removed

temporal-sdk/src/test/java/io/temporal/internal/common/InternalUtilsTest.java
Evanthx: Removed as part of another PR comment! So that was handy

temporal-sdk/src/test/java/io/temporal/workflow/nexus/SignalOperationLinkingTest.java
Evanthx: Removed - thank you for the pointer!

temporal-sdk/src/main/java/io/temporal/nexus/WorkflowRunOperationImpl.java
```diff
@@ -40,38 +33,9 @@ public OperationStartResult<R> start(
 
     NexusStartWorkflowResponse nexusStartWorkflowResponse =
         handle.getInvoker().invoke(nexusRequest);
-    WorkflowExecution workflowExec = nexusStartWorkflowResponse.getWorkflowExecution();
 
-    // If the start workflow response returned a link use it, otherwise
```
Quinn-With-Two-Ns: Where did all this logic move?

temporal-sdk/src/main/java/io/temporal/internal/nexus/InternalNexusOperationContext.java
```diff
@@ -14,7 +17,23 @@ public class InternalNexusOperationContext {
...
+  // StartOperationResponse so each RPC the handler issued gets a corresponding link on the caller
+  // workflow's history event.
+  //
+  // This context is only safe for use from the single thread that runs the operation handler (the
+  // Nexus task executor's thread). The mutators below assert this contract; a stray cross-thread
+  // call fails fast rather than silently corrupting the ArrayList.
+  private final List<Link> responseBacklinks = new ArrayList<>();
+  // Captured at construction (on the Nexus task executor's thread) and used to fail fast on any
+  // cross-thread mutation. See note on responseBacklinks.
+  private final Thread ownerThread;
```
Quinn-With-Two-Ns: Don't think this is needed anymore?

temporal-sdk/src/main/java/io/temporal/internal/nexus/NexusTaskHandlerImpl.java
```diff
@@ -284,6 +285,10 @@ private StartOperationResponse handleStartOperation(
             .setCallbackUrl(task.getCallback())
             .setRequestId(task.getRequestId());
     task.getCallbackHeaderMap().forEach(operationStartDetails::putCallbackHeader);
+    // Stash the inbound links in common.v1.Link form on the operation context so that signal
```
Quinn-With-Two-Ns: Nit: We are stashing the links so that `RPC's like signal` can use them. More future proof IMO

temporal-sdk/src/main/java/io/temporal/internal/nexus/NexusTaskHandlerImpl.java
```diff
@@ -296,7 +301,23 @@ private StartOperationResponse handleStartOperation(
...
+              // LinkConverter only returns a WorkflowEvent-shaped common.v1.Link; nexus links of
+              // other shapes (e.g. non-temporal URLs) come back null and are intentionally not
+              // forwarded onto SignalWorkflowExecutionRequest.links, which requires the
+              // WorkflowEvent variant. Log so a debugging session can see what was dropped.
+              io.temporal.api.common.v1.Link commonLink =
+                  LinkConverter.nexusLinkToWorkflowEvent(link);
+              if (commonLink != null) {
+                inboundCommonLinks.add(commonLink);
+              } else {
+                log.warn(
```
Quinn-With-Two-Ns, line 313: Nit: I would remove "signal propagation" here to account for more use cases

temporal-sdk/src/main/java/io/temporal/internal/nexus/NexusTaskHandlerImpl.java
```diff
@@ -349,6 +388,8 @@ private StartOperationResponse handleStartOperation(
             new RuntimeException("Unknown operation state: " + e.getState()));
       }
       startResponseBuilder.setFailure(dataConverter.exceptionToFailure(temporalFailure));
+    } catch (Throwable failure) {
+      convertKnownFailures(failure);
```
Quinn-With-Two-Ns: Why did we need to add this? Not saying it is wrong, just curious?

temporal-sdk/src/main/java/io/temporal/internal/nexus/InternalNexusOperationContext.java
```diff
@@ -60,12 +80,36 @@ public NexusOperationContext getUserFacingContext() {
...
+   * StartOperationResponse.
+   */
+  public void addBacklink(Link link) {
+    if (link != null) {
+      this.responseBacklinks.add(link);
+    }
+  }
+
+  /**
+   * Backlinks from every outbound RPC the handler issued. Never null; may be empty. Returned as an
```
Quinn-With-Two-Ns:
```suggestion
   * Backlinks from every outbound RPC the handler issued. Returned as an
```
Annotate with `@Nonnull`

temporal-sdk/src/main/java/io/temporal/internal/nexus/InternalNexusOperationContext.java
```diff
@@ -60,12 +80,36 @@ public NexusOperationContext getUserFacingContext() {
...
+   * to RPCs issued by the operation handler.
+   */
+  public void setNexusOperationLinks(List<Link> links) {
+    this.nexusOperationLinks = links == null ? Collections.emptyList() : links;
   }
 
-  public Link getStartWorkflowResponseLink() {
-    return startWorkflowResponseLink;
+  /** Links from the inbound Nexus task; empty if none. Never null. */
+  public List<Link> getNexusOperationLinks() {
```
Quinn-With-Two-Ns: Same for https://github.com/temporalio/sdk-java/pull/2889/changes#r3357302568

temporal-sdk/src/main/java/io/temporal/internal/nexus/InternalNexusOperationContext.java
```diff
@@ -60,12 +80,36 @@ public NexusOperationContext getUserFacingContext() {
...
+    return nexusOperationLinks;
```
Quinn-With-Two-Ns: Why is `getBacklinks` wrapped in `unmodifiableList` and this one not? Are they used differently?

temporal-sdk/src/main/java/io/temporal/internal/nexus/NexusTaskHandlerImpl.java
```diff
@@ -307,10 +328,27 @@ private StartOperationResponse handleStartOperation(
       try {
         OperationStartResult<HandlerResultContent> result =
             startOperation(context, operationStartDetails.build(), input.build());
+        // If signal or signalWithStart RPCs the handler issued returned backlinks, propagate them
```
Quinn-With-Two-Ns: I think this comment is wrong, `getBacklinks` is also set by start workflow, I left this comment a few other places, but I would review all all the code comments and make sure they aren't being specific to signal since the code isn't now and will only get more general as we add more Nexus features.

temporal-sdk/src/main/java/io/temporal/nexus/WorkflowRunOperationImpl.java
Evanthx: These changes happened around when we talked about moving the nexus specific backlink logic into a generic backlink logic.

So the logic to capture the start-workflow link from the response is now in RootWorkflowClientInvoker.start - it see the addBackLink call. And attaching the link to the OperationStartResult is in NexusTaskHandlerImpl.

Seemed to work ok, and made everything generic.

temporal-sdk/src/main/java/io/temporal/internal/nexus/InternalNexusOperationContext.java
Evanthx: Removed!

temporal-sdk/src/main/java/io/temporal/internal/nexus/NexusTaskHandlerImpl.java
Evanthx: Good call!

temporal-sdk/src/main/java/io/temporal/internal/nexus/NexusTaskHandlerImpl.java
Evanthx: It was left from when I consolidated the double try block and then reverted it - I missed this one! I'd focussed on making sure the loop was restored and no changes showed but missed this bit after the loop

temporal-sdk/src/main/java/io/temporal/internal/nexus/NexusTaskHandlerImpl.java
Evanthx: Changed

temporal-sdk/src/main/java/io/temporal/internal/nexus/InternalNexusOperationContext.java
Evanthx: Done

temporal-sdk/src/main/java/io/temporal/internal/nexus/InternalNexusOperationContext.java
Evanthx: Added it!

temporal-sdk/src/main/java/io/temporal/nexus/WorkflowRunOperationImpl.java
Evanthx: The server always returned a link so I didn't have the code to create one if it wasn't present - I added that back in though it is now in RootWorkflowClientInvoker

temporal-sdk/src/main/java/io/temporal/internal/nexus/NexusTaskHandlerImpl.java
Evanthx: Done

Quinn-With-Two-Ns (approved):
LGTM

maciejdudko (approved):

PR #2900, opened 2026-06-04
Add initial extstore types

cconstable:
## What was changed

Adds the `@Experimental` External Storage API surface under `io.temporal.payload.storage`. Type
  definitions only. 

  Types:

  - `StorageDriver`: driver contract (async `store`/`retrieve`, `getName`/`getType`).
  - `StorageDriverClaim`: claim-check token for an externally stored payload.
  - `StorageDriverSelector`: chooses a driver per payload, or keeps it inline.
  - `StorageDriverStoreContext` / `StorageDriverRetrieveContext`: driver-call contexts.
  - `StorageDriverTargetInfo` plus `StorageDriverWorkflowInfo` / `StorageDriverActivityInfo`: best-effort,
  store-side target identity.
  - `ExternalStorage` (plus `Builder`): config for drivers, optional selector, and size threshold.

  Notes:

  - No `WorkflowClientOptions` setter yet.
  - Builder validation throws `IllegalStateException`. Stable diagnostic codes are a follow-up.

## Why?

- First part of part of Large Payloads -> External Storage work

## Checklist

- Added tests.


temporal-sdk/src/main/java/io/temporal/payload/storage/ExternalStorage.java
```diff
@@ -0,0 +1,95 @@
...
+import java.util.List;
+import java.util.Objects;
+import java.util.Set;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+/** Configuration for offloading large payloads to external storage. */
+@Experimental
+public final class ExternalStorage {
+  private static final int DEFAULT_PAYLOAD_SIZE_THRESHOLD = 256 * 1024;
```
cconstable: I kept this private to be consistent with TS. However, I had to remove `DEFAULT_PAYLOAD_SIZE_THRESHOLD` links from the doc comments so they would actually resolved (and not be broken). Seems reasonable to let users knows what the default is in the doc comments so I added a "Defaults to 256 KiB" but this feels a little worse to me than just making this public. Looking for feedback.

temporal-sdk/src/main/java/io/temporal/payload/storage/ExternalStorage.java
```diff
@@ -0,0 +1,95 @@
...
+      return this;
+    }
+
+    /** Defaults to 256 KiB. Set to {@code null} to store all payloads. */
+    public Builder setPayloadSizeThreshold(@Nullable Integer payloadSizeThreshold) {
+      this.payloadSizeThreshold = payloadSizeThreshold;
+      return this;
+    }
+
+    public ExternalStorage build() {
```
cconstable: Right now, these all throw state exceptions (which are tested). In a subsequent PR, I would like to replace these with stable diagnostic codes.

temporal-sdk/src/main/java/io/temporal/payload/storage/ExternalStorage.java
jmaeagle99: Could we make this package-private (so just remove the `private` keyword) and then generate the docs via javadoc with `-package`? Haven't tried it myself and not sure what else might become docuemented.

temporal-sdk/src/main/java/io/temporal/payload/storage/StorageDriverRetrieveContext.java
```diff
@@ -0,0 +1,7 @@
+package io.temporal.payload.storage;
+
+import io.temporal.common.Experimental;
+
+/** Context passed to {@link StorageDriver#retrieve}. */
+@Experimental
+public final class StorageDriverRetrieveContext {}
```
jmaeagle99: Thinking about if making this and `StorageDriverStoreContext` interfaces instead of finalized classes. Might make it easier for us internally define the implementations and thus refactor/update them as necessary. In either case, there is burden on driver authors who write unit tests for their drivers to have to adapt to either changing APIs on classes or interfaces, but interfaces would allow them to implement however they deem necessary to fulfill the context contract.

temporal-sdk/src/main/java/io/temporal/payload/storage/ExternalStorage.java
```diff
@@ -0,0 +1,105 @@
...
+
+  public static final class Builder {
+    private List<StorageDriver> drivers = Collections.emptyList();
+    private StorageDriverSelector driverSelector;
+    private int payloadSizeThreshold = DEFAULT_PAYLOAD_SIZE_THRESHOLD;
+
+    private Builder() {}
+
+    /** At least one driver is required. When more than one is set, a selector is also required. */
+    public Builder setDrivers(@Nonnull List<StorageDriver> drivers) {
```
jmaeagle99: Should we document that when calling `setDrivers` or `setDriver` that the last call wins? Not sure if that is clear, especially in the case of calling `setDriver` multiple times.

temporal-sdk/src/test/java/io/temporal/payload/storage/ExternalStorageTest.java
```diff
@@ -0,0 +1,98 @@
...
+    ExternalStorage.newBuilder().build();
+  }
+
+  @Test(expected = IllegalStateException.class)
+  public void duplicateDriverNamesRejected() {
+    ExternalStorage.newBuilder().setDrivers(Arrays.asList(driver("dup"), driver("dup"))).build();
+  }
+
+  @Test(expected = IllegalStateException.class)
+  public void multipleDriversRequireSelector() {
```
jmaeagle99: Maybe some more tests validating that calling `setDriver` and `setDrivers` in various orders only allows the last call to win.

temporal-sdk/src/main/java/io/temporal/payload/storage/ExternalStorage.java
jmaeagle99: For these in particular, do we need diagnostic codes? The messages are fairly clear as to what the issue is and prevent the use of external storage at startup rather than having to report during execution runtime.

jmaeagle99 (approved):
Left some comments but overall looks good. Please hold off until a Java owner reviews and approves.

temporal-sdk/src/main/java/io/temporal/payload/storage/ExternalStorage.java
cconstable: I am a proponent of best-effort, stable diagnostic codes if the mechanism for adding them very easy. I have a little write-up about this that I wanted to share but the tl;dr is I think we should make a best effort to tag anything that's actionable if for nothing else it supports automated tooling (e.g. AI) and feels like a healthy industry trend (*hand wavy gesture*). I would not have felt this way 4 years ago.

temporal-sdk/src/main/java/io/temporal/payload/storage/ExternalStorage.java
cconstable: That's a good builder pattern call out 😄 I'll update.

temporal-sdk/src/main/java/io/temporal/payload/storage/ExternalStorage.java
cconstable: I think promoting to package is reasonable. One consideration for making it public is being able to reference it in generated docs e.g. on the setThreshold method "The default value is {@link ExternalStorage.DEFAULT_PAYLOAD_SIZE_THRESHOLD" instead of what I have now: "Defaults to 256 KiB".

jmaeagle99: I think you can use `@value` to embed the actual value. I still don't see the value that having an effective public constant has for customers. Sure, it's great they know what the default threshold is, but they aren't going to write code specifically with that in mind. So we shouldn't provide it in code. Documentation, absolutely.

temporal-sdk/src/main/java/io/temporal/payload/storage/ExternalStorage.java
```diff
@@ -0,0 +1,108 @@
+package io.temporal.payload.storage;
```
cconstable: Should the package be located here? or should we consider something like `temporal.common.extstore`? Here in `temporal.payload` it's next to the codec which seems like the right place.

maciejdudko: The current `io.temporal.payload.storage` makes sense.

temporal-sdk/src/main/java/io/temporal/payload/storage/ExternalStorage.java
```diff
@@ -0,0 +1,108 @@
...
+import java.util.HashSet;
+import java.util.List;
+import java.util.Objects;
+import java.util.Set;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+/** Configuration for offloading large payloads to external storage. */
+@Experimental
+public final class ExternalStorage {
```
maciejdudko: Rename to `ExternalStorageOptions` for consistency with other similar classes.

maciejdudko (approved):

temporal-sdk/src/main/java/io/temporal/payload/storage/ExternalStorage.java
jmaeagle99: Why rename to *Options? These don't carry options but the user executable pieces of external storage.

PR #2901, opened 2026-06-04
AWS Lambda (Java)

eamsden:
## What was changed
Add contrib support for running Java workers on AWS Lambda

## Why?
Towards SDK parity for AWS Lambda support for serverless workers

## Checklist

1. Closes (NA)

2. How was this tested:
Full E2E deploy and workflow run of a Lambda worker, with and without OTel tracing and metrics

3. Any docs updates needed?
- Update serverless docs for Java SDK support
- Flag that tracing format will change when we replace OpenTracing shim with native OpenTelemetry support

### DIFF
--- a/contrib/temporal-aws-lambda/README.md
+++ b/contrib/temporal-aws-lambda/README.md
@@ -14,13 +14,13 @@ import io.temporal.common.WorkerDeploymentVersion;
 
 public final class Handler implements RequestHandler<Object, Void> {
   private static final RequestHandler<Object, Void> WORKER =
-      LambdaWorker.run(
-          new WorkerDeploymentVersion("orders-worker", "2026-06-02"),
-          options ->
-              options
-                  .setTaskQueue("orders")
-                  .registerWorkflowImplementationTypes(OrderWorkflowImpl.class)
-                  .registerActivitiesImplementations(new OrderActivitiesImpl()));
+	      LambdaWorker.run(
+	          new WorkerDeploymentVersion("orders-worker", "2026-06-02"),
+	          builder ->
+	              builder
+	                  .setTaskQueue("orders")
+	                  .registerWorkflowImplementationTypes(OrderWorkflowImpl.class)
+	                  .registerActivitiesImplementations(new OrderActivitiesImpl()));
 
   @Override
   public Void handleRequest(Object input, Context context) {
@@ -33,7 +33,7 @@ public final class Handler implements RequestHandler<Object, Void> {
 
 Connection options are loaded with `temporal-envconfig` when the handler is constructed during Lambda cold start. The Lambda worker checks `TEMPORAL_CONFIG_FILE` first, then readable `$LAMBDA_TASK_ROOT/temporal.toml`, then readable `./temporal.toml`, then falls back to the envconfig defaults and Temporal environment variables. The `configure` callback also runs during handler construction, so non-invocation configuration is prepared once and reused.
 
-If you need to assemble options outside the `run` callback, call `LambdaWorkerOptions.fromEnvironment()`, mutate the returned options, and pass them to `LambdaWorker.newHandler(...)`.
+If you need to assemble options outside the `run` callback, call `LambdaWorkerOptions.newBuilderFromEnvironment()`, configure the returned builder, call `build()`, and pass the options to `LambdaWorker.newHandler(...)`.
 
 Dynamic workflow and activity implementations can be registered with `registerDynamicWorkflowImplementationType(...)` and `registerDynamicActivityImplementation(...)`. Java SDK worker rules still apply: only one dynamic workflow implementation type and one dynamic activity implementation can be registered per worker.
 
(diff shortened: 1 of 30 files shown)


contrib/temporal-aws-lambda/README.md
```diff
@@ -0,0 +1,80 @@
...
+  @Override
+  public Void handleRequest(Object input, Context context) {
+    return WORKER.handleRequest(input, context);
+  }
+}
+```
+
+`TEMPORAL_TASK_QUEUE` can provide the task queue. If it is not set, call `setTaskQueue`.
+
+Connection options are loaded with `temporal-envconfig` when the handler is constructed during Lambda cold start. The Lambda worker checks `TEMPORAL_CONFIG_FILE` first, then readable `$LAMBDA_TASK_ROOT/temporal.toml`, then readable `./temporal.toml`, then falls back to the envconfig defaults and Temporal environment variables. The `configure` callback also runs during handler construction, so non-invocation configuration is prepared once and reused.
```
Sushisource:
```suggestion
Connection options are loaded with `temporal-envconfig` when the handler is constructed during Lambda cold start. The Lambda worker checks `TEMPORAL_CONFIG_FILE` first, then `$LAMBDA_TASK_ROOT/temporal.toml`, then `./temporal.toml`, then falls back to the envconfig defaults and Temporal environment variables. The `configure` callback also runs during handler construction, so non-invocation configuration is prepared once and reused.
```

I found "readable" is just noise here.

Also we mention the configure callback here, but it's not immediately clear that that's referring to the lambda that's the second parameter to `run`

contrib/temporal-aws-lambda/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+Use the per-invocation configuration overload when final options depend on the Lambda `Context` or on resources opened for one invocation. The cold-start callback still runs once; the invocation callback runs before Temporal service stubs, client, and worker are created:
```
Sushisource: Same here, would be good to clarify which is which

contrib/temporal-aws-lambda/src/main/java/io/temporal/aws/lambda/OtelLambdaWorker.java
```diff
@@ -0,0 +1,343 @@
...
+import io.temporal.opentracing.OpenTracingWorkerInterceptor;
+import io.temporal.worker.WorkerFactoryOptions;
+import java.time.Duration;
+import java.util.Arrays;
+import java.util.Map;
+import java.util.Objects;
+import java.util.function.Consumer;
+
+/** OpenTelemetry helper for {@link LambdaWorker}. */
+public final class OtelLambdaWorker {
```
Sushisource: This isn't really a `Worker` itself. More like a `ConfigurationHelper` or something? Open to a better name but I think the `Worker` postfix isn't quite right.

contrib/temporal-aws-lambda/src/main/java/io/temporal/aws/lambda/LambdaWorker.java
```diff
@@ -0,0 +1,527 @@
...
+
+  /**
+   * Returns an AWS Lambda Java handler that creates, starts, and shuts down one Temporal worker per
+   * invocation.
+   *
+   * @param version worker deployment version to advertise for this worker.
+   * @param configure callback invoked once while the Lambda handler is constructed.
+   */
+  public static RequestHandler<Object, Void> run(
+      WorkerDeploymentVersion version, Consumer<LambdaWorkerOptions.Builder> configure) {
```
Sushisource: Since `configure` is required non-null we should use the `@Nonnull` (or `@Notnull`) I forget which is the right one for IDEs (applies everywhere)

contrib/temporal-aws-lambda/src/main/java/io/temporal/aws/lambda/LambdaWorker.java
```diff
@@ -0,0 +1,527 @@
...
+    void sleep(Duration duration) throws InterruptedException;
+  }
+
+  interface NanoClock {
+    long nanoTime();
+  }
+
+  private static NanoClock systemNanoClock() {
+    return System::nanoTime;
+  }
```
Sushisource, line 186: Do we need all our own stuff here for this? I know there's https://docs.oracle.com/javase/8/docs/api/java/time/Clock.html

contrib/temporal-aws-lambda/src/main/java/io/temporal/aws/lambda/LambdaWorkerOptions.java
```diff
@@ -0,0 +1,673 @@
...
+      this.taskQueue = taskQueue;
+      this.workerOptions = workerOptions;
+      this.gracefulShutdownTimeout = gracefulShutdownTimeout;
+      this.shutdownDeadlineBuffer = shutdownDeadlineBuffer;
+      this.registrations = Collections.unmodifiableList(registrations);
+      this.shutdownHooks = Collections.unmodifiableList(shutdownHooks);
+    }
+  }
+
+  static final class Prepared {
```
Sushisource: These two separate prepared/materialized classes feel duplicative when all that's changing is the identity. One could just extend / compose the other or something.

contrib/temporal-aws-lambda/src/main/java/io/temporal/aws/lambda/LambdaWorker.java
Sushisource: Saw this got renamed but still curious if there's anything off the shelf we could've used here (fine if not)

Sushisource (commented):
Ah - another thing that comes to mind here, I implemented the OTel support as a plugin in the other langs, we should probably do the same here (looks like Justin mentioned this in the .NET one too)

contrib/temporal-aws-lambda/src/main/java/io/temporal/aws/lambda/LambdaWorker.java
eamsden: I must have messed up the rebase somehow. This was supposed to have been dealt with. Let me check.

eamsden: Nevermind I forgot a note I wrote to myself, for NanoClock/MonotonicClock there's no narrower standard library interface than `LongSupplier` which is...any 0-argument function that returns `long`. So I think `Sleeper` can go in favor of `Consumer<Duration>` (very few things you can do with a void-valued function that takes a duration other than sleep, unless it's a setter or something), but `MonotonicClock` should stay. These interfaces exist to inject not-actually-sleeping methods for the Lambda-specific timers, solving a similar problem to workflow time-skipping.

contrib/temporal-aws-lambda/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+```java
+import com.amazonaws.services.lambda.runtime.Context;
+import com.amazonaws.services.lambda.runtime.RequestHandler;
+import io.temporal.aws.lambda.LambdaWorker;
+import io.temporal.common.WorkerDeploymentVersion;
+
+public final class Handler implements RequestHandler<Object, Void> {
+  private static final RequestHandler<Object, Void> WORKER =
+	      LambdaWorker.run(
```
Sushisource: I'm actually wondering if we should call this `define` instead of `run`.

In the other langs the run still looks like actually running, even if it is ultimately doing the same thing semantically. Here it's more obvious that the worker isn't really running, per se, but is rather set up to be run by the handler.

Sushisource (approved):
Overall I think this makes sense, maybe with one possible renaming suggestion here.

Plz get a Java owner approval before merging.

contrib/temporal-aws-lambda/README.md
eamsden: Yeah it's handing off a thunk, effectively, to be dispatched by the lambda runtime. `define` or `prepare` both make sense.

contrib/temporal-aws-lambda/src/main/java/io/temporal/aws/lambda/LambdaWorker.java
eamsden: Followup on this: `Sleeper` needs to be able to throw `InterruptedException` which `Consumer` can't do.

maciejdudko (approved):

Issue #2884, opened 2026-05-19
Allow payload visitors to visit memo payloads in aggregate - Java


jmaeagle99:
Parent Issue: https://github.com/temporalio/features/issues/782

PR #2902, opened 2026-06-05
Message and concurrent payload visitors

### LINKED ISSUE #2881: Enable concurrent payload visiting - Java
Parent Issue: https://github.com/temporalio/features/issues/772

jmaeagle99:
## What was changed

Adds an SDK-internal payload/message visitor for Temporal API proto messages (`io.temporal.internal.payload.visitor` in `temporal-sdk`).

- `PayloadVisitors.visit(message, options)` — invokes a callback on every `Payload` reachable in any API message; the callback returns replacements, so payloads can be transformed in place.
- `MessageVisitors.visit(message, options)` — invokes a callback on each payload-bearing message (no per-payload visiting), for per-message-type logic.
- Per-message context seam (`MessageVisitor`) that threads a scoped value down the tree — e.g. tag payloads with the enclosing command's info.
- Bounded concurrency: sequential by default; `concurrency > 1` requires a caller-supplied `Executor`.
- Handles all payload shapes, `google.protobuf.Any`, and cyclic messages (`Failure`).
- Visitor exceptions propagate unchanged to the caller.

The traversal code is generated at build time from the proto descriptors (new `payloadVisitorGenerator` source set + Gradle task; output under `build/generated`, not committed).

## Why?

Provides the shared traversal foundation the upcoming external storage feature needs (concurrently offload large payloads), and the planned payload-limits validator.

## Checklist

1. Closes #2881 #2884

2. How was this tested: Unit tests under `io.temporal.internal.payload.visitor` (`PayloadVisitorTest`, `MessageVisitorTest`) covering traversal/mutation across every field shape, `Any`, cyclic messages, scoped context, concurrency, and error propagation

3. Any docs updates needed? No, internal implementation.


temporal-sdk/src/main/java/io/temporal/internal/payload/visitor/MessageVisitor.java
```diff
@@ -0,0 +1,23 @@
+package io.temporal.internal.payload.visitor;
+
+import com.google.protobuf.MessageOrBuilder;
+
+/**
+ * Callback invoked when traversal enters a proto message. The returned value becomes the contextual
+ * value in scope for that message and everything within it, and is restored to the enclosing value
+ * once traversal leaves the message. The message is provided as a builder and may be inspected or
+ * mutated.
```
GregoryTravis, line 9: Do these also possibly run concurrently?

jmaeagle99: Message traversal is serial and non-concurrent.

temporal-sdk/src/main/java/io/temporal/internal/payload/visitor/PayloadVisitorOptions.java
```diff
@@ -0,0 +1,141 @@
...
+    private int concurrency = 1;
+    private Executor executor;
+
+    private Builder() {}
+
+    /** Required. The payload visitor. */
+    public Builder<C> setPayloadVisitor(@Nonnull PayloadVisitor<C> payloadVisitor) {
+      this.payloadVisitor = payloadVisitor;
+      return this;
+    }
```
cconstable: This is minor but it seems odd to have a required field (or something so core to the abstraction) be an optional setter on a builder. I would rather express requirements like this in the type system either as a required arg on the builder constructor itself or lifting this up to the PayloadVisitors.visit interface.

temporal-sdk/src/main/java/io/temporal/internal/payload/visitor/PayloadVisitorContext.java
```diff
@@ -0,0 +1,33 @@
...
+  @Nullable
+  public C getContext() {
+    return context;
+  }
+
+  /** The message that directly contains the payloads being visited. */
+  @Nonnull
+  public MessageOrBuilder getParent() {
+    return parent;
+  }
```
cconstable: Is this needed? Also this safe for multiple threads to be referencing at once?

cconstable: looks like these are not thread-safe https://protobuf.dev/reference/java/java-generated/

jmaeagle99: Removed this accessor method, which then obviated the need for the `PayloadVisitorContext`, which has also been removed and replaced with the context value.

temporal-sdk/src/main/java/io/temporal/internal/payload/visitor/PayloadVisitorOptions.java
jmaeagle99: Moved to be a required parameter of both builders

temporal-sdk/src/main/java/io/temporal/internal/payload/visitor/PayloadVisitor.java
```diff
@@ -0,0 +1,26 @@
...
+@FunctionalInterface
+interface PayloadVisitor<C> {
+  /**
+   * Visits a sequence of payloads and returns their replacements.
+   *
+   * @param context the contextual value in scope
+   * @param payloads the payloads found at this location
+   * @return the replacement payloads
+   */
+  List<Payload> visit(C context, List<Payload> payloads);
```
jmaeagle99:
```suggestion
  CompletableFuture<List<Payload>> visit(C context, List<Payload> payloads);
```
I think we need the visitor to return `CompletableFuture` so that the visitor doesn't have to join when calling external storage. Since external storage will be the only consumer of this at this time and is internal, I think it's fine to optimize the API for it.

FYI @cconstable

temporal-sdk/src/payloadVisitorGenerator/java/io/temporal/internal/payload/visitor/gen/ProtoClosure.java
```diff
@@ -0,0 +1,138 @@
...
+/**
+ * Shared proto-descriptor model for the build-time generators: the message closure reachable from a
+ * set of seed services, and which of those messages can transitively contain a {@code Payload}.
+ */
+final class ProtoClosure {
+
+  /** All non-map-entry messages in the closure, in discovery order. */
+  final List<Descriptor> allMessages;
+
+  private final Set<String> reaches;
```
GregoryTravis, line 27: This could use a descriptive comment.

temporal-sdk/src/payloadVisitorGenerator/java/io/temporal/internal/payload/visitor/gen/ProtoClosure.java
```diff
@@ -0,0 +1,138 @@
...
+          case MAP_ANY:
+            direct = true;
+            break;
+          case MESSAGE_SINGLE:
+          case MESSAGE_REPEATED:
+          case MAP_MESSAGE:
+            refs.add(plan.child);
+            break;
+          default:
+            break;
```
GregoryTravis, line 113: It might be good to fail on an unhandled type here in case the Kind set is ever changed.

GregoryTravis (approved):
Looks great, just a request for a field comment and a suggestion about the unlilely possibility of the Kind classification changing.

PR #2903, opened 2026-06-08
Add option to let activities heartbeat during worker shutdown

### LINKED ISSUE #2075: Add the ability to keep heartbeating while the worker is shutting down

The `WorkerFactory.shutdown` call starts killing the heartbeats like it says in the documentation with the `ActivityWorkerShutdownException` but what is the point of waiting N units of time with `WorkerFactory.awaitTermination` if right when the worker shutdown is called the heartbeats die and then the activities die off because the heartbeats are no longer going....

I want ability to keep heartbeating while the worker is shutting down so my activities can actually finish like `WorkerFactory.awaitTermination` promises me. 

For example: If I tell the `awaitTermination` call to wait 25 minutes my heartbeat calls should still be valid for 25 minutes and then the heartbeat call can start reporting the `ActivityWorkerShutdownException`

Call ActivityExecutionContext.useLocalManualCompletion() and then use the ManualActivityCompletionClient to heartbeat.

**Additional context**

See the Temporal Java SDK slack thread here: https://temporalio.slack.com/archives/CTT84KXK9/p1716488319315099

baekgyu-kim:
## What was changed

Added an experimental worker option, `WorkerOptions.Builder#setAllowActivityHeartbeatDuringShutdown(boolean)` (default `false`, preserving the existing behavior).

- When **enabled**, on a graceful worker shutdown the activity heartbeat executor is shut down **only after all outstanding activity tasks have finished executing**, so in-flight activities keep heartbeating through the existing mechanism — no separate heartbeat code path.
- When **disabled** (default), or whenever `shutdownNow` is used, the behavior is unchanged: the heartbeat executor is shut down first, so `ActivityExecutionContext.heartbeat()` throws `ActivityWorkerShutdownException`.
- Gated by `allowActivityHeartbeatDuringShutdown && !interruptTasks`, so non-graceful shutdown (`WorkerFactory.shutdownNow`) always behaves as if the option were disabled.

Touched: `WorkerOptions`, `SingleWorkerOptions`, `Worker`, `SyncActivityWorker`.

## Why?

- **Before:** once a graceful shutdown was requested, `heartbeat()` threw `ActivityWorkerShutdownException` for the rest of the `awaitTermination` grace period. An activity that wanted to run to completion during that window could not refresh its server-side heartbeat deadline → the server timed it out and retried it → duplicate executions, even though the worker deliberately gave the activity time to finish.
- **After (opt-in):** with the option enabled, in-flight activities keep heartbeating during the grace period, so each call refreshes the heartbeat deadline and the server does not prematurely time out / retry the activity.
- **Design:** reuses the existing heartbeat path by only reordering executor shutdown — no new heartbeat logic. The default stays `false`, so existing behavior is preserved, and `shutdownNow` is intentionally excluded. Originates from #2075.

Note: with the option enabled, activities are no longer notified of shutdown via `ActivityWorkerShutdownException`, so they are expected to complete within the termination grace period on their own. This is documented on the setter.

## Checklist

1. Closes #2075

2. How was this tested:

   New `HeartbeatDuringWorkerShutdownTest` (standalone activity + two-semaphore handshake; gated behind `SDKTestWorkflowRule.useExternalService`):
   - `testHeartbeatingActivityCompletesDuringShutdown` — option enabled + graceful `shutdown()`: the activity heartbeats and runs to completion.
   - `testHeartbeatingActivityFailsDuringShutdownNow` — `shutdownNow()`: the option is ignored, and the heartbeat fails the activity.

   `WorkerOptionsTest` — copy-builder round-trips the new option.

   ```
   # the integration tests require an external server
   USE_EXTERNAL_SERVICE=true TEMPORAL_SERVICE_ADDRESS=localhost:7233 \
     ./gradlew :temporal-sdk:test --tests "io.temporal.worker.shutdown.HeartbeatDuringWorkerShutdownTest"
   ./gradlew :temporal-sdk:test --tests "io.temporal.worker.WorkerOptionsTest"
   ./gradlew :temporal-sdk:spotlessJavaCheck
   ```

3. Any docs updates needed?

   - The behavior is documented in the `WorkerOptions.Builder#setAllowActivityHeartbeatDuringShutdown` Javadoc.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/worker/SyncActivityWorker.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/worker/SyncActivityWorker.java
@@ -26,6 +26,7 @@ public class SyncActivityWorker implements SuspendableWorker {
   private final ScheduledExecutorService heartbeatExecutor;
   private final ActivityTaskHandlerImpl taskHandler;
   private final ActivityWorker worker;
+  private final boolean allowActivityHeartbeatDuringShutdown;
 
   public SyncActivityWorker(
       WorkflowClient client,
@@ -38,6 +39,7 @@ public class SyncActivityWorker implements SuspendableWorker {
     this.identity = options.getIdentity();
     this.namespace = namespace;
     this.taskQueue = taskQueue;
+    this.allowActivityHeartbeatDuringShutdown = options.getAllowActivityHeartbeatDuringShutdown();
 
     this.heartbeatExecutor =
         Executors.newScheduledThreadPool(
@@ -89,16 +91,31 @@ public class SyncActivityWorker implements SuspendableWorker {
 
   @Override
   public CompletableFuture<Void> shutdown(ShutdownManager shutdownManager, boolean interruptTasks) {
-    return shutdownManager
-        // we want to shut down heartbeatExecutor before activity worker, so in-flight activities
-        // could get an ActivityWorkerShutdownException from their heartbeat
-        .shutdownExecutor(heartbeatExecutor, this + "#heartbeatExecutor", Duration.ofSeconds(5))
-        .thenCompose(r -> worker.shutdown(shutdownManager, interruptTasks))
-        .exceptionally(
-            e -> {
-              log.error("[BUG] Unexpected exception during shutdown", e);
-              return null;
-            });
+    CompletableFuture<Void> shutdownFuture;
+    if (allowActivityHeartbeatDuringShutdown && !interruptTasks) {
+      // we want to shut down heartbeatExecutor only after all outstanding activity tasks have
+      // finished executing, so in-flight activities can keep heartbeating during the shutdown
+      shutdownFuture =
+          worker
+              .shutdown(shutdownManager, interruptTasks)
+              .thenCompose(r -> shutdownHeartbeatExecutor(shutdownManager));
+    } else {
+      // we want to shut down heartbeatExecutor before activity worker, so in-flight activities
+      // could get an ActivityWorkerShutdownException from their heartbeat
+      shutdownFuture =
+          shutdownHeartbeatExecutor(shutdownManager)
+              .thenCompose(r -> worker.shutdown(shutdownManager, interruptTasks));
+    }
+    return shutdownFuture.exceptionally(
+        e -> {
+          log.error("[BUG] Unexpected exception during shutdown", e);
+          return null;
+        });
+  }
+
+  private CompletableFuture<Void> shutdownHeartbeatExecutor(ShutdownManager shutdownManager) {
+    return shutdownManager.shutdownExecutor(
+        heartbeatExecutor, this + "#heartbeatExecutor", Duration.ofSeconds(5));
   }
 
   @Override
--- a/temporal-sdk/src/main/java/io/temporal/worker/WorkerFactory.java
+++ b/temporal-sdk/src/main/java/io/temporal/worker/WorkerFactory.java
@@ -358,7 +358,9 @@ public final class WorkerFactory {
    * activity tasks are executed. <br>
    * After the shutdown, calls to {@link
    * io.temporal.activity.ActivityExecutionContext#heartbeat(Object)} start throwing {@link
-   * io.temporal.client.ActivityWorkerShutdownException}.<br>
+   * io.temporal.client.ActivityWorkerShutdownException}, unless {@link
+   * WorkerOptions.Builder#setAllowActivityHeartbeatDuringShutdown(boolean)} is enabled, in which
+   * case heartbeats keep working until the activity tasks finish executing.<br>
    * This method does not wait for the shutdown to complete. Use {@link #awaitTermination(long,
    * TimeUnit)} to do that.<br>
    * Invocation has no additional effect if already shut down.
--- a/temporal-sdk/src/main/java/io/temporal/worker/WorkerOptions.java
+++ b/temporal-sdk/src/main/java/io/temporal/worker/WorkerOptions.java
@@ -77,6 +77,7 @@ public final class WorkerOptions {
     private PollerBehavior workflowTaskPollersBehavior;
     private PollerBehavior activityTaskPollersBehavior;
     private PollerBehavior nexusTaskPollersBehavior;
+    private boolean allowActivityHeartbeatDuringShutdown;
 
     private Builder() {}
 
@@ -112,6 +113,7 @@ public final class WorkerOptions {
       this.workflowTaskPollersBehavior = o.workflowTaskPollersBehavior;
       this.activityTaskPollersBehavior = o.activityTaskPollersBehavior;
       this.nexusTaskPollersBehavior = o.nexusTaskPollersBehavior;
+      this.allowActivityHeartbeatDuringShutdown = o.allowActivityHeartbeatDuringShutdown;
     }
 
     /**
@@ -524,6 +526,28 @@ public final class WorkerOptions {
       return this;
     }
 
+    /**
+     * If true, activities can keep heartbeating during graceful worker shutdown (see {@link
+     * io.temporal.worker.WorkerFactory#shutdown WorkerFactory.shutdown}). Defaults to false, which
+     * means that after graceful shutdown is requested, calling {@link
+     * io.temporal.activity.ActivityExecutionContext#heartbeat ActivityExecutionContext.heartbeat}
+     * does not send a heartbeat and instead throws {@link
+     * io.temporal.client.ActivityWorkerShutdownException ActivityWorkerShutdownException}. This
+     * option is ignored by non-graceful shutdown (see {@link
+     * io.temporal.worker.WorkerFactory#shutdownNow WorkerFactory.shutdownNow}).
+     *
+     * <p>Note that with this option enabled, activities are no longer notified of the worker
+     * shutdown by the {@link io.temporal.client.ActivityWorkerShutdownException
+     * ActivityWorkerShutdownException} exception, so they are expected to complete within the
+     * termination grace period on their own.
+     */
+    @Experimental
+    public Builder setAllowActivityHeartbeatDuringShutdown(
+        boolean allowActivityHeartbeatDuringShutdown) {
+      this.allowActivityHeartbeatDuringShutdown = allowActivityHeartbeatDuringShutdown;
+      return this;
+    }
+
     public WorkerOptions build() {
       return new WorkerOptions(
           maxWorkerActivitiesPerSecond,
@@ -553,7 +577,8 @@ public final class WorkerOptions {
           deploymentOptions,
           workflowTaskPollersBehavior,
           activityTaskPollersBehavior,
-          nexusTaskPollersBehavior);
+          nexusTaskPollersBehavior,
+          allowActivityHeartbeatDuringShutdown);
     }
 
     public WorkerOptions validateAndBuildWithDefaults() {
@@ -685,7 +710,8 @@ public final class WorkerOptions {
           deploymentOptions,
           workflowTaskPollersBehavior,
           activityTaskPollersBehavior,
-          nexusTaskPollersBehavior);
+          nexusTaskPollersBehavior,
+          allowActivityHeartbeatDuringShutdown);
     }
   }
 
@@ -717,6 +743,7 @@ public final class WorkerOptions {
   private final PollerBehavior workflowTaskPollersBehavior;
   private final PollerBehavior activityTaskPollersBehavior;
   private final PollerBehavior nexusTaskPollersBehavior;
+  private final boolean allowActivityHeartbeatDuringShutdown;
 
   private WorkerOptions(
       double maxWorkerActivitiesPerSecond,
@@ -746,7 +773,8 @@ public final class WorkerOptions {
       WorkerDeploymentOptions deploymentOptions,
       PollerBehavior workflowTaskPollersBehavior,
       PollerBehavior activityTaskPollersBehavior,
-      PollerBehavior nexusTaskPollersBehavior) {
+      PollerBehavior nexusTaskPollersBehavior,
+      boolean allowActivityHeartbeatDuringShutdown) {
     this.maxWorkerActivitiesPerSecond = maxWorkerActivitiesPerSecond;
     this.maxConcurrentActivityExecutionSize = maxConcurrentActivityExecutionSize;
     this.maxConcurrentWorkflowTaskExecutionSize = maxConcurrentWorkflowTaskExecutionSize;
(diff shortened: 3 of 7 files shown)


maciejdudko (changes requested):
Hi @baekgyu-kim, thank you for your contribution! It's great to see someone taking on these long standing issues. However, this is not the right implementation.

There should be a new worker option to enable heartbeating during shutdown, It should default to disabled, and when disabled, the behavior should be identical to existing behavior for backward compatibility purposes.

When the option is enabled, the heartbeat behavior should be identical to what happens during normal heartbeat when the worker is not shutting down. There should be no additional code path that calls `sendHeartbeatRequest` a different way, the existing mechanism should be used. The way to achieve that is to modify `SyncActivityWorker.shutdown` so that `heartbeatExecutor.shutdown` is only called after all outstanding activity tasks have finished executing.

If you need assistance with implementation, feel free to reach out on [community Slack](https://t.mp/slack), either message me directly or post on `#java-sdk` channel.

baekgyu-kim:
Hi @maciejdudko,
Thank you for the thoughtful review! I've reworked the PR as suggested.

It now adds an experimental `WorkerOptions.Builder#setActivityHeartbeatDuringShutdown` option. (default `false`, preserving the existing behavior)
When enabled, the heartbeat executor is shut down only after all outstanding activity tasks have finished executing, so heartbeats go through the existing mechanism without any separate code path. 

Whenever you have a chance, I'd appreciate another look. Thanks again!

temporal-sdk/src/main/java/io/temporal/worker/WorkerOptions.java
```diff
@@ -77,6 +77,7 @@ public static final class Builder {
     private PollerBehavior workflowTaskPollersBehavior;
     private PollerBehavior activityTaskPollersBehavior;
     private PollerBehavior nexusTaskPollersBehavior;
+    private boolean activityHeartbeatDuringShutdown;
```
maciejdudko: The field should be named `allowActivityHeartbeatDuringShutdown`, the options getter should be named `getAllowActivityHeartbeatDuringShutdown`, and the builder setter should be named `setAllowActivityHeartbeatDuringShutdown`. Apply this change consistently throughout the PR.

temporal-sdk/src/main/java/io/temporal/internal/worker/SyncActivityWorker.java
```diff
@@ -89,16 +91,31 @@ public boolean start() {
...
-        // could get an ActivityWorkerShutdownException from their heartbeat
-        .shutdownExecutor(heartbeatExecutor, this + "#heartbeatExecutor", Duration.ofSeconds(5))
-        .thenCompose(r -> worker.shutdown(shutdownManager, interruptTasks))
-        .exceptionally(
-            e -> {
-              log.error("[BUG] Unexpected exception during shutdown", e);
-              return null;
-            });
+    CompletableFuture<Void> shutdownFuture;
+    if (activityHeartbeatDuringShutdown) {
```
maciejdudko: When `interruptTasks` is true (`shutdownNow` was called instead of `shutdown`), it should behave as if heartbeat during shutdown was disabled.
```suggestion
    if (allowActivityHeartbeatDuringShutdown && !interruptTasks) {
```

temporal-sdk/src/main/java/io/temporal/worker/WorkerFactory.java
```diff
@@ -375,7 +377,9 @@ public synchronized void shutdown() {
    * interrupts may never terminate.<br>
    * After the shutdownNow calls to {@link
    * io.temporal.activity.ActivityExecutionContext#heartbeat(Object)} start throwing {@link
-   * io.temporal.client.ActivityWorkerShutdownException}.<br>
+   * io.temporal.client.ActivityWorkerShutdownException}, unless {@link
+   * WorkerOptions.Builder#setActivityHeartbeatDuringShutdown(boolean)} is enabled, in which case
+   * heartbeats keep working until the activity tasks finish executing.<br>
```
maciejdudko: `shutdownNow` behavior stays the same, see comment in `SyncActivityWorker`.

```suggestion
   * io.temporal.client.ActivityWorkerShutdownException}.<br>
```

temporal-sdk/src/main/java/io/temporal/worker/WorkerOptions.java
```diff
@@ -524,6 +526,29 @@ public Builder setNexusTaskPollersBehavior(PollerBehavior pollerBehavior) {
...
+     * shutdown by an {@link io.temporal.client.ActivityWorkerShutdownException} thrown from {@code
+     * heartbeat}, so they are expected to complete within the termination grace period on their
+     * own.
+     *
+     * <p>Defaults to false, meaning that after shutdown is requested, {@link
+     * io.temporal.activity.ActivityExecutionContext#heartbeat(Object)} stops sending heartbeats and
+     * throws {@link io.temporal.client.ActivityWorkerShutdownException}.
+     */
+    @Experimental
+    public Builder setActivityHeartbeatDuringShutdown(boolean activityHeartbeatDuringShutdown) {
```
maciejdudko: We don't want to document implementation details.

```suggestion
    /**
     * If true, activities can keep heartbeating during graceful worker shutdown (see {@link
     * io.temporal.worker.WorkerFactory#shutdown WorkerFactory.shutdown}). Defaults to false,
     * which means that after graceful shutdown is requested, calling {@link
     * io.temporal.activity.ActivityExecutionContext#heartbeat ActivityExecutionContext.heartbeat}
     * does not send a heartbeat and instead throws {@link
     * io.temporal.client.ActivityWorkerShutdownException ActivityWorkerShutdownException}. This
     * option is ignored by non-graceful shutdown (see {@link
     * io.temporal.worker.WorkerFactory#shutdownNow WorkerFactory.shutdownNow}).
     *
     * <p>Note that with this option enabled, activities are no longer notified of the worker
     * shutdown by the {@link io.temporal.client.ActivityWorkerShutdownException
     * ActivityWorkerShutdownException} exception, so they are expected to complete within the
     * termination grace period on their own.
     */
    @Experimental
    public Builder setAllowActivityHeartbeatDuringShutdown(boolean allowActivityHeartbeatDuringShutdown) {
```

temporal-sdk/src/test/java/io/temporal/worker/shutdown/HeartbeatDuringWorkerShutdownTest.java
```diff
@@ -0,0 +1,122 @@
...
+   * ActivityWorkerShutdownException}.
+   */
+  @Test
+  public void testHeartbeatingActivityCompletesDuringShutdown()
+      throws ExecutionException, InterruptedException {
+    TestWorkflowReturnString workflow =
+        testWorkflowRule.newWorkflowStub(TestWorkflowReturnString.class);
+    WorkflowExecution execution = WorkflowClient.start(workflow::execute);
+    started.get();
+    testWorkflowRule.getTestEnvironment().shutdown();
```
maciejdudko: There's a race condition here - `shutdown()` call can go through before activity worker receives the task, which will prevent the activity from running and the test will fail.

This feature will be easier to test using a standalone activity. It should work like this:

1. Test starts activity.
2. Test blocks on a semaphore 1 until activity starts.
4. Activity signals semaphore 1.
5. Activity blocks on semaphore 2 until shutdown is triggered.
6. Test calls `shutdown()`.
7. Test signals semaphore 2.
8. Activity heartbeats then returns. (An exception will fail the activity.)
9. Test calls `result()` on activity handle to ensure it succeeded. (Failure will throw exception and fail the test.)

temporal-sdk/src/test/java/io/temporal/worker/shutdown/HeartbeatDuringWorkerShutdownTest.java
```diff
@@ -0,0 +1,122 @@
...
+          .build();
+
+  /**
+   * Tests that when {@link WorkerOptions.Builder#setActivityHeartbeatDuringShutdown(boolean)} is
+   * enabled, {@code Activity#heartbeat} keeps working normally after {@link
+   * WorkerFactory#shutdown()} and the activity runs to completion instead of getting an {@code
+   * ActivityWorkerShutdownException}.
+   */
+  @Test
+  public void testHeartbeatingActivityCompletesDuringShutdown()
```
maciejdudko: Also add a test case for when `shutdownNow` is called instead of `shutdown`.

maciejdudko (changes requested):

baekgyu-kim:
Hi @maciejdudko,
Thanks again for your review. 
Sorry for making you review this twice, and I really appreciate the detailed feedback. 

I've addressed all of your comments:
- **Naming**: Renamed consistently to `allowActivityHeartbeatDuringShutdown` (field, `getAllowActivityHeartbeatDuringShutdown()` getter, `setAllowActivityHeartbeatDuringShutdown(...)` setter) across `WorkerOptions`, `SingleWorkerOptions`, and `Worker`.
- **`shutdownNow` behavior**: Changed the condition to `allowActivityHeartbeatDuringShutdown && !interruptTasks`, so `shutdownNow` behaves exactly as if the option were disabled (heartbeat executor shut down first).
- **`WorkerFactory.shutdown` javadoc**: Reverted. It no longer references the option, so both `shutdown` and `shutdownNow` keep their original documented behavior.
- **`WorkerOptions` setter javadoc**: Replaced with your suggested wording, dropping the implementation details.
- **Test race condition**: Reworked into a standalone activity with the two-semaphore handshake you outlined (wait for the activity to start → trigger shutdown → release the activity to heartbeat and return), eliminating the race where `shutdown` could land before the task is picked up.
- **`shutdownNow` test case**: Added `testHeartbeatingActivityFailsDuringShutdownNow`, which asserts the heartbeat fails the activity instead of letting it complete.

Thanks again for the careful review!

temporal-sdk/src/main/java/io/temporal/worker/WorkerOptions.java
baekgyu-kim: Renamed consistently to `allowActivityHeartbeatDuringShutdown` (field), `getAllowActivityHeartbeatDuringShutdown()` (getter), and `setAllowActivityHeartbeatDuringShutdown(...)` (setter) across WorkerOptions, SingleWorkerOptions, and Worker.

temporal-sdk/src/main/java/io/temporal/internal/worker/SyncActivityWorker.java
baekgyu-kim: Updated the condition to `allowActivityHeartbeatDuringShutdown && !interruptTasks`, so `shutdownNow` now behaves exactly as if the option were disabled.

temporal-sdk/src/main/java/io/temporal/worker/WorkerFactory.java
baekgyu-kim: Reverted. The `WorkerFactory.shutdown` Javadoc no longer references the option, so `shutdownNow` keeps its original documented behavior.

temporal-sdk/src/main/java/io/temporal/worker/WorkerOptions.java
baekgyu-kim: Applied your suggested wording and dropped the implementation details.

temporal-sdk/src/test/java/io/temporal/worker/shutdown/HeartbeatDuringWorkerShutdownTest.java
baekgyu-kim: Thanks for the detailed outline. 
Reworked the test to use a standalone activity with the two-semaphore handshake you described: it waits for the activity to start, triggers shutdown, then releases the activity to heartbeat and return. 
This removes the race where `shutdown()` could land before the task was picked up.

temporal-sdk/src/test/java/io/temporal/worker/shutdown/HeartbeatDuringWorkerShutdownTest.java
baekgyu-kim: Added `testHeartbeatingActivityFailsDuringShutdownNow`, which calls `shutdownNow()` and asserts the heartbeat fails the activity instead of letting it complete, confirming the option is ignored for non-graceful shutdown.

temporal-sdk/src/test/java/io/temporal/worker/shutdown/HeartbeatDuringWorkerShutdownTest.java
```diff
@@ -0,0 +1,136 @@
...
+    ActivityHandle<String> handle = startHeartbeatingActivity();
+
+    assertTrue(
+        "Activity did not start within 30s", activityStarted.tryAcquire(30, TimeUnit.SECONDS));
+    testWorkflowRule.getTestEnvironment().shutdownNow();
+    shutdownTriggered.release();
+
+    // if heartbeating was incorrectly allowed, the activity would complete successfully and this
+    // assertion would fail
+    assertThrows(ActivityFailedException.class, handle::getResult);
```
maciejdudko: Also assert that cause is `ActivityWorkerShutdownException`.

maciejdudko (changes requested):
Almost there! Just one small change to a test remaining. Also, I see you removed the doc change for `WorkerFactory.shutdown`. I liked that one; I believe we should mention that `shutdown` behavior is different when `allowActivityHeartbeatDuringShutdown` is true (unlike `shutdownNow` which is the same both ways).

baekgyu-kim:
Hi @maciejdudko,
Thank you for the thorough reviews, and sorry for the extra rounds. 

Both points are addressed:

  - WorkerFactory.shutdown javadoc: Restored the note about heartbeats continuing when `allowActivityHeartbeatDuringShutdown` is enabled.
  - Test: `testHeartbeatingActivityFailsDuringShutdownNow` now asserts the cause, checking the failure type matches `ActivityWorkerShutdownException`.

Thanks again for the careful reviews. Have a great day.

temporal-sdk/src/test/java/io/temporal/worker/shutdown/HeartbeatDuringWorkerShutdownTest.java
baekgyu-kim: * Restored the `WorkerFactory.shutdown` javadoc note that heartbeats keep working until the activity tasks finish when `allowActivityHeartbeatDuringShutdown` is enabled.
* `testHeartbeatingActivityFailsDuringShutdownNow` now also asserts the cause by its type.

maciejdudko (approved):

PR #2905, opened 2026-06-09
Add tests for temporal-kotlin extension APIs

vikas0686:
## What changed

Added test coverage for previously untested Kotlin extension APIs in the `temporal-kotlin` module.

### Added tests for

* `RpcRetryOptionsExt`
* `WorkerOptionsExt`
* `WorkerFactoryOptionsExt`
* `ContinueAsNewOptionsExt`
* `ChildWorkflowOptionsExt`
* `WorkflowImplementationOptionsExt`
* `WorkflowStubExt`
* `SagaExt`

### Test coverage includes

* DSL builder behavior matches equivalent Java builder configuration
* `copy {}` extension behavior correctly preserves and overrides fields
* Nested DSL usage (`setRetryOptions {}`)
* Vararg activity option mappings in `WorkflowImplementationOptions`
* Reified `WorkflowStub.getResult<T>()` and `getResultAsync<T>()` extensions
* `Saga {}` DSL compensation execution behavior

### Notes

`ContinueAsNewOptions` and `WorkerFactoryOptions` do not implement `equals()`, so those tests validate field values through getters rather than object equality.

### Validation

```bash
./gradlew :temporal-kotlin:test
```

All tests pass successfully.

Thank you for your time and feedback.

### DIFF
--- a/temporal-kotlin/src/test/kotlin/io/temporal/workflow/SagaExtTest.kt
+++ b/temporal-kotlin/src/test/kotlin/io/temporal/workflow/SagaExtTest.kt
@@ -1,0 +1,81 @@
+package io.temporal.workflow
+
+import io.temporal.activity.ActivityInterface
+import io.temporal.activity.ActivityMethod
+import io.temporal.activity.ActivityOptions
+import io.temporal.client.WorkflowClientOptions
+import io.temporal.client.WorkflowOptions
+import io.temporal.common.converter.DefaultDataConverter
+import io.temporal.common.converter.JacksonJsonPayloadConverter
+import io.temporal.common.converter.KotlinObjectMapperFactory
+import io.temporal.testing.internal.SDKTestWorkflowRule
+import org.junit.Assert.assertTrue
+import org.junit.Rule
+import org.junit.Test
+import java.time.Duration
+import java.util.concurrent.atomic.AtomicBoolean
+
+class SagaExtTest {
+
+  companion object {
+    val compensated = AtomicBoolean(false)
+  }
+
+  @Rule
+  @JvmField
+  val testWorkflowRule: SDKTestWorkflowRule = SDKTestWorkflowRule.newBuilder()
+    .setWorkflowTypes(SagaWorkflowImpl::class.java)
+    .setActivityImplementations(CompensationActivityImpl())
+    .setWorkflowClientOptions(
+      WorkflowClientOptions.newBuilder()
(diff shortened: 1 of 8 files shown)


temporal-kotlin/src/test/kotlin/io/temporal/workflow/SagaExtTest.kt
```diff
@@ -0,0 +1,81 @@
...
+        saga.addCompensation { activity.compensate() }
+        throw RuntimeException("simulated failure")
+      } catch (e: Exception) {
+        saga.compensate()
+      }
+    }
+  }
+
+  @Test
+  fun `Saga DSL extension should build Saga with options and run compensations`() {
```
cconstable, line 77: I think this test would be more complete if we actually assert the options being set (setParallelCompensation).

temporal-kotlin/src/test/kotlin/io/temporal/workflow/SagaExtTest.kt
```diff
@@ -0,0 +1,81 @@
...
+    compensated.set(false)
+    val client = testWorkflowRule.workflowClient
+    val stub = client.newWorkflowStub(
+      SagaWorkflow::class.java,
+      WorkflowOptions.newBuilder().setTaskQueue(testWorkflowRule.taskQueue).build()
```
cconstable: While we are at it, we could update this to use the `WorkflowOptions` DSL

cconstable:
@vikas0686 thanks for the contribution, this is a really great PR! I left comments on one of the tests but other than that this looks great.

temporal-kotlin/src/test/kotlin/io/temporal/workflow/SagaExtTest.kt
vikas0686: Updated the test to add two numbered compensations and assert they run in reverse order ([2, 1]), which directly verifies that setParallelCompensation(false) is working as expected.

temporal-kotlin/src/test/kotlin/io/temporal/workflow/SagaExtTest.kt
vikas0686: Done, replaced `WorkflowOptions.newBuilder()...build() ` with the Kotlin DSL `WorkflowOptions { setTaskQueue(...)   }.`

cconstable (approved):
👍 thank you!

PR #2907, opened 2026-06-10
Add extstore s3 driver

cconstable:
## What changed?
- Added the experimental S3 external storage driver and AWS SDK v2 adapter at `contrib/temporal-payload-storage-s3/src/main/java/io/temporal/payload/storage/s3/`
- Added shared `PayloadHasher` for drivers to consistently hash payload content.

## Why?
- Enables Temporal payloads to be offloaded to S3 through the external storage driver API.

## Checklist
- Added focused S3 driver tests covering storage, retrieval, key format, error handling, validation, and README examples.

### DIFF
--- a/contrib/temporal-payload-storage-s3driver/src/main/java/io/temporal/payload/storage/s3driver/CompletableFutures.java
+++ b/contrib/temporal-payload-storage-s3driver/src/main/java/io/temporal/payload/storage/s3driver/CompletableFutures.java
@@ -1,0 +1,56 @@
+package io.temporal.payload.storage.s3driver;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.atomic.AtomicInteger;
+
+final class CompletableFutures {
+  private CompletableFutures() {}
+
+  /**
+   * Completes with the results in input order once every future succeeds. Fails fast with the first
+   * failure's cause as soon as any future fails, without waiting for the rest.
+   */
+  static <T> CompletableFuture<List<T>> allAsList(List<CompletableFuture<T>> futures) {
+    CompletableFuture<List<T>> result = new CompletableFuture<>();
+    if (futures.isEmpty()) {
+      result.complete(new ArrayList<>());
+      return result;
+    }
+    AtomicInteger remaining = new AtomicInteger(futures.size());
+    for (CompletableFuture<T> future : futures) {
+      future.whenComplete(
+          (value, ex) -> {
+            if (ex != null) {
+              result.completeExceptionally(unwrap(ex));
+            } else if (remaining.decrementAndGet() == 0) {
+              List<T> results = new ArrayList<>(futures.size());
+              for (CompletableFuture<T> completed : futures) {
+                results.add(completed.join());
+              }
+              result.complete(results);
+            }
+          });
+    }
+    result.whenComplete(
+        (value, ex) -> {
+          if (ex != null) {
+            for (CompletableFuture<T> future : futures) {
+              future.cancel(true);
+            }
+          }
+        });
+    return result;
+  }
+
+  static Throwable unwrap(Throwable t) {
+    while ((t instanceof CompletionException || t instanceof ExecutionException)
+        && t.getCause() != null) {
+      t = t.getCause();
+    }
+    return t;
+  }
+}
--- a/contrib/temporal-payload-storage-s3driver/src/main/java/io/temporal/payload/storage/s3driver/README.md
+++ b/contrib/temporal-payload-storage-s3driver/src/main/java/io/temporal/payload/storage/s3driver/README.md
@@ -1,0 +1,117 @@
+# AWS S3 Driver
+
+Temporal's S3 Driver for External Storage. Uses the official [AWS S3 Java SDK](https://github.com/aws/aws-sdk-java-v2).
+
+## Usage
+
+Construct the S3 storage driver:
+
+```java
+import io.temporal.payload.storage.s3driver.S3StorageDriver;
+import io.temporal.payload.storage.s3driver.awssdkv2.S3AsyncClientAdapter;
+import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.services.s3.S3AsyncClient;
+
+S3AsyncClient s3Client =
+  S3AsyncClient.builder().region(Region.US_EAST_1).build();
+
+S3StorageDriver driver =
+  S3StorageDriver.newBuilder()
+    .setClient(new S3AsyncClientAdapter(s3Client))
+    .setBucket("temporal-payloads")
+    .build();
+```
+
+Register the driver in external storage config:
+
+```java
+import io.temporal.payload.storage.ExternalStorage;
+
+ExternalStorage externalStorage =
+  ExternalStorage.newBuilder()
+    .setDriver(driver)
+    .build();
+```
+
+Use `setBucketResolver(...)` instead of `setBucket(...)` when bucket selection must vary per
+payload.
+
+## S3 Storage Key Specification
+
+All Temporal S3 drivers generate S3 keys in a consistent manner.
+
+### Key format
+
+Workflow key:
+```text
+v0/ns/{namespace}/wt/{workflow-type}/wi/{workflow-id}/ri/{run-id}/d/{hash-algorithm}/{hex-digest}
+```
+
+Activity key:
+```text
+v0/ns/{namespace}/at/{activity-type}/ai/{activity-id}/ri/{run-id}/d/{hash-algorithm}/{hex-digest}
+```
+
+Fallback key (unknown target):
+```text
+v0/d/{hash-algorithm}/{hex-digest}
+```
+
+- If no namespace, workflow, or activity information is available, the fallback is used.
+- Dynamic path segments are percent-encoded (rules below).
+- Missing values (including a missing `run-id`) are encoded as `null`.
+- `hex-digest` is lower-case SHA-256 hex (64 characters).
+
+### Percent-encoding rules
+
+The Temporal SDKs escape anything that isn't listed in S3's safe character set: https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html
+
+Safe Characters:
+```text
+Alphanumeric characters	
+  0-9
+  a-z
+  A-Z
+
+Special characters	
+  Exclamation point (!)
+  Hyphen (-)
+  Underscore (_)
+  Period (.)
+  Asterisk (*)
+  Single quotation mark (')
+  Opening parenthesis (()
+  Closing parenthesis ())
+```
+
+### Examples
+
+Workflow key example:
+
+```text
+input:
+  namespace=payments prod
+  workflow-type=ChargeWorkflow
+  workflow-id=order+123=abc
+  run-id=3f1d6c7a-8b2e-4f7a-9d0a-87a6f95e4d31
+  hash-algorithm=sha256
+  hex-digest=9f86d081884c7d659a2feaa0c55ad015a3bf4f1b2b0b822cd15d6c15b0f00a08
+
+output:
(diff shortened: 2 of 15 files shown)


contrib/temporal-payload-storage-s3driver/src/main/java/io/temporal/payload/storage/s3driver/README.md
```diff
@@ -0,0 +1,115 @@
...
+```text
+v0/d/{hash-algorithm}/{hex-digest}
+```
+
+- If no namespace, workflow, or activity information is available, the fallback is used.
+- Dynamic path segments are percent-encoded (rules below).
+- Missing values (including a missing `run-id`) are encoded as `null`.
+- `hex-digest` is lower-case SHA-256 hex (64 characters).
+
+### Percent-encoding rules
```
cconstable, line 65: These are encoding rules that all temporal s3 drivers should conform to. If we have a better place to put  information like this I will move it, otherwise I'll put something similar in the READMEs of the other SDKs.

temporal-sdk/src/main/java/io/temporal/payload/storage/PayloadHasher.java
```diff
@@ -0,0 +1,31 @@
+package io.temporal.payload.storage;
+
+import io.temporal.common.Experimental;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import javax.annotation.Nonnull;
+
+/** Computes payload hashes shared by external storage drivers. */
+@Experimental
+public final class PayloadHasher {
```
cconstable: Any driver we add that creates content-addressable keys will need to hash the content somehow. I opted to put this in a shared place. Happy to rename or move if needed.

cconstable:
The actual changes here are only a few hundred lines. The vast majority of the diff is tests, comments, and java ceremony. `storage/s3/S3StorageDriver.java` is the big change.

contrib/temporal-payload-storage-s3/src/main/java/io/temporal/payload/storage/s3/S3Client.java
```diff
@@ -0,0 +1,43 @@
...
+
+import io.temporal.common.Experimental;
+import java.util.Collections;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import javax.annotation.Nonnull;
+
+/** Interface for S3 {@link S3StorageDriver} operations: upload, existence check, and download. */
+@Experimental
+public interface S3Client {
```
jmaeagle99: S3StorageDriverClient

contrib/temporal-payload-storage-s3/src/main/java/io/temporal/payload/storage/s3/BucketResolver.java
```diff
@@ -0,0 +1,18 @@
+package io.temporal.payload.storage.s3;
```
jmaeagle99: `package io.temporal.payload.storage.s3driver` to avoid collisions with the AWS S3 SDK package naming?

contrib/temporal-payload-storage-s3/src/main/java/io/temporal/payload/storage/s3/S3AsyncClientAdapter.java
```diff
@@ -0,0 +1,89 @@
...
+import software.amazon.awssdk.services.s3.model.NoSuchKeyException;
+import software.amazon.awssdk.services.s3.model.PutObjectRequest;
+import software.amazon.awssdk.services.s3.model.S3Exception;
+
+/**
+ * {@link S3Client} backed by the AWS SDK for Java v2 {@link S3AsyncClient}. The wrapped client must
+ * be configured with credentials and a region by the caller.
+ */
+@Experimental
+public final class S3AsyncClientAdapter implements S3Client {
```
jmaeagle99: This makes the `io.temporal.payload.storage.s3` package have a hard dependency on the AWS S3 SDK libraries, which is different than how we did it for Go and Python, which use a separate module for the client implementation.

temporal-sdk/src/main/java/io/temporal/payload/storage/PayloadHasher.java
jmaeagle99: I'd move this to the s3 package for now. As it is, it's publicly available to anyone using the SDK, and I think this should be an implementation detail.

contrib/temporal-payload-storage-s3/src/main/java/io/temporal/payload/storage/s3/S3StorageKey.java
```diff
@@ -0,0 +1,69 @@
...
+import io.temporal.payload.storage.StorageDriverWorkflowInfo;
+import java.nio.charset.StandardCharsets;
+
+/**
+ * Builds the content-addressable S3 object key. The key format and percent-encoding rules are the
+ * cross-SDK specification documented in this package's {@code README.md}.
+ */
+final class S3StorageKey {
+  private static final String KEY_VERSION = "v0";
+  private static final String PATH_SEGMENT_UNRESERVED = "-_.~$&+:=@";
```
jmaeagle99: Did you validate that these are acceptable characters in S3 object keys?

contrib/temporal-payload-storage-s3driver/src/main/java/io/temporal/payload/storage/s3driver/S3StorageDriver.java
```diff
@@ -0,0 +1,295 @@
...
+      this.name = Objects.requireNonNull(name, "name");
+      return this;
+    }
+
+    /**
+     * Maximum serialized payload size in bytes the driver accepts. Defaults to 50 MiB. Storing a
+     * larger payload fails the {@code store} call.
+     */
+    public Builder setMaxPayloadSize(int maxPayloadSize) {
+      this.maxPayloadSize = maxPayloadSize;
```
jmaeagle99, line 299: range check here instead of in builder? not sure what's idiomatic in Java

contrib/temporal-payload-storage-s3/src/main/java/io/temporal/payload/storage/s3/S3StorageDriver.java
```diff
@@ -0,0 +1,295 @@
...
+      this.client = Objects.requireNonNull(client, "client");
+      return this;
+    }
+
+    /**
+     * Stores every payload in a fixed bucket. Mutually exclusive with {@link #setBucketResolver}:
+     * setting both before {@link #build()} is an error.
+     */
+    public Builder setBucket(@Nonnull String bucket) {
+      this.staticBucket = Objects.requireNonNull(bucket, "bucket");
```
jmaeagle99: Can you remove the staticBucket field and provide an annonymous implementation for the bucket resolver? Eliminates the field and the XOR check.

contrib/temporal-payload-storage-s3/src/main/java/io/temporal/payload/storage/s3/S3AsyncClientAdapter.java
cconstable: I'll split out to a `temporal.payload.storage.s3driver.awssdkv2`

contrib/temporal-payload-storage-s3driver/src/main/java/io/temporal/payload/storage/s3driver/S3StorageDriver.java
cconstable: I'm not either. I looked it up and it seems like putting it in here is more idiomatic. Updating.

contrib/temporal-payload-storage-s3/src/main/java/io/temporal/payload/storage/s3/S3StorageKey.java
cconstable: Good callout. I followed Go's implementation for this but I did some reading and interestingly both Go and Python allow at least one character the S3 docs say to avoid (`~`) https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html

I'm going to update the Java SDK to diverge from both the existing implementations, update the key encoding "spec" in the README, and then patch the other two SDKs to follow AWS's recommendations. That seems like the correct thing to do here.

contrib/temporal-payload-storage-s3driver/src/main/java/io/temporal/payload/storage/s3driver/S3StorageDriver.java
```diff
@@ -0,0 +1,313 @@
...
+                  exists -> {
+                    if (exists) {
+                      return CompletableFuture.<Void>completedFuture(null);
+                    }
+                    CompletableFuture<Void> uploadRequest = client.putObject(bucket, key, data);
+                    inFlightRequest.set(uploadRequest);
+                    return withFailureContext(uploadRequest, "upload failed " + location);
+                  })
+              .thenApply(ignored -> claimFor(bucket, key, hexDigest));
+      cancelRequestWhenCancelled(claimFuture, inFlightRequest);
```
cconstable, line 107: What I had here previously did not properly cancel the upstream client request. Now cancellation should propagate correctly:

1. `client.objectExists(bucket, key);` creates a future that eagerly executes.
2. that is wrapped in a future that provides failure context via `withFailureContext`
3. `thenCompose` chains the `client.putObject` request
4. `cancelRequestWhenCancelled(claimFuture, inFlightRequest);` listens for a cancel on `claimFuture` and then reaches upstream and cancels the current inFlightRequest.

If there is a more elegant way to do this in Java please let me know. An alternative solution is returning an opaque cancellation token/context but now you've got two things to cancel and that might be confusing.

cconstable:
Merging the old base into master jumbled up the commits in this PR. Doing a rebase --onto and then force pushing to clean this up.

contrib/temporal-payload-storage-s3driver/src/main/java/io/temporal/payload/storage/s3driver/CompletableFutures.java
```diff
@@ -0,0 +1,56 @@
...
+                results.add(completed.join());
+              }
+              result.complete(results);
+            }
+          });
+    }
+    result.whenComplete(
+        (value, ex) -> {
+          if (ex != null) {
+            for (CompletableFuture<T> future : futures) {
```
jmaeagle99, line 41: After issuing cancellations, do we need to wait for all of the futures to actually finish? Thinking if the futures are using resources that are provided by the caller that we might want to wait for their cancellations to finish before returning to the caller. This _might_ be difficult because of how `CompletableFuture` works. I don't have a specific scenario in mind.

contrib/temporal-payload-storage-s3driver/src/main/java/io/temporal/payload/storage/s3driver/S3StorageDriver.java
```diff
@@ -0,0 +1,313 @@
...
+      throw new S3StorageException("unsupported hash algorithm \"" + algorithm + "\"");
+    }
+    String expectedHash = claimData.get(CLAIM_HASH_VALUE);
+    if (expectedHash == null) {
+      throw missingField(CLAIM_HASH_VALUE);
+    }
+    String actualHash = PayloadHasher.sha256Hex(data);
+    if (!actualHash.equals(expectedHash)) {
+      throw new S3StorageException(
+          "integrity check failed [bucket="
```
jmaeagle99, line 167: Should we use `storageLocation` here and when catching `InvalidProtocolBufferException`?

jmaeagle99 (approved):
I would definitely get someone with Java expertise to review this, especially around the completable future chaining and cancellations.

contrib/temporal-payload-storage-s3driver/src/main/java/io/temporal/payload/storage/s3driver/CompletableFutures.java
maciejdudko: Interrupt is true so the time spent hogging up resources will be minimal. I'd say better not to wait.

maciejdudko (approved):

PR #2908, opened 2026-06-10
Fix flaky test `NexusWorkflowTest.testNexusOperationTimeout_AfterStart`

maciejdudko:
## What was changed

- Changed `NexusWorkflowTest.testNexusOperationTimeout_AfterStart` to handle Nexus operation timeout before operation start and to use larger timeout.
- Bumped CLI version in CI to `v1.7.1-standalone-nexus-operations`

## Why?

Server has introduced small regression in handling of very short schedule-to-close timeouts for Nexus operations, which exposed a bug in this test - it only handled the case where the operation starts successfully and only times out later. It didn't handle the case where the timeout happened before the operation starts, but we have no way to control whether timeout happens before or after start.

The new CLI version is needed by several pending PRs, and contains the regression. The version was bumped here so we can validate the fixed test against the new server version.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,7 +84,7 @@ jobs:
 
       - name: Start CLI server
         env:
-          TEMPORAL_CLI_VERSION: 1.7.0
+          TEMPORAL_CLI_VERSION: 1.7.1-standalone-nexus-operations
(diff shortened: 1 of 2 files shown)


mjameswh (approved):
lgtm

PR #2911, opened 2026-06-11
Add GZIP compression defaulting to on

Sushisource:
## What was changed
Add gzip compression, defaulting on with opt-out

## Why?
Save on data! Parity with other SDKs

## Checklist

1. Closes 

2. How was this tested:
Added simple unit test that headers are set properly. More comprehensive testing was performed in Core to validate server actually compresses bytes. Existing integ tests are verifying nothing is breaking, since this value defaults on.

3. Any docs updates needed?

### DIFF
--- a/temporal-serviceclient/src/main/java/io/temporal/serviceclient/ChannelManager.java
+++ b/temporal-serviceclient/src/main/java/io/temporal/serviceclient/ChannelManager.java
@@ -161,6 +161,7 @@ final class ChannelManager {
 
     return ClientInterceptors.intercept(
         channel,
+        new GrpcCompressionInterceptor(options.getGrpcCompression()),
         MetadataUtils.newAttachHeadersInterceptor(headers),
         new SystemInfoInterceptor(serverCapabilitiesFuture));
   }
@@ -205,6 +206,8 @@ final class ChannelManager {
     } else {
       builder.useTransportSecurity();
     }
+
+    builder.decompressorRegistry(options.getGrpcCompression().getDecompressorRegistry());
 
     // Disable built-in idleTimer until https://github.com/grpc/grpc-java/issues/8714 is resolved.
     // jsdk force-idles channels often anyway, so this is not needed until we stop doing
--- a/temporal-serviceclient/src/main/java/io/temporal/serviceclient/ChannelManager.java
+++ b/temporal-serviceclient/src/main/java/io/temporal/serviceclient/ChannelManager.java
@@ -159,9 +159,13 @@ final class ChannelManager {
       }
     }
 
+    if (options.getGrpcCompression() != GrpcCompression.NONE) {
+      channel =
+          ClientInterceptors.intercept(
+              channel, new GrpcCompressionInterceptor(options.getGrpcCompression()));
+    }
     return ClientInterceptors.intercept(
         channel,
-        new GrpcCompressionInterceptor(options.getGrpcCompression()),
         MetadataUtils.newAttachHeadersInterceptor(headers),
         new SystemInfoInterceptor(serverCapabilitiesFuture));
   }
@@ -206,8 +210,6 @@ final class ChannelManager {
     } else {
       builder.useTransportSecurity();
     }
-
-    builder.decompressorRegistry(options.getGrpcCompression().getDecompressorRegistry());
 
     // Disable built-in idleTimer until https://github.com/grpc/grpc-java/issues/8714 is resolved.
     // jsdk force-idles channels often anyway, so this is not needed until we stop doing
--- a/temporal-serviceclient/src/main/java/io/temporal/serviceclient/GrpcCompression.java
+++ b/temporal-serviceclient/src/main/java/io/temporal/serviceclient/GrpcCompression.java
@@ -1,31 +1,23 @@
 package io.temporal.serviceclient;
 
-import io.grpc.Codec;
-import io.grpc.DecompressorRegistry;
 import javax.annotation.Nullable;
 
-/** Selects transport-level gRPC compression for service calls. */
+/** Selects outbound transport-level gRPC compression for service calls. */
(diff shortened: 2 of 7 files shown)


temporal-serviceclient/src/main/java/io/temporal/serviceclient/GrpcCompression.java
```diff
@@ -0,0 +1,31 @@
+package io.temporal.serviceclient;
+
+import io.grpc.Codec;
+import io.grpc.DecompressorRegistry;
+import javax.annotation.Nullable;
+
+/** Selects transport-level gRPC compression for service calls. */
+public enum GrpcCompression {
```
maciejdudko, line 6: Using an enum locks us out of supporting custom compressors if gRPC ever decides to stabilize it. Instead, make it a regular class with private constructor and a single static member `GZIP` (no compression should be represented by null). This is functionally equivalent to current design, and we still don't expose custom compressors at this point, but it will make future changes easier.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/ChannelManager.java
```diff
@@ -206,6 +207,8 @@ private ManagedChannel prepareChannel() {
       builder.useTransportSecurity();
     }
 
+    builder.decompressorRegistry(options.getGrpcCompression().getDecompressorRegistry());
```
maciejdudko: Compression, decompression and advertising should be 3 independent settings. There are valid scenarios to support decompression without enabling compression, or to support decompression without advertising it (for compatibility with servers that ignore Message-Accept-Encoding). Disabling compression should not impact the others.

Decompressor registry APIs are experimental in gRPC right now. I think we should not call them at all, and not expose the settings to control it. I could be convinced to allow disabling advertising as a separate boolean option marked experimental. But I don't think we should allow disabling decompression entirely until the relevant upstream APIs are stabilized.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/ChannelManager.java
```diff
@@ -161,6 +161,7 @@ private Channel applyHeadStandardInterceptors(Channel channel) {
 
     return ClientInterceptors.intercept(
         channel,
+        new GrpcCompressionInterceptor(options.getGrpcCompression()),
```
maciejdudko: Nit: I would say `GrpcCompressionInterceptor` should only be added to the chain if compression is enabled, but it's not a big deal one way or the other.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/ServiceStubsOptions.java
```diff
@@ -460,6 +479,7 @@ public static class Builder<T extends Builder<T>> {
     private Collection<ClientInterceptor> grpcClientInterceptors;
     private Scope metricsScope;
     private boolean apiKeyProvided;
+    private GrpcCompression grpcCompression = GrpcCompression.GZIP;
```
maciejdudko, line 482: We should have one release cycle where compression is available but not enabled by default.

maciejdudko (changes requested):

temporal-serviceclient/src/main/java/io/temporal/serviceclient/GrpcCompression.java
Sushisource: I don't think we're ever going to need to support a custom compressor. After all, the server has to support it. Maybe we'll add support for zstd or snappy or something, but they can easily be added as variants to this enum (with options).

temporal-serviceclient/src/main/java/io/temporal/serviceclient/ServiceStubsOptions.java
Sushisource: Haven't done that with the other SDKs. This change is I think quite safe, and if we have it off by default we basically won't get any usage. Turning it off I think is sufficient for anyone who happens to have some kind of broken proxy setup that messes things up.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/ChannelManager.java
Sushisource: Ah, OK fair point about the API being experiemental. My preference then would be for the no-compression option to simply not set the compressor and otherwise not change anything, rather than having more knobs

temporal-serviceclient/src/main/java/io/temporal/serviceclient/GrpcCompression.java
maciejdudko: I'm thinking more of custom OSS setups where the user augments both the server and the client with custom compression algorithm.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/ChannelManager.java
maciejdudko: Yes, that would work. If no compression then simply do nothing at all. If GZIP compression then set `callOptions.withCompression` and do nothing else. We don't have to add the other knobs at this time.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/ServiceStubsOptions.java
maciejdudko: Fair enough.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/GrpcCompression.java
Sushisource: I'm not worried about supporting that scenario. We've never once got a request for it in all these years.

maciejdudko (approved):

PR #2912, opened 2026-06-11
Add temporal-workflowstreams contrib module

brianstrauch:
## What was changed

Adds a new `temporal-workflowstreams` contrib module: a durable, multi-topic pub/sub log hosted inside a workflow, mirroring the workflow streams contrib packages in the [Go](https://github.com/temporalio/sdk-go), [Python](https://github.com/temporalio/sdk-python), and [TypeScript](https://github.com/temporalio/sdk-typescript) SDKs. All APIs are marked `@Experimental`.

**Wire protocol (cross-SDK contract):** external publishers send batches via the `__temporal_workflow_stream_publish` signal, subscribers long-poll via the `__temporal_workflow_stream_poll` update, and the `__temporal_workflow_stream_offset` query exposes the current offset. The JSON envelope field names and the per-item payload encoding (base64 of the serialized `temporal.api.common.v1.Payload`) match the other SDKs exactly, so a Java publisher or subscriber interoperates with a workflow written in any of them and vice versa.

**Workflow side:** `WorkflowStream` registers a typed listener (preferably from a `@WorkflowInit` constructor) and supports publisher dedup (publisher ID + monotonic sequence), ~1 MB poll response paging, truncation, and continue-as-new state carryover via `WorkflowStreamState`.

**Client side:** `WorkflowStreamClient` (also constructible from inside an activity) provides a batching background publisher with retry and sequence-based exactly-once delivery, plus a blocking subscription iterator that follows continue-as-new chains, recovers from truncation, and ends cleanly on terminal workflow states.

**Core SDK change:** like the Go SDK (`isWorkflowStreamReservedName`), the signal/update/query dispatchers and listener registration now permit handler names in the exact `__temporal_workflow_stream_` sub-namespace, which is otherwise reserved under the `__temporal_` prefix. Other `__temporal_` names remain blocked.

## Why?

Brings the Java SDK to parity with the other SDKs' experimental workflow streams support, enabling durable event streams whose cost scales with durable batches rather than message count.

## Checklist

1. Closes: n/a

2. How was this tested:
   - `PayloadWireTest` — wire-format round-trips (base64-of-proto, the cross-SDK contract)
   - `StreamPublisherTest` — publish-path unit tests with an injected signal function (batching, sequence advancement, force flush, drain-on-close, flush timeout)
   - `WorkflowStreamTest` — workflow-side integration tests against the in-process test server (external publish, dedup, topic-filtered polls, truncation errors, `@WorkflowInit` construction, custom payload converters)
   - `SubscribeTest` — subscription integration tests (delivery and offset advancement, topic filtering, truncation reset, clean terminal end, continue-as-new follow across runs)
   - `WorkflowStreamReservedNameTest` — reserved-name exemption works end-to-end; other `__temporal_` names still rejected

3. Any docs updates needed: module README included (`contrib/temporal-workflowstreams/README.md`)

### DIFF
--- a/contrib/temporal-workflowstreams/src/main/java/io/temporal/workflowstreams/internal/StreamPublisher.java
+++ b/contrib/temporal-workflowstreams/src/main/java/io/temporal/workflowstreams/internal/StreamPublisher.java
@@ -30,17 +30,6 @@ public final class StreamPublisher {
     void send(PublishInput input);
   }
 
-  private static final class BufItem {
-
-    final String topic;
-    final Object value;
-
-    BufItem(String topic, Object value) {
-      this.topic = topic;
-      this.value = value;
-    }
-  }
-
   private final SignalFunction signal;
   private final DataConverter dataConverter;
   private final String publisherId;
@@ -49,7 +38,7 @@ public final class StreamPublisher {
   private final long maxRetryDurationMs;
 
   private final Object stateLock = new Object();
-  private List<BufItem> buffer = new ArrayList<>();
+  private List<PublishEntry> buffer = new ArrayList<>();
   private List<PublishEntry> pending;
   private long pendingSeq;
   private long sequence;
@@ -77,14 +66,21 @@ public final class StreamPublisher {
   }
 
   /**
-   * Buffers a value and lazily starts the background flush loop. Triggers an immediate flush on
-   * {@code forceFlush} or once the buffer reaches the max batch size.
+   * Converts and buffers a value, lazily starting the background flush loop. Triggers an immediate
+   * flush on {@code forceFlush} or once the buffer reaches the max batch size.
+   *
+   * <p>Conversion happens here, on the caller's thread, so an unconvertible value fails the {@code
+   * publish} call itself instead of poisoning the buffer and silently wedging every later item
+   * behind it in the background flush loop.
+   *
+   * @throws RuntimeException if no configured payload converter accepts {@code value}
    */
   public void publish(String topic, Object value, boolean forceFlush) {
+    PublishEntry entry = encode(topic, value);
     boolean trigger;
     ScheduledExecutorService toTrigger = null;
     synchronized (stateLock) {
-      buffer.add(new BufItem(topic, value));
+      buffer.add(entry);
       trigger = forceFlush || (maxBatchSize > 0 && buffer.size() >= maxBatchSize);
       if (!closed) {
         ensureStartedLocked();
@@ -132,8 +128,8 @@ public final class StreamPublisher {
   }
 
   /**
-   * Sends the pending batch (retry) or encodes and sends the buffer (new batch). Serialized so
-   * concurrent callers send sequentially.
+   * Sends the pending batch (retry) or the buffered batch (new batch). Serialized so concurrent
+   * callers send sequentially.
    */
   private void doFlush() {
     synchronized (flushLock) {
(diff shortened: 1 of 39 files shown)


diwu-sf:
@brianstrauch we would like to use this, our set up is:
- Temporary worker activity is using Python SDK, running the OpenAI agent sdk, events are published from this activity
- Java spring boot server provide SSE stream to web client
- Java need to be able to subscribe to the agent's LLM completion event stream to send to web client

Can you rebase this PR and get it fully merged?
Thanks

diwu-sf:
One concern with the subscriber API: `WorkflowStreamSubscription` is currently a blocking `Iterator`, which means the application has to dedicate a driver thread to each active subscription.

The iterator loop is ergonomic:

```java
try (WorkflowStreamSubscription sub = client.subscribe(options)) {
  for (WorkflowStreamItem item : sub) {
    ...
  }
}
```

but that caller thread is also what drives the long-poll loop. It blocks while waiting for poll update results, cooldowns, and terminal/continue-as-new handling. So the thread model becomes roughly:

```text
one active workflow-stream subscription ~= one occupied driver thread
```

Could we expose a non-blocking subscriber API as the primary or additional API? For example:

```java
WorkflowStreamSubscription<T> subscribe(
    WorkflowStreamSubscribeOptions<T> options,
    WorkflowStreamListener<T> listener);

interface WorkflowStreamListener<T> {
  CompletionStage<Void> onNext(WorkflowStreamItem<T> item);
  default void onError(Throwable failure) {}
  default void onCompleted() {}
}
```

With that shape, the implementation can run the blocking `startUpdate(..., ACCEPTED)` call on a shared executor, then wait for the long-poll result with `getResultAsync()` instead of keeping a caller thread parked for the duration of the poll. Many subscriptions can share the same executor/scheduler, and `CompletionStage` from `onNext` gives applications a backpressure boundary without blocking the poller thread.

The blocking iterator could still be useful as a convenience for synchronous consumers, but we would prefer not to make it the only subscriber interface, because it bakes in one driver thread per active subscription.

contrib/temporal-workflowstreams/src/main/java/io/temporal/workflowstreams/internal/StreamPublisher.java
```diff
@@ -0,0 +1,280 @@
...
+      synchronized (stateLock) {
+        sequence = seq;
+        pending = null;
+        pendingSeq = 0;
+        pendingStartNanos = 0;
+      }
+    }
+  }
+
+  private List<PublishEntry> encodeBuffer(List<BufItem> items) {
```
xumaple: Here `dataConverter.toPayload(value)` throws `DataConverterException` when no configured converter accepts the value. If, say, a caller narrowed the converter set to byte-array-only and then tried to publish a `String`, the chain would reject. (Or alternatively consider if a default converter was given an unserializable object.)

`encodeBuffer` throws before buffer is reset or pending is set, so `doFlush` leaves the poisoned buffer intact. `backgroundFlush` then catches it as if it were a transient send failure (L129) and retries the same buffer every batchInterval, forever. Net effect: one unconvertible value permanently wedges the publisher, causing every subsequent (valid) item buffers behind it and never ships, and nothing surfaces to a fire-and-forget caller unless they happen to call flush().

Worth contrasting with the workflow side: `WorkflowStream.publishToTopic` uses the same conversion, but there a failure surfaces as a workflow-task failure (visible, retried). On the client it's swallowed and silent.

Consider treating a conversion failure as fatal-to-that-item rather than transient: drop (or quarantine) the offending entry and record it in `deferredError` so `flush()`/`close()` surface it, instead of re-encoding the whole buffer each tick.

brianstrauch: Nice find, will fix across all four languages.

xumaple (approved):

maciejdudko (approved):

diwu-sf:
Thanks!

PR #2913, opened 2026-06-11
Add backoff start for CAN

Sushisource:
## What was changed
See title

## Why?
Expose supported features on backend

## Checklist

1. Closes 

2. How was this tested:
Simple unit test

3. Any docs updates needed?

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/workflow/ContinueAsNewOptions.java
+++ b/temporal-sdk/src/main/java/io/temporal/workflow/ContinueAsNewOptions.java
@@ -37,6 +37,7 @@ public final class ContinueAsNewOptions {
     private String taskQueue;
     private RetryOptions retryOptions;
     private Duration workflowTaskTimeout;
+    private Duration backoffStartInterval;
     private Map<String, Object> memo;
     private Map<String, Object> searchAttributes;
     private SearchAttributes typedSearchAttributes;
@@ -57,6 +58,7 @@ public final class ContinueAsNewOptions {
       this.taskQueue = options.taskQueue;
       this.retryOptions = options.retryOptions;
       this.workflowTaskTimeout = options.workflowTaskTimeout;
+      this.backoffStartInterval = options.backoffStartInterval;
       this.memo = options.getMemo();
       this.searchAttributes = options.getSearchAttributes();
       this.typedSearchAttributes = options.getTypedSearchAttributes();
@@ -82,6 +84,12 @@ public final class ContinueAsNewOptions {
 
     public Builder setWorkflowTaskTimeout(Duration workflowTaskTimeout) {
       this.workflowTaskTimeout = workflowTaskTimeout;
+      return this;
+    }
+
+    /** Sets the delay before the first workflow task of the continued run is scheduled. */
+    public Builder setBackoffStartInterval(Duration backoffStartInterval) {
+      this.backoffStartInterval = backoffStartInterval;
       return this;
     }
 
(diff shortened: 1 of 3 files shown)


temporal-sdk/src/main/java/io/temporal/workflow/ContinueAsNewOptions.java
```diff
@@ -186,10 +196,37 @@ public ContinueAsNewOptions(
...
+        null,
+        memo,
+        searchAttributes,
+        typedSearchAttributes,
+        contextPropagators,
+        versioningIntent,
+        initialVersioningBehavior);
+  }
+
+  public ContinueAsNewOptions(
```
maciejdudko: I consider it a mistake that the original constructor is public; it should be deprecated with a comment directing users to use the builder, and the new constructor should be made private.

temporal-sdk/src/test/java/io/temporal/internal/sync/SyncWorkflowContextTest.java
```diff
@@ -1,25 +1,50 @@
...
+  DeterministicRunner runner;
 
   @Before
   public void setUp() {
+    this.threadPool = Executors.newCachedThreadPool();
     this.context = DummySyncWorkflowContext.newDummySyncWorkflowContext();
     this.context.setReplayContext(mockReplayWorkflowContext);
+    when(mockReplayWorkflowContext.getWorkflowType())
+        .thenReturn(WorkflowType.newBuilder().setName("dummy-workflow").build());
+  }
```
maciejdudko, line 34: `threadPool`, `runner` and the call to `when` should be moved inside the test that uses them.

temporal-sdk/src/test/java/io/temporal/internal/sync/SyncWorkflowContextTest.java
```diff
@@ -32,6 +57,27 @@ public void testUpsertSearchAttributes() {
     verify(mockReplayWorkflowContext, times(1)).upsertSearchAttributes(serializedAttr);
   }
 
+  @Test
+  public void testContinueAsNewBackoffStartInterval() {
```
maciejdudko, line 53: Nit: this test should be moved down as to not split up the two upsert SA tests.

maciejdudko (changes requested):

temporal-sdk/src/test/java/io/temporal/internal/sync/SyncWorkflowContextTest.java
```diff
@@ -22,6 +34,9 @@ public void setUp() {
     this.context.setReplayContext(mockReplayWorkflowContext);
   }
 
+  @After
+  public void tearDown() {}
```
maciejdudko: Nit: empty `tearDown` can be removed.

temporal-sdk/src/main/java/io/temporal/workflow/ContinueAsNewOptions.java
maciejdudko, line 192:
```suggestion
  /**
   * @deprecated This constructor doesn't include all options. Use the builder instead.
   */
  @Deprecated
  public ContinueAsNewOptions(
```

maciejdudko (approved):
Left a suggestion to mark `ContinueAsNewOptions` constructor deprecated, but I don't feel strong about it. Everything else looks good; LGTM!

temporal-sdk/src/test/java/io/temporal/internal/sync/SyncWorkflowContextTest.java
Sushisource: Lol whoops

PR #2917, opened 2026-06-15
Implement nexus-based activity cancels

Sushisource:
## What was changed
* Allow activities to be cancelled via the built-in worker nexus channel without requiring heartbeating
* Adds `ActivityCancellationToken` to the activity context help with this 

## Why?
Parity, improved experience, no need to heartbeat if there are no details to record.

## Checklist

1. Closes 

2. How was this tested:
Added tests

3. Any docs updates needed?


temporal-sdk/src/main/java/io/temporal/activity/ActivityCancellationToken.java
```diff
@@ -0,0 +1,49 @@
+package io.temporal.activity;
+
+import io.temporal.client.ActivityCanceledException;
+import java.util.concurrent.CompletableFuture;
+
+/** Token that allows an Activity implementation to observe cancellation requests. */
+public interface ActivityCancellationToken {
```
maciejdudko, line 9: Add `@Experimental` annotation.

temporal-sdk/src/main/java/io/temporal/activity/ActivityCancellationToken.java
```diff
@@ -0,0 +1,49 @@
...
+  void throwIfCancellationRequested() throws ActivityCanceledException;
+
+  /**
+   * Future that completes when cancellation has been requested for this Activity Execution.
+   *
+   * <p>The future completes normally. Activity code should still call {@link
+   * #throwIfCancellationRequested()} or otherwise report cancellation if it wants the Activity
+   * Execution to complete as canceled.
+   */
+  CompletableFuture<Void> getCancellationRequest();
```
maciejdudko: I think it would be better if this future completed exceptionally with the actual cancellation exception - that would make it work better when combining with other futures, as well as make it easier to inspect the exception. I'm also not a fan of the name - `getCancellationRequest` makes it sound like it contains info about the actual request to cancel, which it doesn't. `getCancellationFuture` or `throwIfCancellationRequestedAsync` would be better.

temporal-sdk/src/main/java/io/temporal/activity/ActivityExecutionContext.java
```diff
@@ -89,10 +89,18 @@ public interface ActivityExecutionContext {
    */
   byte[] getTaskToken();
 
+  /**
+   * Returns a token that can be used by Activity code to observe cancellation requests without
+   * recording Heartbeats.
+   */
+  default ActivityCancellationToken getCancellationToken() {
```
maciejdudko: Add `@Experimental` annotation. The default impl is not needed because we have `ActivityExecutionContextBase` serving that role.

temporal-sdk/src/main/java/io/temporal/internal/activity/ActivityCancellationTokenImpl.java
```diff
@@ -0,0 +1,33 @@
...
+    ActivityCanceledException exception = cancellationException;
+    if (exception != null) {
+      throw exception;
+    }
+  }
+
+  @Override
+  public CompletableFuture<Void> getCancellationRequest() {
+    return cancellationRequest;
+  }
```
maciejdudko: This allows completing the cancellation future by the user, which will make the token enter invalid state. Instead, this function should return a new `CompletableFuture` on every call.

(Note the suggestion is missing an import.)

```suggestion
  public CompletableFuture<Void> getCancellationRequest() {
    return cancellationRequest.thenApply(Function.identity());
  }
```

temporal-sdk/src/main/java/io/temporal/internal/activity/ActivityCancellationTokenImpl.java
```diff
@@ -0,0 +1,33 @@
...
+
+  void requestCancel(ActivityCanceledException exception) {
+    cancellationException = exception;
+    cancellationRequest.complete(null);
+  }
```
maciejdudko: This should be made thread-safe, even if the downstream code is already thread-safe.

temporal-sdk/src/main/java/io/temporal/internal/activity/ActivityExecutionContextFactory.java
```diff
@@ -5,4 +5,14 @@
...
+
+  /**
+   * Removes a context for a currently running activity identified by task token and optionally
+   * requests cancellation.
+   *
+   * @return true if the activity was found and cleaned up.
+   */
+  default boolean cleanupContext(byte[] taskToken, boolean cancel) {
+    return false;
+  }
```
maciejdudko: Nit: since it's an internal interface, we don't need a default impl.

temporal-sdk/src/test/java/io/temporal/worker/WorkerHeartbeatIntegrationTest.java
```diff
@@ -748,6 +748,7 @@ private List<WorkerHeartbeat> listWorkersForQueue(String taskQueue) {
                       .setNamespace(
                           testWorkflowRule.getWorkflowClient().getOptions().getNamespace())
                       .setQuery("TaskQueue = \"" + taskQueue + "\"")
+                      .setIncludeSystemWorkers(false)
                       .setPageSize(200)
                       .build());
```
maciejdudko, line 752: Looks like a useless change, the value should be false be default?

temporal-sdk/src/main/java/io/temporal/internal/activity/HeartbeatContextImpl.java
```diff
@@ -228,6 +237,32 @@ public void cancelOutstandingHeartbeat() {
...
+  @Override
+  public void asyncCompletionStarted() {
+    lock.lock();
+    try {
+      requestCancelLocked();
+      rejectNewHeartbeats = true;
+    } finally {
+      lock.unlock();
+    }
+  }
```
maciejdudko, line 257: Although it does make sense to reject new heartbeats after activity function returns, it doesn't really hurt anything to continue allowing it. I'm worried there might be some badly written code in the wild that relies on the old behavior; I'd rather avoid breaking it considering we're not really gaining anything by doing so.

Either way, calling `requestCancelLocked` here is wrong - if the activity code catches `ActivityCanceledException`, it will most likely assume the activity was actually canceled, which is not the case. If we were to keep this method, it should not cancel the token, and `rejectNewHeartbeats` should cause `RuntimeException` to be thrown rather than `ActivityCanceledException`.

My suggestion: remove `rejectNewHeartbeats` field, remove `asyncCompletionStarted` method, and remove the `if (rejectNewHeartbeats)` block from `heartbeat` method.

temporal-sdk/src/main/java/io/temporal/internal/activity/LocalActivityExecutionContextImpl.java
```diff
@@ -57,6 +58,12 @@ public byte[] getTaskToken() {
     throw new UnsupportedOperationException("getTaskToken is not supported for local activities");
   }
 
+  @Override
+  public ActivityCancellationToken getCancellationToken() {
+    throw new UnsupportedOperationException(
+        "getCancellationToken is not supported for local activities");
+  }
```
maciejdudko, line 64: Nit: I think we can safely return `ActivityCancellationToken.NONE` here.

```suggestion
  public ActivityCancellationToken getCancellationToken() {
    return ActivityCancellationToken.NONE;
  }
```

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkerCommandTaskHandler.java
```diff
@@ -0,0 +1,129 @@
...
+                        StartOperationResponse.Sync.newBuilder()
+                            .setPayload(
+                                Payload.newBuilder().setData(response.build().toByteString()))))
+            .build());
+  }
+
+  private ExecuteCommandsRequest decodeRequest(NexusTask task) {
+    if (!task.getResponse().hasRequest()
+        || !task.getResponse().getRequest().hasStartOperation()
+        || !task.getResponse().getRequest().getStartOperation().hasPayload()) {
```
maciejdudko: Nit: intermediate `has` checks are not needed. Can be changed to something like:
```
StartOperationRequest request = task.getResponse().getRequest().getStartOperation();
if (!request.hasPayload()) {
```

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkerCommandTaskHandler.java
```diff
@@ -0,0 +1,129 @@
...
+    WorkerCommandResult.Builder result = WorkerCommandResult.newBuilder();
+    if (command.hasCancelActivity()) {
+      byte[] taskToken = command.getCancelActivity().getTaskToken().toByteArray();
+      Boolean found = activityCancelCallback.apply(taskToken);
+      if (!Boolean.TRUE.equals(found)) {
+        log.debug("Activity task token from worker command was not found");
+      }
+      result.setCancelActivity(CancelActivityResult.newBuilder());
+    } else {
+      log.warn("Worker command has no supported type set");
```
maciejdudko: Nit:
```suggestion
      log.warn("Unsupported worker command");
```

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkerCommandTaskHandler.java
```diff
@@ -0,0 +1,129 @@
...
+      throw new IllegalArgumentException("Failed to decode ExecuteCommandsRequest", e);
+    }
+  }
+
+  private WorkerCommandResult handleCommand(WorkerCommand command) {
+    WorkerCommandResult.Builder result = WorkerCommandResult.newBuilder();
+    if (command.hasCancelActivity()) {
+      byte[] taskToken = command.getCancelActivity().getTaskToken().toByteArray();
+      Boolean found = activityCancelCallback.apply(taskToken);
+      if (!Boolean.TRUE.equals(found)) {
```
maciejdudko, line 119:
```suggestion
      if (!activityCancelCallback.apply(taskToken)) {
```

temporal-sdk/src/main/java/io/temporal/worker/Worker.java
```diff
@@ -126,6 +127,10 @@ private static final class TaskSnapshot {
     WorkflowClientOptions clientOptions = client.getOptions();
     String namespace = clientOptions.getNamespace();
     this.namespace = namespace;
+    String workerControlTaskQueue =
+        clientOptions.getWorkerHeartbeatInterval().isNegative()
+            ? null
+            : WorkerCommandTaskHandler.workerControlTaskQueue(namespace, workerGroupingKey);
```
maciejdudko: Arguably we should still allow token-based cancellation even if heartbeats are disabled. The user has to opt into it by checking the token.

temporal-sdk/src/main/java/io/temporal/worker/WorkerFactory.java
```diff
@@ -285,6 +290,33 @@ public synchronized void start() {
...
+              workflowClient.getOptions().getIdentity(),
+              workerGroupingKey,
+              taskToken -> {
+                for (Worker worker : workers.values()) {
+                  if (worker.requestCancelActivity(taskToken)) {
+                    return true;
+                  }
+                }
+                return false;
+              },
```
maciejdudko: How is thread-safety achieved here? Also, I think this lambda is long enough to be a named method.

maciejdudko (changes requested):

temporal-sdk/src/main/java/io/temporal/internal/activity/LocalActivityExecutionContextImpl.java
Sushisource: We could but I think this is much more explicit then silently just doing nothing an implying you _could_ receive a cancel that you never will.

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkerCommandTaskHandler.java
Sushisource: Could technically throw an NPE if found was `null` (though in practice never should). Keeping it for that reason.

temporal-sdk/src/main/java/io/temporal/internal/activity/ActivityCancellationTokenImpl.java
```diff
@@ -0,0 +1,51 @@
...
+    CompletableFuture<Void> result = new CompletableFuture<>();
+    cancellationFuture.whenComplete(
+        (ignored, exception) -> {
+          if (exception == null) {
+            result.complete(null);
+          } else {
+            result.completeExceptionally(unwrapCompletionException(exception));
+          }
+        });
+    return result;
```
maciejdudko, line 36: Nit: this can be simplified:

```suggestion
    return cancellationFuture.exceptionally(e -> throw unwrapCompletionException(e));
```

This will return a new dependent future on each call, so calling `complete()` on it won't mess up token's internal state.

temporal-sdk/src/main/java/io/temporal/internal/activity/LocalActivityExecutionContextImpl.java
maciejdudko: It does make it harder to write an activity that works in both local and non-local context, though. But we can revisit it another time - throwing exception leaves the door open.

maciejdudko (approved):
One nit, otherwise looks great! LGTM!

temporal-sdk/src/main/java/io/temporal/internal/activity/ActivityCancellationTokenImpl.java
Sushisource: Doesn't actually work since the unwrapped exception is checked unfortunately. End up needing to change the unwrapper to be just as long as this more or less.

PR #2923, opened 2026-06-19
Add Kotlin extension for Standalone Activity Client

444am:
## What was changed
Add Kotlin extension funcs for Standalone Activity Client & its options

## Why?
Workflow Client has Kotlin extension funcs in-place, but Standalone Activity Client hasn't got equivalent extensions

## How was this tested:
See the new UT file.


temporal-kotlin/src/main/kotlin/io/temporal/client/ActivityClientExt.kt
```diff
@@ -0,0 +1,786 @@
...
+import io.temporal.workflow.Functions
+import java.lang.reflect.Type
+import java.util.concurrent.CompletableFuture
+
+/**
+ * Creates client for managing standalone activity executions.
+ *
+ * @see ActivityClient.newInstance
+ */
+fun ActivityClient(
```
444am, line 14: As Standalone Activity is yet supported in test-server, this Kotlin extension file is not able to be tested in a way similar to `WorkflowClientExt`.

I was planning to leverage a temp Activity Client impl like [here](https://github.com/temporalio/sdk-java/pull/2923/changes/ccec476d5415eb9801b9ff570ac4f5091811e0c0), but thinking further this means if iface contract changed in the future, the temp client would need to be changed too, which is troublesome.

Eventually I decide to not put UT on this file as of now because Kotlin extension funcs are more about expressiveness than correctness concern. LMK if it's okay for you.

Quinn-With-Two-Ns, line 14: Hi, thank you for the PR I think there some issues with this API approach however. If a user does something like this:

```kotlin
  client.execute(GreetingActivity::greet, options, "Temporal")
```

where greet returns String, Kotlin selects the ProcN overload because a value-
returning function can be coerced to Unit. The call therefore returns Unit and
silently discards the activity result. Likewise, start and executeAsync infer
`ActivityHandle<Void>` and `CompletableFuture<Void>`.

So the primary Kotlin convenience API behaves incorrectly for normal result-
returning method references unless callers manually disambiguate the SAM type
or generic parameters.

Quinn-With-Two-Ns: https://youtrack.jetbrains.com/projects/KT/issues/KT-7052/Select-overload-with-return-value-if-there-is-return-expression-in-lambda

Quinn-With-Two-Ns:
Closing due to lack of activity, if you get back to this feel free to reopen and address the issues in this PR

PR #2931, opened 2026-06-26
Add WorkflowCheck Gradle plugin module

robzienert:
## What changed?

- Added a new `temporal-workflowcheck-gradle-plugin` module
- Published a Gradle plugin with id `io.temporal.workflowcheck`
- The plugin detects the consumer's `io.temporal:temporal-sdk` version from `runtimeClasspath`, resolves the matching `io.temporal:temporal-workflowcheck` analyzer, and registers a `workflowCheck` verification task.
- Wired `workflowCheck` into the Gradle `check` lifecycle.
- Added configuration for source sets, failure behavior, valid-method output, default config suppression, extra config files, ASM override, and forked JVM heap size.
- Added Gradle TestKit coverage and module README usage documentation.

## Why?

`temporal-workflowcheck` is useful for catching non-deterministic calls in Workflow implementations, but Gradle users currently need to wire the analyzer themselves. This module provides a first-party Gradle integration that automatically uses the analyzer version matching the Temporal Java SDK on the consuming project's classpath.

## Breaking changes?

No.

## Server PR

No server changes required.

## Testing

We've been using this across many different project types within Netflix for a few months and have verified that this approach even for projects with massive classpaths.

- `./gradlew --no-daemon :temporal-workflowcheck-gradle-plugin:test`
- `./gradlew --no-daemon :temporal-workflowcheck-gradle-plugin:spotlessCheck`


gradle/publishing.gradle
```diff
@@ -16,13 +16,15 @@ subprojects {
...
     publishing {
         publications {
-            mavenJava(MavenPublication) {
-                afterEvaluate {
-                    plugins.withId('java-platform') {
-                        from components.javaPlatform
-                    }
-                    plugins.withId('java') {
-                        from components.java
+            if (name != 'temporal-workflowcheck-gradle-plugin') {
```
Quinn-With-Two-Ns, line 19: I believe this is breaking the publication of the gradle plugin. The artifact has no POM metadata and is not signed at least in my testing.

temporal-workflowcheck-gradle-plugin/src/main/java/io/temporal/workflowcheck/gradle/WorkflowCheckTask.java
```diff
@@ -0,0 +1,179 @@
...
+      String output = stdout.toString(StandardCharsets.UTF_8.name());
+      File reportFile = getReportFile().getAsFile().get();
+      reportFile.getParentFile().mkdirs();
+      Files.write(reportFile.toPath(), output.getBytes(StandardCharsets.UTF_8));
+
+      if (!output.trim().isEmpty()) {
+        getLogger().lifecycle(stripTrailing(output));
+      }
+
+      if (result.getExitValue() != 0 && getFailOnViolation().get()) {
```
Quinn-With-Two-Ns: `io.temporal.workflowcheck.Main` ca return a non zero value for configuration failures, but if `FailOnViolation` is not set then it will appear like a pass even if the analyzer didn't really run

temporal-workflowcheck-gradle-plugin/README.md
```diff
@@ -0,0 +1,69 @@
...
+A Gradle plugin that runs the [`temporal-workflowcheck`](../temporal-workflowcheck) bytecode analyzer to detect non-deterministic calls in Temporal Workflow code.
+
+## Usage
+
+Apply the plugin to a Java project that depends on the Temporal Java SDK:
+
+```groovy
+plugins {
+    id 'java'
+    id 'io.temporal.workflowcheck'
```
Quinn-With-Two-Ns: NIT: We should have a version here no otherwise this won't actually work if a user just copy and pastes it

temporal-workflowcheck-gradle-plugin/README.md
```diff
@@ -0,0 +1,69 @@
...
+}
+
+dependencies {
+    implementation 'io.temporal:temporal-sdk:<version>'
+}
+```
+
+The plugin automatically:
+
+- Detects the Temporal SDK version from `runtimeClasspath`.
```
Quinn-With-Two-Ns, line 24: Note, we should document the requirement `temporal-sdk >= 1.33.0`

temporal-workflowcheck-gradle-plugin/src/main/java/io/temporal/workflowcheck/gradle/WorkflowCheckTask.java
```diff
@@ -0,0 +1,179 @@
...
+  @TaskAction
+  public void check() {
+    if (getToolClasspath().isEmpty()) {
+      getLogger().warn("WorkflowCheck: Tool classpath is empty. Skipping.");
+      return;
+    }
+
+    List<File> classpathFiles = new ArrayList<>(getClasspath().getFiles());
+    String maxHeap = getMaxHeapSize().getOrNull();
+    getLogger()
```
Quinn-With-Two-Ns, line 116: This will log every classpath we scan. This could cause a very large output for large repos right?  I would consider lowering the severity

temporal-workflowcheck-gradle-plugin/README.md
robzienert: Ah, I didn't realize that! This must be a difference between Netflix's Nebula Gradle and vanilla Gradle 🙏🏻

temporal-workflowcheck-gradle-plugin/src/main/java/io/temporal/workflowcheck/gradle/WorkflowCheckTask.java
robzienert: Yeah, it's definitely quite spammy. Teams inside of Netflix wanted this for reasons; I'll add a config to control it

PR #2935, opened 2026-07-01
Add regression test for GetSystemInfo default capability masking

eamsden:
Add a replay regression test that routes WorkflowService calls through a small proxy which returns UNIMPLEMENTED for GetSystemInfo and forwards all other RPCs.

The test injects VERSION_WAIT_FOR_MARKER into a known interleaved getVersion history, verifies the history replays without the proxy, then expects the same history to replay through the proxy. On current buggy code this fails with TMPRL1100 because default capabilities disable SDK metadata support and mask the recorded SDK flag.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


maciejdudko:
Superseded by #2936

PR #2936, opened 2026-07-01
Respect SDK flags already present in history regardless of server capability detection.

eamsden:
## What was changed

Respect SDK flags already present in history regardless of server capability detection.

Only default `GetSystemInfo` capabilities for true unknown-method `UNIMPLEMENTED` errors.

Retry gzip-compressed unary RPCs once without compression on compression-related `UNIMPLEMENTED` errors, then cache that downgrade per RPC method.

## Why?

A gzip-incompatible gRPC endpoint or proxy can return `UNIMPLEMENTED` for `GetSystemInfo`. That should not be confused with an old server missing the RPC, and default capabilities should not mask SDK flags already recorded in history.

## Checklist

1. Closes N/A.

2. How was this tested:

Added service-client coverage for per-RPC gzip downgrade, method-level downgrade caching, `GrpcCompression.NONE`, and generic `UNIMPLEMENTED` errors that must not downgrade.

Updated capability tests to distinguish unknown-method `GetSystemInfo` failures from gzip/decompression `UNIMPLEMENTED` failures.

Added replay regression coverage for histories whose SDK flags must still be respected when `GetSystemInfo` falls back to default capabilities.

3. Any docs updates needed?

No external docs updates. Javadocs for gRPC compression were updated.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/common/SdkFlags.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/common/SdkFlags.java
@@ -21,12 +21,9 @@ public final class SdkFlags {
   /**
    * Marks a flag as usable regardless of replay status.
    *
-   * @return True, as long as the server supports SDK flags
+   * @return True.
    */
   public boolean setSdkFlag(SdkFlag flag) {
-    if (!supportSdkMetadata) {
-      return false;
-    }
     sdkFlags.add(flag);
     return true;
   }
@@ -35,6 +32,10 @@ public final class SdkFlags {
    * @return True if this flag may currently be used.
    */
   public boolean tryUseSdkFlag(SdkFlag flag) {
+    if (sdkFlags.contains(flag)) {
+      return true;
+    }
+
     if (!supportSdkMetadata) {
       return false;
     }
@@ -52,10 +53,6 @@ public final class SdkFlags {
    * @return True if this flag is set.
    */
   public boolean checkSdkFlag(SdkFlag flag) {
-    if (!supportSdkMetadata) {
-      return false;
-    }
-
     return sdkFlags.contains(flag);
   }
 
--- a/temporal-sdk/src/main/java/io/temporal/internal/common/SdkFlags.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/common/SdkFlags.java
@@ -18,14 +18,9 @@ public final class SdkFlags {
     this.replaying = replaying;
   }
 
-  /**
-   * Marks a flag as usable regardless of replay status.
-   *
-   * @return True.
-   */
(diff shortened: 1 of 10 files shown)


temporal-sdk/src/main/java/io/temporal/internal/common/SdkFlags.java
```diff
@@ -21,12 +21,9 @@ public SdkFlags(boolean supportSdkMetadata, Functions.Func<Boolean> replaying) {
   /**
    * Marks a flag as usable regardless of replay status.
    *
-   * @return True, as long as the server supports SDK flags
+   * @return True.
    */
   public boolean setSdkFlag(SdkFlag flag) {
```
maciejdudko: Nit: arguably this method should return `void`. (Return value is unused anyway.)

temporal-sdk/src/main/java/io/temporal/internal/common/SdkFlags.java
maciejdudko: Nit: this will always return `false` now.

temporal-serviceclient/src/main/java/io/temporal/serviceclient/GrpcCompressionInterceptor.java
```diff
@@ -4,10 +4,21 @@
...
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Locale;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 
 final class GrpcCompressionInterceptor implements ClientInterceptor {
   private final GrpcCompression compression;
+  private final Set<String> compressionUnsupportedMethods =
```
Sushisource: Needs to be protected for thread safety

Sushisource (commented):
Overall semantics make sense to me but will defer to @maciejdudko for more detailed review

temporal-serviceclient/src/main/java/io/temporal/serviceclient/SystemInfoInterceptor.java
```diff
@@ -112,10 +112,25 @@ public static Capabilities getServerCapabilitiesOrThrow(
...
     } catch (StatusRuntimeException ex) {
-      if (Status.Code.UNIMPLEMENTED.equals(ex.getStatus().getCode())) {
+      if (isGetSystemInfoUnknownMethod(ex.getStatus())) {
         return Capabilities.getDefaultInstance();
       }
       throw ex;
     }
   }
+
+  static boolean isGetSystemInfoUnknownMethod(Status status) {
```
maciejdudko, line 124: Nit: add `@VisibleForTesting` annotation

maciejdudko (approved):

PR #2939, opened 2026-07-02
Add structured concurrency wrapper for CompletableFuture

cconstable:
## What was changed

Add a small, nonblocking structured concurrency-like wrapper for `CompletableFuture`. Loosely inspired by Java 25 `StructuredTaskScope` and Swift `TaskGroup`.

Allows for us to write this:

```java
CompletableFuture<List<String>> result =
    TaskScope.withScope( // 1. create a scope
        scope -> {
            scope.attach(startWork()); // 2. add tasks to the scope
            scope.attach(startMoreWork())
                .thenApply(value -> doDownstreamWork(value));

            return scope.awaitAll(); // 3. collect results
        });

result.get();        // 4. get results a usual
// or
result.cancel(true); // 5. cancel (propagates to all child tasks)
```

See the [README](https://github.com/temporalio/sdk-java/pull/2939/changes#diff-48f89babe423b72f12c99b33cfdd32d4b6d06483ac29a305108ec7cc138ea79b) for more details.

## Why?

I'm doing extstore work and fan-in / fan-out + proper cancellation + error handling boilerplate with raw `CompletableFuture` is tedious and error prone. Scoped task creation gives us correctness guarantees and better ergonomics. I noticed other disparate abstractions solving similar problems and provided a general solution.

## Checklist

I wrote and generated a ton of tests for this and also ran some "stress" tests locally. Since this is just a wrapper on top of `CompletableFuture` and we aren't creating/using our own executors or threading abstractions it's not as scary of a thing as it looks.


temporal-sdk/src/main/java/io/temporal/internal/concurrent/structured/DefaultAsyncTask.java
```diff
@@ -0,0 +1,188 @@
...
+              next.complete(value);
+            }
+            return;
+          }
+
+          Throwable unwrapped = unwrap(error);
+          if (unwrapped instanceof CancellationException) {
+            next.completeExceptionally(unwrapped);
+            return;
+          }
```
cconstable, line 88: cancellation is an error but it's not something we can "recover from" so we should just immediately surface cancellations (skip the user supplied transform)

temporal-sdk/src/main/java/io/temporal/common/CancellationToken.java
```diff
@@ -0,0 +1,57 @@
+package io.temporal.common;
+
+import java.util.concurrent.CancellationException;
+import java.util.concurrent.CompletableFuture;
+
+/** Token that allows asynchronous code to observe cancellation requests. */
+@Experimental
+public interface CancellationToken {
```
cconstable: - the only "public" part of this PR
- meant to serve as a general purpose cancellation token
- can replace existing bespoke tokens e.g. `ActivityCancellationToken`

temporal-sdk/src/main/java/io/temporal/internal/concurrent/structured/TaskScope.java
```diff
@@ -0,0 +1,94 @@
...
+   * @return the scope-wide cancellation token (tripped by {@link #cancelAll()} or {@link
+   *     #close()}).
+   */
+  CancellationToken token();
+
+  /**
+   * Attaches an existing asynchronous task to this scope. Scope cancellation requests cancellation
+   * on the attached future.
+   */
+  <U> TaskChain<U> attach(CompletableFuture<U> future);
```
cconstable: This is the main usage point:
```java
scope.attach(someAsyncWork());
```

It lifts a `CompletableFuture` into the structured scope. It's the bridge from the unstructured CF world into the structured one. We return a `TaskChain` rather than a `CompletableFuture` so callers can chain downstream work (via `map`, `recover`, etc) without leaving the scope. It also does not expose blocking operations and foot-guns like `exceptionally` that break cancellation.

temporal-sdk/src/main/java/io/temporal/internal/common/ListUtils.java
```diff
@@ -0,0 +1,18 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+public final class ListUtils {
+
+  private ListUtils() {}
+
+  /** Concatenates a list of lists into a single list, preserving order. */
+  public static <T> List<T> flatten(List<? extends List<? extends T>> lists) {
```
cconstable, line 11: This isn't used in this PR but it's a natural part of the using this abstraction:

```java
scope.awaitAll(ListUtils::flatten);
```

It's also being used in the branch I have this relies on this branch.

temporal-sdk/src/main/java/io/temporal/common/CancellationToken.java
Sushisource: I think we should replace `ActivityCancellationToken` as part of this PR - it's so new probably no one is using it yet and still experimental. I suppose the only issue is that `CancellationException` is not `ActivityCanceledException` which the activity token needs to throw.

Maybe we can parameterize this over exception type?

temporal-sdk/src/main/java/io/temporal/common/CancellationToken.java
```diff
@@ -0,0 +1,57 @@
...
+  /** Throws {@link CancellationException} if cancellation has been requested. */
+  void throwIfCancellationRequested() throws CancellationException;
+
+  /**
+   * Future that completes normally when cancellation has been requested.
+   *
+   * <p>Code waiting on external work can attach a callback to this future to abort in-flight
+   * requests when cancellation is requested.
+   */
+  default CompletableFuture<Void> getCancellationFuture() {
```
Sushisource, line 26: I think probably this should also throw, like was done with the activity version (or, at least, we should have some way for it to work that way)

temporal-sdk/src/main/java/io/temporal/internal/concurrent/structured/AsyncTask.java
```diff
@@ -0,0 +1,72 @@
...
+          fn.accept(value);
+          return null;
+        });
+  }
+
+  /**
+   * Runs a side effect when this task settles. Unlike {@code CompletableFuture.whenComplete}, the
+   * callback receives the unwrapped throwable ({@code null} on success).
+   */
+  AsyncTask<T> whenSettled(BiConsumer<? super T, ? super Throwable> cb);
```
Sushisource: Can only one of these parameters be non-null? IE: Either it completed T or threw?

If so I think a separate `whenSettledExceptionally` method might make more sense, but I don't feel super strongly about that.

temporal-sdk/src/main/java/io/temporal/internal/concurrent/structured/DefaultTaskScope.java
```diff
@@ -0,0 +1,238 @@
...
+  private final List<DefaultAsyncTask<?>> resultTasks = new CopyOnWriteArrayList<>();
+
+  @Override
+  public CancellationToken token() {
+    return scope.token();
+  }
+
+  @Override
+  public <U> AsyncTask<U> attach(CompletableFuture<U> future) {
+    Objects.requireNonNull(future, "future");
```
Sushisource: Can use `@Nonnull` annotation too to help IDEs

temporal-sdk/src/main/java/io/temporal/internal/concurrent/structured/TaskScope.java
```diff
@@ -0,0 +1,94 @@
...
+  /**
+   * Non-blocking fail-fast wait over all tasks currently owned by this scope.
+   *
+   * <p>The returned future completes with {@code resultSupplier.get()} after all tasks complete
+   * successfully. On first failure or cancellation it completes exceptionally and cancels
+   * unfinished tasks.
+   *
+   * <p>This method does not close the scope. Scope lifetime remains caller-owned.
+   */
+  <R> CompletableFuture<R> awaitAll(Supplier<R> resultSupplier);
```
Sushisource: Not sure I get what the point of this is. Seems like the no-param version and the transformer version are the ones that matter. This seems like super thin sugar of the no-param version?

temporal-sdk/src/main/java/io/temporal/internal/concurrent/structured/TaskScope.java
```diff
@@ -0,0 +1,94 @@
...
+   * replaces its parent in the collected result list. On first failure or cancellation it completes
+   * exceptionally and cancels unfinished tasks.
+   *
+   * <p>This method does not close the scope. Scope lifetime remains caller-owned.
+   */
+  <R> CompletableFuture<R> awaitAll(Function<List<T>, R> resultTransformer);
+
+  /**
+   * Non-blocking wait that completes with each collected task's settled outcome in collection
+   * order.
```
Sushisource:
```suggestion
   * Non-blocking wait that completes with each collected task's settled outcome in completion
   * order.
```

Is this true? Unclear to me what collection order means

temporal-sdk/src/main/java/io/temporal/internal/concurrent/structured/Result.java
```diff
@@ -0,0 +1,105 @@
...
+    return status == Status.CANCELLED;
+  }
+
+  /**
+   * @return the value on success; otherwise throws {@link IllegalStateException} (cause attached).
+   */
+  public T get() {
+    if (status == Status.SUCCESS) return value;
+    throw new IllegalStateException("not a success: " + status, cause);
+  }
```
Sushisource, line 65: Might make more sense to just return an `Optional` here?

temporal-sdk/src/main/java/io/temporal/internal/concurrent/structured/README.md
```diff
@@ -0,0 +1,163 @@
...
+        }
+        return scope.awaitAll(ListUtils::flatten);
+    });
+
+result.cancel(true);
+```
+
+The equivalent code using `CompletableFuture` directly has to rebuild the same ownership and cancellation rules by hand:
+
+```java
```
Sushisource, line 65: Hahaha Jesus Christ lol

Sushisource (commented):
The readme example is very motivating.

I don't have any particularly strong comments. Just a few minor things / questions

temporal-sdk/src/main/java/io/temporal/common/CancellationToken.java
cconstable: Yea, we could parameterize it. I'll go ahead and make that change. Personally, I don't think the extra type will be an ergonomic issue.

temporal-sdk/src/main/java/io/temporal/internal/concurrent/structured/TaskScope.java
cconstable: Yep it is just thin sugar. I'll trim it.

temporal-sdk/src/main/java/io/temporal/internal/concurrent/structured/AsyncTask.java
cconstable: this mirrors `CompletableFuture.whenComplete` so no, they both can be null. on success the throwable is `null` however I don't think I'm using this anymore so we can actually just remove this method.

temporal-sdk/src/main/java/io/temporal/internal/concurrent/structured/TaskScope.java
cconstable: It's not completion order it's the order they are `attached`. Updated doc comment to clarify.

Sushisource (approved):
Looking good to me! I think you'll want a review from a Java owner too.

PR #2940, opened 2026-07-03
fix flakey WorkflowUpdateTest.duplicateRejectedUpdate logic where ADMITTED was reported for updates that were actually COMPLETED.

cconstable:
## What was changed

`getStage()` in the test server now checks the terminal `outcome` future before `accepted`, so an update's reported lifecycle stage no longer depends on the order those two futures complete. This fixes the flaky duplicateRejectedUpdate test, which could observe ADMITTED instead of COMPLETED for a rejected update.

essentially, if it's done just say it's done and perform that logic check first.

## Why?

`WorkflowUpdateTest.duplicateRejectedUpdate` intermittently failed, expecting stage
`COMPLETED` but observing `ADMITTED`.

### DIFF
--- a/temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
+++ b/temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
@@ -3172,13 +3172,18 @@ class TestWorkflowMutableStateImpl implements TestWorkflowMutableState {
     }
 
     public UpdateWorkflowExecutionLifecycleStage getStage() {
-      if (!accepted.isDone()) {
-        return UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_ADMITTED;
-      } else if (!outcome.isDone()) {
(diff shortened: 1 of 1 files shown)


temporal-test-server/src/main/java/io/temporal/internal/testservice/TestWorkflowMutableStateImpl.java
```diff
@@ -3172,13 +3172,18 @@ public UpdateWorkflowExecutionLifecycleStage waitForStage(
     }
 
     public UpdateWorkflowExecutionLifecycleStage getStage() {
-      if (!accepted.isDone()) {
-        return UPDATE_WORKFLOW_EXECUTION_LIFECYCLE_STAGE_ADMITTED;
-      } else if (!outcome.isDone()) {
+      // A resolved outcome is terminal (a success result or a rejection/failure), so it always
```
mjameswh, line 3175: nit: comment formatting is weird

mjameswh (approved):
Please fix formatting issue before merging.

PR #2942, opened 2026-07-08
Add preferred version provider to worker options

### LINKED ISSUE #2941: Add preferred version worker option
See https://github.com/temporalio/features/issues/854

Sushisource:
## What was changed
Per a customer request, added a way to specify a "preferred" version to use when making a get version call which may not necessarily be the `maxVersion` which would normally be used when a `getVersion` call is encountered for the first time.

This is accomplished by adding a worker option: `setPreferredVersionProvider` which accepts a callback that can be used to determine the version for a given call.

## Why?
Users would like to have a way to perform a rolling deployment of a worker change that includes an introduction of a `getVersion` call to a workflow. Without this capability, such a rolling deployment means that if a worker with the new code picks up a task, runs the call, and then the workflow replays on the old code, a WFT failure occurs because the old code is missing the call. (The other direction is acceptable). 

This makes rolling deployments noisy, especially if they proceed slowly.

## Checklist

1. Closes https://github.com/temporalio/sdk-java/issues/2941

2. How was this tested:
Added tests

3. Any docs updates needed?

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowContextImpl.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowContextImpl.java
@@ -16,6 +16,8 @@ import io.temporal.internal.common.ProtobufTimeUtils;
 import io.temporal.internal.common.SdkFlag;
 import io.temporal.internal.statemachines.*;
 import io.temporal.internal.worker.SingleWorkerOptions;
+import io.temporal.worker.PreferredVersionProvider;
+import io.temporal.worker.PreferredVersionProviderInput;
 import io.temporal.workflow.Functions;
 import io.temporal.workflow.Functions.Func;
 import io.temporal.workflow.Functions.Func1;
@@ -38,6 +40,7 @@ final class ReplayWorkflowContextImpl implements ReplayWorkflowContext {
   private final @Nullable String fullReplayDirectQueryName;
   private final Scope replayAwareWorkflowMetricsScope;
   private final SingleWorkerOptions workerOptions;
+  private final WorkflowInfoImpl workflowInfo;
 
   /**
    * @param fullReplayDirectQueryName query name if an execution is a full replay caused by a direct
@@ -61,6 +64,7 @@ final class ReplayWorkflowContextImpl implements ReplayWorkflowContext {
     this.replayAwareWorkflowMetricsScope =
         new ReplayAwareScope(workflowMetricsScope, this, workflowStateMachines::currentTimeMillis);
     this.workerOptions = workerOptions;
+    this.workflowInfo = new WorkflowInfoImpl(this);
   }
 
   @Override
@@ -337,7 +341,17 @@ final class ReplayWorkflowContextImpl implements ReplayWorkflowContext {
       int minSupported,
       int maxSupported,
       Functions.Proc2<Integer, RuntimeException> callback) {
-    return workflowStateMachines.getVersion(changeId, minSupported, maxSupported, callback);
+    PreferredVersionProvider preferredVersionProvider = workerOptions.getPreferredVersionProvider();
+    return workflowStateMachines.getVersion(
+        changeId,
+        minSupported,
+        maxSupported,
+        preferredVersionProvider == null
+            ? null
+            : (min, max) ->
+                preferredVersionProvider.getPreferredVersion(
+                    new PreferredVersionProviderInput(workflowInfo, changeId, min, max)),
+        callback);
   }
 
   @Override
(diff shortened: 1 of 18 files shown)


temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowContextImpl.java
```diff
@@ -61,6 +64,7 @@ final class ReplayWorkflowContextImpl implements ReplayWorkflowContext {
     this.replayAwareWorkflowMetricsScope =
         new ReplayAwareScope(workflowMetricsScope, this, workflowStateMachines::currentTimeMillis);
     this.workerOptions = workerOptions;
+    this.workflowInfo = new WorkflowInfoImpl(this);
```
Quinn-With-Two-Ns: I think you could just call `WorkflowInternal.GetInfo`

temporal-sdk/src/main/java/io/temporal/worker/PreferredVersionProvider.java
```diff
@@ -0,0 +1,30 @@
...
+   * Returns the preferred version for the supplied {@code getVersion} call, or {@link
+   * Optional#empty()} to use the SDK default of {@code maxSupported}.
+   *
+   * <p>This method is invoked on the workflow thread. Any exception it throws propagates out of the
+   * {@code getVersion} call and fails the current workflow task, which will be retried; a provider
+   * that always throws will therefore block the workflow. Implementations should be deterministic
+   * and side-effect free.
+   */
+  @Nonnull
+  Optional<VersionPreference> getPreferredVersion(@Nonnull PreferredVersionProviderInput input);
```
Quinn-With-Two-Ns: I'd suggest 
```suggestion
  @Nullable
  VersionPreference getPreferredVersion(@Nonnull PreferredVersionProviderInput input);
```
In general I'd prefer using `@Nullable` over `Optional`. That is what most of the user facing ` @FunctionalInterface` we expose do

Quinn-With-Two-Ns: The Java SDK isn't incredibly consistent in our use of `Optional` , but I tried to stick to the oracle guidance for optional.

> mechanism for library method return types where there needed to be a clear way to represent "no result", and using null for such was overwhelmingly likely to cause errors.

temporal-sdk/src/main/java/io/temporal/internal/statemachines/VersionStateMachine.java
```diff
@@ -253,6 +267,30 @@ State createMarkerExecuting() {
       }
     }
 
+    private int resolveVersionExecuting() {
+      if (preferredVersionProvider == null) {
+        return maxSupported;
+      }
+      Optional<VersionPreference> preference =
+          Objects.requireNonNull(
+              preferredVersionProvider.apply(minSupported, maxSupported),
```
Quinn-With-Two-Ns: This is run in a workflow thread, so I would consider running it in a readonly context so users can't do silly things like try to start an activity, or more likely use things like `Workflow.Random`

Sushisource: Ah, yeah good point, wild.

temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowInternal.java
```diff
@@ -675,6 +675,17 @@ public static <T> T deadlockDetectorOff(Functions.Func<T> func) {
     }
   }
 
+  public static <T> T readOnly(Functions.Func<T> func) {
```
Sushisource, line 677: @Quinn-With-Two-Ns Seemed useful. LMK if you think I should refactor other sites to use this that are doing the same thing by hand.

temporal-sdk/src/main/java/io/temporal/internal/replay/WorkflowInfoImpl.java
```diff
@@ -1,24 +1,23 @@
...
-import io.temporal.internal.replay.ReplayWorkflowContext;
 import io.temporal.workflow.WorkflowInfo;
 import java.time.Duration;
 import java.util.List;
 import java.util.Optional;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
-final class WorkflowInfoImpl implements WorkflowInfo {
+public final class WorkflowInfoImpl implements WorkflowInfo {
```
Quinn-With-Two-Ns: Shouldn't need this change anymore right?

Quinn-With-Two-Ns (approved):

tconley1428 (approved):

PR #2943, opened 2026-07-09
Propagate memo on continue-as-new in the test server

### LINKED ISSUE #2863: TestEnv should propagate Memo along with CAN
This is very similar ask as in https://github.com/temporalio/sdk-java/pull/2731/files

just the test part should also propagate Memo on CAN

aditya-786:
## What changed?

The in-memory test server carries over search attributes and the header on continue-as-new but drops the memo. As a result, a workflow that sets a memo on continue-as-new cannot read it back on the new run under the test environment, even though this works against a real server.

This sets the memo on the `ContinuedAsNew` event attributes (`StateMachines`) and on the new run's start request (`TestWorkflowService`), mirroring the existing search-attribute handling added in #2731. `ContinueAsNewTest` already set a memo on each continue-as-new but never asserted it downstream; it now asserts the memo is visible on the runs that follow.

## Why?

Reported in #2863. We run the Java SDK in production and rely on memo continuity across continue-as-new; the test server behaving differently from the real server makes it impossible to cover that behavior in workflow tests.

## Breaking changes?

No. Test-server only; behavior now matches the real server.

## Server PR

Not required.

Closes #2863

### DIFF
--- a/temporal-sdk/src/test/java/io/temporal/workflow/ContinueAsNewTest.java
+++ b/temporal-sdk/src/test/java/io/temporal/workflow/ContinueAsNewTest.java
@@ -77,6 +77,13 @@ public class ContinueAsNewTest {
         assertEquals(5, Workflow.getInfo().getRetryOptions().getMaximumAttempts());
         assertEquals("foo1", Workflow.getTypedSearchAttributes().get(CUSTOM_KEYWORD_SA));
       }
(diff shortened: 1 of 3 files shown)


maciejdudko (approved):
Hi @aditya-786, thank you for your contribution! The code looks fine and the tests pass. LGTM!

PR #2945, opened 2026-07-10
NEXUS-485: Support Workflow Update as a Nexus Operation

mavemuri:
## What was changed

Adds support for Workflow Update as a Nexus operation

## Why?

Part of effort to expose all Temporal primitives via Nexus operations
## Checklist

1. Closes NEXUS-485

2. How was this tested:

- [x] added integration tests 

3. Any docs updates needed?

- [ ] TBD

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
@@ -521,9 +521,8 @@ public class RootWorkflowClientInvoker implements WorkflowClientCallsInterceptor
                     .setIdentity(clientOptions.getIdentity()))
             .setInput(updateInput);
 
-    // If this update is being issued from inside a Nexus operation handler via
-    // TemporalNexusClientImpl.startWorkflowUpdate, set the fields the server needs
-    // to deliver the Nexus completion callback.
+    // If this update is being issued via TemporalNexusClientImpl.startWorkflowUpdate,
+    // set the fields the server needs to deliver the Nexus completion callback
     if (CurrentNexusOperationContext.isNexusContext()) {
       InternalNexusOperationContext nexusContext = CurrentNexusOperationContext.get();
       // already in a Nexus operation context, dont need to check nexusContext again
@@ -549,10 +548,9 @@ public class RootWorkflowClientInvoker implements WorkflowClientCallsInterceptor
                     nexusOperationMetadata.operationToken,
                     requestLinks))
             .addAllLinks(requestLinks);
-      } else {
-        log.error("unexpected error fetching nexusOperationMetadata");
-        // maybe throw an exception but should never really happen as its all sdk
       }
+      // If no NexusOperationMetadata was found, but there is a NexusContext, then the
+      // update was likely trigger via Operation handler directly
     }
 
     Request request = requestBuilder.build();
--- a/temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
@@ -5,6 +5,7 @@ import static io.temporal.api.workflowservice.v1.ExecuteMultiOperationResponse.R
 import static io.temporal.internal.common.HeaderUtils.intoPayloadMap;
 import static io.temporal.internal.common.WorkflowExecutionUtils.makeUserMetaData;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
 import com.google.common.collect.Iterators;
 import io.grpc.Deadline;
 import io.grpc.Status;
@@ -23,7 +24,11 @@ import io.temporal.common.converter.DataConverter;
 import io.temporal.common.interceptors.WorkflowClientCallsInterceptor;
 import io.temporal.internal.client.external.GenericWorkflowClient;
 import io.temporal.internal.common.HeaderUtils;
+import io.temporal.internal.common.InternalUtils;
 import io.temporal.internal.nexus.CurrentNexusOperationContext;
+import io.temporal.internal.nexus.InternalNexusOperationContext;
+import io.temporal.internal.nexus.NexusOperationMetadata;
+import io.temporal.internal.nexus.OperationTokenUtil;
 import io.temporal.internal.worker.WorkerVersioningProtoUtils;
 import io.temporal.payload.context.WorkflowSerializationContext;
 import io.temporal.serviceclient.StatusUtils;
@@ -473,6 +478,19 @@ public class RootWorkflowClientInvoker implements WorkflowClientCallsInterceptor
       }
     } while (updateNotYetDurable(input, result));
 
+    // If triggered by a Nexus Operation, set necessary fields- link, result
+    if (CurrentNexusOperationContext.isNexusContext()) {
+      NexusOperationMetadata nexusOperationMetadata =
+          CurrentNexusOperationContext.get().getNexusOperationMetadata();
+      if (nexusOperationMetadata != null) {
+        if (result.hasLink()) {
+          // add forward links for caller->handler
+          CurrentNexusOperationContext.get().addResponseLink(result.getLink());
+        }
+        nexusOperationMetadata.operationCompleted = result.hasOutcome();
+      }
+    }
+
     return toUpdateHandle(input, result, dataConverterWithWorkflowContext);
   }
 
@@ -495,14 +513,49 @@ public class RootWorkflowClientInvoker implements WorkflowClientCallsInterceptor
             .setName(input.getUpdateName());
     inputArgs.ifPresent(updateInput::setArgs);
 
-    Request request =
+    Request.Builder requestBuilder =
         Request.newBuilder()
             .setMeta(
                 Meta.newBuilder()
                     .setUpdateId(input.getUpdateId())
                     .setIdentity(clientOptions.getIdentity()))
-            .setInput(updateInput)
-            .build();
+            .setInput(updateInput);
+
+    // If this update is being issued from inside a Nexus operation handler via
+    // TemporalNexusClientImpl.startWorkflowUpdate, set the fields the server needs
+    // to deliver the Nexus completion callback.
+    if (CurrentNexusOperationContext.isNexusContext()) {
+      InternalNexusOperationContext nexusContext = CurrentNexusOperationContext.get();
+      // already in a Nexus operation context, dont need to check nexusContext again
+      NexusOperationMetadata nexusOperationMetadata = nexusContext.getNexusOperationMetadata();
+      if (nexusOperationMetadata != null) {
+        try {
+          nexusOperationMetadata.operationToken =
+              OperationTokenUtil.generateWorkflowUpdateOperationToken(
+                  clientOptions.getNamespace(),
+                  input.getWorkflowExecution().getWorkflowId(),
+                  input.getWorkflowExecution().getRunId(),
+                  input.getUpdateId());
+        } catch (JsonProcessingException e) {
+          throw new IllegalStateException("failed to generate update operation token", e);
+        }
+        List<Link> requestLinks = nexusContext.getRequestLinks();
+        requestBuilder
+            .setRequestId(nexusOperationMetadata.requestId)
+            .addCompletionCallbacks(
+                InternalUtils.buildNexusCallback(
+                    nexusOperationMetadata.callbackHeaders,
+                    nexusOperationMetadata.callbackUrl,
+                    nexusOperationMetadata.operationToken,
+                    requestLinks))
+            .addAllLinks(requestLinks);
+      } else {
+        log.error("unexpected error fetching nexusOperationMetadata");
+        // maybe throw an exception but should never really happen as its all sdk
+      }
+    }
+
+    Request request = requestBuilder.build();
 
     return UpdateWorkflowExecutionRequest.newBuilder()
         .setNamespace(clientOptions.getNamespace())
--- a/temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
@@ -5,7 +5,6 @@ import static io.temporal.api.workflowservice.v1.ExecuteMultiOperationResponse.R
 import static io.temporal.internal.common.HeaderUtils.intoPayloadMap;
 import static io.temporal.internal.common.WorkflowExecutionUtils.makeUserMetaData;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
 import com.google.common.collect.Iterators;
 import io.grpc.Deadline;
 import io.grpc.Status;
@@ -535,7 +534,7 @@ public class RootWorkflowClientInvoker implements WorkflowClientCallsInterceptor
                   input.getWorkflowExecution().getWorkflowId(),
                   input.getWorkflowExecution().getRunId(),
                   input.getUpdateId());
-        } catch (JsonProcessingException e) {
+        } catch (Exception e) {
           throw new IllegalStateException("failed to generate update operation token", e);
         }
         List<Link> requestLinks = nexusContext.getRequestLinks();
--- a/temporal-sdk/src/main/java/io/temporal/internal/common/InternalUtils.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/common/InternalUtils.java
@@ -157,6 +157,31 @@ public final class InternalUtils {
     return name.startsWith(WORKFLOW_STREAM_RESERVED_PREFIX);
   }
 
+  /** Helper to build a Nexus Callback from the provided input. */
+  public static Callback buildNexusCallback(
+      Map<String, String> callbackHeaders,
+      String callbackUrl,
+      String operationToken,
+      List<Link> links) {
+    Map<String, String> headers =
+        callbackHeaders.entrySet().stream()
+            .collect(
+                Collectors.toMap(
+                    (k) -> k.getKey().toLowerCase(),
+                    Map.Entry::getValue,
+                    (a, b) -> a,
+                    TreeMap::new));
+    headers.put(Header.OPERATION_TOKEN.toLowerCase(), operationToken);
+    Callback.Builder cbBuilder =
+        Callback.newBuilder()
+            .setNexus(
+                Callback.Nexus.newBuilder().setUrl(callbackUrl).putAllHeader(headers).build());
+    if (links != null) {
+      cbBuilder.addAllLinks(links);
+    }
+    return cbBuilder.build();
+  }
+
   /** Check the method name for reserved prefixes or names. */
   public static void checkMethodName(POJOWorkflowMethodMetadata methodMetadata) {
     boolean workflowStreamExempt =
--- a/temporal-sdk/src/main/java/io/temporal/internal/common/LinkConverter.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/common/LinkConverter.java
@@ -19,6 +19,7 @@ public class LinkConverter {
 
   private static final Logger log = LoggerFactory.getLogger(LinkConverter.class);
 
+  private static final String temporalUrlScheme = "temporal";
   private static final String linkPathFormat = "temporal:///namespaces/%s/workflows/%s/%s/history";
   private static final String nexusOperationLinkPathFormat =
       "temporal:///namespaces/%s/nexus-operations/%s/%s/details";
@@ -35,6 +36,7 @@ public class LinkConverter {
       Link.WorkflowEvent.getDescriptor().getFullName();
   private static final String nexusOperationLinkType =
       Link.NexusOperation.getDescriptor().getFullName();
+  private static final String workflowLinkType = Link.Workflow.getDescriptor().getFullName();
 
   public static io.temporal.api.nexus.v1.Link workflowEventToNexusLink(Link.WorkflowEvent we) {
     try {
@@ -91,6 +93,24 @@ public class LinkConverter {
       log.error("Failed to encode Nexus link URL", e);
     }
     return null;
+  }
+
+  public static io.temporal.api.nexus.v1.Link workflowLinkToNexusLink(Link.Workflow w) {
+    try {
+      String namespace = URLEncoder.encode(w.getNamespace(), StandardCharsets.UTF_8.toString());
+      String workflowId =
+          URLEncoder.encode(w.getWorkflowId(), StandardCharsets.UTF_8.toString())
+              .replace("+", "%20"); // handle workflowIds supporting spaces
+      String runId = URLEncoder.encode(w.getRunId(), StandardCharsets.UTF_8.toString());
+      String url = String.format(linkPathFormat, namespace, workflowId, runId);
+      return io.temporal.api.nexus.v1.Link.newBuilder()
+          .setUrl(url)
+          .setType(workflowLinkType)
+          .build();
+    } catch (UnsupportedEncodingException e) {
+      log.error("Failed to encode Nexus link URL", e);
+      return null;
+    }
   }
 
   public static Link nexusLinkToWorkflowEvent(io.temporal.api.nexus.v1.Link nexusLink) {
@@ -166,6 +186,43 @@ public class LinkConverter {
     return link.build();
   }
 
+  public static Link nexusLinkToWorkflowLink(io.temporal.api.nexus.v1.Link nexusLink) {
+    Link.Builder link = Link.newBuilder();
+    try {
+      URI uri = new URI(nexusLink.getUrl());
+      if (!uri.getScheme().equals(temporalUrlScheme)) {
+        log.error("Failed to parse Nexus link URL: invalid scheme: {}", uri.getScheme());
+        return null;
+      }
+      StringTokenizer st = new StringTokenizer(uri.getRawPath(), "/");
+      // maybe add constants for "namespaces", "workflows" too
+      if (!st.nextToken().equals("namespaces")) {
+        log.error("Failed to parse Nexus link URL: invalid path: {}", uri.getRawPath());
+        return null;
+      }
+      String namespace = URLDecoder.decode(st.nextToken(), StandardCharsets.UTF_8.toString());
+      if (!st.nextToken().equals("workflows")) {
+        log.error("Failed to parse Nexus link URL: invalid path: {}", uri.getRawPath());
+        return null;
+      }
+      String workflowID = URLDecoder.decode(st.nextToken(), StandardCharsets.UTF_8.toString());
+      if (!st.hasMoreTokens()) {
+        log.error("Failed to parse Nexus link URL: invalid path: {}", uri.getRawPath());
+        return null;
+      }
+      String runID = URLDecoder.decode(st.nextToken(), StandardCharsets.UTF_8.toString());
+      link.setWorkflow(
+          Link.Workflow.newBuilder()
+              .setNamespace(namespace)
+              .setWorkflowId(workflowID)
+              .setRunId(runID));
+    } catch (Exception e) {
+      log.error("Failed to parse Nexus link URL", e);
+      return null;
+    }
+    return link.build();
+  }
+
   /**
    * Dispatches on the oneof variant of {@code commonLink} and converts to the matching {@link
    * io.temporal.api.nexus.v1.Link}. Returns {@code null} if no variant is set or encoding fails.
@@ -176,6 +233,10 @@ public class LinkConverter {
     }
     if (commonLink.hasNexusOperation()) {
       return nexusOperationToNexusLink(commonLink.getNexusOperation());
+    }
+    // check for workflow link only if a more specific variant is not set
+    if (commonLink.hasWorkflow()) {
+      return workflowLinkToNexusLink(commonLink.getWorkflow());
     }
     return null;
   }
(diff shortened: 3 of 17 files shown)


temporal-sdk/src/main/java/io/temporal/internal/common/LinkConverter.java
```diff
@@ -177,6 +234,10 @@ public static io.temporal.api.nexus.v1.Link linkToNexusLink(Link commonLink) {
     if (commonLink.hasNexusOperation()) {
       return nexusOperationToNexusLink(commonLink.getNexusOperation());
     }
+    // check for workflow link only if a more specific variant is not set
```
Quinn-With-Two-Ns: Nit: I don't think this comment makes sense, we are just checking variants here and there can only be one

temporal-sdk/src/test/java/io/temporal/workflow/nexus/UpdateWorkflowOperationTest.java
```diff
@@ -0,0 +1,274 @@
...
+import io.temporal.workflow.shared.TestWorkflows;
+import java.time.Duration;
+import java.util.UUID;
+import org.junit.AfterClass;
+import org.junit.Assert;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Test;
+
+public class UpdateWorkflowOperationTest extends BaseNexusTest {
```
Quinn-With-Two-Ns, line 32: Shouldn't need to inherit from `BaseNexusTest`, we should probably mark it as deprecated or remove it. (not saying do that in this PR)

temporal-sdk/src/main/java/io/temporal/nexus/TemporalNexusClientImpl.java
```diff
@@ -265,4 +268,269 @@ private <R> TemporalOperationResult<R> invokeAndReturn(WorkflowHandle<R> handle)
...
+      UpdateOptions<R> options, UpdateCommand<R> updateWrapper) throws OperationException {
+
+    UpdateOptions.Builder<R> effectiveOptsBuilder = UpdateOptions.newBuilder(options);
+    String requestId = operationStartDetails.getRequestId();
+    if (Strings.isNullOrEmpty(options.getUpdateId())) {
+      // if updateId is unset, use requestId - consistent with other SDKs
+      effectiveOptsBuilder.setUpdateId(requestId);
+    }
+    if (options.getWaitForStage() == null) {
+      effectiveOptsBuilder.setWaitForStage(WorkflowUpdateStage.ACCEPTED);
```
Quinn-With-Two-Ns: Oh `WaitForStage` should be explicitly set, that is the clients behaviour so we should try to match that here

temporal-sdk/src/test/java/io/temporal/workflow/nexus/UpdateWorkflowOperationTest.java
```diff
@@ -0,0 +1,274 @@
...
+                    .getWorkflowClient()
+                    .newWorkflowStub(HandlerWorkflow.class, input.getTargetWorkflowId());
+            UpdateOptions.Builder<String> optionsBuilder =
+                UpdateOptions.newBuilder(String.class)
+                    .setUpdateName("setValue")
+                    .setWaitForStage(WorkflowUpdateStage.ACCEPTED);
+            if (input.getUpdateId() != null) {
+              optionsBuilder.setUpdateId(input.getUpdateId());
+            }
+            return client.startWorkflowUpdate(
```
Quinn-With-Two-Ns, line 235: The Nexus client wasn't intended to take the stubs. The proposed API was to take the unbound method reference like all new Java SDK APIs like
``` java
  public <T, A1, R> TemporalOperationResult<R> updateWorkflow(
      String workflowId,
      Class<T> workflowClass,
       unctions.Func2<T, A1, R>> updateMethod,
      A1 arg);
```
We should discuss if the proposed API does not work

temporal-sdk/src/main/java/io/temporal/internal/nexus/OperationToken.java
```diff
@@ -18,22 +18,49 @@ public class OperationToken {
...
+  @JsonInclude(JsonInclude.Include.NON_NULL)
+  // only set for updates and activities
+  private final String runId;
+
+  @JsonProperty("uid")
+  @JsonInclude(JsonInclude.Include.NON_NULL)
+  // only set for updates
+  private final String updateId;
+
   public OperationToken(
```
Quinn-With-Two-Ns, line 31: Could consider splitting the `OperationToken` into different Java types for the different `t`. Should be possible using `jackson`. Although I would be fine defering this until we have activities here as well and then doing the refactor then

temporal-sdk/src/main/java/io/temporal/nexus/TemporalNexusClientImpl.java
```diff
@@ -265,4 +268,269 @@ private <R> TemporalOperationResult<R> invokeAndReturn(WorkflowHandle<R> handle)
...
+        new NexusOperationMetadata(
+            requestId, callbackUrl, operationStartDetails.getCallbackHeaders());
+    // set the nexusOperationMetadata and capture operationCompleted
+    nexusContext.setNexusOperationMetadata(nexusOperationMetadata);
+    try {
+      WorkflowUpdateHandle<R> handle = updateWrapper.triggerUpdate(options);
+      if (nexusOperationMetadata.operationCompleted) {
+        try {
+          R value = handle.getResult();
+          return TemporalOperationResult.sync(value);
```
Evanthx, line 559: This vranch doesn't call asyncOperationStarted.set(false) - does it need to? It looks like it is left set to true.

temporal-sdk/src/main/java/io/temporal/nexus/CancelUpdateWorkflowExecutionInput.java
```diff
@@ -0,0 +1,40 @@
...
+import io.temporal.common.Experimental;
+import java.util.Objects;
+import javax.annotation.Nullable;
+
+/**
+ * Input to {@link TemporalOperationHandler#cancelUpdateWorkflow} describing the workflow update to
+ * cancel.
+ */
+@Experimental
+public final class CancelUpdateWorkflowExecutionInput {
```
Quinn-With-Two-Ns:
```suggestion
public final class CancelUpdateWorkflowInput {
```
I thought we would call it `CancelUpdateWorkflowInput `,open to other ideas though, regardless we should make sure we are consistent across all the SDKs and docs here.

temporal-sdk/src/main/java/io/temporal/nexus/CancelUpdateWorkflowInput.java
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  /** Returns the workflow ID extracted from the operation token. */
+  public String getWorkflowId() {
+    return workflowId;
+  }
+
+  /** Returns the run ID extracted from the operation token, or null if not present. */
+  @Nullable
+  public String getRunId() {
```
Quinn-With-Two-Ns, line 29: I think for `CancelUpdateWorkflowInput ` this is always non null

temporal-sdk/src/main/java/io/temporal/nexus/TemporalNexusClientImpl.java
```diff
@@ -265,4 +268,269 @@ private <R> TemporalOperationResult<R> invokeAndReturn(WorkflowHandle<R> handle)
...
+  }
+
+  /**
+   * @throws HandlerException if the options provided are invalid like missing
+   *     UpdateName/WorkflowID/etc
+   */
+  private <R> void checkNexusUpdateOptionsValid(UpdateOptions<R> options) {
+    if (options.getWaitForStage() != WorkflowUpdateStage.ACCEPTED) {
+      throw new HandlerException(
+          HandlerException.ErrorType.BAD_REQUEST,
```
Evanthx: Go returns this as a nexus.OperationError - should this be consistent?

temporal-sdk/src/main/java/io/temporal/internal/common/LinkConverter.java
```diff
@@ -93,6 +95,24 @@ public static io.temporal.api.nexus.v1.Link workflowEventToNexusLink(Link.Workfl
...
   }
 
+  public static io.temporal.api.nexus.v1.Link workflowLinkToNexusLink(Link.Workflow w) {
+    try {
+      String namespace = URLEncoder.encode(w.getNamespace(), StandardCharsets.UTF_8.toString());
+      String workflowId =
+          URLEncoder.encode(w.getWorkflowId(), StandardCharsets.UTF_8.toString())
+              .replace("+", "%20"); // handle workflowIds supporting spaces
+      String runId = URLEncoder.encode(w.getRunId(), StandardCharsets.UTF_8.toString());
+      String url = String.format(linkPathFormat, namespace, workflowId, runId);
```
Evanthx, line 105: String.format can throw IllegalFormatException which is uncaught - should you just catch Exception instead of limiting it?

temporal-sdk/src/main/java/io/temporal/internal/common/LinkConverter.java
```diff
@@ -166,6 +186,43 @@ public static Link nexusLinkToWorkflowEvent(io.temporal.api.nexus.v1.Link nexusL
...
+        return null;
+      }
+      String runID = URLDecoder.decode(st.nextToken(), StandardCharsets.UTF_8.toString());
+      link.setWorkflow(
+          Link.Workflow.newBuilder()
+              .setNamespace(namespace)
+              .setWorkflowId(workflowID)
+              .setRunId(runID));
+    } catch (Exception e) {
+      log.error("Failed to parse Nexus link URL", e);
```
Evanthx: Might still want to log the bad URL, like you did above

Evanthx (changes requested):

temporal-sdk/src/main/java/io/temporal/internal/common/LinkConverter.java
mavemuri: fixed - misread it as hasWorkflowID initially

temporal-sdk/src/test/java/io/temporal/workflow/nexus/UpdateWorkflowOperationTest.java
mavemuri: Any reason for removing it? looks useful as a utility

temporal-sdk/src/main/java/io/temporal/nexus/TemporalNexusClientImpl.java
mavemuri: Ah ok wanted to see if we could make it easier with a default callout - stricter makes sense, will change

temporal-sdk/src/main/java/io/temporal/internal/nexus/OperationToken.java
mavemuri: Thought about it with a base class that does common encode/decodes but left it as overloads give hints and the remaining is mostly similar 
But agree on maybe changing after activities- if changing here, will track in other SDKs too for consistency

temporal-sdk/src/main/java/io/temporal/nexus/CancelUpdateWorkflowExecutionInput.java
mavemuri: makes sense, fixed

temporal-sdk/src/main/java/io/temporal/nexus/CancelUpdateWorkflowInput.java
mavemuri: There are 2 cases - 
1. Right now, we are generating it based on the input - if user provides no runID, there wont be one in the token 
Not sure if we need to grab it from a handle either - intent of user could be that they want to just run an update on current workflow run - but for cancel its not clear if the cancel should also "preserve" that original intent - i thought it should 
eg. run update- without specifying runID- while workflow is on runID1, get token. Now, when workflow is on runID2, still want to cancel the update => the current token will work 
OTOH, if they specifically had a runID, then it wont work but that is stricter and is fully consistent with what they need 
This is consistent across the go/python/java SDK PRs 
But we can discuss if this preserving original intent means getting runID via handle if empty- or if thats decided already 
2. user just wants to cancel it, crafts the encoded token themselves without runID to cancel current one.  Required for above case as well if token was initially minted with a specific runID 

changed it to empty string default but i think its still same functionally

temporal-sdk/src/main/java/io/temporal/nexus/TemporalNexusClientImpl.java
mavemuri: The asyncOperationStarted is more of a guard against multiple invocations using the same ephemeral nexusClient. Whether it completed sync or async is the result and isnt tied to how the req was made - because in either case, request itself was always an asyncReq 
Only case where it resets to false is if the op can be retried safely- due to real issues like op failing due to invalid params etc

temporal-sdk/src/main/java/io/temporal/nexus/TemporalNexusClientImpl.java
mavemuri: agree, fixed

temporal-sdk/src/test/java/io/temporal/workflow/nexus/UpdateWorkflowOperationTest.java
Quinn-With-Two-Ns: All the utility is already in the SDK rule

temporal-sdk/src/main/java/io/temporal/nexus/CancelUpdateWorkflowInput.java
Quinn-With-Two-Ns: We should not be generating based on the input, we should generate based on the **output** of the update response. Updates do not move between runs so please use the run ID of the update.

Quinn-With-Two-Ns: Please also make sure you fix this in Go as well this needs to encode where the workflow update actually is so this is blocker for the Go SDK as well

mavemuri: > Updates do not move between runs 

Missed this, thanks for pointing out. So a workflows history will look different if its re-run even with the same inputs? Im not 100% clear on the implications there- eg, shopping cart was updated to different quantity etc and bill generated based on it(not paid, just bill generated in workflow)- whats the guard in that case? 
Will address in all the SDKs

Quinn-With-Two-Ns: Sorry by Run here I mean in the Temporal definition of a run https://docs.temporal.io/workflow-execution/workflowid-runid, let me know if you want to have a quick call about this

temporal-sdk/src/test/java/io/temporal/workflow/nexus/UpdateWorkflowOperationTest.java
```diff
@@ -0,0 +1,274 @@
...
+        "immediate-value",
+        second.execute(targetWorkflowId + "|immediate-value|" + fixedUpdateId + "|dedup"));
+  }
+
+  @Test
+  public void asyncUpdateWorkflowOperationCompletes() {
+    // only run this test if the server supports update completion callbacks
+    // currently only for local server manual checks until the in-memory test
+    // server can support update completion callbacks
+    assumeTrue(
```
Quinn-With-Two-Ns, line 114: We should just skip the Update tests for the in-memory test server we should always be running these tests when using a real server for test. You can look at how we tested other things like Stand Alone Activities

Quinn-With-Two-Ns: So we won't need `UPDATE_CALLBACKS_SUPPORTED`

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
```diff
@@ -495,14 +513,49 @@ private <R> UpdateWorkflowExecutionRequest toUpdateWorkflowExecutionRequest(
...
+            .setRequestId(nexusOperationMetadata.requestId)
+            .addCompletionCallbacks(
+                InternalUtils.buildNexusCallback(
+                    nexusOperationMetadata.callbackHeaders,
+                    nexusOperationMetadata.callbackUrl,
+                    nexusOperationMetadata.operationToken,
+                    requestLinks))
+            .addAllLinks(requestLinks);
+      } else {
+        log.error("unexpected error fetching nexusOperationMetadata");
```
Quinn-With-Two-Ns: This is an expected case if the user makes a regular update call from the regular Temporal client inside a Nexus handler, we need to make sure that still works like before and we didn't break anything there

temporal-sdk/src/test/java/io/temporal/workflow/nexus/UpdateWorkflowOperationTest.java
mavemuri: `SDKTestWorkflowRule.useExternalService` this wasnt enough as the test also needs update callbacks to be enabled via dynamic config 
Didnt want to rely on just local server flag because someone could run server without those flags - they arent required anyway 
Added this way to ensure that if someone runs this against their own server- for testing other SDK changes etc- itll only get tested iff the server supports it 
Or can add a note if someone runs into this that theyd need to enable flags on their real server

Quinn-With-Two-Ns: Yes you should assume the users server has this DC set or fail the test

Quinn-With-Two-Ns: To run SDK test against a local server you have to set the require DC, that is true is all SDKs CI

temporal-sdk/src/main/java/io/temporal/nexus/TemporalNexusClientImpl.java
```diff
@@ -265,4 +271,350 @@ private <R> TemporalOperationResult<R> invokeAndReturn(WorkflowHandle<R> handle)
...
+        options,
+        effective ->
+            WorkflowClient.startUpdate(
+                () -> updateMethod.apply(stub, arg1, arg2, arg3), effective));
+  }
+
+  @Override
+  public <T, A1, A2, A3, A4> TemporalOperationResult<Void> startWorkflowUpdate(
+      Class<T> workflowClass,
+      String workflowId,
```
Quinn-With-Two-Ns, line 465: I think we need overloads that take the workflow execution as well to target the specific run ID , @maciejdudko thoughts?

temporal-sdk/src/main/java/io/temporal/nexus/TemporalNexusClientImpl.java
```diff
@@ -265,4 +271,350 @@ private <R> TemporalOperationResult<R> invokeAndReturn(WorkflowHandle<R> handle)
...
+  }
+
+  /**
+   * @throws OperationException if the options provided are invalid like missing
+   *     UpdateName/WorkflowID/etc
+   */
+  private <R> void checkNexusUpdateOptionsValid(UpdateOptions<R> options)
+      throws OperationException {
+    if (options.getWaitForStage() != WorkflowUpdateStage.ACCEPTED) {
+      throw OperationException.failed(
```
Quinn-With-Two-Ns: I don't think this should fail the Nexus operation, I thought we already decided that in the Go SDK PR

temporal-sdk/src/main/java/io/temporal/nexus/CancelUpdateWorkflowInput.java
```diff
@@ -0,0 +1,38 @@
...
+@Experimental
+public final class CancelUpdateWorkflowInput {
+
+  private final String workflowId;
+  private final String runId;
+  private final String updateId;
+
+  public CancelUpdateWorkflowInput(String workflowId, String runId, String updateId) {
+    this.workflowId = Objects.requireNonNull(workflowId);
+    this.runId = Strings.nullToEmpty(runId);
```
Quinn-With-Two-Ns: Note still need to update this to require

temporal-sdk/src/test/java/io/temporal/workflow/nexus/UpdateWorkflowOperationTest.java
mavemuri: Got it - will change and just add a comment

temporal-sdk/src/main/java/io/temporal/nexus/TemporalNexusClientImpl.java
mavemuri: updated to handler error to retry

temporal-sdk/src/main/java/io/temporal/nexus/TemporalNexusClientImpl.java
mavemuri: added one just to align on the shape here https://github.com/temporalio/sdk-java/pull/2945/changes#diff-5f53049561675309c6d8a1a6b09638b2e6e9096a5b614047e2d96d4491ed5a73R776 

Just so we're on the same page, if we want to do this, there will be 15 more(additional, not replacement) that will have workflow execution instead of the workflowID right?

Quinn-With-Two-Ns: Yeah it is a lot of overloads , but we need to be able to target by workflow ID or (Workflow ID, Run ID) pair

Quinn-With-Two-Ns (approved):

temporal-sdk/src/main/java/io/temporal/nexus/TemporalNexusClientImpl.java
Evanthx: Might have gotten unfixed. :-D It's still BAD_REQUEST!

temporal-sdk/src/main/java/io/temporal/internal/common/LinkConverter.java
Evanthx: Same as before, might have gotten unfixed!

Evanthx (changes requested):

mavemuri:
> Might have gotten unfixed. :-D It's still BAD_REQUEST!

This was an intentional change - the idea was that we keep retrying and the user will deploy a new version of the handler that will fix it. Still up for discussion before it goes GA but for the preview, this is consistent across go/python/java
https://github.com/temporalio/sdk-java/pull/2945#discussion_r3590665145

> Same as before, might have gotten unfixed!

This was more just to be in-line with the existing convention - dont fail if unparsable  https://github.com/temporalio/sdk-java/blob/6d09a341c14daeddf9b3795cd43721de770fec2c/temporal-sdk/src/main/java/io/temporal/internal/common/LinkConverter.java#L162 
Should we fix for all? 

Can you also please comment on this thread https://github.com/temporalio/sdk-java/pull/2945#discussion_r3590660858 - unsure about adding 14/15 more overloads but there might be a valid need for it - targeting by runID  
@Evanthx

Quinn-With-Two-Ns:
`BAD_REQUEST ` is not retried by default

mavemuri:
> BAD_REQUEST  is not retried by default

Yes, but was also setting the retry explicitly for that reason- iiuc, this will ensure it will retry 
https://github.com/temporalio/sdk-java/pull/2945/changes#diff-b859b54dda09bacf7f8d2fddcb6c9f6727315ceb62f22d472a5cc889bd457507R607

This case is bit odd because it is a bad request- but is retriable so this seemed like the right way to do it

Quinn-With-Two-Ns:
I'd probably capture the validation failure as `INTERNAL`, but if we did `BAD_REQUEST` in other SDKs then I am fine with it for now

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
```diff
@@ -495,14 +513,47 @@ private <R> UpdateWorkflowExecutionRequest toUpdateWorkflowExecutionRequest(
...
+        try {
+          nexusOperationMetadata.operationToken =
+              OperationTokenUtil.generateWorkflowUpdateOperationToken(
+                  clientOptions.getNamespace(),
+                  input.getWorkflowExecution().getWorkflowId(),
+                  input.getWorkflowExecution().getRunId(),
+                  input.getUpdateId());
+        } catch (JsonProcessingException e) {
+          throw new IllegalStateException("failed to generate update operation token", e);
+        }
```
maciejdudko: 1. Under what circumstances can `JsonProcessingException` happen? Is it user error or server/implementation error? If it's user error then we ought to have better message.

2. This doesn't catch unchecked exceptions. `catch (Exception e)` would be a better choice.

temporal-sdk/src/main/java/io/temporal/internal/nexus/OperationTokenType.java
```diff
@@ -5,7 +5,8 @@
 
 public enum OperationTokenType {
   UNKNOWN(0),
-  WORKFLOW_RUN(1);
+  WORKFLOW_RUN(1),
+  WORKFLOW_UPDATE(3);
```
maciejdudko: Value 2 is missing. If it's intentional, a comment would be helpful.

temporal-sdk/src/main/java/io/temporal/nexus/TemporalNexusClient.java
```diff
@@ -507,4 +510,482 @@ <R> TemporalOperationResult<R> startWorkflow(
...
+      A2 arg2,
+      A3 arg3,
+      A4 arg4,
+      A5 arg5,
+      A6 arg6,
+      UpdateOptions<R> options)
+      throws OperationException;
+
+  /**
+   * TBD: determine if all the other 14 func+proc+void are required and add if needed Starts a
```
maciejdudko: 1. Change TBD to TODO.
2. TBD/TODO should not be part of the docstring. Make it a regular comment before docstring.

temporal-sdk/src/main/java/io/temporal/nexus/TemporalNexusClient.java
```diff
@@ -507,4 +510,482 @@ <R> TemporalOperationResult<R> startWorkflow(
...
+   * six-argument workflow update on the provided workflow execution. See {@link
+   * #startWorkflowUpdate(Class, String, Functions.Func1, UpdateOptions)} for the full behavior
+   * contract.
+   *
+   * @param <T>
```
maciejdudko: Doc tag contents are missing for this overload.

maciejdudko (changes requested):
Just a few small things, then it should be ready to merge.

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
mavemuri: It should only be a server error unless the user is trying to manually craft a new token- but that is not the normal case

maciejdudko (approved):

temporal-sdk/src/main/java/io/temporal/nexus/TemporalNexusClientImpl.java
```diff
@@ -265,4 +274,605 @@ private <R> TemporalOperationResult<R> invokeAndReturn(WorkflowHandle<R> handle)
...
+   */
+  private <R> void checkNexusUpdateOptionsValid(UpdateOptions<R> options)
+      throws OperationException {
+    if (options.getWaitForStage() != WorkflowUpdateStage.ACCEPTED) {
+      throw new HandlerException(
+          HandlerException.ErrorType.INTERNAL,
+          "invalid update request",
+          new IllegalArgumentException(
+              "nexus op workflow updates only support WorkflowUpdateStageAccepted for async updates"),
+          RetryBehavior.RETRYABLE);
```
Evanthx, line 607: If this fails, it will always fail, won't it? So the error shouldn't be RETRYABLE

temporal-sdk/src/main/java/io/temporal/internal/common/InternalUtils.java
```diff
@@ -157,6 +157,31 @@ public static boolean isWorkflowStreamReservedName(String name) {
     return name.startsWith(WORKFLOW_STREAM_RESERVED_PREFIX);
   }
 
+  /** Helper to build a Nexus Callback from the provided input. */
+  public static Callback buildNexusCallback(
```
Evanthx, line 161: Lines 105 to 136 in this same file construct a Nexus callback. Can that be refactored to use this method so we don't have duplicate logic that might go out of sync later?

Evanthx (commented):
A few small things ...

temporal-sdk/src/main/java/io/temporal/nexus/TemporalNexusClientImpl.java
mavemuri: This was the one discussed here - https://github.com/temporalio/sdk-java/pull/2945#issuecomment-5048855283
The idea was that a new version of the handler could be deployed which fixes the wid and it would then start passing 
This is consistent with the go/python SDKs https://github.com/temporalio/sdk-go/pull/2417#discussion_r3516221714

temporal-sdk/src/main/java/io/temporal/internal/common/InternalUtils.java
mavemuri: Attempted the refactor but the [existing code](https://github.com/temporalio/sdk-java/blob/2f6dcac6d2da32152bcc48777d13b7ac25163f78/temporal-sdk/src/main/java/io/temporal/internal/common/InternalUtils.java#L119) refers to a deprecated field and adding deprecated fields in new code is causing CI to [fail](https://github.com/temporalio/sdk-java/actions/runs/30594825192/job/91044809401?pr=2945) 

Im hesitant to risk removing that field completely in this pr - it shouldnt be *required anymore but would like to hold off in case some edge case still needs it

Evanthx (approved):
LGTM

PR #2953, opened 2026-07-15
Auto-enroll pollers into autoscaling on PollerAutoscalingAutoEnroll (Java SDK)

veeral-patel:
Mirrors the Go SDK ([temporalio/sdk-go#2442](https://github.com/temporalio/sdk-go/pull/2442)).

## What

When a namespace advertises the `PollerAutoscalingAutoEnroll` capability, workers automatically use **poller autoscaling** for any poller type left at its default — i.e. where the user set neither `MaxConcurrent<Type>TaskPollers` nor a `<Type>TaskPollerBehavior`. Explicitly configured pollers are left alone. Applies to workflow, activity, and nexus pollers.

## How

- `Worker` records, per poller type, whether it was left at its default (read from the raw `WorkerOptions`, before defaults are applied).
- At `start()` — once namespace capabilities are known — each eligible poller's behavior is swapped to autoscaling before its poller is created.
- Auto-enroll also implies scale-down support, so `NamespaceCapabilities` treats it as enabling `pollerAutoscaling` too.

## Proto

Bumps the `temporal/api` submodule to pick up the `poller_autoscaling_auto_enroll` capability (api #803).

## Testing

- `PollerAutoscalingAutoEnrollTest` — the enroll decision (default + capability → autoscaling; capability off, explicit count, or explicit behavior → unchanged).
- `WorkerPollerAutoEnrollEligibilityTest` — `Worker` marks defaults eligible and explicit config ineligible, per type.
- Existing poller/worker tests still pass.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/worker/NamespaceCapabilities.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/worker/NamespaceCapabilities.java
@@ -19,8 +19,10 @@ public final class NamespaceCapabilities {
     if (capabilities.getPollerAutoscalingAutoEnroll()) {
       pollerAutoscalingAutoEnroll.set(true);
     }
-    // Auto-enroll implies full poller autoscaling support, including scaling down, so it also
-    // enables pollerAutoscaling (which drives serverSupportsAutoscaling in PollScaleReportHandle).
+    // The auto-enroll capability implies the server speaks the full poller-scaling protocol, so
+    // also enable pollerAutoscaling: it is the flag AsyncPoller passes to PollScaleReportHandle as
+    // serverSupportsAutoscaling, which lets the poller count follow the server's scaling
+    // suggestions.
     if (capabilities.getPollerAutoscaling() || capabilities.getPollerAutoscalingAutoEnroll()) {
       pollerAutoscaling.set(true);
     }
--- a/temporal-sdk/src/main/java/io/temporal/internal/worker/NamespaceCapabilities.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/worker/NamespaceCapabilities.java
@@ -19,11 +19,7 @@ public final class NamespaceCapabilities {
     if (capabilities.getPollerAutoscalingAutoEnroll()) {
       pollerAutoscalingAutoEnroll.set(true);
     }
-    // The auto-enroll capability implies the server speaks the full poller-scaling protocol, so
-    // also enable pollerAutoscaling: it is the flag AsyncPoller passes to PollScaleReportHandle as
-    // serverSupportsAutoscaling, which lets the poller count follow the server's scaling
-    // suggestions.
-    if (capabilities.getPollerAutoscaling() || capabilities.getPollerAutoscalingAutoEnroll()) {
+    if (capabilities.getPollerAutoscaling()) {
       pollerAutoscaling.set(true);
     }
     if (capabilities.getWorkerPollCompleteOnShutdown()) {
--- a/temporal-sdk/src/main/java/io/temporal/internal/worker/NamespaceCapabilities.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/worker/NamespaceCapabilities.java
@@ -10,12 +10,18 @@ import java.util.concurrent.atomic.AtomicBoolean;
  */
 public final class NamespaceCapabilities {
   private final AtomicBoolean pollerAutoscaling = new AtomicBoolean(false);
+  private final AtomicBoolean pollerAutoscalingAutoEnroll = new AtomicBoolean(false);
   private final AtomicBoolean gracefulPollShutdown = new AtomicBoolean(false);
   private final AtomicBoolean workerHeartbeats = new AtomicBoolean(false);
   private final AtomicBoolean workerCommands = new AtomicBoolean(false);
 
   public void setFromCapabilities(Capabilities capabilities) {
-    if (capabilities.getPollerAutoscaling()) {
+    if (capabilities.getPollerAutoscalingAutoEnroll()) {
+      pollerAutoscalingAutoEnroll.set(true);
+    }
+    // Auto-enroll implies full poller autoscaling support, including scaling down, so it also
+    // enables pollerAutoscaling (which drives serverSupportsAutoscaling in PollScaleReportHandle).
+    if (capabilities.getPollerAutoscaling() || capabilities.getPollerAutoscalingAutoEnroll()) {
       pollerAutoscaling.set(true);
     }
     if (capabilities.getWorkerPollCompleteOnShutdown()) {
@@ -31,6 +37,10 @@ public final class NamespaceCapabilities {
 
   public boolean isPollerAutoscaling() {
     return pollerAutoscaling.get();
+  }
+
+  public boolean isPollerAutoscalingAutoEnroll() {
+    return pollerAutoscalingAutoEnroll.get();
   }
 
   public boolean isGracefulPollShutdown() {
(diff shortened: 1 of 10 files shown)


Sushisource (approved):
Looking good to me!

temporal-sdk/src/main/java/io/temporal/internal/worker/NamespaceCapabilities.java
```diff
@@ -10,12 +10,18 @@
...
   private final AtomicBoolean workerHeartbeats = new AtomicBoolean(false);
   private final AtomicBoolean workerCommands = new AtomicBoolean(false);
 
   public void setFromCapabilities(Capabilities capabilities) {
-    if (capabilities.getPollerAutoscaling()) {
+    if (capabilities.getPollerAutoscalingAutoEnroll()) {
+      pollerAutoscalingAutoEnroll.set(true);
+    }
+    // Auto-enroll implies full poller autoscaling support, including scaling down, so it also
+    // enables pollerAutoscaling (which drives serverSupportsAutoscaling in PollScaleReportHandle).
```
yuandrew: not entirely sure the purpose for this comment, why is "scaling down" relevant here?

temporal-sdk/src/main/java/io/temporal/worker/Worker.java
```diff
@@ -132,6 +132,27 @@ private static final class TaskSnapshot {
     Map<String, String> tags =
         new ImmutableMap.Builder<String, String>(1).put(MetricsTag.TASK_QUEUE, taskQueue).build();
     Scope taggedScope = metricsScope.tagged(tags);
+
+    // Determine, per poller type, whether the user left it at its default (set neither a fixed
+    // poller count nor a poller behavior). Eligible types are auto-enrolled into poller autoscaling
+    // at start() when the namespace advertises the PollerAutoscalingAutoEnroll capability. This
+    // must
+    // be read from the raw (pre-defaulting) options, since validateAndBuildWithDefaults() fills in
+    // a
```
yuandrew: spacing is weird on this comment

temporal-sdk/src/main/java/io/temporal/worker/Worker.java
```diff
@@ -132,6 +132,27 @@ private static final class TaskSnapshot {
...
+    // fixed default poller count that would otherwise look like an explicit choice.
+    boolean workflowTaskAutoEnrollEligible =
+        options == null
+            || (options.getWorkflowTaskPollersBehavior() == null
```
yuandrew: I think we need to preserve a "setter was explicitly called" as private/internal metadata through build, validation, and copy-building, then transfer that into the internal `PollerOptions`. Only `WorkerOptions.Builder` knows whether the user actually called either poller setter.

temporal-sdk/src/test/java/io/temporal/worker/WorkerPollerAutoEnrollEligibilityTest.java
```diff
@@ -0,0 +1,138 @@
+package io.temporal.worker;
```
yuandrew, line 1: Can we add tests showing that pollers remain eligible when default options come from `WorkerOptions.getDefaultInstance()`, `validateAndBuildWithDefaults()`, or a copy of either? And a test that explicitly sets the count to the numerical default of 5 proving its ineligible.

It would also be useful to have one startup-path test that enables the namespace capability, starts workflow/activity/Nexus workers, and verifies their effective poller behavior becomes autoscaling. I think we're missing a full E2E wiring test here.

temporal-sdk/src/main/java/io/temporal/internal/worker/NamespaceCapabilities.java
veeral-patel: Removed the comment, thank you

temporal-sdk/src/main/java/io/temporal/worker/Worker.java
veeral-patel: Fixed this

temporal-sdk/src/main/java/io/temporal/worker/Worker.java
veeral-patel: Good idea, this is done.

I added a private {workflow,activity,nexus}TaskPollersConfigured flag, set to true inside both setters for each type (setMaxConcurrent<Type>TaskPollers and set<Type>TaskPollersBehavior).

temporal-sdk/src/test/java/io/temporal/worker/WorkerPollerAutoEnrollEligibilityTest.java
veeral-patel: > Can we add tests showing that pollers remain eligible when default options come from WorkerOptions.getDefaultInstance(), validateAndBuildWithDefaults(), or a copy of either? And a test that explicitly sets the count to the numerical default of 5 proving its ineligible.

Done, added the following test cases under `WorkerPollerAutoEnrollEligibilityTest`:

| Test case | Summary | Expected |
|---|---|---|
| `defaultInstanceMakesEveryPollerTypeEligible` | `getDefaultInstance()` carries the numeric default count (5), but no setter was called — so intent-tracking keeps it eligible | Eligible (all types) |
| `validateAndBuildWithDefaultsMakesEveryPollerTypeEligible` | Same for `newBuilder().validateAndBuildWithDefaults()`: defaulting fills in 5 without marking the pollers configured | Eligible (all types) |
| `copyOfDefaultInstanceIsEligible` | Copying a default-options object (via both `build()` and `validateAndBuildWithDefaults()`) preserves the "no setter called" state, so the copy stays eligible too | Eligible (all types) |
| `explicitCountEqualToNumericDefaultIsIneligible` | User explicitly calls `setMaxConcurrentWorkflowTaskPollers(5)` — same value as the default, but now the setter was invoked, so it must count as configured | Ineligible (workflow); others eligible |

> It would also be useful to have one startup-path test that enables the namespace capability, starts workflow/activity/Nexus workers, and verifies their effective poller behavior becomes autoscaling. I think we're missing a full E2E wiring test here.

Done, added this test: `WorkerPollerAutoEnrollStartupTest.autoEnrollAtStartupSwitchesPollersToAutoscaling`.

It starts a Worker (with a NamespaceCapabilities advertising the capability) that has workflow, activity, and nexus implementations registered, then asserts all three pollers' effective behavior is PollerBehaviorAutoscaling after start(). This covers the full path: WorkerOptions → eligibility → start-time enrollment → AsyncPoller.

temporal-sdk/src/test/java/io/temporal/worker/WorkerPollerAutoEnrollEligibilityTest.java
```diff
@@ -123,6 +124,53 @@ public void explicitMaxConcurrentPollersOnOneTypeLeavesOthersEligible() {
     assertTrue(nexusEligible(worker));
   }
 
+  @Test
+  public void defaultInstanceMakesEveryPollerTypeEligible() {
+    // getDefaultInstance() carries the numeric default poller count (5), but the user never called
+    // a
```
yuandrew: another weird spacing comment

yuandrew (approved):

veeral-patel:
Tested this change manually successfully.

Setup: Built temporal server from my branch, 1 worker, 2 namespaces. Examined the logs in order to check.

default namespace:
=== server capability PollerAutoscalingAutoEnroll=false
=== worker started …
=== done for namespace=default
No async-poller lines → fixed pollers (MultiThreadedPoller, PollerBehaviorSimpleMaximum).

autoscale-ns namespace:
=== server capability PollerAutoscalingAutoEnroll=true
INFO io.temporal.internal.worker.AsyncPoller - Starting async poller: StickyWorkflowPollTask
INFO io.temporal.internal.worker.AsyncPoller - Starting async poller: NormalWorkflowPollTask
INFO io.temporal.internal.worker.AsyncPoller - Starting async poller: AsyncActivityPollTask
Three async pollers → auto-enrolled into autoscaling for workflow (sticky + normal) and activity.

temporal-sdk/src/test/java/io/temporal/worker/WorkerPollerAutoEnrollEligibilityTest.java
veeral-patel: Fixed spacing

PR #2960, opened 2026-07-22
Include WorkflowType in wrapFailure error message

DanielPBak:
## Summary
- `WorkflowWorker$TaskHandlerImpl.wrapFailure` builds a `RuntimeException` message ("Failure processing workflow task. WorkflowId=..., RunId=..., Attempt=...") when a workflow task fails processing, but omits the workflow type.
- `PollWorkflowTaskQueueResponse` (already available on `task.getResponse()`) carries the workflow type, so this adds `WorkflowType=` to the message using that existing data — no new lookups or API surface required.
- This makes the error immediately actionable: today, debugging one of these failures requires a separate lookup (e.g. `temporal workflow describe`) just to find out which workflow type failed.

## Test plan
- [ ] Existing `WorkflowWorker`/task-handler tests still pass
- [ ] No behavior change other than the added field in the exception message

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
@@ -607,6 +607,8 @@ final class WorkflowWorker implements SuspendableWorker {
               + execution.getWorkflowId()
               + ", RunId="
               + execution.getRunId()
+              + ", WorkflowType="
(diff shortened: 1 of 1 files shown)


Quinn-With-Two-Ns:
I will say, in general, that workflow task failures are generally something that happens to a specific workflow, not a specific workflow type. That being said there are cases like bad code or an NDE where workflow type would be valuable

Quinn-With-Two-Ns (approved):

PR #2963, opened 2026-07-23
Bump jacoco and simplify code coverage CI

### LINKED ISSUE #2962: Code coverage causing commit failures on main
The Code Coverage GitHub Actions workflow is consistently failing and the README coverage badge is stale. Possible causes and resolutions, according to codex:

  ## Possible causes

  - Tests run on Java 23, but JaCoCo is pinned to 0.8.9, which does not support Java 23 class files. Logs report:
```
    Unsupported class file major version 67
    Execution failed for task ':testCodeCoverageReport'
    Error while creating report
```

  - Report generation depends on the module test tasks, so the suite runs twice in separate Gradle invocations. The first run has continue-on-error: true, but failures from the second run are not ignored. Flaky tests can therefore also fail the workflow.

  ## Possible resolutions

  Upgrade JaCoCo to a Java 23-compatible release and avoid running the tests twice.

dplyukhin:
## What was changed
- Bump JaCoCo to 0.8.15 for Java 23 compatibility
- Avoid feeding duplicate classes to jacoco
- Simplify code coverage CI workflow

## Why?
Code coverage is failing in CI

## Checklist

1. Closes #2962 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -25,14 +25,7 @@ jobs:
       - name: Set up Gradle
         uses: gradle/actions/setup-gradle@ac396bf1a80af16236baf54bd7330ae21dc6ece5 # v6
 
-      - name: Run Tests
-        run: ./gradlew test -x spotlessCheck -x spotlessApply -Pjacoco -PtestJavaVersion=23
-        continue-on-error: true
-
-      - name: Run Test Coverage
-        run: ./gradlew testCodeCoverageReport -Pjacoco
-
-      - name: Publish Coverage
+      - name: Run and Publish Test Coverage
         env:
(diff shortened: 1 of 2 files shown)


.github/workflows/coverage.yml
```diff
@@ -3,12 +3,16 @@ on:
...
+  pull_request:
+    branches:
+      - main
 
 permissions:
   contents: read
 
 jobs:
   code-coverage:
+    if: github.event_name == 'push' || github.head_ref == 'fix-temporal-sdk-java-main-code-coverage'
```
dplyukhin: I'll revert this file after we've verified that code coverage is working properly

GregoryTravis (approved):

PR #2976, opened 2026-07-29
feat(extstore): implement core extstore logic that uses payload visitor to batch and store/retrieve payloads.

cconstable:
## What was changed
- Added `internal/payload/storage/ExternalStorageMessageConverter.java` to transform payloads in messages.
- Added `internal/payload/storage/ExternalStoragePayloadConverter.java` to handle the actual transformation.
- Refactored to use new `TaskScope` for handling parent-child and sibling task cancellation.

## Why?
- Unblocks extstore integration

## Checklist
- Added new tests


temporal-sdk/src/main/java/io/temporal/internal/payload/storage/ExternalStoragePayloadConverter.java
```diff
@@ -0,0 +1,282 @@
...
+                + "'");
+      }
+      batches.computeIfAbsent(driver.getName(), name -> new Batch<>(driver)).add(i, payload);
+    }
+    return batches;
+  }
+
+  private CompletableFuture<List<IndexedValue<Payload>>> runStoreDrivers(
+      Map<String, Batch<Payload>> batches, @Nullable StorageDriverTargetInfo target) {
+    return TaskScope.withScope(
```
cconstable: The previous logic here needed to manually account for external cancellation and handle sibling cancellation. The `TaskScope` handles that automatically. The `scope.awaitAll()` at the end has fail-fast semantics so it will immediately cancel other siblings on first error.

temporal-sdk/src/main/java/io/temporal/internal/payload/storage/ExternalStorageMessageConverter.java
```diff
@@ -0,0 +1,48 @@
...
+
+  <T extends Message> CompletableFuture<T> retrieve(T message) {
+    PayloadVisitorOptions<Object> options =
+        PayloadVisitorOptions.newBuilder(
+                (PayloadVisitor<Object>) (context, payloads) -> payloadConverter.retrieve(payloads))
+            .setConcurrency(payloadVisitConcurrency)
+            .setSkipSearchAttributes(true)
+            .build();
+    return PayloadVisitors.visit(message, options);
+  }
```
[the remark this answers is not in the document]
cconstable: This doesn't appear to be valid feedback.

temporal-sdk/src/main/java/io/temporal/internal/payload/storage/ExternalStorageReferences.java
```diff
@@ -0,0 +1,83 @@
...
+
+    ParsedReference(String driverName, StorageDriverClaim claim) {
+      this.driverName = driverName;
+      this.claim = claim;
+    }
+  }
+
+  static boolean isReference(@Nonnull Payload payload) {
+    return payload.getExternalPayloadsCount() > 0;
+  }
```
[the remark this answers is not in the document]
cconstable: I think the correct thing to do it replace isReference() + fromReferencePayload() with a single  `tryParseReference(Payload)` that checks both `external_payloads > 0` and `messageType == temporal.api.sdk.v1.ExternalStorageReference` and returns null instead of throwing. Will update this soon.

cconstable: Going to replace this. See thread in conversation tab.

temporal-sdk/src/main/java/io/temporal/internal/payload/storage/ExternalStorageReferences.java
```diff
@@ -0,0 +1,83 @@
...
+  static ParsedReference fromReferencePayload(@Nonnull Payload payload) {
+    ByteString messageType = payload.getMetadataMap().get(EncodingKeys.METADATA_MESSAGE_TYPE_KEY);
+    if (messageType == null || !REFERENCE_MESSAGE_TYPE.equals(messageType.toStringUtf8())) {
+      throw new IllegalArgumentException(
+          "Payload is not an external storage reference; expected messageType '"
+              + REFERENCE_MESSAGE_TYPE
+              + "' but was '"
+              + (messageType == null ? "" : messageType.toStringUtf8())
+              + "'");
+    }
```
cconstable: Going to replace this. See thread in conversations tab.

temporal-sdk/src/main/java/io/temporal/internal/payload/storage/ExternalStoragePayloadConverter.java
```diff
@@ -0,0 +1,282 @@
...
+import java.util.Map;
+import java.util.concurrent.CancellationException;
+import java.util.concurrent.CompletableFuture;
+import javax.annotation.Nullable;
+
+/**
+ * Converts one payload list between inline payloads and external-storage references by routing
+ * entries to storage drivers.
+ */
+final class ExternalStoragePayloadConverter {
```
jmaeagle99: I might call this something like `ExternalStoragePayloadTransformer` or something to that effect. Avoid `PayloadConverter` since that is an actual concept already that is tangentially related.

temporal-sdk/src/main/java/io/temporal/internal/payload/storage/ExternalStorageReferences.java
```diff
@@ -0,0 +1,82 @@
...
+                .build())
+        .build();
+  }
+
+  /**
+   * Returns the reference encoded in {@code payload}, or null if the payload is not an external
+   * storage reference this SDK understands.
+   */
+  static @Nullable ParsedReference tryParseReference(@Nonnull Payload payload) {
+    if (payload.getExternalPayloadsCount() == 0) {
```
jmaeagle99: Remove this check. This is not a requirement for exchanging a reference for the externally stored payload.

temporal-sdk/src/main/java/io/temporal/internal/payload/storage/StorageDriverStoreContextImpl.java
```diff
@@ -0,0 +1,33 @@
...
+
+import io.temporal.common.CancellationToken;
+import io.temporal.payload.storage.StorageDriverStoreContext;
+import io.temporal.payload.storage.StorageDriverTargetInfo;
+import java.util.Objects;
+import java.util.concurrent.CancellationException;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+final class StorageDriverStoreContextImpl implements StorageDriverStoreContext {
```
jmaeagle99, line 11: Any particular reason this one is in its own separate file but the `StorageDriverRetrieveContextImpl` is a nested static class of `ExternalStoragePayloadConverter`?

temporal-sdk/src/main/java/io/temporal/payload/storage/StorageDriver.java
```diff
@@ -29,6 +29,9 @@ public interface StorageDriver {
   /**
    * Stores {@code payloads} and returns one {@link StorageDriverClaim} per payload, in the same
    * order. The returned list must be the same length as {@code payloads}.
+   *
+   * <p>Drivers should use {@link StorageDriverStoreContext#getCancellationToken()} to abort
```
jmaeagle99: This comment reads as "I should call getCancellationToken to abort in-flight requests". I think it should be more like "call getCancellationToken() to get a token to observe when cancellation is requested". Could use better wording.

temporal-sdk/src/main/java/io/temporal/payload/storage/StorageDriverRetrieveContext.java
```diff
@@ -9,4 +12,8 @@
  * production code, only when constructing instances for their own tests.
  */
 @Experimental
-public interface StorageDriverRetrieveContext {}
+public interface StorageDriverRetrieveContext {
```
jmaeagle99, line 16: I looked over the other SDKs and this is the first one where a user would have to implement an interface (that isn't structural) in their testing. And I totally acknowledge that I suggested it be as such. However, that means if we ever needed to add something to the interface after GA, it would technically be a breaking change. Wondering if we should document it as such that we are free to add more to it over time or if we should convert it to a class that allows users to instantiate for testing. Maybe not too much of an issue for Java since it has default interface implementations. Just something to think through for maintenance purposes.

temporal-sdk/src/main/java/io/temporal/internal/payload/storage/ExternalStoragePayloadConverter.java
```diff
@@ -0,0 +1,282 @@
...
+      int payloadSizeThreshold) {
+    this.driversByName = driversByName;
+    this.selector = selector;
+    this.payloadSizeThreshold = payloadSizeThreshold;
+  }
+
+  CompletableFuture<List<Payload>> store(
+      @Nullable StorageDriverTargetInfo target, List<Payload> payloads) {
+    StorageDriverStoreContext context =
+        new StorageDriverStoreContextImpl(target, CancellationToken.none());
```
jmaeagle99: I think cancellation token should be plumbed through the store method and used here. Or are we not doing that because there are no callers that can offer cancellation? Might still do that so the API has the correct shape and then have the caller pass `CancellationToken.none()`.

temporal-sdk/src/main/java/io/temporal/internal/payload/storage/ExternalStorageMessageConverter.java
```diff
@@ -0,0 +1,48 @@
...
+  private final ExternalStoragePayloadConverter payloadConverter;
+  private final int payloadVisitConcurrency;
+
+  ExternalStorageMessageConverter(
+      ExternalStoragePayloadConverter payloadConverter, int payloadVisitConcurrency) {
+    this.payloadConverter = payloadConverter;
+    this.payloadVisitConcurrency = payloadVisitConcurrency;
+  }
+
+  <T extends Message> CompletableFuture<T> store(
```
jmaeagle99: Any possible efficiency in allowing support of Message.Builder as well? Might avoid a bunch of copying.

temporal-sdk/src/main/java/io/temporal/internal/payload/storage/ExternalStorageReferences.java
```diff
@@ -0,0 +1,82 @@
...
+      return null;
+    }
+    ByteString messageType = payload.getMetadataMap().get(EncodingKeys.METADATA_MESSAGE_TYPE_KEY);
```
jmaeagle99: Should we also be checking encoding == ENCODING_PROTOBUF_JSON?

temporal-sdk/src/main/java/io/temporal/internal/payload/storage/ExternalStoragePayloadConverter.java
jmaeagle99: I would imagine that at least worker shutdown has something to plumb into, if not some kind of cancellation per workflow or task.

temporal-sdk/src/main/java/io/temporal/internal/payload/storage/ExternalStorageReferences.java
cconstable: swapping back to the original implementation

temporal-sdk/src/main/java/io/temporal/internal/payload/storage/StorageDriverStoreContextImpl.java
cconstable: No, but I'll make them consistent

jmaeagle99 (approved):

PR #2978, opened 2026-07-29
exstore/pipeline integration

cconstable:
Client + Nexus + Worker external storage integration

## What was changed

- enabled extstore via `WorkflowClientOptions.setExternalStorage(...)`, `worker/WorkerFactoryOptions`, `internal/worker/SingleWorkerOptions`,  and `worker/Worker`
- client/worker with no external storage configured fails with `TMPRL-1105`
- Client: integration via new decorator `ExternalStorageGenericWorkflowClient` that is used in `client/WorkflowClientInternalImpl`
- Workers: integration via `internal/worker/{WorkflowWorker,ActivityWorker,NexusWorker}` and `internal/replay/{ServiceWorkflowHistoryIterator,ReplayWorkflowTaskHandler}`
- Activity completion + heartbeat integration.

## Why?
- makes extstore "work" in client, worker, and nexus pipelines

## Checklist
- Added new tests

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/client/WorkflowClientInternalImpl.java
+++ b/temporal-sdk/src/main/java/io/temporal/client/WorkflowClientInternalImpl.java
@@ -18,12 +18,15 @@ import io.temporal.common.interceptors.WorkflowClientInterceptor;
 import io.temporal.internal.WorkflowThreadMarker;
 import io.temporal.internal.client.*;
 import io.temporal.internal.client.NexusStartWorkflowResponse;
+import io.temporal.internal.client.external.ExternalStorageGenericWorkflowClient;
 import io.temporal.internal.client.external.GenericWorkflowClient;
 import io.temporal.internal.client.external.GenericWorkflowClientImpl;
 import io.temporal.internal.client.external.ManualActivityCompletionClientFactory;
 import io.temporal.internal.common.PluginUtils;
+import io.temporal.internal.payload.storage.ExternalStorageMessageConverter;
 import io.temporal.internal.sync.StubMarker;
 import io.temporal.internal.worker.HeartbeatManager;
+import io.temporal.payload.storage.ExternalStorageOptions;
 import io.temporal.serviceclient.MetricsTag;
 import io.temporal.serviceclient.WorkflowServiceStubs;
 import io.temporal.serviceclient.WorkflowServiceStubsPlugin;
@@ -106,7 +109,17 @@ final class WorkflowClientInternalImpl implements WorkflowClient, WorkflowClient
             .getOptions()
             .getMetricsScope()
             .tagged(MetricsTag.defaultTags(options.getNamespace()));
-    this.genericClient = new GenericWorkflowClientImpl(workflowServiceStubs, metricsScope);
+    GenericWorkflowClient genericClient =
+        new GenericWorkflowClientImpl(workflowServiceStubs, metricsScope);
+    ExternalStorageOptions externalStorage = options.getExternalStorage();
+    if (externalStorage != null) {
+      genericClient =
+          new ExternalStorageGenericWorkflowClient(
+              genericClient,
+              ExternalStorageMessageConverter.create(externalStorage, 1),
+              options.getNamespace());
+    }
+    this.genericClient = genericClient;
     this.interceptors = options.getInterceptors();
     this.workflowClientCallsInvoker = initializeClientInvoker();
     this.manualActivityCompletionClientFactory =
--- a/temporal-sdk/src/main/java/io/temporal/client/WorkflowClientInternalImpl.java
+++ b/temporal-sdk/src/main/java/io/temporal/client/WorkflowClientInternalImpl.java
@@ -111,7 +111,7 @@ final class WorkflowClientInternalImpl implements WorkflowClient, WorkflowClient
             .tagged(MetricsTag.defaultTags(options.getNamespace()));
     ExternalStorageOptions externalStorage = options.getExternalStorage();
     ExternalStorageMessageConverter externalStorageConverter =
-        externalStorage == null ? null : ExternalStorageMessageConverter.create(externalStorage, 1);
+        externalStorage == null ? null : ExternalStorageMessageConverter.create(externalStorage);
     GenericWorkflowClient genericClient =
         new GenericWorkflowClientImpl(workflowServiceStubs, metricsScope);
     if (externalStorageConverter != null) {
--- a/temporal-sdk/src/main/java/io/temporal/client/WorkflowClientInternalImpl.java
+++ b/temporal-sdk/src/main/java/io/temporal/client/WorkflowClientInternalImpl.java
@@ -109,15 +109,17 @@ final class WorkflowClientInternalImpl implements WorkflowClient, WorkflowClient
             .getOptions()
             .getMetricsScope()
             .tagged(MetricsTag.defaultTags(options.getNamespace()));
-    ExternalStorageOptions externalStorage = options.getExternalStorage();
-    ExternalStorageMessageTransformer externalStorageConverter =
-        externalStorage == null ? null : ExternalStorageMessageTransformer.create(externalStorage);
+    ExternalStorageOptions externalStorageOptions = options.getExternalStorage();
+    ExternalStorageMessageTransformer externalStorage =
+        externalStorageOptions == null
+            ? null
+            : ExternalStorageMessageTransformer.create(externalStorageOptions);
     GenericWorkflowClient genericClient =
         new GenericWorkflowClientImpl(workflowServiceStubs, metricsScope);
-    if (externalStorageConverter != null) {
+    if (externalStorage != null) {
       genericClient =
           new ExternalStorageGenericWorkflowClient(
-              genericClient, externalStorageConverter, options.getNamespace());
+              genericClient, externalStorage, options.getNamespace());
     }
     this.genericClient = genericClient;
     this.interceptors = options.getInterceptors();
@@ -128,7 +130,7 @@ final class WorkflowClientInternalImpl implements WorkflowClient, WorkflowClient
             options.getNamespace(),
             options.getIdentity(),
             options.getDataConverter(),
-            externalStorageConverter);
+            externalStorage);
 
     java.time.Duration heartbeatInterval = options.getWorkerHeartbeatInterval();
     if (!heartbeatInterval.isNegative()) {
--- a/temporal-sdk/src/main/java/io/temporal/client/WorkflowClientInternalImpl.java
+++ b/temporal-sdk/src/main/java/io/temporal/client/WorkflowClientInternalImpl.java
@@ -23,7 +23,7 @@ import io.temporal.internal.client.external.GenericWorkflowClient;
 import io.temporal.internal.client.external.GenericWorkflowClientImpl;
 import io.temporal.internal.client.external.ManualActivityCompletionClientFactory;
 import io.temporal.internal.common.PluginUtils;
-import io.temporal.internal.payload.storage.ExternalStorageMessageTransformer;
+import io.temporal.internal.payload.storage.ExternalStorage;
 import io.temporal.internal.sync.StubMarker;
 import io.temporal.internal.worker.HeartbeatManager;
 import io.temporal.payload.storage.ExternalStorageOptions;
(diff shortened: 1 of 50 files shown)


temporal-sdk/src/main/java/io/temporal/internal/client/ActivityClientHelper.java
```diff
@@ -30,17 +33,23 @@ public static RecordActivityTaskHeartbeatResponse sendHeartbeatRequest(
...
+      @Nullable StorageDriverTargetInfo storageTarget) {
+    RecordActivityTaskHeartbeatRequest.Builder builder =
         RecordActivityTaskHeartbeatRequest.newBuilder()
             .setTaskToken(ByteString.copyFrom(taskToken))
             .setNamespace(namespace)
             .setIdentity(identity);
-    payloads.ifPresent(request::setDetails);
+    payloads.ifPresent(builder::setDetails);
+    RecordActivityTaskHeartbeatRequest request = builder.build();
+    if (externalStorage != null) {
```
cconstable: This covers both worker and manual heartbeats

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
```diff
@@ -94,7 +100,12 @@ private Result handleWorkflowTaskWithQuery(
       logWorkflowTaskToBeProcessed(workflowTask, createdNew);
 
       ServiceWorkflowHistoryIterator historyIterator =
-          new ServiceWorkflowHistoryIterator(service, namespace, workflowTask, metricsScope);
+          new ServiceWorkflowHistoryIterator(
```
cconstable, line 104: is this used by the worker?

_update:_ yes, this is called by the workers. the iterator handles fetching additional pages of history for large histories.

temporal-sdk/src/main/java/io/temporal/internal/nexus/NexusTaskHandlerImpl.java
```diff
@@ -136,8 +137,18 @@ public Result handle(NexusTask task, Scope metricsScope) throws TimeoutException
               (Throwable) null);
       }
     } catch (HandlerException e) {
+      ExternalStorageNotConfiguredException externalStorageFailure =
+          ExternalStorageNotConfiguredException.find(e);
+      if (externalStorageFailure != null) {
+        throw externalStorageFailure;
+      }
```
cconstable: lets look at how other errors are being handled here and make sure we aren't special-casing extstore errors

temporal-sdk/src/main/java/io/temporal/internal/worker/ActivityWorker.java
```diff
@@ -392,16 +399,21 @@ public Throwable wrapFailure(ActivityTask t, Throwable failure) {
...
-                .setNamespace(namespace)
-                .setWorkerVersion(options.workerVersionStamp())
-                .build();
+            storeOutboundPayloads(
+                taskCompleted.toBuilder()
+                    .setTaskToken(taskToken)
+                    .setIdentity(options.getIdentity())
+                    .setNamespace(namespace)
+                    .setWorkerVersion(options.workerVersionStamp())
+                    .build(),
```
cconstable, line 433: @jmaeagle99 suggests looking at passing through builders to visitor. are there perf concerns? are we building messages twice? @cconstable will investigate.

cconstable: I'd like to leave this as a follow-up to avoid adding more changes to this PR. Created https://github.com/temporalio/sdk-java/issues/2999 to track this.

temporal-sdk/src/main/java/io/temporal/internal/nexus/NexusTaskHandlerImpl.java
cconstable: No longer special casing these.

temporal-sdk/src/main/java/io/temporal/internal/replay/ServiceWorkflowHistoryIterator.java
```diff
@@ -64,7 +77,11 @@ public boolean hasNext() {
       // true.
       GetWorkflowExecutionHistoryResponse response = queryWorkflowExecutionHistory();
 
-      current = response.getHistory().getEventsList().iterator();
+      History history = response.getHistory();
+      if (externalStorage != null) {
+        history = externalStorage.retrieveBlocking(history);
```
cconstable, line 84: errors here are caught by the replay handlers try-catch.

temporal-sdk/src/main/java/io/temporal/client/WorkflowClientInternalImpl.java
```diff
@@ -106,15 +109,28 @@ public static WorkflowClient newInstance(
             .getOptions()
             .getMetricsScope()
             .tagged(MetricsTag.defaultTags(options.getNamespace()));
-    this.genericClient = new GenericWorkflowClientImpl(workflowServiceStubs, metricsScope);
+    ExternalStorageOptions externalStorageOptions = options.getExternalStorage();
+    ExternalStorageMessageTransformer externalStorage =
```
jmaeagle99: nit: I would move this to inside the if-then a few lines below and update the condition to check if `externalStorageOptions` is not null.

temporal-sdk/src/main/java/io/temporal/common/converter/PayloadAndFailureDataConverter.java
```diff
@@ -71,6 +72,10 @@ public <T> T fromPayload(Payload payload, Class<T> valueClass, Type valueType)
       return (T) new RawValue(payload);
     }
 
+    if (payload.getExternalPayloadsCount() > 0) {
```
jmaeagle99: Doesn't this need to check if external storage is not configured AND there are externally stored payloads?

temporal-sdk/src/main/java/io/temporal/internal/client/ActivityClientHelper.java
cconstable: Correction: this path only covers activity heartbeats with user payloads. Automatic worker heartbeats are handled by `HeartbeatManager`.

temporal-sdk/src/main/java/io/temporal/internal/client/external/ExternalStorageGenericWorkflowClient.java
```diff
@@ -0,0 +1,345 @@
...
+  public CompletableFuture<GetWorkflowExecutionHistoryResponse> getWorkflowExecutionHistoryAsync(
+      @Nonnull GetWorkflowExecutionHistoryRequest request) {
+    return next.getWorkflowExecutionHistoryAsync(request)
+        .thenCompose(externalStorage::retrieveAsync);
+  }
+
+  @Override
+  public ListWorkflowExecutionsResponse listWorkflowExecutions(
+      ListWorkflowExecutionsRequest listRequest) {
+    return externalStorage.retrieveBlocking(next.listWorkflowExecutions(listRequest));
```
jmaeagle99: I'm slightly concerned about these iterable results that we will be fetching everything within the result set but a caller wants to inspect the non-payload-bearing fields and pick out only a few items. We should then resolve the payloads on demand. It might not be this proto that has that, but there are several. We do this deferral in the Go SDK specifically to avoid this potential performance penalty. This might not be doable due to the structure of the workflow client interface and might need to be refactored.

temporal-sdk/src/main/java/io/temporal/internal/worker/ActivityWorker.java
```diff
@@ -354,7 +361,7 @@ private ActivityTaskHandler.Result handleActivity(ActivityTask task, Scope metri
       }
 
       try {
-        sendReply(taskToken, result, metricsScope);
+        sendReply(taskToken, result, metricsScope, activityStorageTarget(pollResponse));
```
jmaeagle99, line 382: This needs to pivot the target based on whether it's a standalone activity or a workflow activity. Workflow activities should have a target that is the workflow information.

temporal-sdk/src/main/java/io/temporal/common/converter/PayloadAndFailureDataConverter.java
jmaeagle99: Also, it shouldn't be checking for the `external_payloads` field on the payload, it should be checking the metadata of the payload, probably using ExternalStorageReferences.tryParseReference or similar.

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
jmaeagle99, line 552: Missing the "per-command" targeting information for RespondWorkflowTaskCompletedResponse; it's only using the current workflow information, which is not correct for several commands. This is different than the TypeScript SDK because payload converter and codec are run within the workflow code execution in Java, so the serialization context targeting is done per invocation, whereas in TypeScript or other core-based SDKs (not that it has anything to do with core, just coincidentally) codecs are applied after the workflow code yields.

Probably need to set the `PayloadVisitorOptions.setMessageVisitor` with a callback that pivots the target based on the visited command.

temporal-sdk/src/main/java/io/temporal/internal/client/external/ManualActivityCompletionClientImpl.java
```diff
@@ -176,16 +195,18 @@ public void fail(@Nonnull Throwable exception) {
 
   @Override
   public void recordHeartbeat(@Nullable Object details) throws CanceledFailure {
+    Optional<Payloads> payloads = dataConverterWithActivityExecutionContext.toPayloads(details);
```
jmaeagle99: This should be inside the try block.

temporal-sdk/src/main/java/io/temporal/internal/client/external/ExternalStorageGenericWorkflowClient.java
```diff
@@ -0,0 +1,345 @@
...
+    UpdateWorkflowExecutionRequest stored =
+        externalStorage.storeBlocking(
+            updateParameters, workflowTarget(updateParameters.getWorkflowExecution(), null));
+    return externalStorage.retrieveBlocking(next.update(stored, deadline));
+  }
+
+  @Override
+  public CompletableFuture<PollWorkflowExecutionUpdateResponse> pollUpdateAsync(
+      @Nonnull PollWorkflowExecutionUpdateRequest request, @Nonnull Deadline deadline) {
+    return next.pollUpdateAsync(request, deadline).thenCompose(externalStorage::retrieveAsync);
```
jmaeagle99: The `thenCompose` calls are probably running on the gRPC completion thread, thus blocking any other gRPC processing from moving forward while external storage is processing.

temporal-sdk/src/main/java/io/temporal/internal/client/external/ExternalStorageGenericWorkflowClient.java
```diff
@@ -0,0 +1,345 @@
...
+
+  @Override
+  public CountWorkflowExecutionsResponse countWorkflowExecutions(
+      CountWorkflowExecutionsRequest request) {
+    return next.countWorkflowExecutions(request);
+  }
+
+  @Override
+  public CreateScheduleResponse createSchedule(CreateScheduleRequest request) {
+    return next.createSchedule(externalStorage.storeBlocking(request, null));
```
jmaeagle99, line 164: All of the schedule calls should pull the target information from the request instead of passing null.

temporal-sdk/src/test/java/io/temporal/internal/payload/storage/ExternalStorageTest.java
```diff
@@ -158,4 +160,89 @@ public synchronized CompletableFuture<List<Payload>> retrieve(
       return CompletableFuture.completedFuture(payloads);
     }
   }
+
+  private static final class ControlledDriver implements StorageDriver {
```
jmaeagle99: Are all of these dead code?

jmaeagle99 (changes requested):
In addition to these comments, I think there isn't much in terms of failure handling, as in if external storage fails for some reason, we should be failing the workflow/activity/nexus task retryably; I'd like to see integration tests around that too. Also, I don't see much for validating that the correct target information is being produced for each system; integration tests would help here as well.

temporal-sdk/src/main/java/io/temporal/common/converter/PayloadAndFailureDataConverter.java
cconstable: > it should be checking the metadata of the payload
 
 yes, updated
 
 > Doesn't this need to check if external storage is not configured AND there are externally stored payloads?

No because if external storage is configured, it will have already walked the messages and resolved the references before getting here. If no driver can retrieve the message that error should be thrown before this point. This is just a single place to catch "external storage was not configured and we have a reference payload still".

Ideally, I'd really like to remove external storage from this class but that either requires doing a bunch of eager message walking up front or adding lazy checks to a bunch of random places in the sdk which seems error prone and architecturally unsound.

temporal-sdk/src/test/java/io/temporal/internal/payload/storage/ExternalStorageTest.java
cconstable: Removed

PR #2982, opened 2026-07-30
Add dev server downloader & runner to testing package

Sushisource:
## What changed

Added an experimental dev-server API to temporal-testing, including configurable CLI download/cache support, readiness checks, process ownership, TestWorkflowEnvironment integration, and JUnit rule/extension support.

Repository tests now use the same Java-owned server through -PtestServer=dev-server. Gradle only prepares and activates the profile, while CI and contributor documentation use the new commands. Server lifecycle coverage runs against the pinned real CLI.

## Why?

This makes local dev-server ownership reusable by SDK users and tools such as flake verification, while keeping sdk-java tests on the same implementation instead of maintaining a separate Gradle process runner.

## Breaking changes?

None. The unreleased useTemporalCli profile and preparation task were replaced without compatibility aliases.

## Server PR

None.

## Testing

- temporal-testing tests and Javadocs
- Java 11 dev-server profile test
- Java 21 virtual-thread tests
- Real-CLI lifecycle integration tests


temporal-testing/src/main/java/io/temporal/testing/TestWorkflowRule.java
```diff
@@ -266,6 +278,41 @@ public Builder setActivityImplementations(Object... activityImplementations) {
...
+      return useDevServer(TemporalDevServerOptions.getDefaultInstance());
+    }
+
+    /**
+     * Uses an owned local Temporal dev server with the supplied options.
+     *
+     * <pre>{@code
+     * TestWorkflowRule.newBuilder()
+     *     .useDevServer(
+     *         TemporalDevServerOptions.newBuilder().setDownloadVersion("1.7.2").build())
```
Quinn-With-Two-Ns:
```suggestion
     *         TemporalDevServerOptions.newBuilder().setDownloadVersion("v1.7.2").build())
```

gradle/temporalCli.gradle
```diff
@@ -0,0 +1,139 @@
+import org.gradle.api.services.BuildService
+import org.gradle.api.services.BuildServiceParameters
+import org.gradle.api.tasks.JavaExec
+import org.gradle.api.tasks.testing.Test
+import org.gradle.jvm.toolchain.JavaLanguageVersion
+
+import java.util.concurrent.TimeUnit
+
+abstract class TemporalDevServerTestService
```
Quinn-With-Two-Ns, line 9: Think it would be simpler to just move this into the SDK test rule?

temporal-testing/src/main/java/io/temporal/testing/TemporalDevServer.java
```diff
@@ -0,0 +1,74 @@
...
+    return start("default", TemporalDevServerOptions.getDefaultInstance());
+  }
+
+  /** Starts a dev server in namespace {@code default} with the supplied options. */
+  public static TemporalDevServer start(TemporalDevServerOptions options) {
+    return start("default", options);
+  }
+
+  /** Starts a dev server for the supplied namespace and options. */
+  public static TemporalDevServer start(String namespace, TemporalDevServerOptions options) {
```
Quinn-With-Two-Ns: Nit: could use some non null annotations

Quinn-With-Two-Ns:
We should make a follow up issue to update our docs and samples to show how to do this

temporal-testing/src/main/java/io/temporal/testing/TestWorkflowRule.java
```diff
@@ -266,6 +278,41 @@ public Builder setActivityImplementations(Object... activityImplementations) {
...
+     * }</pre>
+     */
+    @Experimental
+    public Builder useDevServer(TemporalDevServerOptions options) {
+      if (options == null) {
+        throw new NullPointerException("options");
+      }
+      this.useDevServer = true;
+      this.useExternalService = false;
+      this.target = null;
```
Quinn-With-Two-Ns, line 327: Should we document these are mutually exclusive?

gradle/temporalCli.gradle
Sushisource: I had it that way at one point but this allowed us to use just one instance for the whole integ test run when gradle is orchestrating.

temporal-testing/src/main/java/io/temporal/testing/TestWorkflowRule.java
Sushisource: ~I was actually intending to just make this an enum and somehow that didn't happen, so, I'll just do that.~

Derr, didn't realize what part of the code this was when I wrote that. We can't since use external's already public. Will doc.

temporal-testing/src/main/java/io/temporal/testing/TemporalDevServer.java
Sushisource: HAVE ALL THE ANNOTATIONS!!!

Quinn-With-Two-Ns (approved):

PR #2988, opened 2026-08-03
Adding SDK Ergonomics Query link

Evanthx:
The one thing to note is that there was some duplication around link converters and still is. I would like to fix that separately, and have created an epic to check over the link converters for each SDK which I am planning on doing in the next sprint.
https://temporalio.atlassian.net/browse/NEXUS-578

The proto bump brings in the link field.

### DIFF
--- a/temporal-sdk/src/test/java/io/temporal/workflow/nexus/QueryOperationTest.java
+++ b/temporal-sdk/src/test/java/io/temporal/workflow/nexus/QueryOperationTest.java
@@ -47,7 +47,6 @@ import org.junit.function.ThrowingRunnable;
  * <p>Covers the value round trip plus the failure modes a caller can observe: an unknown workflow,
  * a query handler that throws, and a Query rejected by the client's reject condition. All of these
  * must fail the caller's Nexus operation rather than hanging or returning a default.
- *
  */
 public class QueryOperationTest {
 
--- a/temporal-sdk/src/test/java/io/temporal/workflow/nexus/QueryOperationTest.java
+++ b/temporal-sdk/src/test/java/io/temporal/workflow/nexus/QueryOperationTest.java
@@ -1,12 +1,18 @@
 package io.temporal.workflow.nexus;
 
+import static org.junit.Assume.assumeTrue;
+
 import io.nexusrpc.Operation;
 import io.nexusrpc.Service;
 import io.nexusrpc.handler.HandlerException;
 import io.nexusrpc.handler.OperationHandler;
 import io.nexusrpc.handler.OperationImpl;
 import io.nexusrpc.handler.ServiceImpl;
+import io.temporal.api.common.v1.Link;
+import io.temporal.api.enums.v1.EventType;
 import io.temporal.api.enums.v1.QueryRejectCondition;
+import io.temporal.api.history.v1.History;
+import io.temporal.api.history.v1.HistoryEvent;
 import io.temporal.client.WorkflowClient;
 import io.temporal.client.WorkflowClientOptions;
 import io.temporal.client.WorkflowFailedException;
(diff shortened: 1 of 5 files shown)


Quinn-With-Two-Ns:
>duplication around link converters and still is

What duplication? I am not aware of any except for what we had in the Go SDK/Server

temporal-sdk/src/test/java/io/temporal/workflow/nexus/QueryOperationTest.java
```diff
@@ -0,0 +1,316 @@
...
+  private static final int BUMPS = 2;
+
+  @Rule
+  public SDKTestWorkflowRule testWorkflowRule =
+      SDKTestWorkflowRule.newBuilder()
+          .setWorkflowTypes(QueryCallerWorkflowImpl.class, CounterWorkflowImpl.class)
+          .setNexusServiceImplementation(new QueryingNexusServiceImpl())
+          // The workflow being queried parks on a signal, so time skipping would fast-forward it
+          // into its execution timeout and it would be gone before the Query lands.
+          .setUseTimeskipping(false)
```
Quinn-With-Two-Ns, line 69: You'll need an assume in this class too, to stop these test from running against the test server in CI

temporal-sdk/src/test/java/io/temporal/workflow/nexus/QueryOperationTest.java
```diff
@@ -0,0 +1,316 @@
...
+/**
+ * A Nexus operation backed by a workflow Query. A Query is always synchronous and writes nothing to
+ * history, so the handler simply queries and returns the result; there is no operation token and no
+ * completion callback.
+ *
+ * <p>Covers the value round trip plus the failure modes a caller can observe: an unknown workflow,
+ * a query handler that throws, and a Query rejected by the client's reject condition. All of these
+ * must fail the caller's Nexus operation rather than hanging or returning a default.
+ *
+ * <p>The response link the server attaches to {@code QueryWorkflowResponse} is verified in {@link
```
Quinn-With-Two-Ns: This test should assert on the link , we will need to publish a real server before we merge this anyway

Quinn-With-Two-Ns:
Other then the tests the actual SDK changes LGTM

Quinn-With-Two-Ns:
FWIW I also ran a Codex review and it only had one finding

```
  - Medium — nexusLinkToWorkflowLink (https://github.com/temporalio/sdk-java/blob/accc671abce2496605da46ec28c1966613e8b220/temporal-sdk/src/main/java/io/temporal/internal/common/LinkConverter.java#L232-L238) uses form-style URLDecoder for path
    segments, turning literal + into a space. A Go-produced workflow ID a+b would therefore become a b, potentially pointing to the wrong execution. The new test currently codifies this bug
    (https://github.com/temporalio/sdk-java/blob/accc671abce2496605da46ec28c1966613e8b220/temporal-sdk/src/test/java/io/temporal/internal/common/LinkConverterTest.java#L855-L883). Use a path-segment percent decoder that preserves +, then invert
    that assertion. The Go reference converter uses PathEscape/PathUnescape (https://github.com/temporalio/api-go/blob/85042fb4769bb46c3ed1addac9de526af51f4d83/temporalnexus/link_converter.go#L336-L405).


```

temporal-sdk/src/test/java/io/temporal/workflow/nexus/QueryOperationTest.java
Evanthx: True. Test added

temporal-sdk/src/test/java/io/temporal/workflow/nexus/QueryOperationTest.java
Evanthx: Added

Quinn-With-Two-Ns (approved):

PR #2990, opened 2026-08-06
Fix recordHeartbeat: swallowed cancel/reset/pause exceptions

### LINKED ISSUE #2983: Manual completion client's recordHeartbeat swallows cancellation, reset and pause exceptions
## Expected Behavior

`ManualActivityCompletionClient.recordHeartbeat` should throw `ActivityCanceledException`,
`ActivityResetException` or `ActivityPausedException` when the server reports cancellation, reset or
pause. The method already builds those exceptions.

The declaration on the public interface should change too. `recordHeartbeat` is declared
`throws CanceledFailure`, a type the method never throws. It should name what callers really have to
handle:

```java
// io.temporal.activity.ManualActivityCompletionClient
void recordHeartbeat(@Nullable Object details) throws ActivityCompletionException;
```

## Actual Behavior

`ActivityCanceledException`, `ActivityResetException` and `ActivityPausedException` never reach the
caller. The `try` block wraps the response checks along with the RPC, so `catch (Exception e)` catches
all three. `processException` then turns each one into `ActivityCompletionFailureException`.

```java
// ManualActivityCompletionClientImpl.java, same on main as of d310594f
try {
  RecordActivityTaskHeartbeatResponse status = ActivityClientHelper.sendHeartbeatRequest(...);
  if (status.getCancelRequested()) {
    throw new ActivityCanceledException();   // caught below, never reaches the caller
  } else if (status.getActivityReset()) {
    throw new ActivityResetException();      // same
  } else if (status.getActivityPaused()) {
    throw new ActivityPausedException();     // same
  }
  ...
} catch (Exception e) {
  processException(e);
}
```

The caller now sees the same exception for a cancelled activity and for a failed heartbeat RPC. The only
way to tell them apart is to unwrap `getCause()`.

The `recordActivityTaskHeartbeatById` branch has the same problem. The worker-side path does not.
`HeartbeatContextImpl.sendHeartbeatRequest` checks the flags outside its catch block, so
`ActivityExecutionContext.heartbeat` throws the specific type.

## Specifications

  - Version: 1.37.0
  - Platform:

### LINKED ISSUE #2984: Retry transient errors in ActivityCompletionClient.heartbeat
`ActivityCompletionClient.heartbeat` fails on the first transient error from the server — most often `RESOURCE_EXHAUSTED` from namespace rate limiting, but also `DEADLINE_EXCEEDED` and `UNAVAILABLE`. For async completion, where an external process holds the task token and the heartbeat is the only thing keeping a long-running activity alive, a single rate-limited
heartbeat can cost us the activity via heartbeat timeout.

The worker-side path already handles this. `HeartbeatContextImpl.doHeartBeatLocked` catches `StatusRuntimeException`, logs a warning and reschedules after `HEARTBEAT_RETRY_WAIT_MILLIS` instead of failing the activity.

Wrap `recordHeartbeat` in the same `GrpcRetryer` the other three reply calls in
`ManualActivityCompletionClientImpl` already use — `complete`, `fail` and `reportCancellation`
all run inside `grpcRetryer.retry(...)` with `replyGrpcRetryerOptions`, and `recordHeartbeat` is
the only one that doesn't. That gives heartbeat the same `RESOURCE_EXHAUSTED` handling
(including the message-too-large carve-out) for free.

A shorter expiration than `replyGrpcRetryerOptions` may make more sense here, since retrying past
the next heartbeat interval is pointless.

Open to another approach if you have one in mind — making the retryability decision public so
callers can apply their own policy would also unblock us.

**Additional context**

brainlimitexceeded:
## Summary

Closes #2983. Fixes `ManualActivityCompletionClientImpl`'s `recordHeartbeat` to raise `ActivityCanceledException`/`ActivityResetException`/`ActivityPausedException` instead of an incorrect `ActivityCompletionFailureException`.

## What changed

- `ActivityHeartbeatResponse`: New wrapper class to reduce code duplication. Technically a breaking change, but only to internals that users shouldn't be relying on.
- Updated documentation for `ManualActivityCompletionClient.recordHeartbeat`
- Restructured `ManualActivityCompletionClientImpl` to propagate exceptions correctly
- Added regression tests in `ManualActivityCompletionClientImplTest` for each possible thrown exception


dplyukhin:
Thanks for the contrib @brainlimitexceeded! I removed the retry part of the PR because that's complicated logic I want to handle separately. Can you please sign the CLA?

temporal-sdk/src/main/java/io/temporal/activity/ManualActivityCompletionClient.java
```diff
@@ -30,8 +30,14 @@ public interface ManualActivityCompletionClient {
    * Records heartbeat for an activity
    *
    * @param details to record with the heartbeat
+   * @throws ActivityCompletionException if the server reports the activity was cancelled, reset, or
```
Quinn-With-Two-Ns, line 33:
```suggestion
   * @throws ActivityCompletionException if activity should stop executing
```
nit: This is the docs we use elsewhere. It is easier to maintain and then move the details here into `ActivityCompletionException` since there are a lot of methods that would need all these detauils

temporal-sdk/src/main/java/io/temporal/activity/ManualActivityCompletionClient.java
```diff
@@ -30,8 +30,14 @@ public interface ManualActivityCompletionClient {
...
+   *     paused ({@link io.temporal.client.ActivityCanceledException}, {@link
+   *     io.temporal.client.ActivityResetException}, {@link
+   *     io.temporal.client.ActivityPausedException}), or if the heartbeat RPC fails ({@link
+   *     io.temporal.client.ActivityCompletionFailureException}, {@link
+   *     io.temporal.client.ActivityNotExistsException}).
    */
-  void recordHeartbeat(@Nullable Object details) throws CanceledFailure;
+  void recordHeartbeat(@Nullable Object details) throws ActivityCompletionException;
```
Quinn-With-Two-Ns, line 40: I'll note both the old code and this are not really best practice since both are `RuntimeException`. We throw `ActivityCompletionException` from the other heart beating code it isn't the end of the world but is not the most idiomatic

Quinn-With-Two-Ns:
Note this would be pretty breaking change for anyone doing `ManualActivityCompletion` + `Heartbeating` I think at a minimum we need to call this out with big warnings in the SDK release notes .

temporal-sdk/src/main/java/io/temporal/activity/ManualActivityCompletionClient.java
dplyukhin: Hmm, I'll just delete the unnecessary doc and base it on `ActivityCompletionClient`. I don't think there's much point in having `ActivityCompletionException` document its own subclasses.

temporal-sdk/src/main/java/io/temporal/activity/ManualActivityCompletionClient.java
dplyukhin: I could get rid of it, but we do the same thing in `ActivityCompletionClient`. I feel like it's still useful to describe as part of the contract, even if the exception isn't checked

PR #2986, opened 2026-07-31
Add Temporal-backed Java SDK release automation

eamsden:
## What changed?

- Added a Java release Workflow backed by Temporal Cloud. GitHub Actions runs transient Workers on release-specific Task Queues.
- Bound each release to its tag, full source SHA, release-note hash, exact GitHub Actions artifacts, and trusted automation SHA.
- Separated unprivileged build work from approved publication. Publication Activities recheck the Workflow execution, approval, ownership, artifact manifest, and remote state.
- Added zero-input approval for active `temporalio/sdk` members, durable pause/resume/handoff, and scheduled runner recovery.
- Made Maven and GitHub publication reconcile exact existing state. GitHub releases stay draft until all assets are verified.
- Kept `prepare-release.yml` as the only manual release procedure. It takes durable Temporal ownership and refuses to publish Maven after automatic Maven submission has started.

## Why?

The current Actions-only process loses orchestration state with its runner. Temporal keeps control state durable; GitHub Actions artifacts hold transient binary payloads so later jobs can resume the exact release.

Sonatype cannot atomically create a staging repository and persist its ID in Temporal. An ambiguous creation therefore stops for inspection and explicit authorization before another submission generation.

## Breaking changes?

No SDK API changes. The automation remains inactive until its Temporal Cloud, GitHub environment, signing, and Maven Central settings are configured.

## Server PR

None.

No release, test release, Maven publication, tag push, or GitHub release was performed while testing this change.

PR #2995, opened 2026-08-10
Add Temporal-backed release automation

eamsden:
## What changed?

- Automate release with one Temporal workflow implemented in Python.
- Start and complete a release from the GitHub Actions run triggered when an SDK owner merges one valid release-note file.
- Build the fixed native matrix and signed Maven payload, freeze their GitHub Actions artifact identities in Workflow state, publish Maven with bounded ambiguity recovery, then create the GitHub draft, attach assets, and make it public.
- Keep release- and Maven-generation-specific task queues, GitHub App publication authentication, global publication concurrency, and credential-free candidate compilation.
- Remove timers, separate approval and ownership workflows, manual takeover and recovery UI, and the former manual release workflow.

## Why?

The current manu-matic Java release process contains footguns which have led to incidents. There is also enough fragility in the GitHub and Maven release process that dogfooding Temporal here seems reasonable. The Python implementation keeps the necessary orchestration and reconciliation in one readable operational path.

## Breaking changes?

No SDK API changes.

The former manual prepare-release path is intentionally removed. Merging a valid release-note file is the release authorization and execution trigger.

## Verification

- 8 focused local tests using the Temporal ephemeral test server and mocked publication activities
- Ruff formatting and lint
- Strict mypy
- uv lock consistency
- Shell syntax and deterministic native/Maven archive smoke checks
- Workflow YAML parsing, pinned Action validation, and permission-boundary checks
- No scheduled or manual dispatch entrypoint in the Temporal release workflow

No release, test release, Maven publication, tag, GitHub release, or external Temporal operation was run while developing or testing this change.

## Server PR

None.

Supersedes #2986.


.github/release-automation/release_automation/release.py
```diff
@@ -0,0 +1,910 @@
...
+        current, view = self.value.generations[-1], await self.view()
+        repository = current.repository or view.find(self.description(current.number))
+        if not repository:
+            for prior in self.value.generations[:-1]:
+                old = prior.repository or view.find(self.description(prior.number))
+                profiles, manual = view.repository(old) if old else ([], [])
+                _, portal = self.repository_state(profiles, manual)
+                portal_id = prior.portal or portal
+                state = await self.portal_state(portal_id) if portal_id else ""
+                if profiles or manual and (manual[0].get("state") != "released" or state != "FAILED"):
```
cconstable: `and` has a higher precedence than `or` so this is evaluated as `profiles or (manual and ...)` which I don't think was the intention

.github/release-automation/release_automation/release.py
```diff
@@ -0,0 +1,910 @@
...
+        finally:
+            session.temp.cleanup()
+
+    @activity.defn(name="discoverArtifacts")
+    async def discover(self, value: ReleaseInput) -> list[Artifact]:
+        """Freeze deterministic native and Maven Actions artifact identities."""
+        session = Session(value, self.source, self.environment)
+        try:
+            artifacts = [
+                await session.artifact(
```
cconstable: this can throw a `ReleaseError` (L458) but no except to capture it. looks like that would cause retries until the start_to_close_timeout hits.

.github/release-automation/release_automation/release.py
```diff
@@ -0,0 +1,910 @@
...
+                )
+                for platform in PLATFORMS
+            ]
+            artifacts.append(await session.artifact(f"sdk-java-release-maven-{value.candidate.id}", "maven-payload.tar"))
+            return artifacts
+        finally:
+            session.temp.cleanup()
+
+    @activity.defn(name="inspectMaven")
+    async def inspect_maven(self, value: ReleaseInput) -> Inspection:
```
cconstable, line 830: I haven't seen any heartbeats in the activities so far. Might be good to add so we can distinguish between a worker dying vs the activity just taking a long time. We would need to add a `heartbeat_timeout` to the `execute_activity` options.

.github/release-automation/release_automation/release.py
```diff
@@ -0,0 +1,910 @@
...
+        return cast(Inspection, await self._run(value, "inspect"))
+
+    @activity.defn(name="publishRelease")
+    async def publish(self, value: ReleaseInput) -> ReleaseResult:
+        """Reconcile Maven, then the GitHub draft, then public GitHub state."""
+        session = Session(value, self.source, self.environment)
+        try:
+            root, records = await session.materialize()
+            await session.maven(root, records)
+            return await session.github_release()
```
cconstable: is it possible to make these separate activities? in the event that gh is having issues `publish` would fail we would have to re-materialize. probably fine but something to think about.

.github/workflows/README.md
```diff
@@ -1,38 +0,0 @@
-# sdk-java Github Workflows
```
cconstable, line 1: A new README would be helpful as I'm still not quite sure I've got the whole flow mapped out.

cconstable (commented):
Initial review. Didn't get through everything, will take another pass.

.github/release-automation/release_automation/release.py
cconstable: Maybe would be better as a follow-up.

.github/release-automation/release_automation/release.py
```diff
@@ -0,0 +1,910 @@
...
+        if status != 200 or not isinstance(response, dict) or response.get("deploymentId") != identity:
+            raise RuntimeError("Publisher Portal deployment is unavailable.")
+        return cast(str, response.get("deploymentState", ""))
+
+    @staticmethod
+    def repository_state(profiles: list[dict[str, Any]], manual: list[dict[str, Any]]) -> tuple[str, str | None]:
+        """Derive the staging state and Portal ID from joined repository rows."""
+        if manual:
+            return str(manual[0].get("state", "")), cast(str | None, manual[0].get("portal_deployment_id"))
+        return ("open", None) if profiles else ("absent", None)
```
cconstable: are there other states here like `closed` or `already_published`? I don't know if we can get into a state where the staging state is actually one of those but we report `open` anyways and then try to push stuff to it.

cconstable (changes requested):
Left a few comments. A few might need changes.

.github/release-automation/release_automation/release.py
eamsden: Good catch. I replaced this expression with a shared `replaceable()` predicate used by both workflow recovery and publication, and there are tests covering the logic now.

.github/release-automation/release_automation/release.py
eamsden: Discovery is now implemented in the same activity wrapper as inspection and publication, which turns `ReleaseError` into a non-retryable `ApplicationError`

.github/release-automation/release_automation/release.py
eamsden: Open to this as a follow-up, I was pushing very hard for concision in this attempt at automating this workflow.

.github/release-automation/release_automation/release.py
eamsden: Added heartbeats and `heartbeat_timeout`

.github/workflows/README.md
eamsden: New README added.

.github/release-automation/release_automation/release.py
eamsden: This question helped Codex catch a bug, so thank you :) States are now checked explicitly and there is a fixed set of acceptable states. Profile states other than `open`, `closed`, or `released` will fail with `ReleaseError`.

cconstable (approved):
Outside of evaluating the interaction with sonatype (which I know little about) this lgtm.

PR #3013, opened 2026-08-18
Implement operator commands for Standalone Activities

GregoryTravis:
# Implement operator commands for standalone activities

Adds pause, unpause, reset, and update-options to standalone activities, plus
the describe surface needed to observe their effects.

Standalone activities already supported start, result, describe, cancel, and
terminate. This adds the four operator commands the server exposes for them, so
an operator can hold, resume, restart, and retune a running activity without
going through a workflow.

## Describe: payload fields are opt-in

`DescribeActivityExecutionRequest` gates four payload-bearing fields behind
per-call flags (api#792). All four are now plumbed through
`DescribeActivityOptions` and default to false.


temporal-sdk/src/main/java/io/temporal/client/ActivityExecutionDescription.java
```diff
@@ -27,28 +28,32 @@
 @Experimental
 public final class ActivityExecutionDescription extends ActivityExecutionMetadata {
 
+  private final DescribeActivityExecutionResponse response;
   private final ActivityExecutionInfo info;
```
maciejdudko: `info` field is now redundant, remove it and change methods to use `response.getInfo()`

temporal-sdk/src/main/java/io/temporal/client/ActivityExecutionDescription.java
```diff
@@ -217,6 +261,125 @@ public <V> Optional<V> getHeartbeatDetails(Class<V> valueType, Type genericType)
...
+  /**
+   * Deserializes the activity's first input argument. Returns {@link Optional#empty()} if no input
+   * is present (the activity took no arguments, or {@code includeInput} was false).
+   *
+   * <p>For a multi-argument activity this returns only the first argument; use {@link
+   * #getInput(int, Class)} to read the rest, and {@link #getInputCount()} for how many there are.
+   *
+   * @param valueType the class to deserialize the input into
+   */
+  public <V> Optional<V> getInput(Class<V> valueType) {
```
maciejdudko: `getInput` should have only one 0-arg overload that returns `EncodedValues`. After that change, we should also remove `getInputCount`.

While we're at it, I think we should change `getHeartbeatDetails` to return `EncodedValues` too.

temporal-sdk/src/main/java/io/temporal/client/ActivityExecutionDescription.java
```diff
@@ -27,28 +28,32 @@
...
-        info.getStatus(),
-        info.getTaskQueue(),
-        SearchAttributesUtil.decodeTyped(info.getSearchAttributes()));
-    this.info = info;
+        response.getInfo().getStatus(),
+        response.getInfo().getTaskQueue(),
+        SearchAttributesUtil.decodeTyped(response.getInfo().getSearchAttributes()));
+    this.response = response;
+    this.info = response.getInfo();
     this.dataConverter = dataConverter;
```
maciejdudko: We should add `ActivitySerializationContext` to `dataConverter` here.

temporal-sdk/src/main/java/io/temporal/client/ActivityExecutionDescription.java
```diff
@@ -217,6 +261,125 @@ public <V> Optional<V> getHeartbeatDetails(Class<V> valueType, Type genericType)
...
+
+  /**
+   * Deserializes the activity's success result. Returns {@link Optional#empty()} if no result is
+   * present (activity still running, closed with a failure, or {@code includeOutcome} was false).
+   *
+   * @param valueType the class to deserialize the result into
+   */
+  public <V> Optional<V> getResult(Class<V> valueType) {
+    return getResult(valueType, valueType);
+  }
```
maciejdudko, line 294: We should match what we do in `ActivityClient.startActivity`.
```suggestion
  public <V> Optional<V> getResult(Class<V> valueType) {
    return getResult(valueType, null);
  }
```

temporal-sdk/src/main/java/io/temporal/client/ActivityExecutionDescription.java
```diff
@@ -217,6 +261,125 @@ public <V> Optional<V> getHeartbeatDetails(Class<V> valueType, Type genericType)
...
+
+  /**
+   * Whether the activity closed with a successful result. {@code false} while the activity is still
+   * running, when it closed with a failure, or when the description was requested without {@link
+   * DescribeActivityOptions.Builder#setIncludeOutcome(boolean)}.
+   */
+  public boolean hasResult() {
+    return response.hasOutcome() && response.getOutcome().hasResult();
+  }
```
maciejdudko, line 284: Protobuf getters are null-coallescing. Fix other similar lines too.
```suggestion
  public boolean hasResult() {
    return response.getOutcome().hasResult();
  }
```

temporal-sdk/src/main/java/io/temporal/client/ActivityExecutionDescription.java
```diff
@@ -217,6 +261,125 @@ public <V> Optional<V> getHeartbeatDetails(Class<V> valueType, Type genericType)
...
+
+  /**
+   * The failure the activity closed with, as an exception. {@code null} if the activity did not
+   * close with a failure or if {@code includeOutcome} was false on the describe call.
+   *
+   * <p>This is the terminal outcome; {@link #getLastFailure()} is the failure of the most recent
+   * attempt, which may be set while the activity is still retrying.
+   */
+  @Nullable
+  public Exception getFailure() {
```
maciejdudko: I'd rename this method to `getOutcomeFailure` or something to better differentiate it from `getLastFailure`.

We should also change return type of both this and `getLastFailure` to `RuntimeException`.

temporal-sdk/src/main/java/io/temporal/client/ActivityExecutionOptions.java
```diff
@@ -0,0 +1,136 @@
...
+import javax.annotation.Nullable;
+
+/**
+ * The resolved options of a standalone activity execution, returned by {@link
+ * UntypedActivityHandle#updateOptions(UpdateActivityOptions)}.
+ *
+ * <p>Reflects the activity's options as the server resolved them after the update was applied.
+ */
+@Experimental
+public final class ActivityExecutionOptions {
```
maciejdudko, line 17: `UpdateActivityExecutionOptionsRequest` and `UpdateActivityOptionsResponse` use the same type to store options, which means the available fields will always match. Instead of adding yet another class, we can return `UpdateActivityOptions`.

temporal-sdk/src/main/java/io/temporal/client/UpdateActivityOptions.java
```diff
@@ -0,0 +1,268 @@
...
+
+/**
+ * Options for {@link UntypedActivityHandle#updateOptions(UpdateActivityOptions)}.
+ *
+ * <p>Only the fields that are explicitly set are sent to the server; a derived field mask ensures
+ * that unset fields are left unchanged (a partial update).
+ *
+ * <p>{@link Builder#setRestoreOriginal(boolean)} is mutually exclusive with every other field: an
+ * instance that sets {@code restoreOriginal} together with any other option is rejected by {@link
+ * Builder#build()} before any request is sent.
```
maciejdudko: Consider alternative design: `UpdateActivityOptions` does not have `restoreOriginal` field. Instead, `ActivityHandle` has an additional method `restoreOriginalOptions`.

temporal-sdk/src/main/java/io/temporal/client/UntypedActivityHandle.java
```diff
@@ -118,12 +118,22 @@ <R> CompletableFuture<R> getResultAsync(
...
   ActivityExecutionDescription describe();
 
+  /**
+   * Describes the current state of the activity execution.
+   *
+   * @param options which payload-bearing fields to include in the description. These are opt-in
+   *     because they can be arbitrarily large.
+   * @return detailed information about the activity
+   */
+  ActivityExecutionDescription describe(DescribeActivityOptions options);
```
maciejdudko, line 135: I think `DescribeActivityOptions.java` file is missing.

temporal-sdk/src/main/java/io/temporal/common/interceptors/ActivityClientCallsInterceptor.java
```diff
@@ -339,6 +386,185 @@ public String getReason() {
...
+
+  @Experimental
+  final class PauseActivityOutput {}
+
+  @Experimental
+  final class UnpauseActivityInput {
+    private final String id;
+    private final @Nullable String runId;
+    private final @Nullable String reason;
+    private final @Nullable Duration jitter;
```
maciejdudko:
```suggestion
    private final UnpauseActivityOptions options;
```

temporal-sdk/src/main/java/io/temporal/common/interceptors/ActivityClientCallsInterceptor.java
```diff
@@ -339,6 +386,185 @@ public String getReason() {
...
+  final class UnpauseActivityOutput {}
+
+  @Experimental
+  final class ResetActivityInput {
+    private final String id;
+    private final @Nullable String runId;
+    private final boolean keepPaused;
+    private final @Nullable Duration jitter;
+    private final boolean restoreOriginalOptions;
+    private final boolean resetHeartbeat;
```
maciejdudko:
```suggestion
    private final ResetActivityOptions options;
```

temporal-sdk/src/main/java/io/temporal/common/interceptors/ActivityClientCallsInterceptor.java
```diff
@@ -339,6 +386,185 @@ public String getReason() {
...
+    }
+
+    public boolean isRestoreOriginal() {
+      return restoreOriginal;
+    }
+  }
+
+  @Experimental
+  final class UpdateActivityOptionsOutput {
+    private final ActivityOptions activityOptions;
```
maciejdudko: `UpdateActivityOptionsOutput` should have the final options object, not Proto object. The conversion should happen inside the root interceptor.

temporal-sdk/src/main/java/io/temporal/internal/client/RootActivityClientInvoker.java
```diff
@@ -339,14 +339,18 @@ public DescribeActivityOutput describeActivity(DescribeActivityInput input) {
...
+            .setIncludeOutcome(input.getOptions().isIncludeOutcome())
+            .setIncludeHeartbeatDetails(input.getOptions().isIncludeHeartbeatDetails())
+            .setIncludeLastFailure(input.getOptions().isIncludeLastFailure());
     if (input.getRunId() != null) {
       req.setRunId(input.getRunId());
     }
     DescribeActivityExecutionResponse response = genericClient.describeActivity(req.build());
     return new DescribeActivityOutput(
         new ActivityExecutionDescription(
-            response.getInfo(), clientOptions.getDataConverter(), clientOptions.getNamespace()));
```
maciejdudko, line 352: We should remove payload fields that were not requested (to support older/buggy servers).

temporal-sdk/src/test/java/io/temporal/internal/client/ActivityHandleOperatorCommandsTest.java
```diff
@@ -0,0 +1,123 @@
...
+
+  @Test
+  public void unobservableRequestFields() {
+    // updateActivityOptions returns a non-void response; stub so handle.updateOptions doesn't NPE.
+    when(genericClient.updateActivityOptions(any()))
+        .thenReturn(UpdateActivityExecutionOptionsResponse.getDefaultInstance());
+
+    UntypedActivityHandle handle = newHandle();
+
+    handle.pause("because");
```
maciejdudko: lol

maciejdudko (changes requested):
Looking good, although there are a few design choices that warrant more discussion.

temporal-sdk/src/main/java/io/temporal/client/UpdateActivityOptions.java
GregoryTravis: Done

temporal-sdk/src/main/java/io/temporal/common/interceptors/ActivityClientCallsInterceptor.java
GregoryTravis: @maciejdudko I also added one for Pause, even though there is just one field, for future proofing.

temporal-sdk/src/main/java/io/temporal/client/PauseActivityOptions.java
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(reason);
+  }
+
+  @Override
+  public String toString() {
+    return "PauseActivityOptions{" + "reason='" + reason + "'" + '}';
```
maciejdudko, line 84: Nit:

```suggestion
    return "PauseActivityOptions{" + "reason='" + reason + "'}";
```

temporal-sdk/src/main/java/io/temporal/client/ActivityExecutionOptions.java
```diff
@@ -0,0 +1,219 @@
...
+import javax.annotation.Nullable;
+
+/**
+ * Options for {@link UntypedActivityHandle#updateOptions(ActivityOptionsUpdate<?>...)}.
+ *
+ * <p>Only the fields that are explicitly set are sent to the server; a derived field mask ensures
+ * that unset fields are left unchanged (a partial update).
+ */
+@Experimental
+public final class ActivityExecutionOptions {
```
maciejdudko, line 17: Documentation is inaccurate with the new design.

temporal-sdk/src/main/java/io/temporal/client/ActivityExecutionOptions.java
```diff
@@ -0,0 +1,219 @@
...
+
+  public static Builder newBuilder() {
+    return new Builder();
+  }
```
maciejdudko, line 21: With the new design, this class doesn't need a builder. Instead it should have a package-visibility constructor that takes proto `ActivityOptions`.

temporal-sdk/src/main/java/io/temporal/client/ActivityOptionsKeys.java
maciejdudko, line 1: I think it would be less confusing if `ActivityOptionsKey` was an inner class of `ActivityOptionsUpdate`, and the static constants in `ActivityOptionsKeys` were also moved into `ActivityOptionsUpdate`. I'd also remove the methods `valueSet` and `valueUnset` from `ActivityOptionsUpdate` and keep them only on `ActivityOptionsKey`. (and rename them to `set` and `unset` for brevity.)

temporal-sdk/src/main/java/io/temporal/common/interceptors/ActivityClientCallsInterceptor.java
```diff
@@ -339,6 +379,121 @@ public String getReason() {
...
+
+  @Experimental
+  final class UnpauseActivityOutput {}
+
+  @Experimental
+  final class UpdateActivityOptionsInput {
+    private final String id;
+    private final @Nullable String runId;
+    private final ActivityOptions activityOptions;
+    private final FieldMask updateMask;
```
maciejdudko, line 445: `UpdateActivityOptionsInput` should receive `List<ActivityOptionsUpdate>` so that it can be intercepted, and the computation of `activityOptions` and `updateMask` should happen in `RootActivityClientInvoker`.

temporal-sdk/src/main/java/io/temporal/internal/client/ActivityHandleImpl.java
```diff
@@ -130,4 +155,110 @@ public void terminate(@Nullable String reason) {
...
+  /**
+   * Writes one option's value onto the request. The cast is safe because every key is created by
+   * {@link ActivityOptionsKeys} with the value type its path expects.
+   */
+  private static void applyUpdate(
+      ActivityOptions.Builder options, ActivityOptionsKey<?> key, Object value) {
+    switch (key.getName()) {
+      case "task_queue.name":
+        options.setTaskQueue(TaskQueue.newBuilder().setName((String) value).build());
+        break;
```
maciejdudko, line 225: The logic how to set the field in proto should be a method of `ActivityOptionsKey` itself, so that we don't have to match on the exact string value of the path. (Each key constant would have its own implementation.)

temporal-sdk/src/main/java/io/temporal/client/ActivityOptionsKey.java
```diff
@@ -0,0 +1,48 @@
...
+ * Typed key for one updatable activity option.
+ *
+ * <p>Use the keys on {@link ActivityOptionsKeys} rather than constructing these directly.
+ *
+ * @param <T> type of the option's value
+ */
+@Experimental
+public final class ActivityOptionsKey<T> {
+
+  private final String name;
```
maciejdudko, line 16: Rename `name` to `path`.

CHANGELOG.md
maciejdudko, line 1: Java SDK hasn't adopted the CHANGELOG.md file yet. Remove this.

maciejdudko (changes requested):

temporal-sdk/src/main/java/io/temporal/internal/client/ActivityHandleImpl.java
```diff
@@ -130,4 +144,60 @@ public void terminate(@Nullable String reason) {
...
+  }
+
+  @Override
+  public ActivityExecutionOptions updateOptions(ActivityOptionsUpdate<?>... updates) {
+    List<ActivityOptionsUpdate<?>> list = new ArrayList<>(updates.length);
+    for (ActivityOptionsUpdate<?> update : updates) {
+      if (update != null) {
+        list.add(update);
+      }
+    }
```
maciejdudko: Nit:
```suggestion
    List<ActivityOptionsUpdate<?>> list = new ArrayList<>(updates);
```

temporal-sdk/src/test/java/io/temporal/client/functional/StandaloneActivityTest.java
```diff
@@ -873,10 +872,8 @@ public void testDescribeLastFailureIsPopulatedDuringRetryBackoff() {
                             .setIncludeLastFailure(true)
                             .build());
             ActivityExecutionDescription desc =
-                new ActivityExecutionDescription(
-                    raw.getInfo(),
-                    DefaultDataConverter.STANDARD_INSTANCE,
-                    SDKTestWorkflowRule.NAMESPACE);
+                handle.describe(
+                    DescribeActivityOptions.newBuilder().setIncludeLastFailure(true).build());
```
maciejdudko: Leftover old query. Remove.

maciejdudko (approved):
2 small comments, but otherwise very good. Approved, LGTM!

PR #3014, opened 2026-08-18
Fix custom slot supplier SlotInfo fields

### LINKED ISSUE #2650: Ensure tests exist to confirm custom slot supplier slot info has proper fields

See https://github.com/temporalio/features/issues/672

baekgyu-kim:
## What was changed
- Corrected `WorkflowSlotInfo` to report the normal task queue name for both normal and sticky workflow tasks.
- Ensured synchronous sticky polling builds `WorkflowSlotInfo` from the request that was actually polled.
- Corrected worker build ID reporting for deployment options, legacy worker version capabilities, and binary checksum configurations.
- Added custom slot supplier coverage for workflow, activity, local activity, and Nexus `SlotInfo` fields, including synchronous/asynchronous and normal/sticky workflow polling.

## Why?
Custom slot suppliers could receive an empty or incorrect task queue, an incorrect sticky queue flag, or a missing/wrong worker build ID. These changes ensure that all supported `SlotInfo` types expose accurate worker and task information.

## Checklist

1. Closes #2650

2. How was this tested:

   - `./gradlew --offline spotlessApply`
   - `git diff --check`
   - `WorkflowSlotInfoTest`
   - `SlotInfoTest`
   - `SlotSupplierTest`
   - `WorkerVersioningTest`
   - `WorkerOptionsTest`

   External-server-only versioning cases were skipped by their existing test conditions.

3. Any docs updates needed?
   No. This change corrects existing `SlotInfo` behavior and adds test coverage without changing the public API surface.

### DIFF
--- a/temporal-sdk/src/test/java/io/temporal/internal/worker/WorkflowSlotInfoTest.java
+++ b/temporal-sdk/src/test/java/io/temporal/internal/worker/WorkflowSlotInfoTest.java
@@ -1,0 +1,227 @@
+package io.temporal.internal.worker;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertSame;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+import com.google.common.util.concurrent.Futures;
+import com.google.protobuf.ByteString;
+import com.uber.m3.tally.NoopScope;
+import io.temporal.api.common.v1.WorkerVersionCapabilities;
+import io.temporal.api.common.v1.WorkflowExecution;
+import io.temporal.api.common.v1.WorkflowType;
+import io.temporal.api.enums.v1.TaskQueueKind;
+import io.temporal.api.taskqueue.v1.TaskQueue;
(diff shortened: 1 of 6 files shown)


temporal-sdk/src/test/java/io/temporal/internal/worker/WorkflowSlotInfoTest.java
```diff
@@ -0,0 +1,238 @@
...
+    assertEquals(WORKER_IDENTITY, slotInfo.getWorkerIdentity());
+    assertEquals(WORKER_BUILD_ID, slotInfo.getWorkerBuildId());
+    if (expectedSticky) {
+      assertTrue(slotInfo.isFromStickyQueue());
+    } else {
+      assertFalse(slotInfo.isFromStickyQueue());
+    }
+  }
+
+  private static final class RecordingSlotSupplier implements SlotSupplier<WorkflowSlotInfo> {
```
Sushisource: There are two copies of this (I think there might actually already be another too, that existed before this PR). We should just have all the tests re-use the same impl.

Sushisource (commented):
Overall this makes sense, thanks for the fix. Just one comment.

temporal-sdk/src/test/java/io/temporal/internal/worker/WorkflowSlotInfoTest.java
baekgyu-kim: Hi @Sushisource, thank you for catching this.
I’ve moved `RecordingSlotSupplier` to a shared test utility so both tests reuse the same implementation. 
Could you please take another look? Thanks again!

Sushisource (approved):

PR #3016, opened 2026-08-18
External Storage Integration: Lazy resolving references, general refactoring

cconstable:
## What was changed
- Renames `ExternalStorageMessageTransformer` to `ExternalStorage`.
- Added a `ExternalStorageDataConverter`: an internal data converter used consolidating extstore usage.

## Why?
These changes support the following PRs:
1. Workflow worker integration https://github.com/temporalio/sdk-java/pull/3017
2. Nexus worker integration https://github.com/temporalio/sdk-java/pull/3018
3. Activity worker and client integration https://github.com/temporalio/sdk-java/pull/3020
4. E2E extstore tests

## Checklist
- Added tests


temporal-sdk/src/main/java/io/temporal/internal/worker/SingleWorkerOptions.java
```diff
@@ -45,6 +47,7 @@ public static final class Builder {
     private boolean allowActivityHeartbeatDuringShutdown;
     private String workerControlTaskQueue;
     private PreferredVersionProvider preferredVersionProvider;
+    private @Nullable ExternalStorage externalStorage;
```
cconstable: This doesn't have any callers in this PR but three of the PRs that build off this will use it. Included it here since those PRs are being up together.

temporal-sdk/src/main/java/io/temporal/common/converter/PayloadAndFailureDataConverter.java
```diff
@@ -71,6 +73,10 @@ public <T> T fromPayload(Payload payload, Class<T> valueClass, Type valueType)
       return (T) new RawValue(payload);
     }
 
+    if (ExternalStorageReferences.isReference(payload)) {
+      throw new ExternalStorageNotConfiguredException();
```
jmaeagle99: IIUC, this looks to be a DataConverter, which means it runs within the workflow code context. This means that this exception is handleable by user code. I think we need to move this to somewhere before the workflow code executes so we can fail the workflow task without allowing the user code to compensate.

temporal-sdk/src/main/java/io/temporal/internal/payload/storage/ExternalStorageResolvingDataConverter.java
```diff
@@ -0,0 +1,112 @@
...
+import io.temporal.payload.context.SerializationContext;
+import java.lang.reflect.Type;
+import java.util.Optional;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+/**
+ * A {@link DataConverter} that resolves external storage reference payloads before deserialization.
+ */
+public final class ExternalStorageResolvingDataConverter implements DataConverter {
```
jmaeagle99: Same concern in https://github.com/temporalio/sdk-java/pull/3016/changes#r3823884199

temporal-sdk/src/main/java/io/temporal/internal/payload/storage/ExternalStorageResolvingDataConverter.java
```diff
@@ -0,0 +1,112 @@
...
+  @Override
+  @Nonnull
+  public RuntimeException failureToException(@Nonnull Failure failure) {
+    return delegate().failureToException(failure);
+  }
+
+  @Nonnull
+  @Override
+  public DataConverter withContext(@Nonnull SerializationContext context) {
+    return new ExternalStorageResolvingDataConverter(delegate, externalStorage, context);
```
jmaeagle99: Shouldn't this call `withContext` on the `delegate` and then determine if a the outer `ExternalStorageResolvingDataConverter` truly needs to be reconstructed?

temporal-sdk/src/main/java/io/temporal/payload/storage/ExternalStorage.java
```diff
@@ -84,10 +98,21 @@ public Builder setPayloadSizeThreshold(int payloadSizeThreshold) {
       return this;
     }
 
+    /**
+     * Maximum number of payload lists visited concurrently while offloading or restoring the
+     * payloads of a single message. Must be at least 1. Defaults to 3.
```
jmaeagle99, line 102: Not sure a user is going to understand what is meant by "a single message". And "a single message" isn't quite the right scope from a implementation perspective. Maybe should describe this in terms of client operations and worker tasks.

temporal-sdk/src/main/java/io/temporal/internal/payload/storage/ExternalStorageRunner.java
```diff
@@ -0,0 +1,162 @@
...
+      Thread.currentThread().interrupt();
+      throw new CancellationException("External storage operation interrupted");
+    } catch (ExecutionException e) {
+      Throwable cause = e.getCause() != null ? e.getCause() : e;
+      Throwables.throwIfUnchecked(cause);
+      throw new CompletionException(cause);
+    }
+  }
+
+  <T extends Message> CompletableFuture<T> store(
```
jmaeagle99: Are there legitimate places where we need to visit on the fully constructed message instead of the build (the next overload)? I presume that the caller already created a builder, constructed the message, then this would effective recreate another builder, and reconstruct the message again. Might be perf issues. I would check to see if we can drop the message overloads and only use the builder overloads to force callers into the better performing algorithm.

jmaeagle99: This will not work for workflow task completions because the nested commands need to change the the context when they are encountered. Having an outer visitor doing that determination and then calling this method is probably okay.

jmaeagle99 (changes requested):

temporal-sdk/src/main/java/io/temporal/internal/payload/storage/ExternalStorageResolvingDataConverter.java
jmaeagle99: Maybe add a comment that this is used in non-workflow contexts.

cconstable: Removed this entirely in favor of just doing the lazy retrieving in the few spots we needed explicitly. After chatting this felt like too much abstraction.

temporal-sdk/src/main/java/io/temporal/common/converter/PayloadAndFailureDataConverter.java
cconstable: The context here is that if we missed integrating external storage into some piece of the SDK or a new feature was added that forgot to integrate external storage, this would catch it and throw a not configured error. It's purely a defensive fallback to make sure that SDK bugs don't appear as "your payloads can't be decoded/transformed" errors. 

I do think `ExternalStorageNotConfiguredException` is probably the wrong error to be throwing here because it's likely that it was configured right and the sdk just isn't using external storage. Should this be a "some feature you are using has not integrated external storage. please file a bug report" type thing?

temporal-sdk/src/main/java/io/temporal/payload/storage/ExternalStorage.java
cconstable: I think this will get cleaned up in the concurrency pass. I believe (technically) a single message is correct but I agree it's not a helpful way to describe it.

cconstable:
Pushed a few updates:
- I realized the names were a little off between this SDK and the Python/TS ones so I renamed `ExternalStorageOptions` to `ExternalStorage` and the previous `ExternalStorage` to `ExternalStorageRunner` (this feels a bit TS-flavored)
- I moved `ExternalStorage` config on to the `DataConverter` and added a default null value for it. Matches the other SDKs.
- Moved the `ExternalStorageGenericWorkflowClient` to the activity/client PR https://github.com/temporalio/sdk-java/pull/3020/changes. It's still awkward that some things in there eagerly fetch payloads and others just pass through but I'm not sure of a good path forward there.
- I removed the lazy data converter decorator in favor of just doing the lazy external storage retrieving for describe and listWorkflowExecutions inline. Seemed like the decorator was more abstraction than necessary.

temporal-sdk/src/main/java/io/temporal/internal/payload/storage/ExternalStorageRunner.java
cconstable: Correct but workflow task completions are routed through a different `store` overload that includes the MessageVisitor so that the context can be changed (it's the one earlier in the file).

```java
public <T extends Message> T storeBlocking(
      T message,
      @Nullable StorageDriverTargetInfo target,
      @Nullable MessageVisitor<StorageDriverTargetInfo> targetVisitor) { ... }
```

Regardless, it is confusing that there are so many overloads. I'll take a pass at consolidating them. I think we could just have one or two public ones. Some of the overloads above organically grew during the refactors and can be removed.

cconstable: 👍

temporal-sdk/src/main/java/io/temporal/common/converter/PayloadAndFailureDataConverter.java
cconstable: I added a new ExternalStorageUnhandledReferenceException for when a reference payload makes it to fromPayload without being retrieved. This is a guard against SDK features that fail or fail to correctly integrate external storage. Normal misconfiguration errors should be caught at a level above fromPayload and are done so in the PRs that build off this.

temporal-sdk/src/main/java/io/temporal/common/converter/DataConverter.java
```diff
@@ -202,6 +204,16 @@ default DataConverter withContext(@Nonnull SerializationContext context) {
     return this;
   }
 
+  /**
+   * External storage offloads large payloads. This should not be used from inside workflow code as
+   * it performs nondeterministic operations.
+   */
+  @Experimental
+  @Nullable
+  default ExternalStorage getExternalStorage() {
```
jmaeagle99: External storage should not be on `DataConverter` but instead on `WorkflowClientOptions`. The `IDataConverter` instance is accessible from within the workflow context, where we do not want external storage be accessible, let alone executable.

temporal-sdk/src/main/java/io/temporal/common/converter/PayloadAndFailureDataConverter.java
```diff
@@ -71,6 +75,10 @@ public <T> T fromPayload(Payload payload, Class<T> valueClass, Type valueType)
       return (T) new RawValue(payload);
     }
 
+    if (ExternalStorageReferences.isReference(payload)) {
```
jmaeagle99: I'm not really convinced this is necessary. If we forget to plumb through retrieval somewhere, then it should likely fail at deserialization time explaining that the ExternalStorageReference message couldn't be deserialized into the target type. I also don't think we should give customers something that they can handle and attempt to compensate for in their executions.

temporal-sdk/src/main/java/io/temporal/internal/payload/storage/ExternalStorageUnhandledReferenceException.java
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * Signals that an external storage reference reached ordinary payload conversion without being
+ * handled by any external storage integration.
+ */
+public final class ExternalStorageUnhandledReferenceException extends DataConverterException {
+  public ExternalStorageUnhandledReferenceException() {
+    super(
+        "[BUG] An external storage reference reached payload conversion without being handled. This"
+            + "is likely an SDK bug. Please file a bug report.");
```
jmaeagle99: nit: Concatenation bug, missing a space

temporal-sdk/src/main/java/io/temporal/internal/payload/storage/ExternalStorageRunner.java
```diff
@@ -0,0 +1,130 @@
...
+      Message.Builder builder,
+      @Nullable StorageDriverTargetInfo target,
+      @Nullable MessageVisitor<StorageDriverTargetInfo> targetVisitor,
+      CancellationToken<CancellationException> cancellationToken) {
+    getOrThrowIfCancelled(
+        PayloadVisitors.visit(builder, storeOptions(target, targetVisitor, cancellationToken)),
+        cancellationToken);
+  }
+
+  public <T extends Message> T retrieve(T message) {
```
jmaeagle99: This and retrieveAsync should passthrough a cancellation token from the caller.

temporal-sdk/src/test/java/io/temporal/common/converter/ExternalStorageReferenceGuardTest.java
```diff
@@ -0,0 +1,49 @@
...
+            .build();
+
+    ExternalStorageUnhandledReferenceException e =
+        assertThrows(
+            ExternalStorageUnhandledReferenceException.class,
+            () -> dataConverter.fromPayload(reference, String.class, String.class));
+  }
+
+  @Test
+  public void rawValueBypassesTheGuard() {
```
jmaeagle99: I don't understand what this test is trying to demonstrate or prove.

temporal-sdk/src/main/java/io/temporal/internal/payload/storage/ExternalStorageRunner.java
```diff
@@ -0,0 +1,130 @@
...
+                      found.completeExceptionally(new ExternalStorageNotConfiguredException());
+                      return found;
+                    }
+                  }
+                  return CompletableFuture.completedFuture(payloads);
+                })
+            .setSkipSearchAttributes(true)
+            .build();
+    try {
+      PayloadVisitors.visit(message.toBuilder(), options).join();
```
jmaeagle99, line 80: nit: Not for this PR, but I wonder if we can generate a non-mutating visitor and use that here.

temporal-sdk/src/main/java/io/temporal/internal/payload/storage/ExternalStorageRunner.java
```diff
@@ -0,0 +1,130 @@
...
+   * Throws {@link ExternalStorageNotConfiguredException} if {@code message} contains any reference
+   * payload. Used at inbound task boundaries when external storage is not configured.
+   */
+  public static void throwIfContainsReference(Message message) {
+    PayloadVisitorOptions<Void> options =
+        PayloadVisitorOptions.<Void>newBuilder(
+                (context, payloads) -> {
+                  for (Payload payload : payloads) {
+                    if (ExternalStorageReferences.isReference(payload)) {
+                      CompletableFuture<List<Payload>> found = new CompletableFuture<>();
```
jmaeagle99: Could this just raise an exception instead of returning an exceptionally completed future so that the walk it terminated? As it is now, everything in the graph is walked even if we saw the first set of payloads are storage references; it's a waste of compute when we already know that it's going to fail.

temporal-sdk/src/main/java/io/temporal/common/converter/DataConverter.java
cconstable: We already have externalStorage on the data converter in both the Python and TS implementations. Also `WorkflowClientOptions` wouldn't be sufficient. I think we would need to independently configure `ActivityClientOptions`, `NexusClientOptions`, etc. I also don't think there is anything preventing a user from just doing `S3Client().getObject(...)` inside a "workflow".

cconstable: After chatting: will move to `WorkflowClientOptions`,  `ActivityClientOptions`, `NexusClientOptions`, `ScheduleClientOptions`.

temporal-sdk/src/main/java/io/temporal/common/converter/PayloadAndFailureDataConverter.java
cconstable: will chat with @Quinn-With-Two-Ns or @maciejdudko

temporal-sdk/src/test/java/io/temporal/common/converter/ExternalStorageReferenceGuardTest.java
cconstable: At the very least, rename. Possibly remove.

temporal-sdk/src/main/java/io/temporal/internal/payload/storage/ExternalStorageRunner.java
cconstable: Yes, will throw ExternalStorageNotConfiguredException

temporal-sdk/src/main/java/io/temporal/common/converter/PayloadAndFailureDataConverter.java
cconstable: Potential directions: 
- wrap the final data converter with a "reference payload checking" data converter. ensure it fails the workflow task.
- remove check entirely

temporal-sdk/src/main/java/io/temporal/client/WorkflowClientOptions.java
```diff
@@ -86,6 +90,17 @@ public Builder setDataConverter(DataConverter dataConverter) {
       return this;
     }
 
+    /**
+     * External storage configuration uses to store/retrieve large payloads.
+     *
+     * <p>Defaults to null.
+     */
+    @Experimental
+    public Builder setExternalStorage(@Nullable ExternalStorage externalStorage) {
```
jmaeagle99, line 101: nit: Is there a way we could possibly mark this hidden from API / doc generation for now and adjust the comment to effectively say this is no-op at this time? I know we have the other PRs as follow ups to enable it, but just in case we need to pause and release Java in between. Obviously remove all of that when it does actually get attached in some usable manner.

temporal-sdk/src/main/java/io/temporal/internal/payload/storage/ExternalStorageReferences.java
```diff
@@ -9,7 +9,7 @@
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
-final class ExternalStorageReferences {
+public final class ExternalStorageReferences {
```
jmaeagle99: This can be non-public again, correct?

temporal-sdk/src/main/java/io/temporal/client/WorkflowClientOptions.java
```diff
@@ -86,6 +90,17 @@ public Builder setDataConverter(DataConverter dataConverter) {
       return this;
     }
 
+    /**
+     * External storage configuration uses to store/retrieve large payloads.
```
jmaeagle99:
```suggestion
     * External storage configuration used to store/retrieve large payloads.
```

temporal-sdk/src/main/java/io/temporal/internal/payload/storage/ExternalStorageDataConverter.java
```diff
@@ -0,0 +1,124 @@
...
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+/**
+ * A {@link DataConverter} that stores/retrieves payloads to/from external storage.
+ *
+ * <p>This This is an internal class that is not exposed to users or workflow code. The intent is to
+ * use this data converter to consolidate extstore usage within the SDK.
+ */
+public final class ExternalStorageDataConverter implements DataConverter {
```
jmaeagle99, line 22: Need to implement `Object[] fromPayloads(Optional<Payloads> content, Class<?>[] parameterTypes, Type[] genericParameterTypes)`.

temporal-sdk/src/main/java/io/temporal/internal/payload/storage/ExternalStorageDataConverter.java
```diff
@@ -0,0 +1,124 @@
...
+  private Payload retrieve(Payload payload) {
+    if (!ExternalStorageReferences.isReference(payload)) {
+      return payload;
+    }
+    return retrieveMessage(Payloads.newBuilder().addPayloads(payload).build()).getPayloads(0);
+  }
+
+  private Payloads store(Payloads payloads) {
+    Payloads.Builder builder = payloads.toBuilder();
+    externalStorage.store(builder, storageTarget, null, CancellationToken.none());
```
jmaeagle99, line 131: nit: I _think_ this needs to be lifted to be an instance field so that the owner of the data converter can cancel in flight operations, or maybe provided via something like a thread local so each call site can cancel. Then again, not sure if there is anything on any of the clients that would enable cancellation. Take this as a todo rather than fixing unless others object.

temporal-sdk/src/main/java/io/temporal/internal/payload/storage/ExternalStorageDataConverter.java
```diff
@@ -0,0 +1,124 @@
...
+import io.temporal.payload.storage.StorageDriverTargetInfo;
+import java.lang.reflect.Type;
+import java.util.Optional;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+/**
+ * A {@link DataConverter} that stores/retrieves payloads to/from external storage.
+ *
+ * <p>This This is an internal class that is not exposed to users or workflow code. The intent is to
```
jmaeagle99:
```suggestion
 * <p>This is an internal class that is not exposed to users or workflow code. The intent is to
```

temporal-sdk/src/main/java/io/temporal/client/WorkflowClientInternalImpl.java
```diff
@@ -106,6 +109,9 @@ public static WorkflowClient newInstance(
             .getOptions()
             .getMetricsScope()
             .tagged(MetricsTag.defaultTags(options.getNamespace()));
+    ExternalStorage externalStorageConfig = options.getExternalStorage();
```
jmaeagle99: nit: `externalStorage`

temporal-sdk/src/main/java/io/temporal/client/WorkflowClientInternalImpl.java
```diff
@@ -56,6 +58,7 @@ final class WorkflowClientInternalImpl implements WorkflowClient, WorkflowClient
   private final WorkerFactoryRegistry workerFactoryRegistry = new WorkerFactoryRegistry();
   private final String workerGroupingKey = java.util.UUID.randomUUID().toString();
   private final @Nullable HeartbeatManager heartbeatManager;
+  private final @Nullable ExternalStorageRunner externalStorage;
```
jmaeagle99: nit: `storageRunner` or `externalStorageRunner`

temporal-sdk/src/main/java/io/temporal/internal/worker/SingleWorkerOptions.java
```diff
@@ -252,6 +262,7 @@ public SingleWorkerOptions build() {
   private final boolean allowActivityHeartbeatDuringShutdown;
   private final String workerControlTaskQueue;
   private final PreferredVersionProvider preferredVersionProvider;
+  private final @Nullable ExternalStorageRunner externalStorage;
```
jmaeagle99: nit: `storageRunner` or `externalStorageRunner`

temporal-sdk/src/main/java/io/temporal/internal/client/WorkflowClientInternal.java
```diff
@@ -25,4 +26,7 @@ public interface WorkflowClientInternal {
 
   @Nullable
   HeartbeatManager getHeartbeatManager();
+
+  @Nullable
+  ExternalStorageRunner getExternalStorage();
```
jmaeagle99: `getStorageRunner` or `getExternalStorageRunner`

temporal-sdk/src/test/java/io/temporal/internal/payload/storage/ExternalStorageDataConverterTest.java
```diff
@@ -0,0 +1,169 @@
...
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import org.junit.Test;
+
+public class ExternalStorageDataConverterTest {
```
jmaeagle99, line 36: When you do implement the array form of fromPayloads, please add a test.

jmaeagle99, line 36: This we could add a test that wraps CodecDataConverter with a basic "encrypting/decrypting" codec to validate that drivers receive "encrypted" payloads?

temporal-sdk/src/test/java/io/temporal/internal/payload/storage/ExternalStorageRunnerTest.java
```diff
@@ -0,0 +1,363 @@
...
+    StartChildWorkflowExecutionCommandAttributes attrs =
+        stored.getStartChildWorkflowExecutionCommandAttributes();
+    assertNotNull(ExternalStorageReferences.tryParseReference(attrs.getInput().getPayloads(0)));
+    Payload indexed = attrs.getSearchAttributes().getIndexedFieldsOrThrow("k");
+    assertNull(ExternalStorageReferences.tryParseReference(indexed));
+    assertEquals(payload("indexed-value"), indexed);
+  }
+
+  @Test
+  public void throwIfContainsReferenceThrowsOnReference() throws Exception {
```
jmaeagle99, line 144: While this is a worthwhile test, I think there should be a test that does the same thing but for a non-`Payload`/`Payloads` message type. Maybe a WFT completion with a start activity task command that has a storage reference input.

jmaeagle99 (approved):
Most are nits except the `ExternalStorageDataConverter` decorator should implement all methods and some more test coverage. I also pointed out some naming inconsistencies, but please do take a pass over them.

temporal-sdk/src/main/java/io/temporal/client/WorkflowClientOptions.java
cconstable: AFAIK there isn't a way to hide this but I did add a n.b. note to the doc comment saying it's a no-op until integration is complete.

temporal-sdk/src/main/java/io/temporal/internal/payload/storage/ExternalStorageReferences.java
cconstable: Yes. I believe there may be a downstream test that is using this I'll have to update but I don't think this needs to be public.

temporal-sdk/src/main/java/io/temporal/internal/worker/SingleWorkerOptions.java
cconstable: refactoring leftover. agreed and updated.

temporal-sdk/src/main/java/io/temporal/internal/payload/storage/ExternalStorageDataConverter.java
cconstable: Yea, AFAICT we don't have anything that could actually enable this cancellation right now but eventually, when we do, I think the token will need to be passed in with a `withCancellationToken` or something.

PR #3017, opened 2026-08-18
External Storage Integration: WorkflowWorker, replay, history

cconstable:
## What was changed
- Integrated external storage into the replay handler and history iterator.
- `WorkflowWorker` handle + send response now use external storage.

## Why
- Workflow workers should utilize external storage and fetch payloads on replay.

## Checklist

- Added tests.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
@@ -23,6 +23,7 @@ import io.temporal.api.workflowservice.v1.*;
 import io.temporal.common.converter.DataConverter;
 import io.temporal.internal.common.ProtobufTimeUtils;
 import io.temporal.internal.common.WorkflowExecutionUtils;
+import io.temporal.internal.payload.storage.ExternalStorageRunner;
 import io.temporal.internal.worker.*;
 import io.temporal.payload.context.WorkflowSerializationContext;
 import io.temporal.serviceclient.MetricsTag;
@@ -77,6 +78,12 @@ public final class ReplayWorkflowTaskHandler implements WorkflowTaskHandler {
     String workflowType = workflowTask.getWorkflowType().getName();
     Scope metricsScope =
         options.getMetricsScope().tagged(ImmutableMap.of(MetricsTag.WORKFLOW_TYPE, workflowType));
+    ExternalStorageRunner externalStorage = options.getExternalStorage();
+    if (externalStorage == null) {
+      ExternalStorageRunner.throwIfContainsReference(workflowTask);
+    } else {
+      workflowTask = externalStorage.retrieve(workflowTask);
+    }
     return handleWorkflowTaskWithQuery(workflowTask.toBuilder(), metricsScope);
   }
 
@@ -94,7 +101,8 @@ public final class ReplayWorkflowTaskHandler implements WorkflowTaskHandler {
       logWorkflowTaskToBeProcessed(workflowTask, createdNew);
 
       ServiceWorkflowHistoryIterator historyIterator =
-          new ServiceWorkflowHistoryIterator(service, namespace, workflowTask, metricsScope);
+          new ServiceWorkflowHistoryIterator(
+              service, namespace, workflowTask, metricsScope, options.getExternalStorage());
       boolean finalCommand;
       Result result;
 
--- a/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
@@ -166,7 +166,7 @@ public final class ReplayWorkflowTaskHandler implements WorkflowTaskHandler {
       }
 
       return result;
-    } catch (InterruptedException e) {
+    } catch (InterruptedException | CancellationException e) {
       throw e;
     } catch (Throwable e) {
       // Note here that the executor might not be in the cache, even when the caching is on. In that
--- a/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
@@ -20,6 +20,7 @@ import io.temporal.api.sdk.v1.WorkflowTaskCompletedMetadata;
 import io.temporal.api.taskqueue.v1.StickyExecutionAttributes;
 import io.temporal.api.taskqueue.v1.TaskQueue;
 import io.temporal.api.workflowservice.v1.*;
+import io.temporal.common.CancellationToken;
 import io.temporal.common.converter.DataConverter;
 import io.temporal.internal.common.ProtobufTimeUtils;
 import io.temporal.internal.common.WorkflowExecutionUtils;
@@ -82,7 +83,7 @@ public final class ReplayWorkflowTaskHandler implements WorkflowTaskHandler {
     if (externalStorage == null) {
       ExternalStorageRunner.throwIfContainsReference(workflowTask);
     } else {
-      workflowTask = externalStorage.retrieve(workflowTask);
+      workflowTask = externalStorage.retrieve(workflowTask, CancellationToken.none());
     }
     return handleWorkflowTaskWithQuery(workflowTask.toBuilder(), metricsScope);
   }
@@ -407,7 +408,7 @@ public final class ReplayWorkflowTaskHandler implements WorkflowTaskHandler {
       if (externalStorage == null) {
         ExternalStorageRunner.throwIfContainsReference(getHistoryResponse);
       } else {
-        getHistoryResponse = externalStorage.retrieve(getHistoryResponse);
+        getHistoryResponse = externalStorage.retrieve(getHistoryResponse, CancellationToken.none());
       }
       workflowTask
           .setHistory(getHistoryResponse.getHistory())
--- a/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
@@ -403,6 +403,12 @@ public final class ReplayWorkflowTaskHandler implements WorkflowTaskHandler {
               .blockingStub()
               .withOption(METRICS_TAGS_CALL_OPTIONS_KEY, metricsScope)
               .getWorkflowExecutionHistory(getHistoryRequest);
+      ExternalStorageRunner externalStorage = options.getExternalStorage();
+      if (externalStorage == null) {
+        ExternalStorageRunner.throwIfContainsReference(getHistoryResponse);
+      } else {
+        getHistoryResponse = externalStorage.retrieve(getHistoryResponse);
+      }
       workflowTask
           .setHistory(getHistoryResponse.getHistory())
           .setNextPageToken(getHistoryResponse.getNextPageToken());
--- a/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
@@ -104,11 +104,11 @@ public final class ReplayWorkflowTaskHandler implements WorkflowTaskHandler {
     String workflowType = workflowTask.getWorkflowType().getName();
     Scope metricsScope =
         options.getMetricsScope().tagged(ImmutableMap.of(MetricsTag.WORKFLOW_TYPE, workflowType));
-    ExternalStorageRunner externalStorage = options.getExternalStorage();
-    if (externalStorage == null) {
+    ExternalStorageRunner externalStorageRunner = options.getExternalStorageRunner();
+    if (externalStorageRunner == null) {
       ExternalStorageRunner.throwIfContainsReference(workflowTask);
     } else {
-      workflowTask = externalStorage.retrieve(workflowTask, storageCancellation);
+      workflowTask = externalStorageRunner.retrieve(workflowTask, storageCancellation);
     }
     return handleWorkflowTaskWithQuery(workflowTask.toBuilder(), metricsScope);
   }
@@ -132,7 +132,7 @@ public final class ReplayWorkflowTaskHandler implements WorkflowTaskHandler {
               namespace,
               workflowTask,
               metricsScope,
-              options.getExternalStorage(),
+              options.getExternalStorageRunner(),
               storageCancellation);
       boolean finalCommand;
       Result result;
@@ -434,11 +434,12 @@ public final class ReplayWorkflowTaskHandler implements WorkflowTaskHandler {
               .blockingStub()
               .withOption(METRICS_TAGS_CALL_OPTIONS_KEY, metricsScope)
               .getWorkflowExecutionHistory(getHistoryRequest);
-      ExternalStorageRunner externalStorage = options.getExternalStorage();
-      if (externalStorage == null) {
+      ExternalStorageRunner externalStorageRunner = options.getExternalStorageRunner();
+      if (externalStorageRunner == null) {
         ExternalStorageRunner.throwIfContainsReference(getHistoryResponse);
       } else {
-        getHistoryResponse = externalStorage.retrieve(getHistoryResponse, storageCancellation);
+        getHistoryResponse =
+            externalStorageRunner.retrieve(getHistoryResponse, storageCancellation);
       }
       workflowTask
           .setHistory(getHistoryResponse.getHistory())
--- a/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
@@ -292,7 +292,8 @@ public final class ReplayWorkflowTaskHandler implements WorkflowTaskHandler {
         null,
         result.isFinalCommand(),
         eventIdSetHandle,
-        result.getApplyPostCompletionMetrics());
+        result.getApplyPostCompletionMetrics(),
+        result.isContinuedAsNew() ? null : result.getParentWorkflowExecution());
   }
 
   private Result failureToWFTResult(
--- a/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
@@ -104,12 +104,6 @@ public final class ReplayWorkflowTaskHandler implements WorkflowTaskHandler {
     String workflowType = workflowTask.getWorkflowType().getName();
     Scope metricsScope =
         options.getMetricsScope().tagged(ImmutableMap.of(MetricsTag.WORKFLOW_TYPE, workflowType));
-    ExternalStorageRunner externalStorageRunner = options.getExternalStorageRunner();
-    if (externalStorageRunner == null) {
-      ExternalStorageRunner.throwIfContainsReference(workflowTask);
-    } else {
-      workflowTask = externalStorageRunner.retrieve(workflowTask, storageCancellation);
-    }
     return handleWorkflowTaskWithQuery(workflowTask.toBuilder(), metricsScope);
   }
 
@@ -122,6 +116,7 @@ public final class ReplayWorkflowTaskHandler implements WorkflowTaskHandler {
     boolean useCache = stickyTaskQueue != null;
 
     try {
+      workflowTask = retrieveStoredPayloads(workflowTask);
       workflowRunTaskHandler =
           getOrCreateWorkflowExecutor(useCache, workflowTask, metricsScope, createdNew);
       logWorkflowTaskToBeProcessed(workflowTask, createdNew);
@@ -207,6 +202,16 @@ public final class ReplayWorkflowTaskHandler implements WorkflowTaskHandler {
         workflowRunTaskHandler.close();
       }
     }
+  }
+
+  private PollWorkflowTaskQueueResponse.Builder retrieveStoredPayloads(
+      PollWorkflowTaskQueueResponse.Builder workflowTask) {
+    ExternalStorageRunner externalStorageRunner = options.getExternalStorageRunner();
+    if (externalStorageRunner == null) {
+      ExternalStorageRunner.throwIfContainsReference(workflowTask.build());
+      return workflowTask;
+    }
+    return externalStorageRunner.retrieve(workflowTask.build(), storageCancellation).toBuilder();
   }
 
   private Result createCompletedWFTRequest(
--- a/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
@@ -20,7 +20,6 @@ import io.temporal.api.sdk.v1.WorkflowTaskCompletedMetadata;
 import io.temporal.api.taskqueue.v1.StickyExecutionAttributes;
 import io.temporal.api.taskqueue.v1.TaskQueue;
 import io.temporal.api.workflowservice.v1.*;
-import io.temporal.common.CancellationToken;
 import io.temporal.common.converter.DataConverter;
 import io.temporal.internal.common.ProtobufTimeUtils;
 import io.temporal.internal.common.WorkflowExecutionUtils;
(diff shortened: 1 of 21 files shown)


temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
```diff
@@ -381,6 +390,55 @@ public String toString() {
...
+    if (options.getExternalStorage() == null) {
+      return null;
+    }
+    return new StorageDriverWorkflowInfo(
+        namespace, execution.getWorkflowId(), execution.getRunId(), workflowType);
+  }
+
+  static StorageDriverTargetInfo refineStorageTarget(
+      String namespace, StorageDriverTargetInfo current, MessageOrBuilder message) {
+    if (message instanceof ScheduleActivityTaskCommandAttributesOrBuilder) {
```
jmaeagle99: This should be removed; workflow activities should home the storage target to the workflow execution.

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
```diff
@@ -381,6 +390,55 @@ public String toString() {
...
+  private StorageDriverTargetInfo workflowStorageTarget(
+      WorkflowExecution execution, String workflowType) {
+    if (options.getExternalStorage() == null) {
+      return null;
+    }
+    return new StorageDriverWorkflowInfo(
+        namespace, execution.getWorkflowId(), execution.getRunId(), workflowType);
+  }
+
+  static StorageDriverTargetInfo refineStorageTarget(
```
jmaeagle99: Need a branch for CaN and complete workflow execution. CaN should target the continued workflow. Complete workflow should target the parent if it has one and is not a CaN'd workflow, otherwise target self.

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
```diff
@@ -381,6 +391,57 @@ public String toString() {
...
```
cconstable: I'm not confident this covers all the cases. Looking into this.

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
```diff
@@ -61,6 +65,29 @@ public ReplayWorkflowTaskHandler(
...
+  public ReplayWorkflowTaskHandler(
+      String namespace,
+      ReplayWorkflowFactory asyncWorkflowFactory,
+      WorkflowExecutorCache cache,
+      SingleWorkerOptions options,
+      TaskQueue stickyTaskQueue,
+      Duration stickyTaskQueueScheduleToStartTimeout,
+      WorkflowServiceStubs service,
+      LocalActivityDispatcher localActivityDispatcher,
+      CancellationToken<CancellationException> storageCancellation) {
```
jmaeagle99: Instead of passing in a cancellation token to each handler type, can we update `ExternalStorageRunner` to have a shutdown method (or something similar) and have the owner of the runner shut it down from one place?

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
```diff
@@ -381,6 +393,97 @@ public String toString() {
...
+          return new StorageDriverWorkflowInfo(
+              namespace,
+              currentWorkflow.getId(),
+              null,
+              Strings.isNullOrEmpty(workflowType) ? currentWorkflow.getType() : workflowType);
+        }
+        return current;
+      case ATTRIBUTES_NOT_SET:
+      case START_TIMER_COMMAND_ATTRIBUTES:
+      case COMPLETE_WORKFLOW_EXECUTION_COMMAND_ATTRIBUTES:
```
jmaeagle99: This one should be retargeting to the parent if (1) the parent is known, and (2) is not a CaN'd workflow. Just like the other SDKs already do.

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
```diff
@@ -381,6 +393,97 @@ public String toString() {
...
+
+  static StorageDriverTargetInfo deriveStorageTarget(
+      String namespace, StorageDriverTargetInfo current, MessageOrBuilder message) {
+    if (!(message instanceof CommandOrBuilder)) {
+      return current;
+    }
+    CommandOrBuilder command = (CommandOrBuilder) message;
+    // Keep this exhaustive so new command attributes require an explicit target decision.
+    switch (command.getAttributesCase()) {
+      case SCHEDULE_ACTIVITY_TASK_COMMAND_ATTRIBUTES:
```
jmaeagle99: This should not be retargeted to the activity and should remain as the workflow. The external storage is tied to the lifetime of the workflow, not the activity.

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
```diff
@@ -77,6 +104,12 @@ public WorkflowTaskHandler.Result handleWorkflowTask(PollWorkflowTaskQueueRespon
     String workflowType = workflowTask.getWorkflowType().getName();
     Scope metricsScope =
         options.getMetricsScope().tagged(ImmutableMap.of(MetricsTag.WORKFLOW_TYPE, workflowType));
+    ExternalStorageRunner externalStorageRunner = options.getExternalStorageRunner();
```
jmaeagle99: These are executed outside of the try-catch that would proactively fail the workflow task; otherwise, an external storage failure is going to just end task processing and wait for the server to timeout.

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
```diff
@@ -381,6 +393,97 @@ public String toString() {
...
+      com.google.protobuf.Message.Builder builder,
+      @Nullable StorageDriverTargetInfo target,
+      @Nullable MessageVisitor<StorageDriverTargetInfo> targetVisitor) {
+    ExternalStorageRunner externalStorageRunner = options.getExternalStorageRunner();
+    if (externalStorageRunner == null) {
+      return;
+    }
+    try {
+      externalStorageRunner.store(builder, target, targetVisitor, storageCancellation);
+    } catch (Throwable e) {
```
jmaeagle99: Ignore the catch if it's cancelled?

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
```diff
@@ -381,6 +393,97 @@ public String toString() {
         options.getIdentity(), namespace, taskQueue);
   }
 
+  private void storeOutboundPayloads(
+      com.google.protobuf.Message.Builder builder, @Nullable StorageDriverTargetInfo target) {
+    storeOutboundPayloads(builder, target, null);
+  }
+
+  private void storeOutboundPayloads(
```
jmaeagle99, line 406: Not sure all of the callers of these two overloads are handling the storage failure case appropriately, but could be wrong.

temporal-sdk/src/test/java/io/temporal/internal/replay/ReplayWorkflowRunTaskHandlerTaskHandlerTests.java
```diff
@@ -121,6 +136,67 @@ public void workflowTaskFailOnIncompleteHistory() throws Throwable {
         result.getTaskFailed().getFailure().getMessage());
   }
 
+  @Test
+  public void resolvesExternalStorageReferencesInFetchedFullHistory() throws Throwable {
```
jmaeagle99, line 297: This seems to test successful external storage retrieval on history fetching. Could we add a test for when retrieval fails?

jmaeagle99, line 297: Can we add a similar test for when external storage retrieval runs on the workflow task itself? And for external storage failures?

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
jmaeagle99, line 980: Should have test coverage for this.

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
```diff
@@ -704,25 +867,30 @@ private void sendTaskFailed(
...
           () ->
               service
                   .blockingStub()
                   .withOption(METRICS_TAGS_CALL_OPTIONS_KEY, workflowTypeMetricsScope)
-                  .respondWorkflowTaskFailed(taskFailed.build()),
+                  .respondWorkflowTaskFailed(request),
           grpcRetryOptions);
     }
 
     private void sendDirectQueryCompletedResponse(
```
jmaeagle99, line 1020: Should have test coverage for this.

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
```diff
@@ -546,7 +673,37 @@ public void handle(WorkflowTask task) throws Exception {
                       currentTask.getTaskToken(),
                       taskFailedBuilder,
                       result.getRequestRetryOptions(),
-                      workflowTypeScope);
+                      workflowTypeScope,
+                      workflowStorageTarget(workflowExecution, workflowType));
+                } catch (ExternalStorageTaskFailure e) {
+                  if (currentTask.getAttempt() > 1) {
```
jmaeagle99: This is behaviorally different than the other SDKs. I think this came from duplicating the gRPC response path. However, they are different causes, where too large gRPC is not fixable but external storage failures likely are (missing replicated payloads, service temporarily faulted, etc).

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
```diff
@@ -61,6 +65,29 @@ public ReplayWorkflowTaskHandler(
...
+      String namespace,
+      ReplayWorkflowFactory asyncWorkflowFactory,
+      WorkflowExecutorCache cache,
+      SingleWorkerOptions options,
+      TaskQueue stickyTaskQueue,
+      Duration stickyTaskQueueScheduleToStartTimeout,
+      WorkflowServiceStubs service,
+      LocalActivityDispatcher localActivityDispatcher,
+      CancellationToken<CancellationException> storageCancellation) {
+    this.storageCancellation = storageCancellation;
```
jmaeagle99: I don't think there is test coverage for the shutdown / cancellation behavior.

temporal-sdk/src/test/java/io/temporal/internal/replay/ReplayWorkflowRunTaskHandlerTaskHandlerTests.java
jmaeagle99, line 297: This test probably passed even if external storage just passed the payload through without modification. So the test isn't demonstrating any capability of external storage.

temporal-sdk/src/test/java/io/temporal/internal/replay/ReplayWorkflowRunTaskHandlerTaskHandlerTests.java
```diff
@@ -231,4 +307,41 @@ private ReplayWorkflowFactory setUpMockWorkflowFactory() throws Throwable {
     when(mockWorkflow.getWorkflowContext()).thenReturn(mockWorkflowContext);
     return mockFactory;
   }
+
+  private static final class InMemoryStorageDriver implements StorageDriver {
```
jmaeagle99: There are several test driver implementations that nearly do the same thing. Can we make one and alter the behavior via parameters / options?

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
cconstable: Hmm yea I think this would simplify this. Will update unless I can think of a reason why we would need the token

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
cconstable: Yes, should be checking if it's a cancellation here.

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
cconstable: That makes sense. Will remove this check.

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
cconstable: Ok I think we do need to check for cancellation here but we also need we also need to throw the cancellation and then check it somewhere up the chain (`sendTaskCompleted`) and explicitly handle cancellation there (e.g. if the worker is shutting down we want to ignore the error). This was not very easy to understand.

cconstable: We can't just ignore it here because that would cause the completion event to be sent in `sendTaskCompleted`.

cconstable: Another issue: when the task fails from this it just says `Failed to send workflow task completion: External storage store failed` even if it was cancellation. That seems misleading.

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
cconstable: This is going to require some plumbing. I don't think we have access to that information here

temporal-sdk/src/main/java/io/temporal/internal/replay/ReplayWorkflowTaskHandler.java
cconstable: Thought about this a bit and I don't think we could add a shutdown method. A shutdown method would cancel everything and external storage is shared by across clients and workers (that is how we will eventually get our concurrency controls). One worker shutting down shouldn't cause external storage to shutdown in the others.

cconstable: We could potentially put a token on the worker options that can be used to then cancel all inflight extstore operations and get rid of the threading.

cconstable: We also still need to retain the ability to cancel individual extstore operations for things like heartbeats (they can be canceled)

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
```diff
@@ -381,6 +388,118 @@ public String toString() {
...
+    ExternalStorageRunner externalStorageRunner = options.getExternalStorageRunner();
+    if (externalStorageRunner == null) {
+      return;
+    }
+    try {
+      externalStorageRunner.store(builder, target, targetVisitor, options.getStorageCancellation());
+    } catch (CancellationException e) {
+      // if the worker is shutting down, extstore will throw a CancellationException and we need to
+      // rethrow it here so the handle() method can decide what to do.
+      throw e;
```
cconstable, line 419: Rather than absorbing a cancellation exception it's just rethrown. I think this covers all the cases we want and it allows the caller (`handle()`) to determine whether or not to actually send a task failure to the server... if we are in the middle of a forced shutdown it will see this cancellation error and just absorb it instead of throwing (which would result in a task failure being emitted).

temporal-sdk/src/main/java/io/temporal/internal/worker/BasePoller.java
```diff
@@ -176,6 +177,8 @@ static boolean shouldIgnoreDuringShutdown(Throwable ex) {
     ex instanceof RejectedExecutionException
         // if the worker thread gets InterruptedException - it's normal during shutdown
         || ex instanceof InterruptedException
+        || ex instanceof CancellationException
+        || ex.getCause() instanceof CancellationException
```
cconstable, line 181: During shutdown we shouldn't emit task failures due to things being cancelled because of the shutdown. At the moment, I don't think this actually catches any external storage stuff (since this check happens at a higher level than where the external storage exceptions are caught) but I think it's the "correct" behavior to add regardless and would prevent future issues. Would like some feedback on this @Quinn-With-Two-Ns or @maciejdudko

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
```diff
@@ -546,9 +690,41 @@ public void handle(WorkflowTask task) throws Exception {
...
+                      workflowTypeScope,
+                      workflowStorageTarget(workflowExecution, workflowType));
                 }
               }
+            } catch (CancellationException e) {
+              if (!options.getStorageCancellation().isCancellationRequested()) {
+                throw e;
+              }
+              log.trace("Abandoned a workflow task while the worker was shutting down", e);
+              return;
```
cconstable, line 766: When extstore throws a cancellation, if we are not forcefully shutting down then rethrow the error and report a failure. Otherwise, we should accept that the cancellation is just a normal part of the shutdown. The same logic was added below. It's possible that this cancellation exception might not originate from external storage in the future but I still think this is the correct course of action.

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
```diff
@@ -453,7 +572,26 @@ public void handle(WorkflowTask task) throws Exception {
...
+                      storageFailure(
+                          workflowExecution.getWorkflowId(), e, "Failed to send query response");
+                  RespondQueryTaskCompletedRequest.Builder queryFailedBuilder =
+                      RespondQueryTaskCompletedRequest.newBuilder()
+                          .setTaskToken(currentTask.getTaskToken())
+                          .setNamespace(namespace)
+                          .setCompletedType(QueryResultType.QUERY_RESULT_TYPE_FAILED)
+                          .setErrorMessage(failure.getMessage())
+                          .setFailure(failure);
+                  sendDirectQueryCompletedResponse(
```
jmaeagle99: I think we should send this without offloading to external storage. Recovering from an external storage error to attempt to send a failure message that external storage failed previous and use external storage on that message is likely going to fail again. This would also make it consistent with Go and Python. It should also clear `taskFailedCause` so that metrics are not emitted for a failure that was not actually reported.

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
```diff
@@ -546,9 +690,41 @@ public void handle(WorkflowTask task) throws Exception {
...
+                          "Failed to send workflow task %s",
+                          taskFailed == null ? "completion" : "failure");
+                  RespondWorkflowTaskFailedRequest.Builder storageFailedBuilder =
+                      RespondWorkflowTaskFailedRequest.newBuilder()
+                          .setFailure(
+                              storageFailure(workflowExecution.getWorkflowId(), e, messagePrefix))
+                          .setCause(
+                              WorkflowTaskFailedCause
+                                  .WORKFLOW_TASK_FAILED_CAUSE_WORKFLOW_WORKER_UNHANDLED_FAILURE);
+                  sendTaskFailed(
```
jmaeagle99: I think we should send this without offloading to external storage. Recovering from an external storage error to attempt to send a failure message that external storage failed previous and use external storage on that message is likely going to fail again. This would also make it consistent with Go and Python. It should also clear `taskFailedCause` so that metrics are not emitted for a failure that was not actually reported.

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
```diff
@@ -473,7 +611,10 @@ public void handle(WorkflowTask task) throws Exception {
                           .setErrorMessage(failure.getMessage())
                           .setFailure(failure);
                   sendDirectQueryCompletedResponse(
```
jmaeagle99: This should be wrapped in try-catch `ExternalStorageTaskFailure` and recover by just logging and abandoning sending to the server. See https://github.com/temporalio/sdk-go/blob/57cc5a7d619493f2e67166e2200e8e3758a72ae0/internal/internal_task_pollers.go#L764-L803 as an example. It should also clear `taskFailedCause` so that metrics are not emitted for a failure that was not actually reported.

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
jmaeagle99: This should be wrapped in try-catch `ExternalStorageTaskFailure` and recover by just logging and abandoning sending to the server. See https://github.com/temporalio/sdk-go/blob/57cc5a7d619493f2e67166e2200e8e3758a72ae0/internal/internal_task_pollers.go#L764-L803 as an example. It should also clear `taskFailedCause` so that metrics are not emitted for a failure that was not actually reported.

jmaeagle99 (approved):

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
jmaeagle99: I've called out in new comments where these should be handled.

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
cconstable: From the link you sent it looks like `emitFailMetric = true` is set and metrics are emitted for a failure in Go. Based on your feedback I'm guessing that is a bug. Also, we still could send a result to the server but just not use external storage... it would just be sending a message too large result right?

jmaeagle99: Yeah, that would be a bug. We shouldn't emit metrics when we fail to send message.

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
```diff
@@ -872,13 +868,23 @@ private RespondWorkflowTaskCompletedResponse sendTaskCompleted(
...
     }
 
-    @SuppressWarnings("deprecation")
     private void sendTaskFailed(
         ByteString taskToken,
         RespondWorkflowTaskFailedRequest.Builder taskFailed,
         RpcRetryOptions retryOptions,
         Scope workflowTypeMetricsScope,
-        @Nullable StorageDriverTargetInfo storageTarget) {
+        StorageDriverTargetInfo storageTarget) {
```
jmaeagle99: `@Nullable` annotation should be added back. `workflowStorageTarget` does return `null` when external storage is not configured. Not sure if the Java build is warning about that misalignment or not.

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
```diff
@@ -907,9 +912,17 @@ private void sendDirectQueryCompletedResponse(
         ByteString taskToken,
         RespondQueryTaskCompletedRequest.Builder queryCompleted,
         Scope workflowTypeMetricsScope,
-        @Nullable StorageDriverTargetInfo storageTarget) {
-      queryCompleted.setTaskToken(taskToken).setNamespace(namespace);
+        StorageDriverTargetInfo storageTarget) {
```
jmaeagle99: `@Nullable` annotation should be added back. `workflowStorageTarget` does return `null` when external storage is not configured. Not sure if the Java build is warning about that misalignment or not.

cconstable:
Resolving conflicts...

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
```diff
@@ -464,7 +582,23 @@ public void handle(WorkflowTask task) throws Exception {
               if (queryCompleted != null) {
                 try {
                   sendDirectQueryCompletedResponse(
-                      currentTask.getTaskToken(), queryCompleted.toBuilder(), workflowTypeScope);
+                      currentTask.getTaskToken(),
+                      queryCompleted.toBuilder(),
+                      workflowTypeScope,
+                      workflowStorageTarget(workflowExecution, workflowType));
+                } catch (ExternalStorageTaskFailure e) {
+                  Failure failure =
```
jmaeagle99, line 590: Probably should use `logAbandonedReport` here.

temporal-sdk/src/main/java/io/temporal/internal/worker/BasePoller.java
jmaeagle99: LGTM. Can adjust in future if need be.

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkflowWorker.java
cconstable: Well this is technically not abandoning the response because it calls `sendDirectQueryCompletedResponseWithoutExternalStorage`. I think that should work but I also think it's maybe a little inconsistent... sometimes we respond and sometimes we just log. Should we be sending a failure response when we can? In this case, I think it's safe to send but correct me if I'm wrong.

cconstable: Will defer

PR #3018, opened 2026-08-19
External Storage Integration: NexusWorker

cconstable:
## What was changed
- `NexusWorker` now stores and retrieves payloads.

## Why?
- Nexus workers should use external storage.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/client/NexusClientImpl.java
+++ b/temporal-sdk/src/main/java/io/temporal/client/NexusClientImpl.java
@@ -16,6 +16,9 @@ import io.temporal.internal.client.NexusOperationHandleImpl;
 import io.temporal.internal.client.RootNexusClientInvoker;
 import io.temporal.internal.client.external.GenericWorkflowClient;
 import io.temporal.internal.client.external.GenericWorkflowClientImpl;
+import io.temporal.internal.payload.storage.ExternalStorageDataConverter;
+import io.temporal.internal.payload.storage.ExternalStorageRunner;
+import io.temporal.payload.storage.ExternalStorage;
 import io.temporal.serviceclient.MetricsTag;
 import io.temporal.serviceclient.WorkflowServiceStubs;
 import java.util.List;
@@ -46,6 +49,16 @@ public class NexusClientImpl implements NexusClient {
     workflowServiceStubs =
         new NamespaceInjectWorkflowServiceStubs(workflowServiceStubs, options.getNamespace());
     this.workflowServiceStubs = workflowServiceStubs;
+    ExternalStorage externalStorageConfig = options.getExternalStorage();
+    if (externalStorageConfig != null) {
+      options =
+          NexusClientOptions.newBuilder(options)
+              .setDataConverter(
+                  new ExternalStorageDataConverter(
+                      options.getDataConverter(),
+                      ExternalStorageRunner.create(externalStorageConfig)))
+              .build();
+    }
     this.options = options;
     this.metricsScope =
         workflowServiceStubs
--- a/temporal-sdk/src/main/java/io/temporal/client/NexusClientImpl.java
+++ b/temporal-sdk/src/main/java/io/temporal/client/NexusClientImpl.java
@@ -12,13 +12,11 @@ import io.temporal.common.interceptors.NexusClientCallsInterceptor.ListNexusOper
 import io.temporal.common.interceptors.NexusClientInterceptor;
 import io.temporal.internal.WorkflowThreadMarker;
 import io.temporal.internal.client.NamespaceInjectWorkflowServiceStubs;
+import io.temporal.internal.client.NexusClientResolvedOptions;
 import io.temporal.internal.client.NexusOperationHandleImpl;
 import io.temporal.internal.client.RootNexusClientInvoker;
 import io.temporal.internal.client.external.GenericWorkflowClient;
 import io.temporal.internal.client.external.GenericWorkflowClientImpl;
-import io.temporal.internal.payload.storage.ExternalStorageDataConverter;
-import io.temporal.internal.payload.storage.ExternalStorageRunner;
-import io.temporal.payload.storage.ExternalStorage;
 import io.temporal.serviceclient.MetricsTag;
 import io.temporal.serviceclient.WorkflowServiceStubs;
 import java.util.List;
@@ -33,7 +31,7 @@ public class NexusClientImpl implements NexusClient {
   private static final Logger log = LoggerFactory.getLogger(NexusClientImpl.class);
 
   private final WorkflowServiceStubs workflowServiceStubs;
-  private final NexusClientOptions options;
+  private final NexusClientResolvedOptions options;
   private final GenericWorkflowClient genericClient;
   private final Scope metricsScope;
   private final NexusClientCallsInterceptor nexusClientCallsInvoker;
@@ -42,23 +40,13 @@ public class NexusClientImpl implements NexusClient {
   public static NexusClient newInstance(WorkflowServiceStubs service, NexusClientOptions options) {
     enforceNonWorkflowThread();
     return WorkflowThreadMarker.protectFromWorkflowThread(
-        new NexusClientImpl(service, options), NexusClient.class);
+        new NexusClientImpl(service, options.toResolvedOptions()), NexusClient.class);
   }
 
-  NexusClientImpl(WorkflowServiceStubs workflowServiceStubs, NexusClientOptions options) {
+  NexusClientImpl(WorkflowServiceStubs workflowServiceStubs, NexusClientResolvedOptions options) {
     workflowServiceStubs =
         new NamespaceInjectWorkflowServiceStubs(workflowServiceStubs, options.getNamespace());
     this.workflowServiceStubs = workflowServiceStubs;
-    ExternalStorage externalStorageConfig = options.getExternalStorage();
-    if (externalStorageConfig != null) {
-      options =
-          NexusClientOptions.newBuilder(options)
-              .setDataConverter(
-                  new ExternalStorageDataConverter(
-                      options.getDataConverter(),
-                      ExternalStorageRunner.create(externalStorageConfig)))
-              .build();
-    }
     this.options = options;
     this.metricsScope =
         workflowServiceStubs
--- a/temporal-sdk/src/main/java/io/temporal/client/NexusClientOptions.java
+++ b/temporal-sdk/src/main/java/io/temporal/client/NexusClientOptions.java
@@ -68,6 +68,10 @@ public class NexusClientOptions {
     return dataConverter;
   }
 
+  /**
+   * Get the external storage used to offload large operation payloads, or null if payloads are sent
+   * inline.
+   */
   @Nullable
   public ExternalStorage getExternalStorage() {
     return externalStorage;
@@ -160,6 +164,12 @@ public class NexusClientOptions {
       return this;
     }
 
+    /**
+     * Offload operation payloads that exceed the storage threshold to {@code externalStorage},
+     * sending a reference to the server in their place. The client wraps its {@link DataConverter}
+     * to store outbound payloads and resolve inbound references. Defaults to null, meaning all
+     * payloads are sent inline.
+     */
     public NexusClientOptions.Builder setExternalStorage(
         @Nullable ExternalStorage externalStorage) {
       this.externalStorage = externalStorage;
--- a/temporal-sdk/src/main/java/io/temporal/client/NexusClientOptions.java
+++ b/temporal-sdk/src/main/java/io/temporal/client/NexusClientOptions.java
@@ -4,9 +4,11 @@ import io.temporal.common.Experimental;
 import io.temporal.common.converter.DataConverter;
 import io.temporal.common.converter.GlobalDataConverter;
 import io.temporal.common.interceptors.NexusClientInterceptor;
+import io.temporal.payload.storage.ExternalStorage;
 import java.lang.management.ManagementFactory;
 import java.util.Collections;
 import java.util.List;
+import javax.annotation.Nullable;
 
 /**
  * Options that configure a {@link NexusClient} (and the service-bound clients it produces).
@@ -36,16 +38,19 @@ public class NexusClientOptions {
   private final List<NexusClientInterceptor> interceptors;
   private final DataConverter dataConverter;
   private final String identity;
+  private final @Nullable ExternalStorage externalStorage;
 
   private NexusClientOptions(
       String namespace,
       List<NexusClientInterceptor> interceptors,
       DataConverter dataConverter,
-      String identity) {
+      String identity,
+      @Nullable ExternalStorage externalStorage) {
     this.namespace = namespace;
     this.interceptors = interceptors;
     this.dataConverter = dataConverter;
     this.identity = identity;
+    this.externalStorage = externalStorage;
   }
 
   /** Get the namespace this client will operate on. */
@@ -61,6 +66,11 @@ public class NexusClientOptions {
   /** Get the data converter used to serialize Nexus operation inputs and deserialize results. */
   public DataConverter getDataConverter() {
     return dataConverter;
+  }
+
+  @Nullable
+  public ExternalStorage getExternalStorage() {
+    return externalStorage;
   }
 
   /**
@@ -101,6 +111,7 @@ public class NexusClientOptions {
     private List<NexusClientInterceptor> interceptors = Collections.emptyList();
     private DataConverter dataConverter = GlobalDataConverter.get();
     private String identity;
+    private ExternalStorage externalStorage;
 
     private Builder() {}
 
@@ -112,6 +123,7 @@ public class NexusClientOptions {
       interceptors = options.interceptors;
       dataConverter = options.dataConverter;
       identity = options.identity;
+      externalStorage = options.externalStorage;
     }
 
     /** Set the namespace this client will operate on. */
@@ -148,6 +160,12 @@ public class NexusClientOptions {
       return this;
     }
 
+    public NexusClientOptions.Builder setExternalStorage(
+        @Nullable ExternalStorage externalStorage) {
+      this.externalStorage = externalStorage;
+      return this;
+    }
+
     public NexusClientOptions build() {
       String resolvedIdentity =
           identity == null ? ManagementFactory.getRuntimeMXBean().getName() : identity;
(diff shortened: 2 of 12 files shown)


cconstable:
Missing: NexusClient integration. Adding now.

temporal-sdk/src/main/java/io/temporal/internal/worker/NexusWorker.java
```diff
@@ -512,17 +535,43 @@ private void sendReply(
...
+        ExternalStorageRunner.throwIfContainsReference(built);
+        return task;
+      }
+      return new NexusTask(
+          externalStorage.retrieve(built, storageCancellation.token()),
+          task.getPermit(),
+          task.getCompletionCallback());
+    }
+
+    private void storeOutbound(Message.Builder builder) {
```
Quinn-With-Two-Ns, line 637: Why does this `storeOutbound` work differently then [this one ](https://github.com/temporalio/sdk-java/pull/3020/changes#diff-4fb420884d68649ea5e9624f360f16e9ad3725dc08f692627f44a686af33ce15R88)? Should we just make one common helper ?

Quinn-With-Two-Ns:
Is there any tests showing external storage working for a Nexus operation? or what PR would that be in?

Quinn-With-Two-Ns:
Codex flagged these issues as well

```
I found two actionable issues. Your question about integration tests was
  warranted.

  1. High — Storage failures never produce a retryable Nexus task failure

     ExternalStorageTaskFailure says it is reported as a retryable handler error

     (https://github.com/temporalio/sdk-java/blob/94eaaca69c34a780b00803f1ddac129112ad2107/temporal-sdk/src/main/java/io/temporal/internal/worker/NexusWorker.java#L287-L295),
     but the class is unused. Both inbound retrieval and outbound storage errors

     (https://github.com/temporalio/sdk-java/blob/94eaaca69c34a780b00803f1ddac129112ad2107/temporal-sdk/src/main/java/io/temporal/internal/worker/NexusWorker.java#L553-L574)
     escape the worker. The generic poll executor only forwards them to the
     uncaught-exception handler; it does not call RespondNexusTaskFailed.

     Recovery therefore waits for task timeout/redelivery. For an outbound
     failure, the user handler has already executed, so redelivery can repeat
     its side effects. Catch storage errors and immediately send a small,
     inline, retryable INTERNAL handler failure—without passing that failure
     through external storage again. Add transient retrieve/store failure
     integration tests.

  2. Medium — External retrieval is excluded from the used-slot lifecycle

     markSlotUsed now happens after retrieveInboundPayloads

     (https://github.com/temporalio/sdk-java/blob/94eaaca69c34a780b00803f1ddac129112ad2107/temporal-sdk/src/main/java/io/temporal/internal/worker/NexusWorker.java#L325-L350).
     During a slow retrieval, used-slot metrics and custom supplier state
     under-report active work. If retrieval fails, the completion callback
     releases the permit as taskComplete, but TrackingSlotSupplier has no slot
     info because it was never marked used

     (https://github.com/temporalio/sdk-java/blob/94eaaca69c34a780b00803f1ddac129112ad2107/temporal-sdk/src/main/java/io/temporal/internal/worker/TrackingSlotSupplier.java#L70-L91).

     Mark the slot used before retrieval, or release failed pre-handler
     retrievals through an explicit never-used/error path.

```

temporal-sdk/src/main/java/io/temporal/internal/worker/NexusWorker.java
cconstable: Some callers of `storeOutbound` have already built messages and some have builders. Some also need a target (ActivityWorker) and others also need a visitor (the `WorkflowWorker` one). We could create a MessageOrBuilder type for this (maybe there already is one) and do some refactoring. Personally, I'd like that to be a follow-up if that is the direction we want to head in.

cconstable:
The codex finds are interesting. I think it's correct, it doesn't look like the we are catching failures. Adding some logic to send a nexus task failed so the server can retry instead of waiting for timeout.

cconstable:
Addressed the feedback and issues codex found. One thing to note @Quinn-With-Two-Ns is that when we fail the nexus task (because of an external storage issue which happens AFTER the task is handled + "complete") and respond with a `RespondNexusTaskFailed` this means that the server can retry which means there will be issues if the handler is not doing idempotent work. Is that ok?

Another interesting idea @jmaeagle99 is that we could have a fallback for scenarios like this where we would skip external storage since the nexus task already completed for scenarios where making an idempotent task is more of a detriment than just skipping external storage... so external storage would be configured with a "do your best to offload these payloads but if you cant (because the provider is down, etc) its ok to send them through as regular payloads".

cconstable:
Also added some tests @Quinn-With-Two-Ns to address your feedback.

cconstable:
The other PRs https://github.com/temporalio/sdk-java/pull/3017 and https://github.com/temporalio/sdk-java/pull/3020 also needed to handle the extstore store failures in the same way we were handling here. There is some duplication but I can follow up with a PR to consolidate all that once these PRs are merged.

temporal-sdk/src/main/java/io/temporal/internal/worker/NexusWorker.java
```diff
@@ -512,17 +536,71 @@ private void sendReply(
...
+      log.warn("External storage failed for a nexus task", e);
+      metricsScope
+          .tagged(
+              Collections.singletonMap(
+                  TASK_FAILURE_TYPE, MetricsTag.TASK_FAILURE_VALUE_HANDLER_ERROR_INTERNAL))
+          .counter(MetricsType.NEXUS_EXEC_FAILED_COUNTER)
+          .inc(1);
+      HandlerException handlerException =
+          new HandlerException(HandlerException.ErrorType.INTERNAL, "External storage failed", e);
+      sendReply(
```
jmaeagle99, line 601: I think this needs to handle `ExternalStorageTaskFailure` in case the failure cannot be externally stored and then submit a minimal `NexusTaskHandler.Result` from a small exception instance.

temporal-sdk/src/main/java/io/temporal/internal/worker/NexusWorker.java
```diff
@@ -304,28 +320,38 @@ private String getNexusTaskOperation(PollNexusTaskQueueResponseOrBuilder pollRes
...
+        slotSupplier.markSlotUsed(
+            new NexusSlotInfo(
+                service, operation, taskQueue, options.getIdentity(), options.getBuildId()),
+            task.getPermit());
+
+        try {
+          task = retrieveInboundPayloads(task);
+        } catch (Throwable e) {
+          taskFailed = true;
+          sendStorageFailure(
```
jmaeagle99, line 356: This is going to log "External storage failed for a nexus task" but external storage is not configured, so the log is misleading.

temporal-sdk/src/main/java/io/temporal/client/NexusClientImpl.java
```diff
@@ -46,6 +49,16 @@ public static NexusClient newInstance(WorkflowServiceStubs service, NexusClientO
     workflowServiceStubs =
         new NamespaceInjectWorkflowServiceStubs(workflowServiceStubs, options.getNamespace());
     this.workflowServiceStubs = workflowServiceStubs;
+    ExternalStorage externalStorageConfig = options.getExternalStorage();
+    if (externalStorageConfig != null) {
+      options =
+          NexusClientOptions.newBuilder(options)
```
jmaeagle99: Should this `setExternalStorage(null)`?

temporal-sdk/src/main/java/io/temporal/client/NexusClientOptions.java
```diff
@@ -63,6 +68,11 @@ public DataConverter getDataConverter() {
     return dataConverter;
   }
 
+  @Nullable
+  public ExternalStorage getExternalStorage() {
```
jmaeagle99, line 79: Needs javadoc

temporal-sdk/src/main/java/io/temporal/client/NexusClientOptions.java
```diff
@@ -148,14 +160,21 @@ public NexusClientOptions.Builder setIdentity(String identity) {
       return this;
     }
 
+    public NexusClientOptions.Builder setExternalStorage(
```
jmaeagle99, line 192: Needs javadoc

temporal-sdk/src/test/java/io/temporal/internal/worker/NexusWorkerTest.java
```diff
@@ -0,0 +1,49 @@
...
+import io.temporal.worker.tuning.SlotSupplier;
+import org.junit.Test;
+
+public class NexusWorkerTest {
+
+  @Test
+  public void interruptingShutdownCancelsInFlightStorage() throws Exception {
+    NexusWorker worker = worker();
+
+    worker.shutdown(new ShutdownManager(), true).get();
```
jmaeagle99: These should be closed, probably in a finally, in every test.

jmaeagle99:
> Addressed the feedback and issues codex found. One thing to note @Quinn-With-Two-Ns is that when we fail the nexus task (because of an external storage issue which happens AFTER the task is handled + "complete") and respond with a `RespondNexusTaskFailed` this means that the server can retry which means there will be issues if the handler is not doing idempotent work. Is that ok?
> 
> Another interesting idea @jmaeagle99 is that we could have a fallback for scenarios like this where we would skip external storage since the nexus task already completed for scenarios where making an idempotent task is more of a detriment than just skipping external storage... so external storage would be configured with a "do your best to offload these payloads but if you cant (because the provider is down, etc) its ok to send them through as regular payloads".

I don't think this is a new error mode for Nexus Operation handlers though. They already have at-least-once attempt semantics. The operation's work needs to be idempotent regardless of external storage.

On skipping external storage, if the result is too large, then the task _should_ be failed by the server (have to double check that) in a terminal way. Even if it doesn't, it will likely poison downstream consumers.

temporal-sdk/src/main/java/io/temporal/internal/worker/NexusWorker.java
```diff
@@ -304,28 +320,38 @@ private String getNexusTaskOperation(PollNexusTaskQueueResponseOrBuilder pollRes
...
+        String operation = getNexusTaskOperation(pollResponse);
+        if (!operation.isEmpty()) {
+          MDC.put(LoggerTag.NEXUS_OPERATION, operation);
+          metricsScope =
+              metricsScope.tagged(ImmutableMap.of(MetricsTag.NEXUS_OPERATION, operation));
+        }
+        slotSupplier.markSlotUsed(
+            new NexusSlotInfo(
+                service, operation, taskQueue, options.getIdentity(), options.getBuildId()),
+            task.getPermit());
```
jmaeagle99: This code and above used to be outside of the try block. Not sure it needs to be inside. It _might_ incidentally fix some existing problems. But its not the focus of this PR. If it doesn't need to be in the try to enable external storage, then let's move it back. If it is fixing something separately, then open a different PR for it.

temporal-sdk/src/main/java/io/temporal/client/NexusClientImpl.java
cconstable: I had to come back to this one... I think there is a bigger issue here is that an unwrapped data converter is making it's way to some nexus internals and then we have to decide if, when , and where to wrap it. Plus the external storage is still on the options object which is confusing.

Ideally, I think the nexus client impl and all the things that use the options internally would just see one resolved options object with a wrapped data converter and no superfluous fields.

cconstable: Making a change now to separate "what the user configures for options" and "what the clients actually use for options"

temporal-sdk/src/main/java/io/temporal/internal/worker/NexusWorker.java
cconstable: Moved back. Don't think it does.

temporal-sdk/src/main/java/io/temporal/internal/worker/NexusWorker.java
cconstable: `retrieveInboundPayloads` will throw `ExternalStorageNotConfiguredException` if extstore is not configured. Then `sendStorageFailure` will explicitly check for this and throw the correct message:

```java
String message =
          e instanceof ExternalStorageNotConfiguredException
              ? "Nexus task has externally stored payloads but this worker has no external storage"
                  + " configured"
              : "External storage failed for a nexus task";
```

temporal-sdk/src/main/java/io/temporal/internal/worker/NexusWorker.java
```diff
@@ -512,17 +548,102 @@ private void sendReply(
...
+                  + " configured"
+              : "External storage failed for a nexus task";
+      log.warn(message, e);
+      metricsScope
+          .tagged(
+              Collections.singletonMap(
+                  TASK_FAILURE_TYPE, MetricsTag.TASK_FAILURE_VALUE_HANDLER_ERROR_INTERNAL))
+          .counter(MetricsType.NEXUS_EXEC_FAILED_COUNTER)
+          .inc(1);
+      try {
```
jmaeagle99: We're a little inconsistent across SDKs about the number of recovery attempts. However, most of them make the last attempt without external storage involved.

So I think this should only make one recovery attempt. And the recovery attempt shouldn't use external storage.

temporal-sdk/src/test/java/io/temporal/internal/worker/NexusWorkerTest.java
```diff
@@ -0,0 +1,290 @@
...
+import java.util.UUID;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import javax.annotation.Nonnull;
+import org.junit.Test;
+import org.mockito.stubbing.Answer;
+
+public class NexusWorkerTest {
```
jmaeagle99, line 65: Could you add a test for when the first Nexus operation completion attempt fails due to external storage driver failure and ultimately succeeds to submit the completion?

temporal-sdk/src/test/java/io/temporal/internal/worker/NexusWorkerTest.java
```diff
@@ -0,0 +1,290 @@
...
+    if (inboundPayload != null) {
+      request.setPayload(inboundPayload);
+    }
+    return request;
+  }
+
+  /**
+   * A payload that looks like an external storage reference to {@code throwIfContainsReference}.
+   */
+  private static Payload storageReference() {
```
jmaeagle99, line 393: Unused?

temporal-sdk/src/main/java/io/temporal/internal/worker/NexusWorker.java
```diff
@@ -416,14 +451,14 @@ private boolean handleNexusTask(NexusTask task, Scope metricsScope) {
...
-        boolean supportTemporalFailure =
-            task.getResponse().getRequest().getCapabilities().getTemporalFailureResponses();
-        if (forceOldFailureFormat) {
-          supportTemporalFailure = false;
-        }
-
-        sendReply(taskToken, supportTemporalFailure, result, metricsScope);
+        sendReply(taskToken, supportsTemporalFailure(pollResponse), result, metricsScope);
+      } catch (ExternalStorageTaskFailure e) {
+        sendStorageFailure(
```
jmaeagle99, line 460: I think we're double emitting metrics here. It might be worthwhile to write a test that checks metric emission under external storage error conditions since we attempt recovery and retry.

temporal-sdk/src/main/java/io/temporal/internal/worker/NexusWorker.java
cconstable: I think that is reasonable. Updating.

temporal-sdk/src/test/java/io/temporal/internal/worker/NexusWorkerTest.java
cconstable: updated tests now use this.

temporal-sdk/src/test/java/io/temporal/internal/worker/NexusWorkerTest.java
```diff
@@ -99,31 +117,110 @@ public void storageCancelledByAForcedShutdownIsNotReportedAsATaskFailure() throw
...
+    Eventually.assertEventually(
+        Duration.ofSeconds(10),
+        () ->
+            reporter.assertCounter(
+                MetricsType.NEXUS_EXEC_FAILED_COUNTER,
+                execFailedTags(MetricsTag.TASK_FAILURE_VALUE_OPERATION_FAILED),
+                1));
+    reporter.assertNoMetric(
+        MetricsType.NEXUS_EXEC_FAILED_COUNTER,
+        execFailedTags(MetricsTag.TASK_FAILURE_VALUE_HANDLER_ERROR_INTERNAL));
```
cconstable, line 223: test to make sure we don't emit metrics @jmaeagle99

temporal-sdk/src/main/java/io/temporal/client/NexusClientOptions.java
```diff
@@ -71,6 +88,22 @@ public String getIdentity() {
...
+   * @return a {@link NexusClientResolvedOptions} instance with the resolved options
+   */
+  NexusClientResolvedOptions toResolvedOptions() {
+    DataConverter resolvedDataConverter = dataConverter;
+    if (externalStorage != null) {
+      resolvedDataConverter =
+          new ExternalStorageDataConverter(
+              resolvedDataConverter, ExternalStorageRunner.create(externalStorage));
+    }
+    return new NexusClientResolvedOptions(namespace, interceptors, resolvedDataConverter, identity);
```
cconstable, line 104: @jmaeagle99 the idea here is that the options passed to nexus client impl (and downstream abstractions) don't have user config awkwardly stapled on to them (e.g. `externalStorageOptions`). Putting the logic to resolve/consume things like externalStorage here was the most straightforward place to me. Happy to give this another pass but I think this is better than the alternative (keeping the misleading or confusing fields on the original options object).

jmaeagle99: I think we should invert this: put `from` on `NexusClientResolvedOptions` and let the public types not know about the internals.

jmaeagle99 (approved):

temporal-sdk/src/main/java/io/temporal/client/NexusClientOptions.java
cconstable: Yea that is better, updating.

jmaeagle99: Hmm, maybe not, I think some types would need to become public. So please ignore!

PR #3020, opened 2026-08-19
External Storage Integration: Activity worker, client

cconstable:
## What was changed
- Main integration points are in `WorkflowClientInternalImpl`, `HeartbeatContextImpl`, and `ActivityWorker`.
- Heartbeat store requests are canceled if they pass a timeout (as defined by the caller/heartbeat code).
- Lot's of plumbing in other adjacent files. 

## Why?
- Activity workers and clients should offload payloads with external storage.

## Checklist
- Added lots of tests.


temporal-sdk/src/main/java/io/temporal/internal/client/external/ExternalStorageGenericWorkflowClient.java
```diff
@@ -0,0 +1,344 @@
...
+      @Nonnull GetWorkflowExecutionHistoryRequest request) {
+    return next.getWorkflowExecutionHistoryAsync(request)
+        .thenComposeAsync(externalStorage::retrieveAsync);
+  }
+
+  @Override
+  public ListWorkflowExecutionsResponse listWorkflowExecutions(
+      ListWorkflowExecutionsRequest listRequest) {
+    return next.listWorkflowExecutions(listRequest);
+  }
```
cconstable: Addressed the previous concern of eagerly fetching external payloads https://github.com/temporalio/sdk-java/pull/2978#discussion_r3770732342. They are now loaded lazily.

temporal-sdk/src/main/java/io/temporal/internal/worker/ActivityWorker.java
```diff
@@ -354,7 +379,7 @@ private ActivityTaskHandler.Result handleActivity(ActivityTask task, Scope metri
       }
 
       try {
-        sendReply(taskToken, result, metricsScope);
+        sendReply(taskToken, result, metricsScope, activityStorageTarget(pollResponse));
```
cconstable, line 394: Previous comment about pivoting this target based on standalone vs workflow activity has been addressed https://github.com/temporalio/sdk-java/pull/2978#discussion_r3770775012

`activityStorageTarget` delegates to `storageTargetForActivityTask` (which can be seen above)

temporal-sdk/src/main/java/io/temporal/internal/client/external/ExternalStorageGenericWorkflowClient.java
```diff
@@ -0,0 +1,346 @@
...
+import javax.annotation.Nullable;
+
+/**
+ * Decorates a {@link GenericWorkflowClient} to offload outbound request payloads to external
+ * storage and restore inbound response payloads.
+ *
+ * <p>Only constructed when external storage is configured, so {@code externalStorage} is never
+ * null.
+ */
+public final class ExternalStorageGenericWorkflowClient implements GenericWorkflowClient {
```
cconstable: @jmaeagle99 could you reiterate your feedback on this abstraction? What was the concern?

jmaeagle99: Some functions on this decorator perform external store and retrieve and others do not, even if their responses carry payloads. It's not clear from the caller's perspective whether they need to also apply external storage or not depending on each call. And method additions for future capabilities will (1) likely be missed by the author or (2) be mis-categorized by the author and not accounted for by callers. All of it is more cognitive overhead that is unclear.

My suggestion is to create a `DataConverter` implementation that wraps whatever the callers get and have the caller just use that as they are using the existing `DataConverter` today and remove the workflow client decorator. This removes nearly all ambiguity and the callers do not have to answer "do I perform external storage or not" anymore; they answer the question "do I need to convert values or not", just as they answer it already.

cconstable: Based on the convo we had today I removed this file and added a `ExternalStorageDataConverter`.

temporal-sdk/src/main/java/io/temporal/internal/payload/storage/ExternalStorageDataConverter.java
```diff
@@ -0,0 +1,124 @@
...
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+/**
+ * A {@link DataConverter} that stores/retrieves payloads to/from external storage.
+ *
+ * <p>This This is an internal class that is not exposed to users or workflow code. The intent is to
+ * use this data converter to consolidate extstore usage within the SDK.
+ */
+public final class ExternalStorageDataConverter implements DataConverter {
```
cconstable, line 22: This is the new abstraction that replaces the previous External Storage client decorator. It wraps the data converter supplied to the WorkflowClient in `RootWorkflowClientInvoker`

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
```diff
@@ -50,25 +55,60 @@ public class RootWorkflowClientInvoker implements WorkflowClientCallsInterceptor
...
       WorkerFactoryRegistry workerFactoryRegistry) {
+    this(genericClient, clientOptions, workerFactoryRegistry, null);
+  }
+
+  public RootWorkflowClientInvoker(
+      GenericWorkflowClient genericClient,
+      WorkflowClientOptions clientOptions,
+      WorkerFactoryRegistry workerFactoryRegistry,
+      @Nullable ExternalStorageRunner externalStorage) {
+    this.externalStorage = externalStorage;
```
cconstable: This class inherits the external storage runner from `WorkflowClientInternalImpl` and then uses it to create the new `ExternalStorageDataConverter`

temporal-sdk/src/main/java/io/temporal/client/WorkflowClientInternalImpl.java
cconstable, line 787: This no longer runs through external storage as it was previously being handled by the client decorator. The worker still uses extstore when replaying it's own history (see https://github.com/temporalio/sdk-java/pull/3017/changes#diff-4928a50c4af33d91cd6b57a121d20ed2594cd314e118beb01c259d5e7427728c).

Is this just used for offline replay? Do we need to use extstore here?

Quinn-With-Two-Ns: I think we discussed offline we don't need to? At least that is what other SDKs did so we should be consistent

temporal-sdk/src/main/java/io/temporal/internal/client/external/ManualActivityCompletionClientImpl.java
```diff
@@ -152,7 +176,7 @@ public void fail(@Nonnull Throwable exception) {
       if (activityId == null) {
         throw new IllegalArgumentException("Either activity id or task token are required");
       }
-      RespondActivityTaskFailedByIdRequest request =
+      RespondActivityTaskFailedByIdRequest unstoredRequest =
```
Quinn-With-Two-Ns: nit: this request seems to follow a different pattern then all your others, may be cleaner to follow the convention of alway calling `request = storeOutbound(builder.Build())`, but just a nit feel free to ignore

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
```diff
@@ -50,25 +55,60 @@ public class RootWorkflowClientInvoker implements WorkflowClientCallsInterceptor
   private final WorkflowClientOptions clientOptions;
   private final EagerWorkflowTaskDispatcher eagerWorkflowTaskDispatcher;
   private final WorkflowClientRequestFactory requestsHelper;
+  private final @Nullable ExternalStorageRunner externalStorage;
 
   public RootWorkflowClientInvoker(
       GenericWorkflowClient genericClient,
       WorkflowClientOptions clientOptions,
```
Quinn-With-Two-Ns, line 61: nit: we could get rid of `WorkflowClientOptions` here so we know no one accidentally calls `clientOptions.getDataConverter()` now or in the future

temporal-testing/src/main/java/io/temporal/testing/TestActivityEnvironmentInternal.java
```diff
@@ -100,16 +101,19 @@ public TestActivityEnvironmentInternal(@Nullable TestEnvironmentOptions options)
     this.workflowServiceStubs =
         WorkflowServiceStubs.newServiceStubs(serviceStubsOptionsBuilder.build());
 
+    WorkflowClient client =
```
Quinn-With-Two-Ns, line 107: What is the change for?

Quinn-With-Two-Ns:
Couple nits, generally looks good. Going to do an AI review as well. Would want Justin to look as well of course

Quinn-With-Two-Ns:
Codex identified these two additional issues

```
  1. High — Manual completion uses the wrong storage target for workflow
     activities

     The factory paths

     (https://github.com/temporalio/sdk-java/blob/0c7b3087f75afaf6d30a8cc6034686e9f847a13e/temporal-sdk/src/main/java/io/temporal/internal/client/external/ManualActivityCompletionClientFactoryImpl.java#L49-L117)
     and local-manual-completion path

     (https://github.com/temporalio/sdk-java/blob/0c7b3087f75afaf6d30a8cc6034686e9f847a13e/temporal-sdk/src/main/java/io/temporal/internal/activity/ActivityExecutionContextImpl.java#L154-L170)
     always create StorageDriverActivityInfo. For workflow-scheduled activities,
     the established policy is to use StorageDriverWorkflowInfo, as
     ActivityWorker does

     (https://github.com/temporalio/sdk-java/blob/0c7b3087f75afaf6d30a8cc6034686e9f847a13e/temporal-sdk/src/main/java/io/temporal/internal/worker/ActivityWorker.java#L274-L289).

     This means synchronous and manual completion of the same workflow activity
     can select different drivers or storage layouts. The context/execution
     already contains enough information to distinguish workflow from standalone
     activities. I’d centralize that pivot and add tests covering both activity
     kinds across task-token, by-ID, and local manual completion.

  2. Medium — TestActivityEnvironment heartbeat listeners cannot decode
     offloaded details

     The environment now passes external storage into the heartbeat context

     (https://github.com/temporalio/sdk-java/blob/0c7b3087f75afaf6d30a8cc6034686e9f847a13e/temporal-testing/src/main/java/io/temporal/testing/TestActivityEnvironmentInternal.java#L104-L116),
     so sufficiently large heartbeat details reach the mock service as
     references. However, the heartbeat listener still decodes them with the
     ordinary DataConverter

     (https://github.com/temporalio/sdk-java/blob/0c7b3087f75afaf6d30a8cc6034686e9f847a13e/temporal-testing/src/main/java/io/temporal/testing/TestActivityEnvironmentInternal.java#L135-L152)
     without retrieving them first. With threshold 0, for example,
     setActivityHeartbeatListener will receive an external-storage reference and
     fail conversion. Resolve references before listener conversion and add an
     environment regression test.
```

At a quick glance they look legit, but please let me know.

temporal-testing/src/main/java/io/temporal/testing/TestActivityEnvironmentInternal.java
cconstable: `client.getInternal()` is used below now so we needed to pull out a reference.

cconstable:
The first issue that codex identified led me to do a little refactoring. I believe all the feedback has been addressed here.

temporal-sdk/src/main/java/io/temporal/internal/worker/ActivityWorker.java
```diff
@@ -331,6 +367,7 @@ public void handle(ActivityTask task) throws Exception {
     }
 
     private ActivityTaskHandler.Result handleActivity(ActivityTask task, Scope metricsScope) {
+      task = retrieveInboundPayloads(task);
```
jmaeagle99, line 370: If external storage throws, then there is no recovery. Exceptions here should proactively fail the activity task.

temporal-sdk/src/main/java/io/temporal/internal/worker/ActivityWorker.java
```diff
@@ -354,7 +391,10 @@ private ActivityTaskHandler.Result handleActivity(ActivityTask task, Scope metri
       }
 
       try {
-        sendReply(taskToken, result, metricsScope);
+        sendReply(taskToken, result, metricsScope, activityStorageTarget(pollResponse));
+      } catch (ExternalStorageTaskFailure e) {
+        sendStorageFailure(taskToken, pollResponse, metricsScope, e.getCause());
```
jmaeagle99, line 396: Should be consistent with the other PRs and send the storage failure without using external storage.

temporal-sdk/src/main/java/io/temporal/internal/worker/ActivityWorker.java
```diff
@@ -392,16 +432,20 @@ public Throwable wrapFailure(ActivityTask t, Throwable failure) {
...
-        RespondActivityTaskCompletedRequest request =
+        RespondActivityTaskCompletedRequest.Builder completedBuilder =
             taskCompleted.toBuilder()
                 .setTaskToken(taskToken)
                 .setIdentity(options.getIdentity())
                 .setNamespace(namespace)
-                .setWorkerVersion(options.workerVersionStamp())
-                .build();
+                .setWorkerVersion(options.workerVersionStamp());
+        storeOutboundPayloads(completedBuilder, storageTarget);
```
jmaeagle99, line 447: This and all other call sites should recover and send activity task failure without external storage.

temporal-sdk/src/main/java/io/temporal/internal/activity/HeartbeatContextImpl.java
```diff
@@ -330,16 +343,59 @@ private void checkHeartbeatTimeoutDeadlineLocked() {
...
+            Strings.emptyToNull(info.getWorkflowRunId()),
+            info.getWorkflowType())
+        .build();
+  }
+
+  /**
+   * Offloads large heartbeat payloads aborting if the store call runs longer than the heartbeat
+   * interval or if the activity is cancelled.
+   */
+  private void offloadHeartbeat(RecordActivityTaskHeartbeatRequest.Builder builder) {
```
jmaeagle99, line 369: I _believe_ this runs under the heartbeat lock. And the cancellation below looks to be relying on the heartbeat interval, which has a default of 30 seconds. This is could also starve out other operations like `getHeartbeatDetails` as they queue up. Also, if too many activities get stuck on external storage during heart beating, it might exhaust the thread pool and activity cancels may not run.

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
```diff
@@ -126,8 +152,8 @@ public WorkflowSignalOutput signal(WorkflowSignalInput input) {
         SignalWorkflowExecutionRequest.newBuilder()
             .setSignalName(input.getSignalName())
             .setWorkflowExecution(input.getWorkflowExecution())
-            .setIdentity(clientOptions.getIdentity())
-            .setNamespace(clientOptions.getNamespace())
+            .setIdentity(identity)
+            .setNamespace(namespace)
             .setRequestId(UUID.randomUUID().toString())
             .setHeader(HeaderUtils.toHeaderGrpc(input.getHeader(), null));
```
jmaeagle99, line 158: Need to convert headers. This cannot be done with the `DataConverter` because they are already `Payload`s. Haven't looked for all instances but pointing out this one category.

temporal-sdk/src/main/java/io/temporal/internal/client/WorkflowClientDataConverterFactory.java
```diff
@@ -0,0 +1,44 @@
...
+  private final boolean externalStorageConfigured;
+
+  WorkflowClientDataConverterFactory(
+      WorkflowClientOptions clientOptions, @Nullable ExternalStorageRunner externalStorage) {
+    this.namespace = clientOptions.getNamespace();
+    this.externalStorageConfigured = externalStorage != null;
+    this.baseConverter =
+        externalStorage == null
+            ? clientOptions.getDataConverter()
+            : new ExternalStorageDataConverter(clientOptions.getDataConverter(), externalStorage);
```
jmaeagle99, line 26: I'm thinking we might need to update the `ExternalStroageDataConverter` to be used regardless of it external storage is enabled or not AND perform the TMRPL1005 check when not configured. Otherwise clients calls will not deserialize properly and probably get a not as helpful message.

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
```diff
@@ -792,12 +777,14 @@ public CountWorkflowOutput countWorkflows(CountWorkflowsInput input) {
...
-            input.getQuery(), clientOptions.getNamespace(), input.getPageSize(), genericClient);
+            input.getQuery(), namespace, input.getPageSize(), genericClient);
     iterator.init();
     Iterator<WorkflowExecutionMetadata> wrappedIterator =
         Iterators.transform(
             iterator,
-            info -> new WorkflowExecutionMetadata(info, clientOptions.getDataConverter()));
+            info ->
+                new WorkflowExecutionMetadata(
+                    info, workflowConverter(info.getExecution(), info.getType().getName())));
```
jmaeagle99, line 787: This is gaining a `WorkflowSerializationContext` when it did not have one before. That would be changing the semantics of this API call. It is probably the right thing to do, but shouldn't be slipped in as an undocumented bug fix. So either construct one without the serialization context or we should document it.

temporal-sdk/src/main/java/io/temporal/internal/activity/HeartbeatContextImpl.java
cconstable: It is under the heartbeat lock which I believe is correct however new heartbeats were not properly cancelling the old ones and they were being scheduled via the `heartbeatExecutor` which would essentially do nothing if everything was already blocked.

temporal-sdk/src/main/java/io/temporal/internal/client/RootWorkflowClientInvoker.java
cconstable: this wasn't actually doing anything under the hood since the context that was passed in was never used (which I believe was a bug)

cconstable: now getMemo and getStaticSummary/getStaticDetails actually use the context that is passed in

temporal-sdk/src/main/java/io/temporal/client/WorkflowExecutionMetadata.java
```diff
@@ -123,11 +122,7 @@ public <T> T getMemo(String key, Class<T> valueClass, Type genericType) {
     if (memo == null) {
       return null;
     }
-    return dataConverter
-        .withContext(
```
jmaeagle99, line 127: Same as https://github.com/temporalio/sdk-java/pull/3020/changes#r3973626302

temporal-sdk/src/main/java/io/temporal/client/WorkflowExecutionDescription.java
```diff
@@ -29,15 +29,8 @@ public String getStaticSummary() {
     if (!response.getExecutionConfig().getUserMetadata().hasSummary()) {
       return null;
     }
-    return dataConverter
```
jmaeagle99, line 32: This feels like fixing a separate pre-existing bug. I think it should be restored and then addressed separately.

temporal-sdk/src/main/java/io/temporal/client/WorkflowExecutionDescription.java
```diff
@@ -51,15 +44,8 @@ public String getStaticDetails() {
     if (!response.getExecutionConfig().getUserMetadata().hasDetails()) {
       return null;
     }
-    return dataConverter
```
jmaeagle99, line 54: Same as https://github.com/temporalio/sdk-java/pull/3020/changes#r3973626302

temporal-sdk/src/main/java/io/temporal/internal/activity/HeartbeatContextImpl.java
jmaeagle99, line 319: Many of the external storage errors land here. We should probably handle them separately and fail the activity task immediately without external storage. Can take as a bug and fix in subsequent change.

jmaeagle99 (approved):

PR #3021, opened 2026-08-19
Add payload validation failure factory

bergundy:
## Summary

- add io.temporal.common.converter.PayloadValidationException.newPayloadValidationException
- return a non-retryable PayloadValidationError ApplicationFailure with violations as one details value
- replace positive Nexus test constructions with the public factory while retaining retryable compatibility coverage

## Testing

- temporal-sdk Spotless checks
- PayloadValidationException and PayloadSerializer focused tests
- OperationInputDeserializationErrorPropagationTest
- tests run with the repository Temurin 21 toolchain

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/common/converter/PayloadValidationException.java
+++ b/temporal-sdk/src/main/java/io/temporal/common/converter/PayloadValidationException.java
@@ -10,14 +10,14 @@ public final class PayloadValidationException {
   private PayloadValidationException() {}
 
   /**
-   * Creates a non-retryable failure containing the aggregated validation violations.
+   * Creates a non-retryable failure containing payload validation details.
    *
-   * <p>The violations are stored as a single details value and serialized by the configured {@link
+   * <p>The details are stored as a single value and serialized by the configured {@link
    * DataConverter}.
    *
-   * @param violations aggregated payload validation violations
+   * @param details payload validation details
    */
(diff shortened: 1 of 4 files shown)


temporal-sdk/src/main/java/io/temporal/common/converter/PayloadValidationException.java
```diff
@@ -0,0 +1,23 @@
...
+
+  /**
+   * Creates a non-retryable failure containing payload validation details.
+   *
+   * <p>The details are stored as a single value and serialized by the configured {@link
+   * DataConverter}.
+   *
+   * @param details payload validation details
+   */
+  public static ApplicationFailure create(Object details) {
```
Quinn-With-Two-Ns: This is a more common pattern for these type of factories in the Java SDK 
```suggestion
  public static ApplicationFailure newPayloadValidationException(Object details) {
```

Quinn-With-Two-Ns:
```suggestion
  public static ApplicationFailure create(Object... details) {
```
To be consistent with `ApplicationFailure`

bergundy: SGTM

Quinn-With-Two-Ns (approved):

PR #3024, opened 2026-08-20
Fix NPE when a workflow implementation uses a composed @WorkflowImpl annotation

muhrifqii:
## What was changed

`WorkersTemplate#configureWorkflowImplementationsByWorkerName` now uses `AnnotationUtils.findAnnotation` instead of `Class#getAnnotation` to look up `@WorkflowImpl`.
Added `AutoDiscoveryComposedAnnotationTest` and a dedicated `composedannotation` fixture package that reproduce the bug with a user-defined composed annotation on both a workflow and an activity implementation.

## Why?

If a workflow implementation class is annotated with a custom composed (meta) annotation, for example a user-defined annotation that is itself annotated with `@WorkflowImpl`, rather than `@WorkflowImpl` applied directly - worker auto-discovery threw an NPE.
`Class#getAnnotation` only resolves annotations present directly on the class, so it returned `null`, and the next line (`annotation.workers()`) NPE'd.

`@ActivityImpl`/`@NexusServiceImpl` lookups were already consistent throughout, which is why only workflows registered via `workers()` were affected.

## Checklist

1. Closes N/A - found while implementing `temporal-spring-boot-autoconfigure` in an application with a custom composed annotation.

2. How was this tested:
   Added `AutoDiscoveryComposedAnnotationTest`, confirmed it reproduces the NPE before the fix and passes after
   Ran the full `temporal-spring-boot-autoconfigure` test suite; all tests pass.

3. Any docs updates needed?
   No.

### DIFF
--- a/temporal-spring-boot-autoconfigure/src/test/java/io/temporal/spring/boot/autoconfigure/composedannotation/ComposedAnnotatedWorkflowImpl.java
+++ b/temporal-spring-boot-autoconfigure/src/test/java/io/temporal/spring/boot/autoconfigure/composedannotation/ComposedAnnotatedWorkflowImpl.java
@@ -13,7 +13,7 @@ public class ComposedAnnotatedWorkflowImpl implements ComposedAnnotatedWorkflow 
         Workflow.newActivityStub(
             ComposedAnnotatedActivity.class,
             ActivityOptions.newBuilder()
-                .setStartToCloseTimeout(Duration.ofSeconds(1))
+                .setStartToCloseTimeout(Duration.ofSeconds(10))
                 .validateAndBuildWithDefaults());
     return "composed:" + activity.execute(input);
   }
--- a/temporal-spring-boot-autoconfigure/src/test/java/io/temporal/spring/boot/autoconfigure/composedannotation/ComposedAnnotatedWorkflowImpl.java
+++ b/temporal-spring-boot-autoconfigure/src/test/java/io/temporal/spring/boot/autoconfigure/composedannotation/ComposedAnnotatedWorkflowImpl.java
@@ -1,0 +1,20 @@
+package io.temporal.spring.boot.autoconfigure.composedannotation;
+
+import io.temporal.activity.ActivityOptions;
+import io.temporal.workflow.Workflow;
+import java.time.Duration;
+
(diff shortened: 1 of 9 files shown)


temporal-spring-boot-autoconfigure/src/test/java/io/temporal/spring/boot/autoconfigure/composedannotation/ComposedAnnotatedWorkflowImpl.java
```diff
@@ -0,0 +1,20 @@
...
+@ComposedWorkflowImpl
+public class ComposedAnnotatedWorkflowImpl implements ComposedAnnotatedWorkflow {
+
+  @Override
+  public String execute(String input) {
+    ComposedAnnotatedActivity activity =
+        Workflow.newActivityStub(
+            ComposedAnnotatedActivity.class,
+            ActivityOptions.newBuilder()
+                .setStartToCloseTimeout(Duration.ofSeconds(1))
```
maciejdudko: Timeout needs to be longer so that it works reliably in CI environment.
```suggestion
                .setStartToCloseTimeout(Duration.ofSeconds(10))
```

maciejdudko (dismissed):
Hi @muhrifqii, thank you for your contribution and for reporting the problem. I have reviewed the PR and I think it's good, it just needs one correction to tests before we can merge it (see comment).

temporal-spring-boot-autoconfigure/src/test/java/io/temporal/spring/boot/autoconfigure/composedannotation/ComposedAnnotatedWorkflowImpl.java
muhrifqii: updated ya @maciejdudko , thank you for reviewing

Quinn-With-Two-Ns:
@maciejdudko I will note this change will only effect Springboot users, regular Java SDK users could still not compose annotations. Looks like we already support this for Activities and Nexus Operations so not saying we shouldn't but in principle we could support this for non-Springboot users too

Quinn-With-Two-Ns (approved):

PR #3025, opened 2026-08-20
Workflow streams user publish executor

diwu-sf:
## What was changed

Every `StreamPublisher` instance was creating its own single thread executor to run the background flush. If the worker has high concurrency and many publishers, more and more platform threads will be created (and these mostly just idle).

When using virtual threads, it's also not possible to override this executor with a shared one that uses virtual threads. Follow the pattern from workflow stream client and allow configuring a shared publish executor.

## Why?

To reduce the number of idle platform threads created for each stream publisher instance.

## Checklist

1. Closes 
-

2. How was this tested:

Tests added

3. Any docs updates needed?

Doc updated

### DIFF
--- a/contrib/temporal-workflowstreams/src/main/java/io/temporal/workflowstreams/internal/StreamPublisher.java
+++ b/contrib/temporal-workflowstreams/src/main/java/io/temporal/workflowstreams/internal/StreamPublisher.java
@@ -154,15 +154,23 @@ public final class StreamPublisher {
                 return t;
               });
     }
-    flushTask =
-        scheduler.scheduleWithFixedDelay(
-            this::backgroundFlush, batchIntervalMs, batchIntervalMs, TimeUnit.MILLISECONDS);
+    try {
+      flushTask =
+          scheduler.scheduleWithFixedDelay(
+              this::backgroundFlush, batchIntervalMs, batchIntervalMs, TimeUnit.MILLISECONDS);
+    } catch (RejectedExecutionException e) {
+      // A user-supplied executor was already shut down. Don't fail the publish call with the
+      // executor's own exception: items stay buffered for flush()/close() to drain on the
+      // caller's thread, as they do after a flush timeout stops the loop.
+    }
   }
 
   private void backgroundFlush() {
     synchronized (stateLock) {
-      if (loopStopped) {
-        // A timed-out flush stopped the loop; a task queued before that must not send.
+      if (loopStopped || closed) {
+        // The loop is stopped (a timed-out flush) or the publisher is closed. A task already
+        // queued at that point must not send: with a user-supplied executor nothing purges the
+        // queue, so this is the only thing keeping a flush from running after close() returned.
         return;
       }
     }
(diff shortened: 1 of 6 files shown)


contrib/temporal-workflowstreams/src/main/java/io/temporal/workflowstreams/internal/StreamPublisher.java
```diff
@@ -114,10 +147,15 @@ private void backgroundFlush() {
...
+      ScheduledFuture<?> toCancel;
       ScheduledExecutorService toStop;
       synchronized (stateLock) {
         deferredError = e;
-        toStop = scheduler;
+        toCancel = flushTask;
+        toStop = ownedSchedulerLocked();
+      }
+      if (toCancel != null) {
+        toCancel.cancel(false);
```
brianstrauch, line 193: After a background flush times out, this only cancels the periodic future when a user executor is supplied. Because the shared executor stays live, a later forceFlush=true or maxBatchSize trigger still submits backgroundFlush and sends the next buffered batch before throwDeferred surfaces the original error. I reproduced this with a user executor: let the first batch time out, clear the signal failure, then publish a second item with forceFlush=true; the second item is delivered. This also contradicts the new test comment that later items stay buffered until explicit flush or close. Please gate all triggered background flushes once deferredError is set, without shutting down the caller-owned executor, or otherwise surface the deferred error before another background send.

brianstrauch (changes requested):
Blocking on the publish-executor timeout behavior described in the inline comment. A caller-owned executor remains able to run forced or size-triggered background flushes after a deferred FlushTimeoutException, allowing later data to be sent before the failure is surfaced.

contrib/temporal-workflowstreams/src/main/java/io/temporal/workflowstreams/internal/StreamPublisher.java
brianstrauch: @diwu-sf I validated that this still races on the latest head (`32e3d8e0`) when the supplied publish executor has multiple threads. I added a diagnostic pause immediately after one background task dropped the expired pending batch and released `flushLock`, but before its catch block recorded `loopStopped`/`deferredError`. A second background task that had observed `loopStopped == false` then acquired `flushLock` and sent the later buffered batch. The correctness assertion failed with `a later batch was sent before the deferred timeout was recorded`.

The diagnostic pause only forces a legal scheduler interleaving; it does not alter the state or lock ordering. The timeout transition needs to remain serialized against other background flushes (or `doFlush` needs to recheck the stopped state after acquiring `flushLock`). The current single-thread-executor tests cannot exercise this case.

contrib/temporal-workflowstreams/src/main/java/io/temporal/workflowstreams/internal/StreamPublisher.java
```diff
@@ -81,14 +122,19 @@ public void publish(String topic, Object value, boolean forceFlush) {
...
+      trigger = (forceFlush || (maxBatchSize > 0 && buffer.size() >= maxBatchSize)) && !loopStopped;
       if (!closed) {
         ensureStartedLocked();
         toTrigger = scheduler;
       }
     }
     if (trigger && toTrigger != null) {
-      toTrigger.execute(this::backgroundFlush);
+      try {
+        toTrigger.execute(this::backgroundFlush);
```
brianstrauch, line 133: Every `forceFlush`/max-batch trigger submits another task to the shared executor, but tasks for one publisher serialize by blocking on `flushLock`. With a two-thread shared executor, one slow signal and one additional trigger from publisher A occupy both threads (one signaling, one waiting), so publisher B cannot flush even though only one signal is in flight. A hot publisher can also build an unbounded queue of redundant tasks. I reproduced this with two publishers sharing a two-thread `ScheduledThreadPoolExecutor`. Please coalesce per-publisher background flush requests (or otherwise ensure waiters do not occupy pool threads) so at most one task per publisher is scheduled or running.

PR #3028, opened 2026-08-21
Add Google Cloud Run CloudRunIdPlugin

seanbollin:
## What

Adds an experimental Google Cloud Run worker-identity plugin — the Cloud Run counterpart to the SDK's AWS Lambda module (`contrib/temporal-aws-lambda`) — for both Cloud Run worker pools and services. Cloud Run runs a long-lived container, so this is a small client plugin rather than a worker wrapper. It ships as its own module, `contrib/temporal-gcp-cloud-run-id` (package `io.temporal.gcp.cloudrun.id`).

> **Naming:** the plugin is `CloudRunIDPlugin` and the package is `io.temporal.gcp.cloudrun.id`. It sets the client identity — which the workers then inherit — and the name carries the Cloud Run context without relying on the full package path.

## What it does

Register `CloudRunIDPlugin` on your client:

```java
WorkflowClientOptions options =
    WorkflowClientOptions.newBuilder().setPlugins(new CloudRunIDPlugin()).build();
```

When the client connects, the plugin reads the current Cloud Run instance's metadata — the revision (`CLOUD_RUN_REVISION` / `K_REVISION`), the worker pool or service name (`CLOUD_RUN_WORKER_POOL` / `K_SERVICE`), and the unique instance id from the metadata server (`http://metadata.google.internal/computeMetadata/v1/instance/id`) — and sets the client identity to `<instanceId>@<revision>` when one isn't already set. Workers created from that client inherit it.

`GoogleCloudRunMetadata` / `identity()` are exposed for reading the values directly. Experimental.


contrib/temporal-gcp-cloud-run-worker-id/src/main/java/io/temporal/gcp/cloudrun/workerid/WorkerIdPlugin.java
```diff
@@ -0,0 +1,126 @@
...
+ *             .build());
+ *
+ * WorkerFactory factory = WorkerFactory.newInstance(client);
+ * Worker worker = factory.newWorker("my-task-queue");
+ * }</pre>
+ *
+ * <p><b>Advanced / testing:</b> {@link #WorkerIdPlugin(GoogleCloudRunMetadata)} accepts an
+ * already-resolved {@link GoogleCloudRunMetadata} instance, which skips the lazy fetch entirely.
+ * This is useful when the application fetches the metadata itself (for example to log it) or when a
+ * test injects fixed metadata.
```
Sushisource: This can just be on that constructor overload.

Sushisource (approved):
See note on python PR about possible rename

PR #3029, opened 2026-08-21
💥 Preserve non-cancellable timed promise waits during cancellation

### LINKED ISSUE #3026: Promise.get(timeout, unit) throws a misleading TimeoutException when the workflow is canceled
## Actual Behavior
cancelling a workflow that is blocked in Promise.get(timeout, unit) ends the wait immediately and reports it as a timeout

Attached you can find the reproduction 

[ExternalCancelRacingChildStartTest.java](https://github.com/user-attachments/files/31304312/ExternalCancelRacingChildStartTest.java)

## Specifications

  - Version: temporal sdk  1.38
  - Platform:

Quinn-With-Two-Ns:
## What was changed

This change adds the `DETACH_NON_CANCELLABLE_PROMISE_GET_TIMER` SDK flag and updates timed, non-cancellable `Promise.get(timeout, unit)` calls to run their timeout timer in a detached cancellation scope. The timer is canceled after the promise completes. Regression and replay tests cover promise completion, timeouts, activity cancellation, child workflow cancellation, and combinations with `CANCEL_AWAIT_TIMER_ON_CONDITION`. The activity tests use per-test latches for deterministic synchronization.

## Why?

  `Promise.get(timeout, unit)` is non-cancellable, but cancellation of its enclosing scope previously canceled its timeout timer. This ended the wait early and incorrectly surfaced a `TimeoutException`, potentially hiding the actual activity or child workflow cancellation. The new SDK flag preserves replay compatibility for existing workflow histories while allowing new workflows to use the corrected behaviour.

## 💥 Breaking changes

This change can affect workflow behaviour because timed, non-cancellable `Promise.get(timeout, unit)` calls will no longer return early with a `TimeoutException` when their enclosing cancellation scope is canceled. Instead, they will continue waiting until the promise completes, fails, or reaches its actual timeout. Workflows that currently catch the premature `TimeoutException` and use it to select a control flow path may therefore produce different results. The behaviour is protected by an SDK flag so existing workflow histories retain their original behaviour during replay, while new workflows can adopt the corrected semantics without causing nondeterminism.

## Checklist

1. Closes #3026

2. How was this tested:

3. Any docs updates needed?

---

> [!NOTE]
> **High Risk**
> Changes core workflow cancellation and timed-wait semantics on replay-sensitive promise paths; workflows that relied on premature timeouts after scope cancel may behave differently once the flag is enabled.
> 
> **Overview**
> Fixes incorrect **`TimeoutException`** from non-cancellable **`Promise.get(timeout, unit)`** when a parent **`CancellationScope`** is canceled: the scope was canceling the internal await timer and ending the wait early instead of letting the promise finish or fail.
> 
> Adds SDK flag **`DETACH_NON_CANCELLABLE_PROMISE_GET_TIMER`** (pending enable). When it is on in history, **`CompletablePromiseImpl`** runs the timed await inside a **detached cancellation scope** and cancels that scope when the wait ends. **`WorkflowInternal.checkSdkFlag`** exposes flag checks to sync code. Legacy behavior is unchanged when the flag is off.
> 
> Regression coverage includes parameterized integration tests (with **`CANCEL_AWAIT_TIMER_ON_CONDITION`** combinations) and **12 replay history fixtures** for promise, activity, child, and completion scenarios.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b0b0a89ebb31ca81f19f0c8bfe592c9258f2417a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


temporal-sdk/src/test/resources/promiseGetTimeoutCancellationActivityDetachedAwaitCancel.json
maciejdudko, line 1: Can we put resource files in a subdirectory?

temporal-sdk/src/main/java/io/temporal/internal/common/SdkFlag.java
```diff
@@ -36,6 +36,17 @@ public enum SdkFlag {
    * Bug: https://github.com/temporalio/sdk-java/issues/2796
    */
   VERSION_WAIT_FOR_MARKER(5),
+  /*
+   * Prevents cancellation of the enclosing CancellationScope from canceling the timeout timer
+   * used by non-cancellable Promise.get(timeout, unit).
+   *
+   * Introduced: 1.39.0
```
maciejdudko: Version needs to be updated before merge.

temporal-sdk/src/test/java/io/temporal/workflow/PromiseTimedGetCancellationTest.java
```diff
@@ -0,0 +1,435 @@
...
+  @Before
+  public void setUpSdkFlags() {
+    savedInitialFlags = WorkflowStateMachines.initialFlags;
+    List<SdkFlag> flags = new ArrayList<>(savedInitialFlags);
+    if (detachedTimer) {
+      flags.add(SdkFlag.DETACH_NON_CANCELLABLE_PROMISE_GET_TIMER);
+    }
+    if (cancelAwaitTimer) {
+      flags.add(SdkFlag.CANCEL_AWAIT_TIMER_ON_CONDITION);
+    }
```
maciejdudko, line 92: This needs to remove false flags as well to keep test working after the flags default to true.

maciejdudko (changes requested):

temporal-sdk/src/test/resources/promiseGetTimeoutCancellationActivityDetachedAwaitCancel.json
Quinn-With-Two-Ns: I put them in the folder all the existing json histories are

PR #3031, opened 2026-08-23
Rejecting the blank task queue names in WorkerFactory.newWorker()

### LINKED ISSUE #3030: WorkerFactory.newWorker() accepts blank/whitespace-only task queue names
## Bug found:

`WorkerFactory.newWorker(taskQueue)` validates against null and empty strings but not blank/whitespace-only strings:

Code snippet: 
Preconditions.checkArgument(
    !Strings.isNullOrEmpty(taskQueue), "taskQueue should not be an empty string");

This means `factory.newWorker("   ")` silently passes validation and creates a worker with a whitespace-only task queue name, which will fail at the server level with a cryptic error rather than a clear, early IllegalArgumentException.

Expected Behavior:
`newWorker("   ")` should throw `IllegalArgumentException` with a clear message.

Actual Behavior:
`newWorker("   ")` passes validation silently.

PS:
Other parts of this codebase already handle this correctly — 
`StartNexusOperationOptions.setId()` uses `trim().isEmpty()` for the same reason.

ad-kat:
## What was changed
`WorkerFactory.newWorker()` will now reject blank/whitespace-only task queue names (e.g. `"   "`) in addition to null and empty strings. The error message was updated from "should not be an empty string" to "should not be an empty or blank string" to reflect this. A new test class `WorkerFactoryValidationTest` covers null, empty, and blank task queue rejection using `TestWorkflowEnvironment`.

## Why?
`Strings.isNullOrEmpty("   ")` returns false, so whitespace-only task queue names silently passed validation and were forwarded to the Temporal server, producing a cryptic server-side error instead of a clear, early `IllegalArgumentException` at the call site.

This aligns `WorkerFactory` with the validation pattern already used elsewhere in this codebase. `StartNexusOperationOptions.setId()` uses `trim().isEmpty()` for the same reason.

## Checklist
1. Closes #3030
2. How was this tested: Added `WorkerFactoryValidationTest` with three unit    tests covering null, empty, and blank task queue names. All pass locally    using `TestWorkflowEnvironment` (no external service required).
3. Any docs updates needed? No.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/worker/WorkerFactory.java
+++ b/temporal-sdk/src/main/java/io/temporal/worker/WorkerFactory.java
@@ -181,7 +181,8 @@ public final class WorkerFactory {
    */
   public synchronized Worker newWorker(String taskQueue, WorkerOptions options) {
     Preconditions.checkArgument(
-        !Strings.isNullOrEmpty(taskQueue), "taskQueue should not be an empty string");
+        !Strings.isNullOrEmpty(taskQueue) && !taskQueue.trim().isEmpty(),
+        "taskQueue should not be an empty or blank string");
     Preconditions.checkState(
         state == State.Initial,
         String.format(statusErrorMessage, "create new worker", state.name(), State.Initial.name()));
--- a/temporal-sdk/src/test/java/io/temporal/worker/WorkerFactoryValidationTest.java
+++ b/temporal-sdk/src/test/java/io/temporal/worker/WorkerFactoryValidationTest.java
@@ -1,0 +1,38 @@
+package io.temporal.worker;
+
+import static org.junit.Assert.assertThrows;
+
+import io.temporal.testing.TestWorkflowEnvironment;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+public class WorkerFactoryValidationTest {
+
+  private TestWorkflowEnvironment env;
+
+  @Before
+  public void setUp() {
+    env = TestWorkflowEnvironment.newInstance();
+  }
+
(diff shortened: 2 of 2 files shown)


Quinn-With-Two-Ns:
To be clear you are saying the server reject a blank/whitespace-only task queue name? Can you share the error the server gave? I may want to open an issue on the server to improve the error message in this case

Quinn-With-Two-Ns:
Checking the server code I don't see any logic saying a whitespace-only task queue name is not allowed

ad-kat:
> To be clear you are saying the server reject a blank/whitespace-only task queue name? Can you share the error the server gave? I may want to open an issue on the server to improve the error message in this case

Here is a screenshot of the error I receive:
Screenshot 2026-08-27 204618
I tested this against a local Temporal server (v1.31.2). The server does not reject blank task queue names. The worker starts successfully but silently polls a whitespace-only queue that will never receive tasks, causing a silent misconfiguration with no error or warning.

My fix is a defensive validation at the SDK level to catch this before it reaches the server. I made this addition to be consistent with how StartNexusOperationOptions handles blank string inputs.

ad-kat:
> Checking the server code I don't see any logic saying a whitespace-only task queue name is not allowed

Temporal's own doc (at docs.temporal.io/best-practices/worker) describes this exact failure mode. Here's the screenshot for it (
Screenshot 2026-08-27 215849

I believe a blank task queue name fits this scenario. It'll be best to catch it at SDK level with clear error message before it becomes a silent prod failure."

maciejdudko:
Closing since the parent issue is closed.

PR #3039, opened 2026-08-26
Default Nexus client activity task queue to the worker task queue

JoshuaFrenchwood:
## What was changed

Previously, the taskQueue field was required to be defined in StartActivityOptions for TemporalNexusClient.startActivity.

This PR allows that field to be null, and uses the Workers taskQueue when it is not defined.

## Checklist

1. Closes 

2. How was this tested:

Removed the taskQueue field from the end to end test, and ran it locally.

3. Any docs updates needed?

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/client/StartActivityOptions.java
+++ b/temporal-sdk/src/main/java/io/temporal/client/StartActivityOptions.java
@@ -31,7 +31,7 @@ public final class StartActivityOptions {
 
   public static final class Builder {
     private String id;
-    private String taskQueue;
+    private @Nullable String taskQueue;
     private @Nullable Duration scheduleToCloseTimeout;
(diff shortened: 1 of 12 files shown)


Quinn-With-Two-Ns:
Can you double check any of the Nexus + SAA tests in this repo and if they are setting the task queue to please remove that. Don't want any AI using it as a bad reference

JoshuaFrenchwood:
> Can you double check any of the Nexus + SAA tests in this repo and if they are setting the task queue to please remove that. Don't want any AI using it as a bad reference

Yep, removed from other tests using Nexus and SAA.  Only one test called startActivity_preservesExplicitTaskQueue defines it, to test when its actually being passed in and is not overriden by the worker's taskQueue.

Quinn-With-Two-Ns (approved):

PR #3040, opened 2026-08-27
test: reproducer for terminated workflows retained by the sticky cache

xepozz:
## What was changed

Reproducer: externally terminated workflows are retained by the sticky cache until LRU capacity evicts them. Each dead run keeps its full cached state and its workflow thread — 300 terminated runs × 1 MB `char[]`: heap 18 → 510 MB, 164 workflow threads pinned.

## Why?

The server dispatches no workflow task on termination, so the worker never learns the run is closed; LRU pressure is the only reclaim path. Same behavior as temporalio/sdk-go#2629 (reproducer temporalio/sdk-go#2628). Related: temporalio/features#573, temporalio/sdk-php#635.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


Quinn-With-Two-Ns:
Closing as this is expected behaviour

PR #3045, opened 2026-08-28
Report missing activity timeout as BAD_REQUEST inside a Nexus operation handler

tekkaya:
## What was changed
`NexusTaskHandlerImpl.convertKnownFailures` now maps any `IllegalArgumentException` thrown by
Nexus operation handler code to `HandlerException(BAD_REQUEST)`.

## Why?
`StartActivityOptions.Builder.build()` throws a plain `IllegalArgumentException` when neither
`scheduleToCloseTimeout` nor `startToCloseTimeout` is set. Thrown from inside a Nexus operation
handler, that isn't a recognized Nexus error type, so it fell through to the generic
`HandlerException(INTERNAL)` catch-all in `NexusTaskHandlerImpl.handle()` — the caller then
retries until the operation's schedule-to-close timeout expires, instead of failing fast.

## Checklist

1. Closes: A gap found during Nexus SAA Test Plan execution for Temporal Java SDK

2. How was this tested:
Added `NexusTaskHandlerImplTest.startTaskHandlerMissingBothActivityTimeoutOptions`, which registers
a handler that throws `IllegalArgumentException`, and asserts the response comes back as `HandlerException(BAD_REQUEST)`.

3. Any docs updates needed?
No.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/nexus/NexusTaskHandlerImpl.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/nexus/NexusTaskHandlerImpl.java
@@ -200,6 +200,9 @@ public class NexusTaskHandlerImpl implements NexusTaskHandler {
       }
       throw new HandlerException(HandlerException.ErrorType.BAD_REQUEST, failure);
     }
+    if (failure instanceof IllegalArgumentException) {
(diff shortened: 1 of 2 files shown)


tekkaya:
This is an intended behaviour, closing this

PR #3048, opened 2026-08-28
Propagate Nexus links and request IDs to all activity starts in Nexus handlers

tekkaya:
## What was changed

`RootActivityClientInvoker.startActivity` now propagates the inbound Nexus task’s links and request ID to every activity start made on the operation-handler thread, including starts made through a raw `ActivityClient`.

`NexusOperationMetadata` remains narrowly scoped to the guarded backing start: it is the only mechanism that can attach a completion callback, because only that start is allowed to complete the Nexus operation. When present, its request ID takes precedence over the ambient inbound request ID.

## Why?

`TemporalNexusClient` permits only one guarded primitive activity start per operation invocation. A synchronous handler that starts another activity must use a raw `ActivityClient`; previously, those additional activity starts received neither inbound links nor redelivery-safe request-ID reuse.

Sharing the inbound request ID across all starts is deliberate. A handler that reuses an activity ID within one invocation can resolve a later start to the earlier run; deriving per-start IDs would require assuming the handler repeats starts in the same order after redelivery.

Restricting request-ID reuse to only the guarded start preserves the previous behavior, but does not address additional starts made through a raw `ActivityClient`. This approach aligns with the approved [sdk-go fix #2633](https://github.com/temporalio/sdk-go/pull/2633).

## Checklist

1. Closes the multiple-activity-start gap identified during Java SDK testing of Nexus + Standalone Activities.

2. Tested with:
   - `RootActivityClientInvokerTest` (unit)
   - `ActivityOperationLinkingTest` (functional; requires a real server and is gated by `SDKTestWorkflowRule.useExternalService`)

3. Documentation updates needed: No.


temporal-sdk/src/main/java/io/temporal/internal/client/RootActivityClientInvoker.java
```diff
@@ -64,14 +64,21 @@ public StartActivityOutput startActivity(StartActivityInput input) {
     NexusOperationMetadata nexusOperationMetadata =
         nexusContext == null ? null : nexusContext.getNexusOperationMetadata();
 
+    String requestId;
+    if (nexusOperationMetadata != null
```
Quinn-With-Two-Ns: Why do we need the `nexusOperationMetadata` code path here for links or request ID? Anytime `nexusOperationMetadata` is set `nexusContext` would also be present. I think we can simplify the link and request ID code to just use `nexusContext`. This applies here and where we set the links

tekkaya: Fixed, ptal

Quinn-With-Two-Ns (approved):

Quinn-With-Two-Ns:
Please feel free to merge whenever

Issue #3046, opened 2026-08-28
OpenTelemetry v2 plugin


patbeqo:
**Is your feature request related to a problem? Please describe.**

We want the Java SDK to support OpenTelemetry tracing with replay-safe span IDs, replay-safe span durations, and OpenTelemetry context propagation across Temporal boundaries.  This is the implementation tracker for [#2394: Native OpenTelemetry tracing support](https://github.com/temporalio/sdk-java/issues/2394).

**Describe the solution you'd like**

Add a new opt-in `temporal-opentelemetry-v2` artifact that follows the plugin conventions and replay behavior established by the Go, Python, and TypeScript SDKs. Keeping v2 separate preserves the existing integration and its Workflow History replay contract.

Scope:
- Configure the integration through the Java SDK's `SimplePlugin` API.
- Provide replay-safe span ID generation.
- Provide replay-safe span durations.
- Propagate OpenTelemetry context across Temporal boundaries.

**Describe alternatives you've considered**

- Update the existing `temporal-opentelemetry` artifact in place. That would break existing usages by changing its tracing behavior, so v2 will ship as a separate opt-in artifact.

**Additional context**

Reference implementations:
- Go plugin: https://github.com/temporalio/sdk-go/pull/2475
- Go Workflow prerequisites: https://github.com/temporalio/sdk-go/pull/2521
- Python: https://github.com/temporalio/sdk-python/pull/1314
- TypeScript: https://github.com/temporalio/sdk-typescript/pull/2142

PR #3049, opened 2026-08-28
Add getRandomStream and isSubjectToReplay

patbeqo:
## What changed?

- Added experimental named deterministic `Random` streams through `Workflow.getRandomStream(String)`.
- Streams are independent by name and from existing Workflow randomness.
- Streams replay deterministically, reseed after a Workflow Reset boundary, and use a new sequence for each Run.
- Added experimental `WorkflowUnsafe.isSubjectToReplay()`.
- Added coverage for seed stability, named-stream independence, replay, Reset, Continue-As-New, run and child isolation.

## Why?

OpenTelemetry and other integrations need replay-stable entropy without perturbing a Workflow's application random sequence. This is the prerequisite slice for #3046.

## Breaking changes?

None. The public APIs are additive and experimental.

## Server PR

None.

## Test plan

- [x] `mise exec -- ./gradlew :temporal-sdk:test --offline`
- [x] `mise exec -- ./gradlew test --offline`
- [x] `mise exec -- ./gradlew spotlessCheck`
- [x] `mise exec -- ./gradlew :temporal-sdk:javadoc --offline`


patbeqo:
@codex

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowRandomStreams.java
```diff
@@ -0,0 +1,105 @@
...
+    private int blockOffset;
+    private long counter;
+
+    private Stream(byte[] seed) {
+      reseed(seed);
+    }
+
+    @Override
+    public void nextBytes(byte[] bytes) {
+      Objects.requireNonNull(bytes, "bytes");
```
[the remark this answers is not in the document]
patbeqo: The caller should check if they are in a read only context before drawing.

patbeqo:
@codex

patbeqo:
@codex

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowRandomStreams.java
```diff
@@ -0,0 +1,31 @@
...
+
+  long deriveSeed(@Nonnull String runId, @Nonnull String name) {
+    // The separators keep ("ab", "c") from colliding with ("a", "bc")
+    String seed = String.join("\0", SEED_VERSION, runId, name);
+    HashCode hash = Hashing.sha256().hashString(seed, StandardCharsets.UTF_8);
+    return ByteBuffer.wrap(hash.asBytes()).getLong();
+  }
+
+  Random get(@Nonnull String runId, @Nonnull String name) {
+    return streams.computeIfAbsent(name, key -> new Random(deriveSeed(runId, key)));
```
DABH, line 25: [P2] Preserve more than 48 bits of generator state

Although the seed is hashed with SHA-256 and reduced to a `long`, `new Random(seed)` retains only 48 bits. At roughly 2^24 workflow-run/name streams the birthday probability of at least one duplicate state becomes material, and a collision produces the same entire sequence—not merely one repeated value. The stated consumer is OpenTelemetry IDs, so this can merge otherwise unrelated traces in a large installation; documenting that `Random` is not cryptographically secure does not address global ID uniqueness.

The earlier implementation on this branch used a full-seed SHA-counter stream. Please retain substantially wider deterministic state (for example through a `Random` subclass if the return type must remain `Random`) and pin that sequence in replay tests.

DABH (commented):
Ultrareviewed the exact head and the full stacked use in the OpenTelemetry v2 PRs. I found one scale/correctness issue inline.

I also revisited the resolved read-only-draw thread. The failure mode is real, but the final public Javadoc explicitly makes `WorkflowUnsafe.isSubjectToReplay()` gating a caller precondition, and the downstream ID generator observes it, so I am not treating a query that violates that contract as a separate blocker. A guarded API would still be safer.

Focused random-stream tests and the replay-divergence diagnostic passed locally; the worktree is clean. Current PR checks are green.

temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowInternal.java
```diff
@@ -744,6 +755,10 @@ public static Random newRandom() {
     return getRootWorkflowContext().newRandom();
   }
 
+  public static Random getRandomStream(String name) {
```
DABH, line 758: [P1] Guard getRandomStream against read-only contexts

Unlike `randomUUID()` and `newRandom()` directly above, this method has no `assertNotReadOnly`, and both resulting failure modes are worse than a `ReadOnlyException`.

Acquisition from a query handler fails the workflow task. `WorkflowStateMachines.getRandomStream` calls `checkEventLoopExecuting()`, which calls `WorkflowThread.await(...)` and therefore `currentThreadInternal()`; on the query thread that throws a raw `java.lang.Error("Called from non workflow or workflow callback thread")`. `ReplayWorkflowRunTaskHandler.executeQueries` catches only `Exception`, so the `Error` escapes. Reproduced on this head with a query handler that calls `Workflow.getRandomStream("x")`: history recorded `WORKFLOW_TASK_FAILED_CAUSE_WORKFLOW_WORKER_UNHANDLED_FAILURE`, the task retried, and the client query failed. On the real-server direct-query path the same `Error` fails the query and resets the sticky cache.

Draws in read-only or non-replayed contexts silently desynchronize the stream. A held stream drawn inside a query, a fresh stream drawn inside an `@UpdateValidatorMethod`, and a draw inside `Workflow.sideEffect` each produced live sequences that differ from what a cache-miss replay reconstructs (for example live `[e0, e2]` against replay `[e0, e1]`), with no exception, while `newRandom()` throws `ReadOnlyException` in the validator and side-effect cases. Callers also have no public read-only predicate to check: `WorkflowInternal.isReadOnly()` is package-private, and `isSubjectToReplay()` is a different, opt-in predicate.

Please add `assertNotReadOnly("random stream")` here and return a `Random` subclass whose `next(int)` asserts the same, so held streams are guarded as well. That subclass is also the natural home for the wider generator state requested in the other thread. A test that pins query, validator, and side-effect behavior would keep this from regressing.

temporal-sdk/src/main/java/io/temporal/internal/sync/WorkflowInternal.java
```diff
@@ -929,6 +944,10 @@ static void assertNotReadOnly(String action) {
     }
   }
 
+  public static boolean isSubjectToReplay() {
```
DABH, line 947: [P2] Match isReplaying() off workflow threads

`isReplaying()` uses `currentThreadInternalIfPresent()` and documents that it returns false outside workflow code, but `isSubjectToReplay()` goes through `getRootWorkflowContext()`, so `WorkflowUnsafe.isSubjectToReplay()` called from an activity, Nexus, or arbitrary thread throws a raw `java.lang.Error("Called from non workflow or workflow callback thread")`. The `WorkflowUnsafe` Javadoc does say "Must be called from Workflow code", but the sibling predicate this one is documented against behaves differently, and the downstream `ReplaySafeIdGenerator` in #3089 only avoids the `Error` by checking `isWorkflowThread()` first. Please either return false through `currentThreadInternalIfPresent()` like `isReplaying()` or throw a descriptive exception, and document the off-thread result.

DABH (commented):
Ultrareviewed the exact head d5bb22cf a second time with an independent trace and scratch reproductions. Two new findings are inline: unguarded `getRandomStream` fails the workflow task when called from a query handler and silently desynchronizes streams drawn in read-only or non-replayed contexts (P1), and `isSubjectToReplay()` throws a raw `Error` off workflow threads where `isReplaying()` returns false (P2). I have reopened the earlier Codex read-only thread with the reproductions, since its resolution did not change the code. The 48-bit generator state finding stands, and a `Random` subclass can host both the wider state and the read-only guard.

Seed derivation, memoization, the reset reseed ordering, continue-as-new and child isolation, and the `isSubjectToReplay` context flags all trace correctly. Smaller items, not blocking: `WorkflowRandomStreams.deriveSeed` accepts a null name and hashes the literal "null" instead of failing fast; the `isSubjectToReplay` Javadoc omits the preferred-version provider, and the `getRandomStream` Javadoc suggests gating draws on `isSubjectToReplay()` even though `Workflow.await` conditions are read-only yet subject to replay; no test pins behavior inside queries, validators, or side effects, there is no `WorkflowReplayer` history test, and the reset tests only run against an external service; the comments at `WorkflowRandomStreams.java:18` and `WorkflowRandomStreamTest.java:107-108` do not end with a period. Current PR checks are green.

temporal-sdk/src/main/java/io/temporal/internal/statemachines/WorkflowRandomStreams.java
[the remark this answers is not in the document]
DABH: Reopening this with reproductions on the exact head d5bb22cf, because the current code has no guard at all and the reply relies on a check that callers cannot make. A held stream drawn inside a query, a fresh `getRandomStream(...)` inside an update validator, and a draw inside `Workflow.sideEffect` each produced live sequences that differ from a cache-miss replay of the same run (for example `[e0, e2]` live against `[e0, e1]` on replay) with no exception, whereas `Workflow.newRandom()` throws `ReadOnlyException` in the validator and side-effect cases. There is no public read-only predicate, since `WorkflowInternal.isReadOnly()` is package-private, and `isSubjectToReplay()` is an opt-in and different predicate. Acquisition from a query is worse still: it fails the workflow task with a raw `Error`, detailed in the new [P1] comment on `WorkflowInternal.getRandomStream`. A guarded `Random` subclass fixes both the held-stream and the acquisition cases.

PR #3052, opened 2026-09-01
Report runtime environment information in worker heartbeats

Sushisource:
## What was changed

Worker heartbeats now include an `EnvironmentInfo` message describing the JVM version, detected hosting environments (Docker, Kubernetes, AWS Lambda/ECS, Google Cloud Run/App Engine, Azure App Service/Functions/Container Apps), and the OS platform (Linux/macOS/Windows version, architecture, and Windows CRT).

The information is attached to each worker's heartbeats until the server accepts one, then omitted. `HeartbeatManager` tracks pending workers and clears them only after a successful `recordWorkerHeartbeat`.

A new experimental `WorkflowClientOptions.Builder#setDisableWorkerEnvironmentInfo` turns the reporting off.

## Why?

Port of Core's temporalio/sdk-rust#1457 so all SDKs report the same analytical information about the environments they run in.

## Checklist

1. Closes N/A

2. How was this tested:

New unit tests in `WorkerEnvironmentInfoTest` cover live detection, hosting-environment env vars, cgroup parsing, and Java version parsing. `HeartbeatManagerTest` gained cases for retry-until-accepted delivery and for omitting the environment when disabled.

3. Any docs updates needed?

No.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/worker/HeartbeatManager.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/worker/HeartbeatManager.java
@@ -1,6 +1,5 @@
 package io.temporal.internal.worker;
 
-import io.temporal.api.worker.v1.EnvironmentInfo;
 import io.temporal.api.worker.v1.WorkerHeartbeat;
 import io.temporal.api.workflowservice.v1.RecordWorkerHeartbeatRequest;
 import io.temporal.serviceclient.WorkflowServiceStubs;
@@ -8,7 +7,6 @@ import java.time.Duration;
 import java.util.*;
 import java.util.concurrent.*;
 import java.util.function.Supplier;
-import javax.annotation.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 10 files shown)


temporal-sdk/src/main/java/io/temporal/internal/worker/HeartbeatManager.java
```diff
@@ -37,6 +37,19 @@ public HeartbeatManager(WorkflowServiceStubs service, String identity, Duration
...
+
+  /**
+   * @param onHeartbeatAccepted invoked, from the heartbeat thread, each time a heartbeat produced
+   *     by {@code callback} has been accepted by the server
+   */
+  public void registerWorker(
+      String namespace,
+      String workerInstanceKey,
+      Supplier<WorkerHeartbeat> callback,
+      Runnable onHeartbeatAccepted) {
```
maciejdudko, line 51: Nit: since it's an internal class, we don't need to keep backward compatibility and can add the new argument to the existing constructor.

temporal-sdk/src/main/java/io/temporal/internal/worker/WorkerEnvironmentInfo.java
```diff
@@ -0,0 +1,312 @@
...
+          Runtime.newBuilder()
+              .setType(RuntimeType.RUNTIME_TYPE_JVM)
+              .setVersion(nullToEmpty(System.getProperty("java.version"))));
+      builder.addAllHostingEnvironments(detectHostingEnvironments(System::getenv));
+      Platform platform = detectPlatform();
+      if (platform != null) {
+        builder.setPlatform(platform);
+      }
+    } catch (RuntimeException e) {
+      log.warn("Failed to detect worker environment information, reporting partial results", e);
```
maciejdudko, line 53: Nit: this warning seems uninteresting to users, I'd downgrade it to info log.

temporal-sdk/src/test/java/io/temporal/internal/worker/WorkerEnvironmentInfoTest.java
```diff
@@ -0,0 +1,113 @@
...
+        assertFalse(info.getPlatform().getLinux().getVersion().isEmpty());
+        break;
+      case MACOS:
+        assertEquals(expectedArchitecture, info.getPlatform().getMacos().getArchitecture());
+        break;
+      case WINDOWS:
+        assertEquals(expectedArchitecture, info.getPlatform().getWindows().getArchitecture());
+        break;
+      default:
+        fail("unexpected platform variant " + info.getPlatform().getVariantCase());
```
maciejdudko, line 42: Nit: if this ever gets triggered, it's most likely a user porting Temporal SDK to something wild. A test skip might be better than a test fail.

maciejdudko (approved):
A couple nits but overall looks good. LGTM!

temporal-sdk/src/test/java/io/temporal/internal/worker/WorkerEnvironmentInfoTest.java
Sushisource: Eh, I'm fine with that. They can then either fix it or skip the test as needed.

PR #3054, opened 2026-09-02
Bumping test CLI version

Evanthx:
This bumps the test CLI server version. That caused some failures in tests, so this is to fix the tests. Full comments left in to ensure that I am fixing the tests and not covering up an unwanted change.

Some of the changes needed are in https://github.com/temporalio/sdk-java/pull/3013 - those aren't duplicated here, but maybe they should be?

The CLI server bump is from 1.7.4-standalone-nexus-operations to 1.8.3-server-1.32.0-162.0. Note that doing the same bump in the sample repo makes some of the sample code no longer run. 

--Sample code needs an extra dynamic flag, nexusoperation.enableStandalone. That is why it failed.


GregoryTravis (approved):

mjameswh:
> Note that doing the same bump in the sample repo makes some of the sample code no longer run.

Can you list which samples stopped working after this bump?

A very common cause for this kind of regression on CLI bump is when a test or sample is using trying to use a new/unstable server-side API that need is gated by a dynamic config flag.

mjameswh (changes requested):
Please try and at least clarify why some of the samples stopped working after sever bump. My feeling is that is probably a missing dynamic config flag.

mjameswh (approved):

PR #3055, opened 2026-09-03
Add promise-based workflow await

### LINKED ISSUE #1937: Support a promise-based await
An SDK function that works similarly to [await](https://javadoc.io/doc/io.temporal/temporal-sdk/1.0.0/io/temporal/workflow/Workflow.html#await-java.time.Duration-java.util.function.Supplier-) but is based on promises rather than blocking could be useful to expand developer options for coordinating mixed event classes.  For instance, coordinating an .all() on an invoked activity and receiving a signal.

This interface could be designed in several different ways.  As an example:

```
Promise until(java.time.Duration timeout, java.util.function.Supplier<java.lang.Boolean> predicate)
```
This function would take a timeout and predicate, and return a promise.  Temporal will evaluate the predicate on each state transition (similar to .await), and will resolve the promise with "true" if the predicate returns true. and "false" if the timeout expires.

baekgyu-kim:
## What was changed

- Added `Async.await(Duration, Supplier<Boolean>)`, a non-blocking, promise-based version of `Workflow.await`.
- Runs the existing `Workflow.await` on a separate workflow thread. The promise completes with `true` when the condition is met, `false` on timeout, or exceptionally on cancellation or a predicate exception.
- Added tests for completion, timeout, cancellation, predicate exceptions, non-blocking behavior, and composition with an activity promise.

## Why?

`Workflow.await` blocks the current workflow thread, so its result cannot be combined with other promises. This API allows workflows to use an await condition with `Promise.allOf`, for example when waiting for both a signal and an activity.

## Checklist

1. Closes #1937

2. How was this tested:

   ```bash
   ./gradlew :temporal-sdk:test --offline \
     --tests io.temporal.workflow.AsyncAwaitTest \
     --tests io.temporal.workflow.AwaitTest \
     --tests io.temporal.workflow.cancellationTests.WorkflowAwaitCancelTimerOnConditionTest
   ```

3. Any docs updates needed?

   No. The new API is documented with Javadocs.


temporal-sdk/src/main/java/io/temporal/workflow/Async.java
```diff
@@ -217,6 +218,28 @@ public static <A1, A2, A3, A4, A5, A6> Promise<Void> procedure(
...
+   * completes exceptionally with a {@link io.temporal.failure.CanceledFailure}. An exception thrown
+   * by the condition also completes the promise exceptionally.
+   *
+   * @param timeout time after which the promise completes with {@code false} if the condition is
+   *     not satisfied.
+   * @param unblockCondition condition that completes the promise with {@code true} when satisfied.
+   * @return promise that contains whether the condition was satisfied before the timeout.
+   */
+  public static Promise<Boolean> await(Duration timeout, Supplier<Boolean> unblockCondition) {
+    return AsyncInternal.await(timeout, unblockCondition);
```
Quinn-With-Two-Ns, line 260: Can we also add a overload that take no duration to be consisiten with `Workflow.await`

baekgyu-kim: Hi @Quinn-With-Two-Ns, thanks for taking a look at this issue.

I added a no-timeout overload to match `Workflow.await`, along with documentation and tests. 
Please take a look when you have a chance. Thanks!

PR #3057, opened 2026-09-07
💥 Fix wrong exception thrown by manual activity heartbeats

dplyukhin:
## What was changed

💥 BREAKING CHANGE: `ManualActivityCompletionClientImpl.recordHeartbeat` can now throw `ActivityCanceledException`/`ActivityResetException`/`ActivityPausedException`. Previously, those exceptions were wrapped inside `ActivityCompletionFailureException` and you had to access them with `getCause()`.

Other changes:
- `ActivityHeartbeatResponse`: New wrapper class to reduce code duplication
- Added `ManualActivityCompletionClientImplTest` unit test

## Why?

`ActivityCompletionFailureException` is supposed to be mutually exclusive with `ActivityCanceledException`/`ActivityResetException`/`ActivityPausedException`; it shouldn't wrap them.

## Checklist

1. Closes  #2983

2. How was this tested: Added `ManualActivityCompletionClientImplTest` unit test to check that thrown exceptions have the expected type.

3. Any docs updates needed? No

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/activity/ManualActivityCompletionClient.java
+++ b/temporal-sdk/src/main/java/io/temporal/activity/ManualActivityCompletionClient.java
@@ -1,6 +1,6 @@
 package io.temporal.activity;
 
-import io.temporal.failure.CanceledFailure;
+import io.temporal.client.ActivityCompletionException;
 import javax.annotation.Nonnull;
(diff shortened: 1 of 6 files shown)


dplyukhin:
@Quinn-With-Two-Ns this is a reopened version of #2990 with your review comments incorporated.

PR #3063, opened 2026-09-10
:boom: Use ActivitySerializationContext when starting and getting result of Standalone Activity

maciejdudko:
## 💥 **BREAKING CHANGE**
`ActivitySerializationContext.activityType` and `activityTaskQueue` are now `@Nullable` because they are unavailable when fetching result.

## What was changed

- Added `ActivitySerializationContext` to `DataConverter` used for de/serialization of payloads in `ActivityClient.start` and `ActivityHandle.getResult`.
- Changed `ActivitySerializationContext.activityType` and `activityTaskQueue` to `@Nullable`.

## Why?

Makes it possible to use context-aware data converters with Standalone Activities.

### DIFF
--- a/temporal-sdk/src/main/java/io/temporal/internal/client/RootActivityClientInvoker.java
+++ b/temporal-sdk/src/main/java/io/temporal/internal/client/RootActivityClientInvoker.java
@@ -32,6 +32,7 @@ import io.temporal.internal.common.SearchAttributesUtil;
 import io.temporal.internal.nexus.CurrentNexusOperationContext;
 import io.temporal.internal.nexus.InternalNexusOperationContext;
 import io.temporal.internal.nexus.NexusOperationMetadata;
+import io.temporal.payload.context.ActivitySerializationContext;
 import io.temporal.serviceclient.StatusUtils;
 import java.lang.reflect.Type;
 import java.util.*;
@@ -63,7 +64,18 @@ public class RootActivityClientInvoker implements ActivityClientCallsInterceptor
     if (Strings.isNullOrEmpty(options.getTaskQueue())) {
       throw new IllegalArgumentException("taskQueue must not be null or empty");
     }
-    DataConverter dc = clientOptions.getDataConverter();
+    DataConverter dc =
+        clientOptions
+            .getDataConverter()
+            .withContext(
+                new ActivitySerializationContext(
+                    clientOptions.getNamespace(),
+                    null,
+                    null,
+                    input.getActivityType(),
+                    options.getTaskQueue(),
+                    false));
+
     InternalNexusOperationContext nexusContext =
         CurrentNexusOperationContext.isNexusContext() ? CurrentNexusOperationContext.get() : null;
     NexusOperationMetadata nexusOperationMetadata =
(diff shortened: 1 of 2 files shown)


temporal-sdk/src/main/java/io/temporal/internal/client/RootActivityClientInvoker.java
```diff
@@ -183,7 +195,11 @@ public StartActivityOutput startActivity(StartActivityInput input) {
   public <R> GetActivityResultOutput<R> getActivityResult(GetActivityResultInput<R> input)
       throws TimeoutException {
     String namespace = clientOptions.getNamespace();
-    DataConverter dc = clientOptions.getDataConverter();
+    DataConverter dc =
```
Quinn-With-Two-Ns, line 216: Hmm, technically here we can know the task queue if the handle was created by a start or execute call and we would often know the type if the user created the handle with the typed API

temporal-sdk/src/main/java/io/temporal/payload/context/ActivitySerializationContext.java
```diff
@@ -11,32 +11,34 @@ public class ActivitySerializationContext implements HasWorkflowSerializationCon
...
-   * @param activityType the activity type name; must not be {@code null}
-   * @param activityTaskQueue the task queue for this activity; must not be {@code null}
+   *     standalone activities
+   * @param activityType the activity type name, or {@code null} if unknown. Activity type is
+   *     unknown when getting a Standalone Activity result.
+   * @param activityTaskQueue the task queue for this activity, or {@code null} if unknown. Task
+   *     queue is unknown when getting a Standalone Activity result.
    * @param local {@code true} if this is a local activity
    */
   public ActivitySerializationContext(
```
Quinn-With-Two-Ns, line 30: Interesting we don't have activity ID here 🤔

Quinn-With-Two-Ns:
Shouldn't we have at least one test to show starting a Standalone activity uses the context?

temporal-sdk/src/main/java/io/temporal/internal/client/RootActivityClientInvoker.java
maciejdudko: Standalone Activities can have task queue changed after start. Really we should deprecate task queue in serialization context altogether.

Quinn-With-Two-Ns: I think there is some value in task queue since users can have different data converters on different task queues  so I wouldn't want to deprecate it IMO, but I can understand not wanting to assume it.

temporal-sdk/src/main/java/io/temporal/internal/client/RootActivityClientInvoker.java
dplyukhin: Does the async path need to change too?
```suggestion
    DataConverter dc =
        clientOptions
            .getDataConverter()
            .withContext(
                new ActivitySerializationContext(
                    clientOptions.getNamespace(), null, null, null, null, false));
```

maciejdudko: Yes, good find! Fixed.

Quinn-With-Two-Ns (approved):

PR #3070, opened 2026-09-12
Include inherited fields in service stub options toString

### LINKED ISSUE #2148: Better toString representations on service stub options

* `WorkflowServiceStubsOptions.toString` only includes direct fields not inherited ones
* `OperatorServiceStubsOptions.toString` and `CloudServiceStubsOptions.toString` (as of #2146) do not exist

sdjcse:
## What was changed

`ServiceStubsOptions.toString` now renders its fields through a package-private `toStringFields()`
that each subclass inlines alongside its own fields.

- `WorkflowServiceStubsOptions` rendered only the five fields declared on the subclass, dropping
  every inherited one — target, TLS, timeouts, headers.
- `OperatorServiceStubsOptions` and `CloudServiceStubsOptions` had no `toString` and logged as an
  identity hash.
- Added `apiKeyProvided`, already in `equals`/`hashCode` but missing from `toString`.
- Headers render as names only. `io.grpc.Metadata#toString` prints values in the clear, so
  inheriting the field as-is would have exposed credentials set through `setHeaders`.
- Fixed a missing `, ` before `congestionInitialInterval` in `RpcRetryOptions.toString`.

## Why?

This output is what lands in logs when diagnosing a client. Two of the three option types rendered
nothing usable; the third omitted every inherited field.

## Breaking changes?

None to public API. `ServiceStubsOptions.toString` now emits `headerNames=[...]` where it used to
emit `headers=` with values — deliberate, so the fix does not leak credentials.

## Server PR

N/A.

## Checklist

1. Closes #2148

2. How was this tested:

   Five new tests covering each subclass's `toString`, asserting no API key or header value appears
   in the output, plus a regression test for the separator.

   ```
   ./gradlew --offline :temporal-serviceclient:spotlessCheck
   ./gradlew --offline :temporal-serviceclient:test
   ```

3. Any docs updates needed?

   No.

### DIFF
--- a/temporal-serviceclient/src/main/java/io/temporal/serviceclient/CloudServiceStubsOptions.java
+++ b/temporal-serviceclient/src/main/java/io/temporal/serviceclient/CloudServiceStubsOptions.java
@@ -58,6 +58,11 @@ public final class CloudServiceStubsOptions extends ServiceStubsOptions {
     return Objects.hash(super.hashCode(), version);
   }
 
+  @Override
+  public String toString() {
+    return "CloudServiceStubsOptions{" + toStringFields() + ", version='" + version + '\'' + '}';
+  }
+
   /** Builder is the builder for ClientOptions. */
   public static class Builder extends ServiceStubsOptions.Builder<Builder> {
     private String version;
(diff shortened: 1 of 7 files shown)


Quinn-With-Two-Ns (approved):

Quinn-With-Two-Ns:
Thanks for the contribution, note the CLA still must be signed before we can merge this

PR #3072, opened 2026-09-14
Set WaitForStage to ACCEPTED by default for Nexus Workflow Updates

atol:
## What was changed
Set the `WaitForStage` option to `ACCEPTED` by default for Nexus Workflow Updates.

## Why?
`ACCEPTED` is the only stage supported for async Nexus updates. Instead of requiring users to pass in this value, we set the appropriate default.

## Checklist

1. Closes: N/A

2. How was this tested:
* Added new unit tests
* Updated integration test 

3. Any docs updates needed?
No.


Quinn-With-Two-Ns:
`WaitForStage` is expected to be a required parameter since it is not the obvious behaviour. It is required for regular update as well . The reason is the required is because `ACCEPTED` is a very surprising default

Quinn-With-Two-Ns (changes requested):
I don't think this is something we should do

PR #3075, opened 2026-09-15
First pass at NexusSerializationContext for Java

Evanthx:
The logic for this code change was originally in Python:
https://github.com/temporalio/sdk-python/pull/1828
This is a port of that logic into the Java SDK.

## What was changed
Added NexusSerializationContext for nexus callers and sync handlers. This allows data and failure converters to use nexus endpoint, service and operation to be used for encoding and decoding.

## Why?
This enables codecs to select serialization behavior or encryption keys by Nexus endpoint, service, or operation.

For example, workflows calling two Nexus endpoints can encrypt each endpoint’s payloads with a different key while ensuring that inputs, results, and failures are decoded with the converter selected for the corresponding operation.

How was this tested:
Unit testing


temporal-sdk/src/main/java/io/temporal/common/interceptors/NexusClientCallsInterceptor.java
```diff
@@ -250,18 +251,35 @@ final class GetNexusOperationResultInput<R> {
...
+      this(operationId, runId, deadline, resultClass, resultType, null);
+    }
+
+    public GetNexusOperationResultInput(
+        String operationId,
+        @Nullable String runId,
+        @Nonnull Deadline deadline,
+        Class<R> resultClass,
+        @Nullable Type resultType,
+        @Nullable NexusSerializationContext serializationContext) {
```
Quinn-With-Two-Ns: We shouldn't need to pass this through the interceptor, we don't do this for other `SerializationContext`. We can generally just construct it just in time before use

Quinn-With-Two-Ns: Actually I. will slightly retract my comment since this is just on `GetNexusOperationResultInput`

Quinn-With-Two-Ns: I am debating a if it  would be better to expand it here instead of passing the context but not sure TBH

Quinn-With-Two-Ns:
Overall makes sense I think we just need some integration tests

temporal-sdk/src/main/java/io/temporal/common/interceptors/NexusClientCallsInterceptor.java
Evanthx: I can, but there are two calls to this method and both have a serialization context. So it feels odd to take a class and pass through each individual field of it instead of the class, just so that the method can then create a new copy of that class from the individual fields. Then if we ever add a field that gets weird again.

VegetarianOrc: Not sure it's the exact same situation, but I felt it was the better choice to expand it in Python as well when reviewing this PR https://github.com/temporalio/sdk-python/pull/1828 from @JoshuaFrenchwood.

Evanthx: Thinking about this ... the places making the call are internal and do create the context, so unwrapping it or not just isn't a big deal honestly, so I just did it.

But it makes me wonder - what made this stand out to you two? Not asking to disagree but I wanted to understand - normally I'd go for earlier construction and fewer arguments, it surfaces errors earlier and is more readable. What am I missing that made both of you lean the other way?

Quinn-With-Two-Ns: The thing that stood out to me is the type `NexusSerializationContext` , and all the serialization context, are types that are containers to pass some information from the SDK to the users Data Converter so passing it through the interceptor already stand outs. Now I understand the argument that we are passing this information through the interceptor to support serialization context BUT I feel like that is an implementation detail that authors of interceptors shouldn't need to be aware of.

There is also the argument about what if we added more info to the `NexusSerializationContext` like operation ID? Then we would be passing that information twice and that could be confusing.

temporal-sdk/src/main/java/io/temporal/client/NexusOperationExecutionDescription.java
```diff
@@ -25,11 +25,23 @@ public final class NexusOperationExecutionDescription extends NexusOperationExec
   private final DescribeNexusOperationExecutionResponse response;
   private final NexusOperationExecutionInfo info;
   private final DataConverter dataConverter;
+  // User metadata is attached by the caller without a Nexus serialization context, so it must be
```
Quinn-With-Two-Ns: Is there a reason user metadata isn't suing the context? Is that what we did for stand alone activities or workflows? If so no issue just want to be consistent

temporal-sdk/src/main/java/io/temporal/client/NexusOperationExecutionDescription.java
```diff
@@ -25,11 +25,23 @@ public final class NexusOperationExecutionDescription extends NexusOperationExec
...
+  // decoded without one too. Everything else on a description belongs to the operation and is
+  // decoded with the operation's context.
+  private final DataConverter userMetadataDataConverter;
```
Quinn-With-Two-Ns: Nit: generally, but not always, we have used the pattern `dataConverterWithXContext` where X is the context type

Evanthx: I like this, changed to match it.

temporal-sdk/src/main/java/io/temporal/client/NexusOperationExecutionDescription.java
Evanthx: This seems to be a pretty inconsistent picture. I have bee poking at this for a bit, and if I have this right, in Java workflows have the context for this, but standalone seems to encode without a context but decoded with one? I tested that in code, the client encoded standalone Nexus activities without a context but the worker decoded it with a ActivitySerializationContext and failed.

Going back to this PR, Go and Python is doing this in the same way this PR does, but when digging it seems that Python might encode the summary without a context but decode it with one so there might be a bug there.

SO. For this question - I can make an argument for using the context either way. Maybe the metadata has information that should be encrypted? But then if we encrypt it, does the UI or anything display the metadata? If it's encrypted, we would lose that. Alternately we would have to pass the context everywhere that uses the metadata?

Let me know what you think about that. Right now I am thinking leaving metadata unencrypted, if we do encrypt it then that might be a separate PR that is coordinated across anyplace reading the metadata?

For the bug in Java and Python - let me get this PR merged as it is unrelated to the pre-existing java bug, and I will then fix it and submit that PR. I'll talk to @JoshuaFrenchwood about the Python bug. And I think for these I will also see if as part of this we can improve testing, as it seems like several things went through our existing testing.

Quinn-With-Two-Ns: > standalone Nexus activities without a context but the worker decoded it with a ActivitySerializationContext and failed.

What is a standalone Nexus activities? Do you mean stand alone activities? If so then no worries there is actually another PR for that already.

Quinn-With-Two-Ns (approved):

PR #3076, opened 2026-09-16
Handle poller resubmission during shutdown

frojasg:
## What changed?

Catch `RejectedExecutionException` when a poll loop re-submits itself and its executor has shut down. Rejections from an executor that is still running continue to reach the uncaught exception handler.

## Why?

`MultiThreadedPoller` checks `shouldTerminate()` before it schedules the next poll. Shutdown can start between that check and `execute(this)`. The resulting rejection escapes from `finally`, bypasses the shutdown-aware handler, and produces an ERROR log during normal worker shutdown.

The catch handles that race without changing task handoff or shutdown ordering. It checks the executor's shutdown state, so thread interruption alone cannot suppress an unexpected rejection.

## Validation

The two new shutdown-race tests fail on the original code and pass with this change. Tests also verify that running-executor rejections remain visible, including on an interrupted thread.

On Java 21, formatting, SDK assembly, and all 13 tests across `MultiThreadedPollerTest`, `GracefulPollShutdownTest`, `ShutdownManagerTest`, and `WorkerShutdownTest` passed. The full test suite was not run locally.

## Breaking changes?

None. No public API changes.

## Server PR

None required.

Generated with Codex


Quinn-With-Two-Ns:
Hi, thank you for the contribution. Before we review your PR could you please sign the CLA

frojasg:
Done signing

PR #3084, opened 2026-09-16
Nexus serialization consistency

Evanthx:
This is following https://github.com/temporalio/sdk-java/pull/3075. I did a consistency pass between DotNet and Python and Java, and that resulted in a few changes. Also changed the handling to encode the summary with a context.


temporal-sdk/src/main/java/io/temporal/internal/client/RootNexusClientInvoker.java
```diff
@@ -108,7 +108,17 @@ public StartNexusOperationExecutionOutput startNexusOperationExecution(
     if (options.getSummary() != null) {
       UserMetadata metadata =
           WorkflowExecutionUtils.makeUserMetaData(
-              options.getSummary(), null, clientOptions.getDataConverter());
+              options.getSummary(),
+              null,
+              // User metadata is serialized with the operation's context, the same way workflow
```
Quinn-With-Two-Ns: IMO not a very useful comment since obviously we would always use the converter with context

Quinn-With-Two-Ns (approved):

temporal-sdk/src/main/java/io/temporal/internal/nexus/NexusTaskHandlerImpl.java
```diff
@@ -159,23 +165,45 @@ public Result handle(NexusTask task, Scope metricsScope) throws TimeoutException
...
+   *
+   * <p>Note that servers before 1.30.0 do not report the endpoint the task was addressed to, so the
+   * context is scoped by an empty endpoint there and does not agree with the caller's. Nexus
+   * serialization context on the handler side requires server 1.30.0 or later.
    */
   private void setSerializationContext(String service, String operation) {
     InternalNexusOperationContext nexusContext = CurrentNexusOperationContext.get();
+    String endpoint = nexusContext.getEndpoint();
+    if (endpoint.isEmpty() && warnedMissingEndpoint.compareAndSet(false, true)) {
+      log.warn(
```
Quinn-With-Two-Ns: Bit on the fence about this since it would warn for users who don't care about this feature, but I can see the value

temporal-sdk/src/main/java/io/temporal/internal/worker/NexusWorker.java
```diff
@@ -550,10 +551,19 @@ private void sendReply(
                   .setTaskToken(taskToken)
                   .setIdentity(options.getIdentity())
                   .setNamespace(namespace);
+          // The caller decodes this failure with the operation's context, so it has to be encoded
+          // with the same one. The context rides on the result because it is no longer in scope by
+          // the time the reply is built.
+          NexusSerializationContext serializationContext = response.getSerializationContext();
+          DataConverter failureDataConverter =
```
Quinn-With-Two-Ns:
```suggestion
          DataConverter dataConverterWithContext =
```
Nit: for consistency, but not a strong preference

temporal-sdk/src/main/java/io/temporal/internal/nexus/NexusTaskHandlerImpl.java
Evanthx: It's removed

temporal-sdk/src/main/java/io/temporal/internal/worker/NexusWorker.java
Evanthx: done

